PR #37, opened 2015-07-16
Reporting wrongly-cased file imports in case-insensitive operating systems

doctyper:
This pull request fixes an issue we've dealt with internally. We primarily use OS X as our development OS and deploy to Linux. OS X is case-insensitive, meaning that given a file called `MyUnCoolComponent.jsx`, the following import will work without issue on OS X:

```js
// Note the casing error in the file name
import MyUnCoolComponent from "./MyUncoolComponent.jsx"
```

However, this import will fail on a case-sensitive OS such as Linux. This PR adds functionality to check the casing in the file path before reporting a successful file match.

### DIFF
--- a/lib/core/resolve.js
+++ b/lib/core/resolve.js
@@ -1,9 +1,20 @@
 'use strict'
 
-var path = require('path')
+var fs = require('fs')
+  , path = require('path')
   , resolve = require('resolve')
   , assign = require('object-assign')
 
+// http://stackoverflow.com/a/27382838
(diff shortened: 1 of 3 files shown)


benmosher:
Ah, cool! 

We do the same thing (dev on OS X, run on Linux) and have seen issues with this, but I hadn't thought to check for it in the plugin.

Thanks!

benmosher:
BTW: I really appreciate all the feedback!

Issue #44, opened 2015-08-10
Path resolution doesn't work on Windows


illinar:
fc4f99ecda1d378c277e05fd9f8f7d72d08ffb60 breaks file resolution on Microsoft Windows. The implementation of [fileExistsWithCaseSync](https://github.com/benmosher/eslint-plugin-import/blob/master/src/core/resolve.js#L8) fails to take into account drive letters and will always return false due to path.basename("C:\") returning an empty string.

benmosher:
Bummer! 

Assuming there's not a simple way to fix the issue in the function itself (which I can't test, don't have a Windows box handy), it seems like a decent fix would be to make case-sensitive checking optional and expose it as a rule config option, i.e. `import/exists: [2, 'case-sensitive']`. 

Probably would turn it off by default.

@doctyper, would that still work for you? You'd have to update your config.

doctyper:
Hail-mary stab might work if you just bail out on windows. Otherwise I'm ok with the flag.

doctyper:
Which, actually, would be fine because the case-sensitivity issue is Mac-only.

illinar:
Since NTFS is case-sensitive I am fine with this approach. I would also like to indicate that NTFS path does not have to start with the letter. You may want to keep a note regarding limitations of the current implementation on Windows.

benmosher:
Ah, didn't realize NTFS was case-sensitive. OS X is truly the odd man out, then. Though I did have one machine formatted with case-sensitive HFS+, once upon a time.

I think I'll keep the flag, and try to set up an AppVeyor Windows build at some point just for good measure.

@doctyper: interesting idea to just skip Windows using `os.platform()`, but for some reason I don't love the idea of having a 'win32' floating around in there. Though I'll admit the drive-letter regex is essentially the same thing, stated less obviously. I'll think about it. :smile:

doctyper:
Keeping the flag is fine, but I think it should be opt-out not opt-in. It's value is in warning the developer about casing pitfalls ahead of time.

benmosher:
That's a fair point. I suspect Linux is the build/deploy target for most devs. Or at the very least, something other than OS X.

Would it make sense to only do the case-sensitive extra work for `os.platform() === 'darwin'`?

benmosher:
Published fix via v0.7.8.

PR #45, opened 2015-08-11
optional case sensitivity

benmosher:
Attempted fix for #44, two-fold:

- `case-sensitive` option for `no-unresolved`. defaults to false.
- added an untested drive-letter regex base case to `fileExistsWithCaseSync`

@doctyper and @illinar, looking for comments from you two before I merge this to master and release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,15 +40,6 @@ as defined by standard Node `require.resolve` behavior.
 See [settings](#settings) for customization options for the resolution (i.e.
 additional filetypes, `NODE_PATH`, etc.)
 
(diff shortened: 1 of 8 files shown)


benmosher:
So I changed the strategy to instead detect at startup whether the FS is case-sensitive, and use the vanilla `fs.existsSync` function if it is; otherwise, @doctyper's solution for OSX from SO.

Also set up AppVeyor to run tests on Windows, which now pass (finally).

Any final thoughts?

doctyper:
:shipit:

PR #132, opened 2015-11-30
jsnext:main

### LINKED ISSUE #103: Use Rollup's 'jsnext:main' to collect names
https://github.com/rollup/rollup/wiki/jsnext:main

May need to fork `resolve` (or use Rollup's resolver?) to use this key from `package.json` instead of `main` to find the ES6 source for modules.

benmosher:
Closes #103.

- node resolver defers to `jsnext:main` in `package.json`.
- webpack resolver will as well, in addition to respecting Webpack's `resolve.packageMains` field properly.
- **breaking**: `import/ignore` patterns are checked only once export parsing has yielded no results. 
  This means `node_modules` with `jsnext:main` (i.e. Redux) will have names collected in spite of `node_modules` being ignored by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ build/Release
 # Commenting this out is preferred by some people, see
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git-
 node_modules
+!tests/**/node_modules/
 
 # Users Environment Variables
 .lock-wscript
--- a/.npmignore
+++ b/.npmignore
@@ -1,11 +1,11 @@
 # files
-src/
 reports/
 resolvers/
 
 # config
 .eslintrc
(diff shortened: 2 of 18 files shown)


taion:
Wait, why? webpack 1 can't handle ES modules, so there's no point looking at `jsnext:main`.

benmosher:
It can't, yes, but the original ES6 source is necessary for the plugin to lint correctly

taion:
Ah, for named imports?

You should consider adding `module` there as well, then. It's the slightly more standardized (but still will-never-be-in-Node) successor to `jsnext:main`.

Most libraries are starting to support it now, and it's what webpack 2 will check by default.

benmosher:
Yep, meaning to do that soon 😅

taion:
I'll add it to #475.

PR #136, opened 2015-11-30
Allow 'absolute' specification of webpack config.

nosnickid:
As the title, I have a project structure akin to:

```
node/webpack.config.js
node/package.json
src/module1/submodule/...
src/module2/submodule/...
```
And I alias each src/moduleX folder to something more succinct. Names have been changed to keep identity of real code secret etc ;)

Hence, it would be handy for me to be able to just specify an absolute webpack location. This change makes that possible and tests it.

Thanks for the great plugin!

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -27,12 +27,17 @@ exports.resolveImport = function resolveImport(source, file, settings) {
 
   var webpackConfig
   try {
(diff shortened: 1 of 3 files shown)


benmosher:
Staged for release with Webpack resolver v0.1.4. Thanks for this! 😄

PR #138, opened 2015-12-01
webpack resolver and eslint-loader (fixes #134)

### LINKED ISSUE #134: webpack resolver: issue with eslint-loader
not sure what's happening, works fine from inside ESLint.

xjamundx:
we needed to add support for relative paths. this approach appears to work.


benmosher:
Fix makes sense, tests pass. Thanks for this! 😄

PR #144, opened 2015-12-08
update import/ignore documentation

AsaAyers:
I'm not actually using yaml, so I hope I got the syntax right here. I'm not too sure about the `coffee$`


benmosher:
Thanks for this! I've been trying to clean up and add more examples, this is great,

PR #164, opened 2016-02-03
Adds compiler loading for to-js languages in a similar fashion to Webpack itself

gausie:
As mentioned in https://github.com/benmosher/eslint-plugin-import/issues/102, this brings eslint-plugin-import in line with Webpack's own config loader.

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -3,6 +3,7 @@ var findRoot = require('find-root')
   , resolve = require('resolve')
   , get = require('lodash.get')
   , find = require('array-find')
+  , interpret = require('interpret')
   // not available on 0.10.x
   , isAbsolute = path.isAbsolute || require('is-absolute')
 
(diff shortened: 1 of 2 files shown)


benmosher:
Can you add tests?

An additional extra folder like https://github.com/benmosher/eslint-plugin-import/tree/master/resolvers/webpack/test/files/some with an ES6 config.

gausie:
Sorry that I haven't been able to do this. If someone else is able to (@kentcdodds if you're looking into this also?) that would be great.

Otherwise hopefully I'll have some time this week

kentcdodds:
Sorry, decided that we're actually just going to be upgrading to Node4 this week anyway, so we're not going to be using the `webpack.config.babel.js` filename anymore anyway. Feel free to close #194.

benmosher:
agh, didn't realize this commit wasn't properly part of #278, would've rebased onto it for proper credit. that said, was just published as v0.2.2 of the webpack resolver and you're credited in the change log 😅

gausie:
@benmosher not a problem and thanks!

PR #166, opened 2016-02-04
Allow 'nested' aliases in resolve config

nosnickid:
In my project, I have nested my private aliases in a namespace-ish manner.

My `resolve.alias` looks somewhat like:
```javascript
resolve: {
  alias: {
    'ns/module1': path.resolve('...'),
    'ns/module2': path.resolve('...')
  }
}
```

This functions in webpack but wasn't functioning with the slightly simpler implementation of `matchAlias` in the webpack resolver.

This patch looks for a common prefix in the `source` and `alias` in `matchAlias` and fixes configs as above.

### DIFF
--- a/resolvers/webpack/resolve-alias.js
+++ b/resolvers/webpack/resolve-alias.js
@@ -18,13 +18,23 @@ var sep = '/'
 function matchAlias(source, alias, value) {
   var isExact = (alias[alias.length - 1] === '$')
     , isFile = (path.extname(value) !== '')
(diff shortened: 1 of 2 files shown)


benmosher:
Ah, interesting. I had not considered this given that it was not in the Webpack table.

I wonder if [path.relative](https://nodejs.org/api/path.html#path_path_relative_from_to) could be used for this... something like a `path` version of `indexOf` that is separator-agnostic...

Thanks for this! and thanks for the tests!

Issue #187, opened 2016-02-22
Performance issue


lfrodrigues:
In one of my projects (https://github.com/Seedstars/django-react-redux-jwt-base) eslint validation takes 3/4s to run. 

Adding eslint-plugin-import triples the time to 12s. 

Is there a way to improve run time?

benmosher:
So far, other than #149, I don't have many ideas, and even that will likely only improve performance by about 50%. So I'd predict you'd still be looking at 8s. I have some other ideas about caching a map of imports/exports on the FS, but that would be a pretty major exercise as well, and it's not obvious (without just doing it) how much it would help.

This plugin is so slow--relative to others, at least--because it must parse each imported file into an AST and collect metadata.

Due to the way ESLint is built (with good reason, [see discussion](https://github.com/eslint/eslint/issues/3565)), the file I/O must happen synchronously. My impression is that this is where most of the time is lost.

That said, I can't be sure, so it's risky (in a cost-benefit sense) to attempt any high-effort mission to improve it.

benmosher:
FWIW: [eslint_d](https://www.npmjs.com/package/eslint_d) with the `--cache` option takes me from ~32s to ~10s. 

Just note that there is a known potential gotcha with keeping the export cache in memory for a long time (#200) so it's not perfect, yet. Will try to get this fixed soon.

lencioni:
Using the timing env variable (e.g. `TIMING=1 eslint .`), this is what I'm seeing in a project I am using this plugin in, with only the `no-unresolved` rule enabled:

```
Rule                          | Time (ms) | Relative
:-----------------------------|----------:|--------:
import/no-unresolved          | 18563.065 |    40.6%
react/sort-comp               |  1621.646 |     3.5%
no-irregular-whitespace       |  1386.958 |     3.0%
react/prop-types              |  1197.552 |     2.6%
no-implied-eval               |   829.967 |     1.8%
react/jsx-uses-vars           |   765.702 |     1.7%
keyword-spacing               |   765.665 |     1.7%
no-eval                       |   578.730 |     1.3%
semi-spacing                  |   549.684 |     1.2%
no-whitespace-before-property |   477.285 |     1.0%
```

I think for the `no-unresolved` rule, adding a cache would likely help a good amount.

lencioni:
Glancing at the code, I think that even memoizing some stuff in the `resolve()` path would likely help on full runs like this.

benmosher:
[...]

Issue #149, opened 2015-12-17
cache warmer


benmosher:
Should theoretically cut time by 25% or so by capturing exports while linting; this avoids the second parse operation for files that have already been linted that are then referenced from others.

- [ ] profile as % of total rule execution time:
  - file load time: ___
  - parse time: ___
  - export map building: ___

might be simpler to just move toward an FS cache implementation of the export map since that will be needed anyway

benmosher:
For the tests, out of 983ms:

read: 3.450444ms parse: 99.025113ms full: 159.063341ms

So reading is inconsequential, and parsing takes 10% of the time, and export collection is 15%.

This is probably not a great sample, but does imply that parsing is a decent chunk of the work.

benmosher:
nvm on this, [memo-parser](https://github.com/benmosher/eslint-plugin-import/blob/master/memo-parser/README.md) is a more complete solution

PR #199, opened 2016-03-02
FS cache

benmosher:
work in progress on caching import maps on disk between executions, in the interest of improved performance.

looks like it's still not enough to be worth the correctness risks (improvement from ~35s to ~25s total lint time for test set, with 50% reduction specifically in import rules)

related: #187, #149 

may want to goof around with making ESLint async and see how much that helps.

also may want to recommend `eslint_d` with `--cache` instead. cleaner, no need to create my own files.

need to test what is taking all the time on `--cache` calls, though

### DIFF
--- a/src/core/getExports.js
+++ b/src/core/getExports.js
@@ -7,18 +7,20 @@ import parse from './parse'
 import resolve from './resolve'
 import isIgnored from './ignore'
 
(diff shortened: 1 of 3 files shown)


benmosher:
nvm this, `eslint_d` seems like a better option. certainly simpler and needs to be supported.

PR #228, opened 2016-03-23
remove src and doc from npm package

mathieudutour:
avoid useless files to being send through npm


benmosher:
thanks for this! much simpler than the ever-growing `.npmignore` I was maintaining before!

mathieudutour:
You're welcome ;)

Issue #198, opened 2016-03-01
no-unstaged


benmosher:
flag imports of files that are not under source control

- [ ] should not flag `.gitignore`'d files

could be a mode for `no-unresolved`

benmosher:
this is implemented in benmosher/eslint-plugin-git

PR #234, opened 2016-03-28
Core package

### LINKED ISSUE #233: Expose the resolve function to other plugins
Hi,

I'm writing an eslint plugin that should check a few things with the modules, and I feel like it would be interesting to use the eslint-plugin-import resolver instead of rolling my own (especially since I use eslint-plugin-import too). I could copy/paste your code, but I feel like it would be much better to simply use the "standard implementation" right at the source (it would allow me to have a perfect compatibility with your module as far as the configuration goes). It would be as simple as this:

```js
var resolve = require('eslint-plugin-import').resolve;

module.exports = function (context) {
    resolve('some-module', context);
};
```

Would you mind consider exposing the `resolve` function to the world?

(Worth noting that I have seen that the resolvers are available as standalone packages - my point is mainly about exposing the eslint configuration parser rather than the resolvers themselves)

benmosher:
Pull out `src/core` into `eslint-module-utils`, to be published separately for other plugins to leverage.

Also, a bunch of v2 things.
(closes #233)

- [ ] revert breaking changes to 0.10/0.12 support
- [ ] document (README, etc)
- [ ] extract `hash` to tiny module and/or find existing (`object-hash`, `hash-object`)
- [ ] publish
- [ ] add published version to plugin package `dependencies`
- [ ] bump plugin to 2.0 (losing node 0.x support)
- [x] create `eslint-plugin-git` for #198 
- [x] ~~`babel-preset-argon`~~
- [ ] deal with `doctrine` dependency (either declare or pull out to import plugin, or extract `ExportMap` to third module?)


utils/moduleVisitor.js
```diff
@@ -0,0 +1,109 @@
...
+  function checkCommon(call) {
+    if (call.callee.type !== 'Identifier') return
+    if (call.callee.name !== 'require') return
+    if (call.arguments.length !== 1) return
+
+    const modulePath = call.arguments[0]
+    if (modulePath.type !== 'Literal') return
+    if (typeof modulePath.value !== 'string') return
+
+    checkSourceValue(modulePath)
```
lo1tuma: This function would also apply for `require` function calls when `require` is shadowed. So I think it would make sense to also check the scope where the reference come from.

benmosher: Makes sense, I'll add a note. 

Hadn't run into it so far, and that is part of why it checks the argument count and argument type (also to disambiguate from AMD require)

benmosher:
Superseded by #408!

PR #239, opened 2016-04-07
Add no-namespace rule

singles:
# no-namespace

Reports if namespace import is used.

## Rule Details

Valid:

```js
import defaultExport from './foo'
import { a, b }  from './bar'
import defaultExport, { a, b }  from './foobar'
```

...whereas here imports will be reported:

```js
import * as foo from 'foo';
import defaultExport, * as foo from 'foo';
```

## Motivation:

One should only import names which are needed in current scope, nothing more.

Additonally, users of tools like [rollup.js](http://rollupjs.org/) or [webpack](https://webpack.github.io/) 2.0 which support tree shaking would benefit in a smaller bundle size because unused exports that are not imported can be eliminated.

If all names are required in addition to separate named exports, developer might export default object with all required members:

```js
export const FOO = 1;
export const BAR = 2;

export default {
	FOO,
	BAR
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 ### Added
 - report resolver errors at the top of the linted file
+- add `no-namespace` rule
 
 ## [1.4.0] - 2016-03-25
 ### Added
--- a/README.md
+++ b/README.md
@@ -35,11 +35,13 @@ Style guide:
 * Report AMD `require` and `define` calls. ([`no-amd`])
 * Ensure all imports appear before other statements ([`imports-first`])
 * Report repeated import of the same module in multiple places ([`no-duplicates`])
+* Report namespace imports ([`no-namespace`])
 
 [`no-commonjs`]: ./docs/rules/no-commonjs.md
 [`no-amd`]: ./docs/rules/no-amd.md
 [`imports-first`]: ./docs/rules/imports-first.md
 [`no-duplicates`]: ./docs/rules/no-duplicates.md
+[`no-namespace`]: ./docs/rules/no-namespace.md
 
 Work in progress:
 
--- a/docs/rules/no-namespace.md
+++ b/docs/rules/no-namespace.md
@@ -1,0 +1,24 @@
+# no-namespace
+
+Reports if namespace import is used.
+
+## Rule Details
+
+Valid:
+
+```js
+import defaultExport from './foo'
+import { a, b }  from './bar'
+import defaultExport, { a, b }  from './foobar'
+```
+
+...whereas here imports will be reported:
+
+```js
+import * as foo from 'foo';
+import defaultExport, * as foo from 'foo';
+```
+
+## When Not To Use It
+
+If you want to use namespaces, you don't want to use this rule.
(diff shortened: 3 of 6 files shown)


benmosher:
FWIW: if there are no dynamic references to the namespace, you can still do tree-shaking. I don't believe there should be any performance impact to using a namespace import vs. individual names/default.

I am up for including this rule, but I think the motivation section of the doc should be revised to remove that language and it should be listed in the README under the `Style guide` list of rules?

lo1tuma:
@benmosher There is not always a way to statically analyze all references, for example:

```js
import * as foo from './foo';
import thirdPartyLibrary from 'third-party';

thirdPartyLibrary.doSomething(foo);
```

benmosher:
@lo1tuma fair enough. That would be easy enough to lint explicitly, though. I can imagine some more nuanced `enforce-tree-shaking` rule that reports anything that compromises static analysis:
- passing namespaces as a parameter (as you have shown)
- dynamic references (`foo[someKey]`)
- etc.

I just don't want folks getting the idea that imported namespaces themselves compromise analysis, which I think the current state of the "Motivation" section in this PR pretty strongly implies. 

Also "One should only import names which are needed in current scope, nothing more." is a pretty opinionated statement, IMO.

So I would be inclined to turn "Motivation" into 

```markdown
## When Not To Use It
If you want to use namespaces, you don't want to use this rule.
```

lo1tuma:
@benmosher Agreed, it makes sense to have a separate rule which explicitly enforces patterns that can be optimized by tree-shaking.

singles:
OK, I've updated rule docs and moved rule to the style guide section in README.md. 

Thanks for your feedback guys :)

benmosher:
Looks good! Thanks!

benmosher:
Oh actually, could you add a blurb under the `Unreleased` section of CHANGELOG.md?

singles:
@benmosher no problem, done.

benmosher:
Awesome, thanks!

gajus:
@singles Okay. The rule is built on a premise that discouraging use of _import all_ will enable tree-shaking. However, file size is not the only concern. For the most part, importing just certain properties makes the code easier to read, e.g.

```js
import {FOO, BAR} from './config';

// ...
// Do something with FOO, BAR
```

is a lot easier to scan than

```js
import config from './config';

// ...
// Do something with config.FOO, config.BAR
```

However, at the same time, I might need to import the entire `config`.

This rule encourages to use `export default` whenever there is at least one instance in code where there is need to import the entire config. In effect, promoting use of the pattern in the latter example.

singles:
@gajus tree shaking is one of the examples. Also, it's a good practice to import and use only stuff, that you will be using. 

Also in that case author of library/module can decide, if he want's to expose whole object or not. 

Then 

```js
export const FOO = 123;
export const BAR = 456;
```

allows you to import only specific members (and whole module using `*`)

but if library author decides that we would also like to export whole stuff, (s)he can always do:

```js
export const FOO = 123;
export const BAR = 456;

export default { FOO, BAR };
```

gajus:
@singles that is a good example:

```js
export const FOO = 123;
export const BAR = 456;

export default { FOO, BAR };
```

Thank you. This made me change the setting for `no-namespace-rule` from `0` to `2`. : )

Issue #126, opened 2015-11-27
no-extraneous-dependencies


benmosher:
Report `node_module` imports that are not declared in the sibling `package.json`.

~~Not sure what to name this. `no-missing-dependencies` maybe? or just `dependencies`?~~ (edit: thanks @grabbou for a great suggestion)

Needs an option to check in `dependencies`, `devDependencies`, optional, peer, etc. Just `dependencies` by default.

i.e. given `package.json`:
```json
{ "dependencies": { "x": "^1.0.1" } }
```

then report for 

```js
import x from 'x' // not reported
import y from 'y' // reported: module not in dependencies.
```

inspired by https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-missing-import.md

benmosher:
a la #158, need to allow/disallow core modules based on a flag, probably.

benmosher:
Given the recent Twitter attention, this is what I was imagining as an implementation:

- resolve import path
- if resolved path is not in a `node_modules` folder, ignore
- if no `package.json` is found for the module being linted, do not report (by default, at least)
- if it _is_ in a `node_modules`, verify that the relevant `package.json` for the module being linted (i.e. `path.join(findRoot(context.getFilename()), 'package.json')`) references the module

Options:
- `valid-types`: Array of strings, default is `['dependencies']`. The keys of `package.json` where the module must appear.
  I would envision setting this to `['dependencies', 'devDependencies']` for my `test/` directory, but using the (default) `['dependencies']` for `src/`/`lib/` paths.
- same object as `no-unresolved` that allows enabling/disablng `commonjs` (for `require(x)`) and/or `amd` (for `define`/`require`).

BTW: `valid-types` is a _terrible_ name, IMO. Would be happy with almost anything else.

I can also imagine this being a separate plugin (`eslint-plugin-npm`?) and leveraging the recently-published [eslint-module-utils](https://www.npmjs.com/package/eslint-module-utils). That guy lets you leverage all the `resolve` goodness of this plugin without having to be an explicit part. It knows everything this plugin knows about traversing imports/requires, resolving with Webpack/Node/whatever, etc.

Actually... I think that might be the best option. See [eslint-plugin-git](https://github.com/benmosher/eslint-plugin-git) for an example of how to integrate with the module utils package.

jfmengels:
I'm (as I'm writing) implementing it (and close to done).
[...]

PR #241, opened 2016-04-13
Add `no-extraneous-dependencies` rule

jfmengels:
Add `no-extraneous-dependencies` rule

(kinda) Continuous of https://github.com/jfmengels/eslint-plugin-import-order/pull/9.

I've added two files : `core/importType` and `core/staticRequire`. I use those in `import-order` for the import-order` rule, and will also use them in the no builtin rule probably.

Let me know if there's anything I need to change :)
(If this gets merged, I'll get started on the PRs for `import-order` and no builtin, but this reuses some of the same code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 ### Added
 - [`no-named-as-default-member`] to `warnings` canned config
+- add [`no-extraneous-dependencies`] rule
 
 ## [1.5.0] - 2016-04-18
 ### Added
--- a/README.md
+++ b/README.md
@@ -48,10 +48,12 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 * Ensure all imports appear before other statements ([`imports-first`])
 * Report repeated import of the same module in multiple places ([`no-duplicates`])
 * Report namespace imports ([`no-namespace`])
+* Forbid the use of extraneous packages ([`no-extraneous-dependencies`])
 
 [`imports-first`]: ./docs/rules/imports-first.md
 [`no-duplicates`]: ./docs/rules/no-duplicates.md
 [`no-namespace`]: ./docs/rules/no-namespace.md
+[`no-extraneous-dependencies`]: ./docs/rules/no-extraneous-dependencies.md
 
 
 ## Installation
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -1,0 +1,68 @@
+# Forbid the use of extraneous packages
+
+Forbid the import of external modules that are not declared in the `package.json`'s `dependencies` or `devDependencies`.
+The closest parent `package.json` will be used. If no `package.json` is found, the rule will not lint anything.
+
+### Options
+
+This rule supports the following options:
+
+`devDependencies`: If set to `false`, then the rule will show an error when `devDependencies` are imported. Defaults to `true`.
+
+You can set the options like this:
+
+```js
+"import/no-extraneous-dependencies": ["error", {"devDependencies": false}]
+```
+
+
+## Rule Details
+
+Given the following `package.json`:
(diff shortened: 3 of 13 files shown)


benmosher:
There are a few todos blocked on #234 merging into master, but I made some changes and pushed to another branch: https://github.com/benmosher/eslint-plugin-import/commit/266eb183bf086746911867b6b42eb1e6e1d4a214

Thought it might be simpler to write the code than the English :sweat_smile: 

Fundamentally, I just added a `project` import type and use the resolved path to distinguish the two.

I also changed the `pkgUp` call to use the filepath for the file being linted instead of the CWD of the lint process. This makes a difference for large projects with nested modules, and also gives you a bit more wiggle room for running ESLint from anywhere instead of just inside the project tree.

We may want to add a cache for the resolved package dependencies, but I think there is goodness in #234 that will make that easier too.

I'd also like to get AMD support via #234, but that shouldn't block merging + deploying this.

Things that need to get done either way:
- [ ] docs/rules/no-extraneous-dependencies
- [ ] changelog blurb

benmosher:
BTW I just made you a collaborator, figured that would make this and the `import-order` merge simpler. :smile:

jfmengels:
> I just made you a collaborator

Thanks!

>  I made some changes and pushed to another branch: 266eb18

Should I make a new PR with that new branch to discuss stuff there, or do you want to wait until #234 lands? Sounds like a good idea to me.

> I also changed the pkgUp call to use the filepath for the file being linted

Ah, that sounds like a good idea yeah! :+1: 

> Things that need to get done either way (README...)

My bad, I did them in my project, forgot to do that here. Will do that some time later.

I noticed you changed the `index` type, so that `./importType` gets considered as an index file if it points to an index file.
I don't agree with this change. The name was maybe badly chosen, but that was not the original purpose. The reason I wrote those types was to be able to order imports in `import-order` (see https://github.com/sindresorhus/module-requests/issues/62 for the original request).
To me, an `index` file is the `index` file of the current directory. Anything else, may it point to an index file or not, is not an index.
The whole "beauty" IMO when doing `import './bar'` is that you don't care if it actually points to `'./bar/index.js` or to `./bar.js`, which I find really useful (so much that I like to write a lot of `index` files, which then allows me to scope the internal methods of the module in other files under `./bar`).

Given the following directory:
- foo/
  - bar/
    - baz/
      - index.js
    - index.js
    - a.js (current file)
  - index.js
```js
// foo/bar/a.js
import './baz' // sibling
import './baz/index.js' // sibling
import '../' // parent

import './' // index
import './index' // index
```

In short, `./importType`, as in your test examples, should be a `sibling`, not an `index`.

benmosher:
**New PR**: I think that makes sense, we can hack on it together in this repo. I'm not sure when I'll get a chance to actually do the module-utils stuff (probably not sooner than Tuesday), so you can open the PR whenever feels right.

**On `index` type**: that all makes sense, I definitely misunderstood the intent.

It actually probably makes sense to go ahead and merge and ship the first version of this. I don't think the `index` mischaracterization impacts this rule, and it is already very useful. I don't see a reason to block on anything but the Changelog and a little README blurb--could even put it in the "experimental" block if we're feeling like it might break.

No rush, though, we can just merge it as soon as some minimum of documentation is in place (if you concur).

benmosher:
FYI: just flushed the other pending changes to npm with v1.5.0, ready to ship this as soon as docs are ready (I just didn't have quite enough time to do that and get the rest shipped this morning)

jfmengels:
Sure, will try to add this tonight!

jfmengels:
I've added documentation to your branch (and rebased it).
Let me know if there's anything else to change!

benmosher:
Thanks, looks good!

I need to undo the breaking change I made to the 'index' type, but I'll do that post merge.

jfmengels:
Cool!
For the index type, I undid it in the import-order PR.

benmosher:
Ah, got it. I should have looked there first, had just undone it in `master` locally. I'll dump my commit and merge yours.

benmosher:
Experiment: closes #126

PR #243, opened 2016-04-17
Add rule no-named-as-default-member

### LINKED ISSUE #205: no-named-as-default-member
Given:
```js
// ./whatever.js
export { someThing }
export default class Whatever { /* ... */ }
```
Warn on:
```js
import Whatever from './whatever'
Whatever.someThing // Reported: someThing exported as named export alongside Whatever.
```

Would be unusable/noisy if there are static fields on a default export that match exported fields (i.e. lodash) but that seems fine.

I.e. 
```js
import _, { isEmpty } from 'lodash'
```
would report on `_.isEmpty`, etc. (assuming lodash exported ES modules). so you could only use the default export for chaining (i.e. `_(thing).isEmpty()`) which makes sense, otherwise you'd want `* as _` anyway.

dmnd:
Fixes #205.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Added
 - report resolver errors at the top of the linted file
 - add `no-namespace` rule
+- add `no-named-as-default-member` rule
 
 ## [1.4.0] - 2016-03-25
 ### Added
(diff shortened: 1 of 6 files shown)


benmosher:
Sweet, thanks! I'll review and merge ASAP!

benmosher:
This is a model PR. Thanks for this!

dmnd:
Thanks. Also, thanks for the prompt merge & release!

Issue #244, opened 2016-04-17
Rule proposal: Enforce empty line below imports


sindresorhus:
I'm aware this is purely stylistic, but thought I'd propose it since it seems this plugin will soon expand to stylistic rules with the adoption of https://github.com/jfmengels/eslint-plugin-import-order

## Fail

```js
import foo from 'foo';
import bar from 'bar';
const unicorn = true;
```

```js
const foo = require('foo');
const bar = require('bar');
const unicorn = true;
```

## Pass

```js
import foo from 'foo';
import bar from 'bar';

const unicorn = true;
```

```js
const foo = require('foo');
const bar = require('bar');

const unicorn = true;
```

-

// @jamestalmage @jfmengels

benmosher:
Sounds good to me, will accept a PR.

benmosher:
Oh weird, expected #245 to close this but I guess it didn't mention it explicitly.

PR #245, opened 2016-04-18
Add `newline-after-import` rule

singles:
# newline-after-import (#244)

Reports if there's no new line after last import in group.

## Rule Details

Valid:

```js
import defaultExport from './foo'

const FOO = 'BAR'
```

```js
import defaultExport from './foo'
import { bar }  from 'bar-lib'

const FOO = 'BAR'
```

...whereas here imports will be reported:

```js
import * as foo from 'foo'
const FOO = 'BAR'
```

```js
import * as foo from 'foo'
const FOO = 'BAR'

import { bar }  from 'bar-lib'
```

## When Not To Use It

If you like to visually group module imports with its usage, you don't want to use this rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 * Report namespace imports ([`no-namespace`])
 * Ensure consistent use of file extension within the import path ([`extensions`])
 * Enforce a convention in module import order ([`order`])
+* Enforce a newline after import statements ([`newline-after-import`])
 
 [`imports-first`]: ./docs/rules/imports-first.md
 [`no-duplicates`]: ./docs/rules/no-duplicates.md
--- a/src/rules/newline-after-import.js
+++ b/src/rules/newline-after-import.js
@@ -1,0 +1,33 @@
+/**
+ * @fileoverview Rule to enforce new line after import not followed by another import.
+ * @author Radek Benkel
+ */
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+function getLineDifference(node, nextToken) {
+  return nextToken.loc.start.line - node.loc.start.line
+}
+
+module.exports = function (context) {
+  return {
+    'ImportDeclaration': function (node) {
+      const nextToken = context.getSourceCode(node).getTokenAfter(node)
+
+      if (!nextToken) {
+        return
+      }
+
+      if (getLineDifference(node, nextToken) === 1
+          && nextToken.type === 'Keyword' && nextToken.value !== 'import')
+      {
+        context.report({
+          loc: nextToken.loc.start,
+          message: 'Expected empty line after import statement not followed by another import.',
+        })
+      }
+    },
+  }
+}
--- a/src/rules/newline-after-import.js
+++ b/src/rules/newline-after-import.js
@@ -3,6 +3,8 @@
  * @author Radek Benkel
  */
 
+import isStaticRequire from '../core/staticRequire'
+
 //------------------------------------------------------------------------------
 // Rule Definition
 //------------------------------------------------------------------------------
@@ -11,22 +13,37 @@ function getLineDifference(node, nextToken) {
   return nextToken.loc.start.line - node.loc.start.line
 }
 
+function ensureNoForbiddenKeyword(context, node, tokenToInspect, tokenValue) {
+  if (!tokenToInspect) {
+    return
+  }
+
+  if (getLineDifference(node, tokenToInspect) === 1
+    && tokenToInspect.type === 'Keyword' && tokenToInspect.value !== tokenValue)
+  {
+    context.report({
+      loc: tokenToInspect.loc.start,
+      message: 'Expected empty line after ' + tokenValue +
+        ' statement not followed by another ' + tokenValue + '.',
+    })
+  }
+}
+
 module.exports = function (context) {
   return {
-    'ImportDeclaration': function (node) {
+    ImportDeclaration: function (node) {
       const nextToken = context.getSourceCode(node).getTokenAfter(node)
 
-      if (!nextToken) {
-        return
-      }
+      ensureNoForbiddenKeyword(context, node, nextToken, 'import')
+    },
+    CallExpression: function(node) {
+      if (isStaticRequire(node)) {
+        const nextTokens = context.getSourceCode(node).getTokensAfter(node, 2)
+        const tokenToInspect = nextTokens.length > 1 && nextTokens[0].type === 'Punctuator'
+          ? nextTokens[1]
+          : nextTokens[0]
 
-      if (getLineDifference(node, nextToken) === 1
-          && nextToken.type === 'Keyword' && nextToken.value !== 'import')
-      {
-        context.report({
-          loc: nextToken.loc.start,
-          message: 'Expected empty line after import statement not followed by another import.',
-        })
+        ensureNoForbiddenKeyword(context, node, tokenToInspect, 'require')
       }
     },
   }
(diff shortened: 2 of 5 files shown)


jfmengels:
Looks good to me, but is it possible to also support `require()` at the top of the document? Makes it quite a bit more complicated, but I know I (and the one who requested #244, cc @sindresorhus) would like to have that.

singles:
@jfmengels You mean, you would like it to work for `import foo from 'foo'` and/or `var foo = require('foo')`?

sindresorhus:
@singles Yes :)

README.md
```diff
@@ -48,6 +48,7 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 * Ensure all imports appear before other statements ([`imports-first`])
 * Report repeated import of the same module in multiple places ([`no-duplicates`])
 * Report namespace imports ([`no-namespace`])
+* Enforce new line after import not followed by another import([`newline-after-import`])
```
sindresorhus: Maybe:

```
Enforce a newline after import statements ([`newline-after-import`])
```

singles:
@sindresorhus @jfmengels thank you guys for feedback. 

Regarding `require`. I didn't add support for `require` after reading docs, which says that `This plugin intends to support linting of ES2015+ (ES6+) import/export syntax`. 

Also I checked `test` folder for usage of `require` and none of the current rules support that, while you could think that for example `imports-first` should do that, because `require` is still an import, just other syntax. 

Please don't get me wrong - I'm not trying to be cocky or something, it just matter of consistency for me :)

jfmengels:
You're right that this project doesn't support `require`. There are some rules coming in from another [project](https://github.com/jfmengels/eslint-plugin-import-order) which has rules that does support `require` statements.

From what I understood, this plugin supports rules that can infer something from static analysis using `import`. Using `require`, that becomes harder/not possible, and therefore, most rules can't be applied to uses of `require` (might be wrong, haven't had the chance to look at this plugin in too much detail). In this case (and the rule I mentioned above), it is simply a stylistic issue (which are new to the project too), and can be applied to both "methods".

@benmosher Thoughts on supporting `require` where possible (and first of all, here)?

README.md
singles: Thx, fixed :)

benmosher:
FWIW, `no-unresolved` does support both CommonJS and AMD module syntax, if you are using Browserify or Webpack. Could theoretically support other AMD module loaders if the resolvers were implemented.

For that rule, you may supply an options object like `{ commonjs: true, amd: true }` to enable one or both.

I think it makes sense to support CommonJS to the extent that 

- [x] the community wants it 
- [x] it's technically feasible

I think that is true on both counts in this case.

singles:
Now when #247 is here and based on comments from #246 I think it would be nice to solve this as a part of #247. Just additional option called `newline-after-import-group`, which will give you empty line before any other code, even if you will create one group with all import types. 

What do you think?

benmosher:
Tough call. I might enable this rule without worrying about ordering within the imports themselves, so I'm not sure it makes sense rolled into `order`. (@sindresorhus convinced me with a similar argument in discussion on #253)

I think this makes sense as you've implemented it for `import`s, and I think the `require` version would just ensure that the next line is either another `require` call (possibly an assignment, or could be side-effects only) or a blank line, i.e.

```js
const x = require('x')
require('jqueryui')
var y = require('y')

// ^ line above is enforced
function boringName(someArg) {
  return new Error("the worst function ever") 
}

let z = require('ugh-mutable-module-reference-ftl')

// ^ another enforced line here
function stahp() {
  process.exit(-∞)
}
```

No need to enforce on `require` calls except at module scope†, so you could implement with a `Program` node visitor with a loop over `program.body`.

†: unless @jfmengels or @sindresorhus say otherwise. I don't think it's important for v1 but I'm not going to use it, so what do _I_ know 😅 

I guess you could provide the same visitor for both `Program` and `BlockStatement` nodes, they each have a `body` field that is an array of `Statement`s.

singles:
OK, rebased onto `master` and added support for `require` :)

src/rules/newline-after-import.js
```diff
@@ -0,0 +1,52 @@
...
+      const nextToken = context.getSourceCode(node).getTokenAfter(node)
+
+      if (!nextToken) {
+        return
+      }
+
+      ensureNoForbiddenKeyword(context, node, nextToken, 'import')
+    },
+    CallExpression: function(node) {
+      if (node.callee.name === "require") {
```
jfmengels: Do you mind using single-quotes here? For consistency with the other strings.

src/rules/newline-after-import.js
```diff
@@ -0,0 +1,52 @@
...
+        const nextTokens = context.getSourceCode(node).getTokensAfter(node, 2)
+        const tokenToInspect = nextTokens.length > 1 && nextTokens[0].type === 'Punctuator'
+          ? nextTokens[1]
+          : nextTokens[0]
+
+        if (!tokenToInspect) {
+          return
+        }
```
jfmengels: You could move this check to `ensureNoForbiddenKeyword` for both `import` and `require` cases.

src/rules/newline-after-import.js
```diff
@@ -0,0 +1,52 @@
...
+  return nextToken.loc.start.line - node.loc.start.line
+}
+
+function ensureNoForbiddenKeyword(context, node, tokenToInspect, tokenValue) {
+  if (getLineDifference(node, tokenToInspect) === 1
+    && tokenToInspect.type === 'Keyword' && tokenToInspect.value !== tokenValue)
+  {
+    context.report({
+      loc: tokenToInspect.loc.start,
+      message: 'Expected empty line after import/require ' +
```
jfmengels: Since you know whether it's a `require` or an `import`, might as well indicate which one it is in the error message.

src/rules/newline-after-import.js
```diff
@@ -0,0 +1,52 @@
...
+// Rule Definition
+//------------------------------------------------------------------------------
+
+function getLineDifference(node, nextToken) {
+  return nextToken.loc.start.line - node.loc.start.line
+}
+
+function ensureNoForbiddenKeyword(context, node, tokenToInspect, tokenValue) {
+  if (getLineDifference(node, tokenToInspect) === 1
+    && tokenToInspect.type === 'Keyword' && tokenToInspect.value !== tokenValue)
```
jfmengels: My bad, but I have a hard time understanding the name `ensureNoForbiddenKeyword` (never did any rules that dealt with tokens), and how/what exactly it it testing (except the line difference thing, that I got). Do you mind explaining it to me?

src/rules/newline-after-import.js
singles: No problem, also see #280 :)

src/rules/newline-after-import.js
singles: Good catch, done.

src/rules/newline-after-import.js
singles: Yep, that would be useful, fixed.

src/rules/newline-after-import.js
singles: Role of that function is to check line difference AND if another token is something different that `import/require` - if it is different, then report an error, because only thing that can be directly in next line is either `import` or `require`.

jfmengels:
LGTM :)

singles:
@benmosher no problem, done.

benmosher:
Rebased and added changelog updates. 👍  it's in master, ready for publish via 1.7.0. Not sure when that will be, probably want to get some of @sindresorhus's issues resolved first.

jfmengels:
@benmosher It's not very nice of you to say that @sindresorhus has issues :p

sindresorhus:
It's true. I have a [gadzillion issues](https://github.com/search?q=user%3Asindresorhus&state=open&type=Issues&utf8=%E2%9C%93)...

jfmengels:
You made me found out I had way more issues than I thought... If it makes you feel better, I have a 3.5 hour trip, and I'll spend at least some of it solving your issues ^^

PR #250, opened 2016-04-20
Implement extensions rule

### LINKED ISSUE #204: Rule proposal: disallow no extension imports (folders, file without extension)
I would like to enforce this:

```js
import './file.js';
import './dir/index.js';
```

versus

```js
import './file';
import './dir/';
```

While still being able to require files that have no extension of course.

What do you think?

Is that doable?

### LINKED ISSUE #209: Rule Proposal: Disallow certain extensions in import path
When I transpile my `.jsx` files with babel they will be renamed to have a `.js` extension. That’s why I usually avoid the `.jsx` extentions in the import path, but sometimes other contributors of my project commit code that has import statements with file extensions.

Having an eslint rule would be perfect to identify such problems.

When the rule is configured with: `no-import-file-extensions: [ 2, [ '.js', '.jsx', '.json' ] ]` this would be considered as problems:

```js
import Foo from './Foo.jsx';
import lib from './lib.js';
import data from './data.json';
```

This would not be considered as a problem:

```js
import Foo from './Foo';
import lib from './lib';
import data from './data';
```

lo1tuma:
This fixes #209 and fixes #204.


docs/rules/extensions.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Rule Details
+
+This rule has one option which could be either a string or an object. If it is `"never"` (the default value) the rule forbids the use for any extension. If `"always"` then the rule enforces the use of extensions for all import statements.
+
+By providing an object you can configure each extension separately, so for example `{ "js": "always", "json": "never" }` would always enforce the use of the `.js` extension but never allow the use of the `.json` extension.
+
+### Exception
+
+When disallowing the use of certain extensions this rule makes an exception an allows the use of extension when the file would not be resolvable without extension.
```
jfmengels: `exception an allows` --> `exception and allows`

lo1tuma: Thanks, fixed.

package.json
```diff
@@ -72,7 +72,9 @@
     "es6-symbol": "*",
     "eslint-import-resolver-node": "^0.2.0",
     "lodash.cond": "^4.3.0",
+    "lodash.endswith": "^4.0.1",
     "object-assign": "^4.0.1",
     "pkg-up": "^1.0.0"
+    "object-assign": "^4.0.1"
```
jfmengels: `object-assign` is duplicated (and linked to that: there seems to be missing a comma after `pkg-up`).

package.json
```diff
@@ -72,7 +72,9 @@
     "es6-symbol": "*",
     "eslint-import-resolver-node": "^0.2.0",
     "lodash.cond": "^4.3.0",
+    "lodash.endswith": "^4.0.1",
     "object-assign": "^4.0.1",
-    "pkg-up": "^1.0.0"
+    "pkg-up": "^1.0.0",
+    "object-assign": "^4.0.1"
```
jfmengels: Still duplicated ;)
Just remove the last `object-assign` entry

package.json
lo1tuma: Fixed now.
Seems like I did a bad job at resolving the conflicts after rebasing.

jfmengels:
I think you can also add an entry to the changelog (and thank yourself).

Also, do you mind adding support for `require`? We had a discussion about a similar case here https://github.com/benmosher/eslint-plugin-import/pull/245#issuecomment-211433228.

LGTM up till now otherwise, Nice job @lo1tuma!

benmosher:
@jfmengels is there a utility function in what you've added to filter `CallExpression` to just CommonJS `require` calls? I've got something off in my utils branch, but I thought you had something similar.

I'm imagining something so CommonJS support is as easy as:

```js
return {
  ImportDeclaration: checkFileExtension,
  CallExpression: filterRequire(checkFileExtension),
}
```

Anyway @lo1tuma if you don't want to worry about CommonJS, I'm cool with that, we can do it post-merge.

jfmengels:
Yeah, I have a [function for that](https://github.com/benmosher/eslint-plugin-import/blob/master/src%2Fcore%2FstaticRequire.js), with a use of it [here](https://github.com/benmosher/eslint-plugin-import/blob/master/src%2Frules%2Fno-extraneous-dependencies.js#L61-L64). You can't do something as simple as @benmosher proposed, as the same data (the `source` for instance here) as not at the same property path. It's more test code than code.

> we can do it post-merge

:+1:, I can probably take a look at it if you wish.

benmosher:
Ah, right. I made a `moduleVisitor` to which you pass a function of the `Literal` node, here it is in use for `no-unresolved`: https://github.com/benmosher/eslint-plugin-import/blob/core-package/src/rules/no-unresolved.js

I really need to get that stuff merged back in. CJS vs. ES6 is cross-cutting and common enough that ideally, rules wouldn't need to worry about it explicitly.

lo1tuma:
I’m not very familiar with AMD so I could only provide an implementation for commonjs.

Apart from that, I think it would be better to wait for the `moduleVisitor` and not implementing the same logic in every rule.

jfmengels:
Sounds good to me!

lo1tuma:
I totally forgot about the CHANGELOG. Should be all good now.

benmosher:
Just merged, thanks!

PR #256, opened 2016-04-22
Make imports-first treat directives as an exception

### LINKED ISSUE #255: Ignore directives for imports-first
We currently use a custom directive at the top of most JavaScript files. `imports-first` requires that any imports come before this line, but doing so means that it is no longer treated as a directive. I'm wondering if it would be possible to add an option to `imports-first` to ignore directives (which should be pretty easy to detect, even though I don't think ESLint exposes them as such). Basically, I would love something that lets the following pass:

```js
'expose Foo'

import MyThing from './somewhere'
// ...
```

but continue to fail anything that is not a directive appearing first:

```js
doSomething();

import MyThing from './somewhere'
// ...
```

I'm happy to take a stab at this if it's something you are interested in. Thanks!

lemonmade:
Fixes #255. Per discussion in that issue, directives need to appear at the top of a file in order to be considered directives. This isn't usually a problem because modules remove the need for `'use strict'`, but we use a custom directive as part of our build system. This PR makes `imports-first` understand directives and not throw errors when those appear above import statements.

cc/ @benmosher, let me know if anything else is needed; I didn't see a contributing file in the repo so not sure what the best practices are.

### DIFF
--- a/src/rules/imports-first.js
+++ b/src/rules/imports-first.js
@@ -1,11 +1,24 @@
 module.exports = function (context) {
+  function isPossibleDirective (node) {
+    return node.type === 'ExpressionStatement' &&
+      node.expression.type === 'Literal' &&
+      typeof node.expression.value === 'string'
+  }
+
   return {
     'Program': function (n) {
       const body = n.body
           , absoluteFirst = context.options[0] === 'absolute-first'
-      let last = -1
+      let nonImportCount = 0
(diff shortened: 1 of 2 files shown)


benmosher:
Just tacked a note into the changelog. 

Duly noted about a contribution guide, just opened #258 to that end 😄

PR #258, opened 2016-04-22
first draft of CONTRIBUTING.md

### LINKED ISSUE #257: CONTRIBUTING.md
The time has come for a contributing guide.

Plan:
- [x] find a good example
- [x] copy it with attribution

benmosher:
Given the recent uptick in interest, I figure the time has come to get something like this into `master`. Feels a little goofy to do something like this for a project that is just a plugin for something much larger, but at the same time, this is my primary open source project and it's been getting a lot of great contributions lately.

Comments are encouraged. I've not tried to write anything like this before and in general my written voice can be confusing and pedantic at the best of times.

(closes #257)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,109 @@
+# Contributing
+
+Thanks for your interest in helping out! Here are a **few** _weird_ tricks to
+~~cut your mortgage in half~~ maximize the global net efficiency of your efforts!
+
+## Issues
+
+### Search open + closed issues for similar cases.
+
+  You may find an open issue that closely matches what you are thinking. You
+  may also find a closed issue with discussion that either solves your problem
+  or explains why we are unlikely to solve it in the near future.
+
+  If you find a matching issue that is open, and marked `accepted` and/or `help
+  wanted`, you might want to [open a PR](#prs).
+
+### Open an issue.
+
+  Let's discuss your issue. Could be as simple as unclear documentation or a
+  wonky config file.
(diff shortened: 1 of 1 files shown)


kentcdodds:
You might checkout [makeapullrequest.com](http://makeapullrequest.com) and consider including that.

See also: https://github.com/nayafia/contributing-template

:+1:

jfmengels:
I think it's really good, especially for a first draft. I especially like the tone of it.

It's a big block of text, but it does not feel like a huge todo list (not: I did read it as a reviewer, not a contributor) [this one](https://github.com/kentcdodds/eslint-find-new-rules/blob/master/CONTRIBUTING.md) (which is very good too, especially with the really friendly tone, but there is a huge list of items in the todo list. Even though some are not real todos, but they feel like it).

I think that it might be a good idea to have the checklist at the top of the README though, maybe with a `tl;dr` comment, and putting an extra emphasis (putting it in bold or before the list) on `Remember, you don't need to do it all yourself; any of these are helpful!`. Especially helpful to people who are finishing their PR and want to re-check the todo list.

I'd put that at the top so that readers can quickly see what they need to do contribute. Having a big block of text can scare the reader off, and this could set their expectations quickly.

That said, great tone, great work :)

cc @ta2edchimp who might be interested in this too ;)

ta2edchimp:
The `CONTRIBUTING.md` is really good! Especially including a _Code of Conduct_ is a great idea!

@jfmengels thanks for letting me know of this!

jfmengels:
This still looks very good to me, shall we merge this?

PR #261, opened 2016-04-24
Add `no-nodejs-modules` rule (fixes #158)

### LINKED ISSUE #158: Rule proposal: no-core
I have seen client app code import Node.js modules, e.g.

```js
import querystring from 'querystring';
```

This will potentially import a huge library that will vary depending on the node version used to build the bundle. Bundle should not depend on the version of node used to build it.

This is hard to catch because it will just work until it suddenly stops with no way to traceback the issue in the code itself.

The rule should restrict `import` to only modules declared in `package.json` (or another strategy used to make sure that Node.js modules are not imported).

jfmengels:
Add `no-nodejs-modules` rule (fixes #158)

(I've taken the liberty of naming it this way, as I find it to be more explicit than the proposed `no-core`)


benmosher:
Sweet, looks great! Thanks!

Rule rename makes sense, allows for future no-[nodelikeruntime]-modules rules.

jfmengels:
That was fast! :D
And you're welcome ;)

PR #264, opened 2016-04-25
allow destructuring without init (fixes #249)

### LINKED ISSUE #249: no-named-as-default-member throws an error for destructuring in for-of statements
Given the following ESLint config:

```json
{
  "parser": "babel-eslint",

  "env": {
    "es6": true
  },

  "plugins": [
    "import"
  ],

  "rules": {
    "import/no-named-as-default-member": "error"
  }
}
```

and the following code:

```js
for (const {foo, bar} of baz) {}
```

the following error is thrown:

```
Cannot read property 'type' of null
TypeError: Cannot read property 'type' of null
    at EventEmitter.handleDestructuringAssignment (/Users/chris/Projects/foo/node_modules/eslint-plugin-import/lib/rules/no-named-as-default-member.js:64:70)
    at emitOne (events.js:90:13)
    at EventEmitter.emit (events.js:182:7)
    at NodeEventGenerator.enterNode (/Users/chris/Projects/foo/node_modules/eslint/lib/util/node-event-generator.js:42:22)
    at CodePathAnalyzer.enterNode (/Users/chris/Projects/foo/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:609:23)
    at CommentEventGenerator.enterNode (/Users/chris/Projects/foo/node_modules/eslint/lib/util/comment-event-generator.js:99:23)
    at Controller.traverser.traverse.enter (/Users/chris/Projects/foo/node_modules/eslint/lib/eslint.js:887:36)
    at Controller.__execute (/Users/chris/Projects/foo/node_modules/estraverse/estraverse.js:397:31)
    at Controller.traverse (/Users/chris/Projects/foo/node_modules/estraverse/estraverse.js:501:28)
    at Controller.Traverser.controller.traverse (/Users/chris/Projects/foo/node_modules/eslint/lib/util/traverser.js:38:33)
```

The problem appears to here: https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-named-as-default-member.js#L53, where it's assumed that every declarator has an `init` (`for-of` declarators have `init = null`).

Thanks for the awesome ESLint plugin!

benmosher:


src/rules/no-named-as-default-member.js
```diff
@@ -49,6 +49,8 @@ module.exports = function(context) {
   }
 
   function handleDestructuringAssignment(node) {
+    if (!node.init) return
+
```
jfmengels: You could do

```js
if (!node.init ||
    node.id.type !== 'ObjectPattern' ||
    node.init.type !== 'Identifier') {
  return;
}
```
and remove `isDestructure` below.

benmosher: True. I probably should have put the node.init check after the ObjectPattern test anyway; or something like (node.init && node.init.type === ...) for clarity. 

I generally like to do lots of simple predicates vs. mega-predicates, thus the quick early exit. I'm cool either way, though, feel free to reorder if you'd like.

jfmengels: Nah, I'm fine with it. Style-wise I prefer mine, but clarity-wise yours has an explicit name to the condition, so both are good IMO.

jfmengels:
LGTM :)

src/rules/no-named-as-default-member.js
jfmengels: > I generally like to do lots of simple predicates vs. mega-predicates

When there are too many conditions, I completely agree. Another way that I like, is to create functions and call them in one condition.

```js
function isDestructure(node) {
  return node.id.type === 'ObjectPattern' && node.init.type === 'Identifier';
}

if (!node.init || !isDestructure(node)) {
  return;
}
```

PR #269, opened 2016-04-25
named: should ignore ignored modules for re-exported names

### LINKED ISSUE #251: False positive from import/named when re-exporting.
Given file `lodash.js` (showing 2 different ways to export):
```
import clone from 'lodash/fp/clone';
import filter from 'lodash/fp/filter';
export { default as flow } from 'lodash/fp/flow';
export { default as map } from 'lodash/fp/map';
export { default as sortBy } from 'lodash/fp/sortBy';

export { clone, filter };
```

and a usage in another file like this:
```
import { clone, filter, flow, map, sortBy } from './lodash.js';
```

eslint with `import/named` enabled with show false positives for `flow`, `map`, and `sortBy`.
```
  2:25  error  flow not found in '../lodash'    import/named
  2:31  error  map not found in '../lodash'     import/named
  2:36  error  sortBy not found in '../lodash'  import/named
```

The code executes fine, but the rule isn't detecting that syntax of export as valid.

benmosher:
fixes #251: will assume you know what you're doing when you re-export from a non-ES-module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`imports-first`] now allows directives (i.e. `'use strict'`) strictly before
   any imports ([#256], thanks [@lemonmade])
 
+### Fixed
+- [`named`] now properly ignores the source module if a name is re-exported from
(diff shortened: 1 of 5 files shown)


benmosher:
ugh, appveyor. as soon as this goes green, I'll probably publish 1.6.0.

Issue #272, opened 2016-04-25
Node.js v6: Can't pass null to path module functions


strawbrary:
In Node.js v6, all `path` functions will now throw an error if they receive a non-string path argument. `eslint-import-resolver-node` returns `path: null` for core modules which causes [`fileExistsWithCaseSync`](https://github.com/benmosher/eslint-plugin-import/blob/ea1b0e3d2d0c58287d8513a71e79100b48feed38/src/core/resolve.js#L29) to throw an error.

I wasn't sure if it would be better to make the change in the resolver or the plugin itself so I'm just opening an issue about this for now.

benmosher:
Oh, wild. Good to know.

We probably ought to go ahead and open a PR with v6 in the Travis config and start getting ready! Thanks!

benmosher:
Also, for future reference: the `null` path from the resolver is part of the spec. The plugin should exit early and not pass it to the path functions.

strawbrary:
Got it, thanks for the info. v6 isn't out officially yet so I'm not sure if there's a way to get Travis to build with it, but they're aiming to release it tomorrow I believe. In the meantime the rc builds are here: https://nodejs.org/download/rc/

benmosher:
fixed in #273

PR #273, opened 2016-04-25
Support Node v6

benmosher:
Track progress against this PR. 

Known issues:

- #272: don't pass null to path functions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,8 @@ node_js:
   - 0.10
   - 0.12
   - 4
-  - stable
+  - 5
+  - 6
 
 install:
   - npm -g install npm@3
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ node_js:
   - 5
   - 6
 
+os:
+  - linux
+  - osx
+
 install:
   - npm -g install npm@3
   - npm install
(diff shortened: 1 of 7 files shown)


strawbrary:
v6 is on Travis now so you should be good to retry.

I haven't encountered any issues on v6 besides the null path one.

jfmengels:
I re-ran the tests, and they're good to go :)

benmosher:
Interesting, if no tests are failing, that means there is no test that covers #272? Which seems weird, I would think at least one test would try to resolve a builtin.

So feels like we need at least one test for that breaks a la #272 before merging? (after fixing it, too)

strawbrary:
Hmm that is strange. On my local box I get 36 failures when running on v6.0.0. Could someone else try it and see if they get the same?

jfmengels:
It passes all tests on my machine (removed node_modules, `npm i`ed and `npm test`) :confused:

benmosher:
@strawbrary are you on Windows? or OSX?
@jfmengels: same question?

I just added v6 to the Appveyor config to see how Windows takes it.

jfmengels:
Linux Ubuntu here.

strawbrary:
OSX El Capitan

strawbrary:
It seems like `fileExistsWithCaseSync` is never called with a `null` `filepath` under Linux, while it is under OSX and Windows. Any ideas about that?

benmosher:
Yep, that makes sense. I'm on OSX but haven't tried to run locally yet.

Salakar:
Tested on OSX + node v6, now working, also fixes an issue with Webstorm IDE being unable to parse the `<checkstyle/>` XML - for those finding this error randomly lurking up on them 😆

Issue #259, opened 2016-04-22
remove `absolute-first` option from `imports-first`


benmosher:
It's not documented, and it is very much superseded by the `order` rule.

jfmengels:
As pointed out here https://github.com/benmosher/eslint-plugin-import/pull/277#issuecomment-215059238, it is actually documented.

benmosher:
Yep, I don't know what I was thinking.

benmosher:
This actually feels like a breaking change for breaking's sake. This option is a little lower-intensity than full-on `order`, so I think I'll just make a note in the README and keep it.

PR #277, opened 2016-04-26
Remove `absolute-first` option from `imports-first`

singles:
See #259

### DIFF
--- a/docs/rules/imports-first.md
+++ b/docs/rules/imports-first.md
@@ -14,16 +14,6 @@ initWith(foo)
 import bar from './bar' // <- reported
 ```
 
-Providing `absolute-first` as an option will report any absolute imports (i.e.
-packages) that come after any relative imports:
-
-```js
-import foo from 'foo'
-import bar from './bar'
-
-import * as _ from 'lodash' // <- reported
-```
-
 Notably, `import`s are hoisted, which means the imported modules will be evaluated
 before any of the statements interspersed between them. Keeping all `import`s together
 at the top of the file may prevent surprises resulting from this part of the spec.
--- a/src/rules/imports-first.js
+++ b/src/rules/imports-first.js
@@ -8,28 +8,16 @@ module.exports = function (context) {
   return {
     'Program': function (n) {
       const body = n.body
-          , absoluteFirst = context.options[0] === 'absolute-first'
       let nonImportCount = 0
         , anyExpressions = false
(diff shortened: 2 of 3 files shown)


jfmengels:
LGTM, though I'm not sure when we'd want to merge this. There's a growing number of semver-major issues being opened, so if we can do them all in one swoop, that'd be nice.
Thanks a lot anyway @singles! :D

singles:
Sure, no problem, it's BC break anyway :) 

One question though - do you think I should update docs and in place of `absolute-first` add something like:

> `absolute-first` option is not available anymore - you can get the same results using `order` rule

jfmengels:
I could see a deprecated section from here to v2, but I think I would not keep it. Its useful, but I don't like having relics of the past in the readme/docs. There's a changelog for that.
I'm not totally against though, just a small personal preference.

singles:
You're right, good idea with one `deprecated` section :)

jfmengels:
That'd have to be a new PR then, and we'd put this one on hold until v2 is ready I'd say.

benmosher:
👍 to all points. 

I'm tempted to do it pre-v2 since it was undocumented anyway, but no telling how widely it's in use so probably better to wait. 

Plus it doesn't hurt to have it, short term. This PR serves as a sufficient deprecation notice, I think.

jfmengels:
> no telling how widely it's in use so probably better to wait.

I'd say let's stick with proper semver just in case. And maybe in the meantime we'll found out people don't want to use `order` for some reason and would like to stick with this option.

> This PR serves as a sufficient deprecation notice, I think.

I'd add a `deprecated` section in the documentation of the rule, and add it to the changelog. We can't expect users to find out about it in the GH issues.

singles:
> I'm tempted to do it pre-v2 since it was undocumented anyway

Just a note: Actually, it was https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/imports-first.md:

> Providing `absolute-first` as an option will report any absolute imports (i.e. packages) that come after any relative imports:

benmosher:
Weird, not sure how I hallucinated that it wasn't in there. Definitely should save for v2, then.

sindresorhus:
> I'd say let's stick with proper semver just in case.

Strictly speaking, semver only applies to the public *documented* API, but I see no harm either leaving it for now.

PR #278, opened 2016-04-26
compile webpack.configs

jquense:
from the excellent work by @gausie just trying to get it across the finish line.


benmosher:
published eslint-import-resolver-webpack@0.2.2!

PR #280, opened 2016-04-27
Enable `quotes` rule for eslint

singles:
I missed those quotes while doing other PR: https://github.com/benmosher/eslint-plugin-import/pull/245#discussion_r61135535 :)

Now it should be harder to miss that :)

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -13,3 +13,7 @@ rules:
   comma-dangle: [2, always-multiline]
   eqeqeq: [2, "allow-null"]
   no-shadow: 1
+  quotes:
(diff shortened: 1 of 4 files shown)


jfmengels:
LGTM :)
@benmosher is this something you didn't want to enforce by any chance?

benmosher:
I just recently turned it off, actually. This is one of those rules that (from my perspective) is as controversial as it is arbitrary. 

Which is to say: fine with me to turn it back on. 😁

jfmengels:
Let's merge this then?

benmosher:
👍

PR #289, opened 2016-04-28
Move 'interpret' to webpack loader's package.json.

### LINKED ISSUE #286: eslint-import-resolver-webpack missing 'interpret' as dependency in npm
In my .eslintrc, my settings are
```
"settings": {
    "import/resolver": "webpack"
  }
```
and webpack.config.js lives in the same folder as package.json.

When upgrading from 0.2.1 to 0.2.2 of the webpack resolver, I am getting these:

```
new 'Resolve error: unable to load resolver: "webpack"'
```
Which I haven't seen before. When I downgrade back to 0.2.1, they go away.

jonboiser:
Users downloading the package from npm might have had problems since it wasn't downloading this required dependency.

Fixes #286

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "es6-set": "^0.1.4",
     "es6-symbol": "*",
     "eslint-import-resolver-node": "^0.2.0",
-    "interpret": "^1.0.0",
     "lodash.cond": "^4.3.0",
     "lodash.endswith": "^4.0.1",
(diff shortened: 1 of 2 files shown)


benmosher:
Oh man, makes a ton of sense. Explains why CI didn't catch it. 

Thanks!

benmosher:
Published as v0.2.3. Thanks for your help, couldn't have fixed this so fast without it!

PR #291, opened 2016-04-28
added syntax battery to test utils for future use

### LINKED ISSUE #281: no-named-as-default-member -TypeError: Cannot read property 'name' of undefined
no-named-as-default-member.js:85:42

Unsure what source makes it crash... Narrowed it down to:

```js
function someFunction() {
  switch (someVariable) {
    case 'something': {
      const { // The error is reported at the { on this line
        ...rest
      } = someGlobal;
    }
  }
}
```

benmosher:
+ fixes #281


SimenB:
I just thought to quickly point out that you don't need babel-eslint for object spread.

http://eslint.org/docs/user-guide/configuring#specifying-parser-options (experimentalObjectRestSpread)

benmosher:
good point, probably a good idea to do both

PR #293, opened 2016-04-29
added null check on the `fullPath` to close #292

### LINKED ISSUE #292: resolve: Upgrade to Node 6 tries to resolve `path`
Not sure why, but when I upgrade to node 6, it appears that the `namespace` plugin is trying to check whether the `path` file exists for my `import path from 'path'`.

I've put a bunch of console logs in my node modules and I get to [this line](https://github.com/benmosher/eslint-plugin-import/blob/e69561c027e2e2906e8c6cedfbef94837ad0067b/src/core/resolve.js#L35) where `filepath` is `null`.

I've dug in further and the `eslint-import-resolver` name is `node`.

Looks like when the resolver finds a core module, [it returns](https://github.com/benmosher/eslint-plugin-import/blob/e69561c027e2e2906e8c6cedfbef94837ad0067b/resolvers/node/index.js#L8): `{ found: true, path: null }` and we should probably do a null check [here](https://github.com/benmosher/eslint-plugin-import/blob/e69561c027e2e2906e8c6cedfbef94837ad0067b/src/core/resolve.js#L114) for that.

By adding that null check in my `node_modules`, things work great for me locally. I'll file a pull request in a moment to discuss fixing this :+1:

kentcdodds:
This is a PR to fix #292

### DIFF
--- a/src/core/resolve.js
+++ b/src/core/resolve.js
@@ -111,7 +111,7 @@ export function relative(modulePath, sourceFile, settings) {
     let { path: fullPath, found } = withResolver(resolver, config)
 
     // resolvers imply file existence, this double-check just ensures the case matches
-    if (found && !fileExistsWithCaseSync(fullPath, cacheSettings)) {
+    if (fullPath !== null && found && !fileExistsWithCaseSync(fullPath, cacheSettings)) {
       // reject resolved path
       fullPath = undefined
     }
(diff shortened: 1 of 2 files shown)


kentcdodds:
Ran the tests locally with node 6 and everything passes with flying colors :-)

kentcdodds:
I just added a test. I'm not sure it's doing what we WANT it to do with `null`, but at least we can see what it's actually doing and ensure that wont change :-)

Issue #271, opened 2016-04-25
Better handling of non-existent node_modules directory


sindresorhus:
I ran this plugin in a project without a `node_modules` directory. I hadn't yet run `npm install`.

This is what I got:

```
test.js:2
✖  2:18  Unexpected use of file extension "" for "ava"         import/extensions
✖  2:18  Unable to resolve path to module 'ava'.               import/no-unresolved
✖  3:23  Unexpected use of file extension "" for "read-chunk"  import/extensions
✖  3:23  Unable to resolve path to module 'read-chunk'.        import/no-unresolved
```

I don't think the `import/extensions` rule should report in this case. It's just noise.

benmosher:
Ah, yeah, seems like package refs (even internal ones) should should not trigger `extensions` (`always` mode).

@vvo: did you imagine the `always` mode would disallow package references? I would think that ought to be a separate rule, or at least an option.

i.e.

```js
import foo from './foo.js' // valid
import React from 'react' // valid if points to a package and not `react.js`

import bar from './bar' // invalid if filename is actually `bar.js`
```

vvo:
Dunno about the "always mode" but your example is valid and expected.

vvo:
But

```js
import bar from './bar' // invalid if filename is actually `bar.js`
```

Actually if there's a folder named "bar" I would want to disable requiring it. Requiring folders involves magic and hides the filetree from the code.

benmosher:
Fair enough. So: a flag (allow-packages) that allows a folder reference?

benmosher:
Actually I just realized that the main issue may be that the extensions rule needs to ignore unresolved imports. 

This is a standard behavior that would resolve @sindresorhus's original issue regardless of what mode he put it in.

benmosher:
fixed by #296

PR #296, opened 2016-05-02
Unresolved extensions

### LINKED ISSUE #295: import/extensions reporting empty extension when target file does not exist
> import/extensions Unexpected use of file extension "" for "./../errors/scrape_error"

Subject code:

```js
import ScrapeError from './../errors/scrape_error';
```

Configuration:

```json
"import/extensions": [
    2,
    "never"
],
```

Version:

> eslint-plugin-import@1.6.1

benmosher:
Fixes #295 and #271. 

Actually made the rule work report on violations in spite of file resolution; it will use the source path if the file doesn't exist. I think this is probably ideal, in contrast to other rules that ignore unresolved files (in spite of my previous comments).

I also added an early exit for builtins (i.e. `path`) regardless of the options, built on @jfmengels' `importType` infrastructure. I figure even with `always` mode enabled, it should be possible to import from core modules.

Also, note that `extension` is `""` when the provided path is `null` or has no extension, thus the added `!extension` check for the extension-enforcing branch.

Feedback from @vvo, @lo1tuma, @sindresorhus, @jfmengels appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`newline-after-import`], new rule. ([#245], thanks [@singles])
 - Added an `optionalDependencies` option to [`no-extraneous-dependencies`] to allow/forbid optional dependencies ([#266], thanks [@jfmengels]).
 
+### Fixed
+- [`extensions`]: fallback to source path for extension enforcement if imported
+  module is not resolved. Also, never report for builtins (i.e. `path`). ([#296])
+
 ## resolvers/webpack/0.2.4 - 2016-04-29
 ### Changed
 - automatically find webpack config with `interpret`-able extensions ([#287], thanks [@taion])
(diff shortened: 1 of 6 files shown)


benmosher:
Seeing no feedback (what, not an exciting PR? 😉) I'll probably merge this tomorrow or Thursday.

jfmengels:
It's not a part of the repo I'm familiar with (like most of it), but I'll try and take a look tonight ;)

lo1tuma:
Thanks for fixing this @benmosher. Code LGTM.

jfmengels:
I'm discovering the rule here, but isn't disallowing `import express from 'express/index.js';` an anti-pattern? You're not supposed to know the internal structure of a package.

I would suggest removing it the extension check for packages. And if you're doing that, then you can have an early exit for external type imports, just like for builtins. I know it's not necessarily in the scope, so this can be discussed later, but wanted to mention it as it might simplify things a bit.

jfmengels:
What is the expected behavior if you do `import thing from "non-package.js"` with option `"always"`?

benmosher:
Re: express/index.js: that is the spec for always mode, at the moment. I also think it would make sense to allow externals, at least in some mode, but I figured we could let someone ask for that. 

Re: non-package.js: if it resolved to a file (mysteriously), and the file was non-package.js, it would not report. If it did not resolve, it also wouldn't report. Oh, I see now... Yeah, I think there is a bug there. If it resolved to some index.js, it might not report, but I think it should according to the spec.

PR #297, opened 2016-05-02
import/extensions setting

### LINKED ISSUE #267: Don't try to read imported JSON files
In Node.js, this is valid and a common pattern:

```js
var pkg = require('./package.json');
```

I got errors when I ran this plugin on [`got`](https://github.com/sindresorhus/got):

```
  test/headers.js:3
  ✖  3:17  Parse errors in imported module '../package.json': Line 2: Unexpected token (2:9)  import/namespace
  ✖  3:17  Parse errors in imported module '../package.json': Line 2: Unexpected token (2:9)  import/no-named-as-default-member
  ✖  3:17  Parse errors in imported module '../package.json': Line 2: Unexpected token (2:9)  import/no-named-as-default
```

benmosher:
Parser whitelist of file extensions. Fixes #267 and obviates a lot of awkward uses for `import/ignore`.

Also is a breaking change. I assume a decent chunk of folks are out there using `.jsx` and those references would go silently un-analyzed.

I'd like to create a `no-ignored-extensions` rule to go with this, partially as a migration tool. Wouldn't expect folks to keep it on, necessarily, but would flag references to files that were previously being parsed.

Tasks:
- [x] Implement `import/extensions` and fix #267
- [x] ~~Implement `no-ignored-extensions` rule~~ deferred: #306


benmosher:
Actually: could defer the `semver-major` designation in the short term by making the default do no extension validation (i.e. no default) and set the default to `['js']` in v2.0.

@sindresorhus can you set settings in XO? would that resolve your issue?

sindresorhus:
How does the `settings` thing work? Is it something ESLint provides or is it custom to this plugin?

benmosher:
It'd just be a normal [shared ESLint setting](http://eslint.org/docs/user-guide/configuring#adding-shared-settings) spec'd in an `.eslintrc`, i.e.

```yaml
---
# shown for context. I assume you configure this somewhere
plugins:
  - xo
  - import
  # etc.

# so it would need this too, until the default is set in v2 of import plugin:
settings:
  import/extensions: [ .js ]
```

I'm not sure exactly how XO wraps ESLint, so it's not obvious to me where you're setting up plugins.

sindresorhus:
@benmosher Ok, I didn't know about that ability. Should work for us as we just supply ESLint with a config file: https://github.com/sindresorhus/xo/blob/cc49ed4ec069bcd0885ac54513bc76b964bcdcd3/config/plugins.js

benmosher:
Actually, this might be a better place for it: https://github.com/sindresorhus/eslint-config-xo/blob/master/index.js

Looks like you're leveraging this for the XO executable, right? or is this just a mirror? Either way, probably ought to be there for consistency (and optimally _only_ there for DRYness).

Either way, I'll make the change to have no default for now and open a new issue to default to `['.js']` in v2, so this can go out with `1.8`.

sindresorhus:
@benmosher That's just an XO implementation detail. I don't put plugin stuff in the pure ESLint config, since you're then forcing the user to manually install each plugin. That's why it's in the XO binary instead, as it can just be dependencies there.

But yeah. Sounds good to me.

benmosher:
Ah, sure, that makes sense.

FWIW, the `settings` don't require the plugin to be installed, it's just a map that is shared by all rules (including the baked-in ESLint ones).

I just prefix mine with `import/` to avoid conflicts. So you could put it wherever, though it probably does make sense to colocate it with the plugin reference.

LycanHalcyon:
I hope someone can help with this, it is driving me crazy.  Atom / Eslint will not recognize my ./src/components/header.js import coding..  Attached is a photo of the sim error and such.  

I am very new to this and I have tried to fix it by making a new "rule" but I am not sure if I am doing it right.   Could someone please help me with this walk through?  

What rule do I have to add?

screen shot 2016-10-30 at 10 59 23 am

benmosher:
@LycanHalcyon that should be 
```js
import Header from './src/components/header.js'
```
the syntax error is because you're missing the quotes around the file path.

Issue #282, opened 2016-04-27
Proposal: `new-line-between-groups` option for `order`


singles:
This follow up on discussions we had [here](https://github.com/benmosher/eslint-plugin-import/pull/245#issuecomment-211903386) and #246.

Options for that would be either `always` or `never`. 

Then for `always`  and groups like: `groups: [ [ "builtin", "external" ], [ "index", "sibling", "parent" ] ] }` 

So the following would be considered as a problem:

```js
import url from 'url';
import express from 'express'; // <-- missing new line after this statement
import foo from './foo';
```

This would not be considered as a problem:

```js
import url from 'url';
import express from 'express';

import foo from './foo';
```

`new-line-between-groups`  would allow you to configure new lines inside groups like you want - should be pretty flexible. 

I don't think there should be another rule for achieving that, option for `order` IMHO would be fine.

**NOTE**: Option name could of course be changed to something different, `separate-groups(-with-newline)` maybe.

benmosher:
I will mostly defer to @jfmengels and company on this one, but a few gut reactions: 

I think if you don't set the option at all, it shouldn't enforce any spacing.

Also, if you _do_ set it, to either setting, would it enforce that there are no newlines _within_ a group?

i.e.

```js
import url from 'url';
// assuming this were a blank line, would it be reported? "extra newline inside group" or some such
import express from 'express';
```

Also I think it would be optimal if the option were one or two words, something like `newlines`. Clarity over brevity, but brevity would be ideal.

jfmengels:
> I think if you don't set the option at all, it shouldn't enforce any spacing.

:+1:

>  if you do set it, to either setting, would it enforce that there are no newlines within a group?

I think so yes.

Maybe we could specify the number of blank lines between groups?
If I want no blank lines, I can set 0, if I want an empty line, I can set 1, and if I like bigger spacing, I can set 2. If I don't specify it or set it to false (?), nothing will be enforced.

> if the option were one or two words, something like newlines

I think `group-groups` contains the wanted functionality, but it's a bit... repetitive ^^'

singles:
> I think if you don't set the option at all, it shouldn't enforce any spacing.

Yep, then it's do `whatever you want mode` :)

> if you do set it, to either setting, would it enforce that there are no newlines within a group?

Yes, this is what I had in mind.

> Maybe we could specify the number of blank lines between groups?

Personally I don't think it's necessary, especially at the beginning. We might add it later :)

[...]

PR #298, opened 2016-05-02
Add `newlines-between` option to `order` rule

singles:
Checks existence of new lines between and inside import groups.

Implements #282.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Added
 - [`newline-after-import`], new rule. ([#245], thanks [@singles])
 - Added an `optionalDependencies` option to [`no-extraneous-dependencies`] to allow/forbid optional dependencies ([#266], thanks [@jfmengels]).
+- Added `newlines-between` option to [`order`] rule ([#298], thanks [@singles])
 
 ## resolvers/webpack/0.2.4 - 2016-04-29
 ### Changed
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -73,7 +73,9 @@ var path = require('path');
 
 This rule supports the following options:
 
-`groups`: How groups are defined, and the order to respect. `groups` must be an array of `string` or [`string`]. The only allowed `string`s are: `"builtin"`, `"external"`, `"internal"`, `"parent"`, `"sibling"`, `"index"`. The enforced order is the same as the order of each element in a group. Omitted types are implicitly grouped together as the last element. Example:
+### `groups: [array]`:
+
+How groups are defined, and the order to respect. `groups` must be an array of `string` or [`string`]. The only allowed `string`s are: `"builtin"`, `"external"`, `"internal"`, `"parent"`, `"sibling"`, `"index"`. The enforced order is the same as the order of each element in a group. Omitted types are implicitly grouped together as the last element. Example:
 ```js
 [
   'builtin', // Built-in types are first
@@ -89,3 +91,53 @@ You can set the options like this:
 ```js
 "import/order": ["error", {"groups": ["index", "sibling", "parent", "internal", "external", "builtin"]}]
 ```
+
+### `newlines-between: [always|never]`:
+
+
+Enforces or forbids new lines between import groups:
+
+- If omitted, assertion messages will be neither enforced nor forbidden.
+- If set to `always`, a new line between each group will be enforced, and new lines inside a group will be forbidden.
+- If set to `never`, no new lines are allowed in the entire import section.
+
+With the default group setting, the following will be invalid:
+
+```js
+/* eslint import/order: ["error", {"newlines-between": "always"}] */
+import fs from 'fs';
+import path from 'path';
+import index from './';
+import sibling from './foo';
+```
+
+```js
+/* eslint import/order: ["error", {"newlines-between": "never"}] */
+import fs from 'fs';
+import path from 'path';
+
+import index from './';
+
+import sibling from './foo';
+```
+
+while those will be valid:
+
+```js
+/* eslint import/order: ["error", {"newlines-between": "always"}] */
+import fs from 'fs';
+import path from 'path';
+
+import index from './';
+
+import sibling from './foo';
+```
+
+```js
+/* eslint import/order: ["error", {"newlines-between": "never"}] */
+import fs from 'fs';
+import path from 'path';
+import index from './';
+import sibling from './foo';
+```
+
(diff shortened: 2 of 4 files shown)


benmosher:
LGTM, but @jfmengels should have final say, it's his baby. 😊

benmosher:
Oh, actually, it would be great if you could add a note to the change log. `### Added` section makes sense to me.

src/rules/order.js
```diff
@@ -109,6 +109,37 @@ function convertGroupsToRanks(groups) {
...
+  imported.slice(1).forEach(function(currentImport) {
+    if (newlinesBetweenImports === true) {
+      if (currentImport.rank !== previousImport.rank
+        && getLineDifference(currentImport, previousImport) !== 2)
+      {
+        context.report(
+          previousImport.node, 'There should be one empty line between import groups'
+        )
+      } else if (currentImport.rank === previousImport.rank
+        && getLineDifference(currentImport, previousImport) === 2)
```
jfmengels: `=== 2` --> `>= 2`

docs/rules/order.md
```diff
@@ -89,3 +91,56 @@ You can set the options like this:
 ```js
 "import/order": ["error", {"groups": ["index", "sibling", "parent", "internal", "external", "builtin"]}]
 ```
+
+### `newlines-between: [bool]`:
+
+Checks existence of new lines between and inside import groups. For sake of examples let's assume `groups` settings like:
```
jfmengels: I'm just proposing a new rewording here:
```
Enforces or forbids new lines between import groups.
If omitted, assertion messages will be neither enforced nor forbidden.
If set to true, a new line between each group will be enforced, and new lines inside a group will be forbidden.
If set to false, no new lines are allowed in the entire import section.
```

Btw, starting to think that having a string is more explicit (readable in the config), and more ESLint-like. `newlines-between: "always|never".

docs/rules/order.md
```diff
@@ -89,3 +91,56 @@ You can set the options like this:
...
+groups: [
+	['builtin', 'index'],
+	['sibling'],
+	['parent', 'external'],
+],
+```
+
+**a)** Not setting `newlines-between` option will cause rule not to check any empty lines validity.
+
+**b)** With `newlines-between: true` this will be ok:
```
jfmengels: Do not set a new group setting for the example, let's keep it simple and keep the default setting.
Instead of creating a section for every possibility, set and show it in the example, like

```js
/* eslint import/order: ["error", {"newlines-between": true}] */
var fs = require('fs');
```

Maybe keep examples short? With like 2 groups and 3 imports (suggestion, do what you think renders best here)
And if you need to have comments, add a space after `// ` for consistency :)

jfmengels:
This looks pretty good @singles, thanks a lot for this!
Just some comments on the documentation mostly.

> it's his baby

Man, you have no idea... I still haven't had the heart to deprecate `eslint-plugin-import-order`. It's my most popular repo, and growing in downloads every day... Will do it soon... soon...

benmosher:
I can imagine. But every download here is your download too, now! 😁 I think GitHub itself may be responsible for the recent uptick in downloads, so that's pretty cool.

jfmengels:
I think https://github.com/dustinspecker/awesome-eslint is partially responsible too. Even more so probably.

singles:
OK, new changes:

1. Added info the changelog
2. Fixed https://github.com/benmosher/eslint-plugin-import/pull/298#discussion_r61781477
3. Changed option values from `true|false` to `always|never`
4. Updated docs, less words, more concise

Thx @benmosher @jfmengels for feedback!

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Added
 - [`newline-after-import`], new rule. ([#245], thanks [@singles])
 - Added an `optionalDependencies` option to [`no-extraneous-dependencies`] to allow/forbid optional dependencies ([#266], thanks [@jfmengels]).
+- Added `newlines-between` option to `order` rule ([#298], thanks [@singles])
```
jfmengels: Could you put brackts around `order` so that there is a link there?

docs/rules/order.md
```diff
@@ -89,3 +91,48 @@ You can set the options like this:
...
+### `newlines-between: [always|never]`:
+
+
+Enforces or forbids new lines between import groups:
+
+- If omitted, assertion messages will be neither enforced nor forbidden.
+- If set to `always`, a new line between each group will be enforced, and new lines inside a group will be forbidden.
+- If set to `never`, no new lines are allowed in the entire import section.
+
+For default group setting those will be valid:
```
jfmengels: With the default group setting, the following will be valid:

docs/rules/order.md
```diff
@@ -89,3 +91,48 @@ You can set the options like this:
...
+```
+
+```js
+/* eslint import/order: ["error", {"newlines-between": "never"}] */
+var fs = require('fs');
+var index = require('./');
+var sibling = require('./foo');
+```
+
+while those will be invalid:
```
jfmengels: `while the following will be invalid:`

Actually, if you could swap both sections, the invalid example is always (afaik) put before the valid section.

docs/rules/order.md
```diff
@@ -89,3 +91,48 @@ You can set the options like this:
...
+- If set to `always`, a new line between each group will be enforced, and new lines inside a group will be forbidden.
+- If set to `never`, no new lines are allowed in the entire import section.
+
+For default group setting those will be valid:
+
+```js
+/* eslint import/order: ["error", {"newlines-between": "always"}] */
+var fs = require('fs');
+
+var index = require('./');
```
jfmengels: Could you add
```js
var path = require('path');
```
otherwise it looks like you simply enforce a line between each import.

And since most examples in the repo us the import syntax, might be nice to use that instead?

jfmengels:
Sorry for the nitpicking, looks great otherwise @singles!

CHANGELOG.md
singles: Fixed.

docs/rules/order.md
singles: Fixed.

docs/rules/order.md
singles: Sure, good idea, done.

docs/rules/order.md
singles: No problem, done.

singles:
> Sorry for the nitpicking

No problem :)

jfmengels:
Looks pretty great to me :D

PR #307, opened 2016-05-05
Update readme with `no-mutable-exports`

jfmengels:
Add an entry for `no-mutable-exports` in the readme.


benmosher:
Ah, good catch!

PR #308, opened 2016-05-05
Add `prefer-default-export` rule (fixes #304)

### LINKED ISSUE #304: suggestion to add rule prefer-default-export
I was wondering if you would like me to add a PR with this rule [prefer-default-export](https://github.com/applicaster/eslint-plugin-apprime/blob/master/docs/rules/prefer-default-export.md) as it seams related to this repo.

P.S.
I want to give credit to @ilyavolodin for suggeting this in the eslint Gitter.

gavriguy:
Add `prefer-default-export` rule

When there is only a single export from a module prefer using default export over named export.

### DIFF
--- a/docs/rules/prefer-default-export.md
+++ b/docs/rules/prefer-default-export.md
@@ -1,0 +1,51 @@
+# prefer-default-export
+
+When there is only a single export from a module prefer using default export over named export.
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```javascript
+// bad.js
+
+// There is only a single module export and its a named export.
+export const foo = 'foo';
+
+```
+
+The following patterns are not warnings:
+
+```javascript
+// good1.js
+
+// There is a default export.
+export const foo = 'foo';
+const bar = 'bar';
+export default 'bar';
+```
+
+```javascript
+// good2.js
+
+// There is more thank one named exports in the module.
+export const foo = 'foo';
+export const bar = 'bar';
+```
+
+```javascript
+// good3.js
+
+// There is more thank one named exports in the module
+const foo = 'foo';
+const bar = 'bar';
+export { foo, bar }
+```
+
+```javascript
+// good4.js
+
+// There is a default export.
+const foo = 'foo';
+export { foo as default }
+```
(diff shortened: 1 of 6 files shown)


docs/rules/prefer-default-export.md
```diff
@@ -0,0 +1,33 @@
...
+export const foo = 'foo';
+const bar = 'bar';
+export default 'bar';
+```
+```javascript
+// good2.js
+
+// There is more thank one named exports in the module.
+export const foo = 'foo';
+export const bar 'bar';
```
jfmengels: missing `=` here

jfmengels:
Thanks a lot @gavriguy, looks good!

Could you add a line about it in the README, and add an entry in the changelog? You can thank yourself there too :)

gavriguy:
fixed. thanks for the warm welcome

benmosher:
Just making a note: would like to support 

- [ ] `export { x, y }` as a pair of named exports
- [ ] `expoxt { z as default }` as a valid default export

I think the first is solved by counting [`ExportSpecifier`s](https://github.com/estree/estree/blob/master/es6.md#exportspecifier) instead of `ExportNamedDeclaration`s. T

benmosher:
@gavriguy let me know if you're up for making those changes in this PR. I can also do it post-merge if you'd rather not.

gavriguy:
great ideas ill work on it. thanks

gavriguy:
I've made the changes.
It was a bit more complicated than your suggestion as `ExportSpecifier` isn't called on all named exports - its only called when the named export is inside curly braces.
To solve this - I had to create two counters: `namedExportCount` & `specifierExportCount`.

Anyways now everything works and all tests are passing.

gavriguy:
one of the build failed because of timeout - I think its a travis issue, but I don't see how to rebuild it again (maybe because I don't have the right permissions at Travis) Can you please have a look?

benmosher:
Changes look good, no worries on the build. 

You could rebase against the tip, if you want, but I will probably tweak the change log a little before merging anyway, so you don't need to worry about it. Thanks for the updates! 😄

gavriguy:
Thanks, I rebased against master - please let me know if you need anything more. thanks

benmosher:
Merged via cf45512a1bc3fc476745269fee54713dc4a4ea71. Thanks!

Issue #265, opened 2016-04-25
Webpack 2 support


vanthiyathevan:
[Resolve options](https://gist.github.com/sokra/27b24881210b56bbaff7#resolving-options) in webpack 2.0 is considerably changed. Any way to support that now?

benmosher:
Need a Webpack 2.0 resolver.

Could be a new package (`eslint-import-resolver-webpack2`) or a new major revision of the existing Webpack resolver, though I'd probably opt for the former as I (and I suspect many others) will likely not make the transition to Webpack 2 anytime soon (if even ever!).

benmosher:
FYI, here's the readme section on resolvers: https://github.com/benmosher/eslint-plugin-import#resolvers

The `jspm` resolver is the only community one I'm aware of so far, as I've been stewarding the Node and Webpack (1) resolvers out of this repo. Each resolver is a package unto itself though, so anyone could write and publish one for Webpack 2. 

I'm not planning to implement this any time soon, but a PR with a `resolvers/webpack2` package would be welcome. 😄

benmosher:
Actually, now that I've read through that section again, I can imagine supporting both at once in the Webpack resolver. Would be worth a shot, anyway, might be simpler than forking or trying to share common code.

vanthiyathevan:
Thanks for the updates.

I beg to defer on your opinion on the adoption of webpack 2. I think webpack 2 will be adopted soon rather than never. Yes the config is a pain but the native support for es6 modules will be worth it.

Looking at the config for webpack 2 and what you have for `resolve-alias.js`, shouldn't it be possible to do something similar. The issue I see is that the a specific module can occur anywhere in the tree. 
Ex:

```js
{
  resolve: {
    extensions: ['', '.jsx', '.scss', '.js', '.json', '.styl', 'ts', 'tsx'],
    modules: [
      'web_modules',
      'node_modules',
      'shared',
      'res'
    ]
  },
}
```

The `shared` folder can be in multiple places. My folder tree follows like this
```
- app
  - master
    - components
    - dashboard
      - components
    - revenue
       - components
    - shared
      - common-components-for-master
  - root
  - services
  - root
  - services
  - shared
    - anything-common-for-app
- config 
- shared
  - components
    - some
    - random
    - shared-components
  - utils
```

In my case revenue can import something like `import AComponent from common-components-for-master` from within `components`, `dashboard` or `revenue`. I hope that helps

benmosher:
[...]

PR #319, opened 2016-05-08
Add support for Webpack 2 renamed modules key

pascalduez:
Hi,

just wanted to start the ball rolling, there most likely will be more work needed to fully support Webpack 2, but that very simple addition is already doing it for me on a project with custom import paths.

Rough changelog: https://gist.github.com/sokra/27b24881210b56bbaff7#resolving-options

If you feel that's okay for a start, I'll update the tests ?

Refs #265

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -129,6 +129,7 @@ exports.resolve = function (source, file, settings) {
 
       // http://webpack.github.io/docs/configuration.html#resolve-modulesdirectories
       moduleDirectory: get(webpackConfig, ['resolve', 'modulesDirectories'])
+        || get(webpackConfig, ['resolve', 'modules']) // Webpack >= 2
         || ['web_modules', 'node_modules'],
 
       paths: paths,


benmosher:
I like the simplicity. I don't have any Webpack 2 projects to check against, so I'll take your word for it. 😎

New tests would be much appreciated. Probably a clone of the existing moduleDirectories tests would be a good start, assuming the semantics match.

benmosher:
Heads up: this will probably not be mergeable after I merge #345. 

I think it will make more sense to make a `webpack-2` resolver, see my notes on #345.

michael-wolfenden:
@benmosher 

In webpack2, [resolve.root](https://webpack.github.io/docs/configuration.html#resolve-root) has been merged into the [modules option](https://gist.github.com/sokra/27b24881210b56bbaff7#resolving-options), however the commit doesn't handle this.

webpack
```js
resolve: {
    root: PATHS.appDir
}
```

webpack2
```js
resolve: {
    modules: [
        'node_modules',
        PATHS.appDir,
    ],
},
```

I had to add the following code to make this work, the assumption being that root resolves are absolute paths

```js
// Webpack >= 2
var rootPaths = get(webpackConfig, ['resolve', 'modules'], []).filter(path.isAbsolute);
paths = paths.concat(rootPaths)
```

gausie:
Now that #345 is merged, how are you feeling about this @benmosher?

benmosher:
I think I'd probably opt for a `webpack2` resolver, mostly because there is a factory to build `resolve` from config that would be nice to leverage for ongoing correctness.

I can maintain it here (another PR, anyone?), or someone in the community can maintain it independently. Either way works for me.

I don't think I would keep this duck-typed version. I'm not sure how it would integrate with the current `webpack` resolver.

pascalduez:
Let's close this and do it the proper way :)

PR #321, opened 2016-05-08
Add TomDoc support to no-deprecated rule

### LINKED ISSUE #302: no-deprecated TomDoc support
I know `no-deprecated` is still a WIP, but I was wondering if you'd consider parsing [TomDoc](http://tomdoc.org/) `Deprecated:` annotations? 

Its basically a line like:

``` js
// Deprecated: Resize an object to the given dimensions.
```

I'd totally be willing to code up a PR, just wanted to check if you'd consider this in scope.

Thanks!

josh:
Closes #302.

### DIFF
--- a/src/core/getExports.js
+++ b/src/core/getExports.js
@@ -97,6 +97,12 @@ export default class ExportMap {
       return m // can't continue
     }
 
+    const docstyle = (context.settings && context.settings['import/docstyle']) || ['jsdoc']
+    const docStyleParsers = {}
+    docstyle.forEach(style => {
+      docStyleParsers[style] = availableDocStyleParsers[style]
+    })
+
     // attempt to collect module doc
     ast.comments.some(c => {
       if (c.type !== 'Block') return false
@@ -143,7 +149,7 @@ export default class ExportMap {
     ast.body.forEach(function (n) {
 
       if (n.type === 'ExportDefaultDeclaration') {
-        const exportMeta = captureDoc(n)
+        const exportMeta = captureDoc(docStyleParsers, n)
         if (n.declaration.type === 'Identifier') {
           addNamespace(exportMeta, n.declaration)
         }
@@ -174,11 +180,12 @@ export default class ExportMap {
             case 'FunctionDeclaration':
             case 'ClassDeclaration':
             case 'TypeAlias': // flowtype with babel-eslint parser
-              m.namespace.set(n.declaration.id.name, captureDoc(n))
+              m.namespace.set(n.declaration.id.name, captureDoc(docStyleParsers, n))
               break
             case 'VariableDeclaration':
               n.declaration.declarations.forEach((d) =>
-                recursivePatternCapture(d.id, id => m.namespace.set(id.name, captureDoc(d, n))))
+                recursivePatternCapture(d.id, id =>
+                  m.namespace.set(id.name, captureDoc(docStyleParsers, d, n))))
               break
           }
         }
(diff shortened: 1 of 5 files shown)


josh:
Re: https://github.com/benmosher/eslint-plugin-import/issues/302#issuecomment-216638459

@benmosher what do you think? Does it feel too *ad-hoc* to have a mini tomdoc parser as part of this repo?

benmosher:
Looks good, in theory. I am a little concerned about spurious flagging, though. I think I'd like to have a "style" parameter that defaults to JSDoc but can allow specification of TomDoc.

josh:
Ah, sorry. I forgot you mentioned that in the other thread. Yeah, I'll make
this modal.

benmosher:
It probably needs to be a `setting`, at least for now, though (not a rule parameter--I misspoke above).

Parsing of imported modules can occur during evaluation of any rule, and only happens once.

Something like `import/docstyle` maybe? (actually, I summon @jfmengels to think of a better setting name? 😅)

jfmengels:
> I summon @jfmengels to think of a better setting name

I always delegate that part to others myself :p, but I like `docstyle` ;)

I think it would be nice if this setting accepted an array of strings, like `XX: ["tomdoc", "jsdoc"]`, in case you have multiple doc styles in your repo (while transitionning?), or simply no style (sometimes it's this, sometimes it's that...).

src/core/getExports.js
```diff
@@ -97,6 +97,14 @@ export default class ExportMap {
       return m // can't continue
     }
 
+    const docstyle = (context.settings && context.settings['import/docstyle']) || { jsdoc: true }
```
josh: Is there a better place to define plugin default `settings` so it gets merged correctly with any user config?

benmosher: Not at the moment, nope. I've been doing it as-hoc as you've done here.

josh:
@benmosher what do you think so far? The integration still feels gross, but I don't think I understand enough of the requirements for building a useful abstraction that will allow other doc parsers to hook in.

benmosher:
I think it looks good.

I'm a little torn on specifying parsers as an object vs. just a list of strings. I can't think of any reason one would explicitly disable one (i.e. `{jsdoc:false}` is meaningless vs. just omitting) but it does allow potentially passing an object of options.

So I think it's probably good as-is, to start. Doesn't feel like it's worth publishing the contents of `captureTomDoc` to npm just to say that you can specify any arbitrary docparser.

So, yeah, a good first step towards general support for alternate metadata formats. 😄

jfmengels:
I agree with @benmosher, I'd also prefer a list of strings, would make the settings a bit shorter and more readabe in the config.

josh:
> I'm a little torn on specifying parsers as an object vs. just a list of strings.

👍 I'll make that change

josh:
Sorry, I think I forgot to comment that I updated the PR with @jfmengels "list of strings" suggestion.

benmosher:
Ah, sweet. LGTM

benmosher:
Actually can you document this new functionality in the rule's documentation and add a blurb to the change log?

benmosher:
Can you document the `import/docstyle` setting as well? Probably just in the rule documentation for now, since nothing else references it.

See the main README for examples of existing setting documentation. I normally like to include a snippet of example `.eslintrc` config (in YAML). You don't need to follow suit, just offering as a reference.

Getting close. Sorry this has been dragging, I've been busy IRL. 😅

josh:
> Can you document the import/docstyle setting as well? Probably just in the rule documentation for now, since nothing else references it.

Good call! I actually forgot it was opt in for a second.

> Getting close. Sorry this has been dragging, I've been busy IRL. 😅

No worries! 😄 You've been giving great feedback.

josh:
Thanks!

PR #326, opened 2016-05-09
Fix newline-after-import rule for require

### LINKED ISSUE #318: The `newline-after-import` rule requires newline after every import
The following:

```js
'use strict';
var path = require('path');
var arrify = require('arrify');
var pkgConf = require('pkg-conf');
var deepAssign = require('deep-assign');
var objectAssign = require('object-assign');
var homeOrTmp = require('home-or-tmp');
var multimatch = require('multimatch');
var resolveFrom = require('resolve-from');
var pathExists = require('path-exists');

```

Results in:

```
  options-manager.js:2:1
  ✖    2:1  Expected empty line after require statement not followed by another require.  import/newline-after-import
  ✖    3:1  Expected empty line after require statement not followed by another require.  import/newline-after-import
  ✖    4:1  Expected empty line after require statement not followed by another require.  import/newline-after-import
  ✖    5:1  Expected empty line after require statement not followed by another require.  import/newline-after-import
  ✖    6:1  Expected empty line after require statement not followed by another require.  import/newline-after-import
  ✖    7:1  Expected empty line after require statement not followed by another require.  import/newline-after-import
  ✖    8:1  Expected empty line after require statement not followed by another require.  import/newline-after-import
  ✖    9:1  Expected empty line after require statement not followed by another require.  import/newline-after-import
```

With versions 1.7.0.

I expected it to only report about missing newlines after the last import.

lo1tuma:
This PR fixes #318.

Actually I’m not convinced that this rule really makes sense for commonjs, there are just too many edge-cases and no way to say which to exclude from this rule, e.g. the following:

```js
require('a');
foo(require('b'), require('c'), require('d'));
var assign = Object.assign || require('object-assign');
const nightwatchConfig = { seleniumPath: require('selenium-standalone').path };
```

I’ve tried my best to handle those cases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "lodash.cond": "^4.3.0",
     "lodash.endswith": "^4.0.1",
     "lodash.find": "^4.3.0",
+    "lodash.findindex": "^4.3.0",
     "object-assign": "^4.0.1",
     "pkg-up": "^1.0.0"
   }
--- a/src/rules/newline-after-import.js
+++ b/src/rules/newline-after-import.js
@@ -4,47 +4,103 @@
  */
 
 import isStaticRequire from '../core/staticRequire'
+import findIndex from 'lodash.findindex'
 
 //------------------------------------------------------------------------------
 // Rule Definition
 //------------------------------------------------------------------------------
 
-function getLineDifference(node, nextToken) {
-  return nextToken.loc.start.line - node.loc.start.line
+function containsNodeOrEqual(outerNode, innerNode) {
+    return outerNode.range[0] <= innerNode.range[0] && outerNode.range[1] >= innerNode.range[1]
 }
 
-function ensureNoForbiddenKeyword(context, node, tokenToInspect, tokenValue) {
-  if (!tokenToInspect) {
-    return
-  }
+function getScopeBody(scope) {
+    const { body } = scope.block
 
-  if (getLineDifference(node, tokenToInspect) === 1
-    && tokenToInspect.type === 'Keyword' && tokenToInspect.value !== tokenValue)
-  {
-    context.report({
-      loc: tokenToInspect.loc.start,
-      message: 'Expected empty line after ' + tokenValue +
-        ' statement not followed by another ' + tokenValue + '.',
-    })
-  }
+    if (body.type === 'BlockStatement') {
+        return body.body
+    }
+
+    return body
 }
+
+function findNodeIndexInScopeBody(scope, nodeToFind) {
+    const body = getScopeBody(scope)
+
+    return findIndex(body, (node) => containsNodeOrEqual(node, nodeToFind))
+}
+
+function getLineDifference(node, nextNode) {
+  return nextNode.loc.start.line - node.loc.end.line
+}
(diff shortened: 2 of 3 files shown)


singles:
You should probably update changelog's `Fixed` section :)

benmosher:
@lo1tuma that's fair. 

My default position is generally to support the simplest thing that is useful (like standard assignment and naked function calls), and put the weird stuff in the `Don't Use This Rule If` section of the docs. Don't have to solve every case. :smile:

jfmengels:
I think we should only support requires that are present at the top level.

```js
// Support
require('a');
var b = require('b');

// Do not support
foo(require('b'), require('c'), require('d'));
var assign = Object.assign || require('object-assign');
const nightwatchConfig = { seleniumPath: require('selenium-standalone').path };
```

lo1tuma:
> I think we should only support requires that are present at the top level.

What about modules wrapped in an IIFE?

--------------

The problem is that not supporting certain use-cases is quite hard, when no support means the rule will just ignore them and never report any problems. For that you must actively detect such patterns and ignore them in the rule otherwise you would end up with problems being reported that don’t make any sense.

jfmengels:
> What about modules wrapped in an IIFE?

Ugh, didn't think of those. For what it's worth, most rules in this plugin don't support `require`, and those that do didn't take those cases into account afaik (the `order` rule for instance doesn't).

benmosher:
> most rules in this plugin don't support `require`

Might be a bit too general. I think the ones that can (i.e. not supporting `import`-specific syntax things), do. (i.e. `no-unresolved`)

I still think that IIFE and other cases would be best supported by a visitor that interacts with `BlockStatement`'s `body` array, which is compatible with the `Program` node.

So you only handle `AssigmentExpression` and `CallExpression` if they are a direct descendant of `Program` or `BlockStatement`.

So you'd also enforce newlines in any function body, i.e. 
```js
function foo() {
  var y = require('whatever') // AssignmentExpression
  require('something-else') // CallExpression
  
  // ^ this newline is enforced
  /* ... code ... */
}
```

Would work inside `if` and anything else that uses a block.

lo1tuma:
> So you only handle AssigmentExpression and CallExpression if they are a direct descendant of Program or BlockStatement.

I thought about the same, but unfortunately an `AssignmentExpression` is never a direct descendant of a `Program` or `BlockStatement` node, it is wrapped either in an `ExpressionStatement` or in `VariableDeclarator` within of an `VariableDeclaration`. Which leads to another edge-case, if you define multiple variable bindings in one var statement, for example:

```js
var a = require('a'),
    b = require('b'),
    c = require('c');
var d = require('d');
```

benmosher:
Ahh, yeah, that makes sense.

So, now that we've all gone on this discussion adventure: anyone opposed to merging this as-is? Seems to be strictly an improvement and fixes #318? I think we're all in agreement, just double-checking.

benmosher:
Seeing no feedback, will merge as is. Thanks, @lo1tuma!

jfmengels:
Thanks @lo1tuma :D

PR #329, opened 2016-05-10
Fix `order`s `newline-between` for multiline imports

### LINKED ISSUE #313: `order`'s `newlines-between` doesn't work with multiline import statements
Settings `newlines-between: always`

Not working examples:

```js
import FurtherStepsHeadline // invalid error report here: There should be no empty line within import group  import/order
   from '../../../../../../app/components/pages/contentFunnel/donePartials/FurtherStepsHeadline';
```

```js
import { expect } from 'chai';

import RangeType, { // invalid error report here: There should be no empty line within import group  import/order
    isFullRange,
    isPartialRange,
} from '../../../../app/lib/dateRange';
```

```js
local: {
	launch_url: 'http://localhost:${APPLICATION_PORT}',
	selenium_host: 'localhost',
	selenium_port: 4444,
	selenium: {
		start_process: true,
		server_path: require('selenium-server-standalone-jar').path, // invalid error report here: There should be no empty line within import group  import/order
		cli_args: {
			'webdriver.chrome.driver': require('chromedriver').path
		}
	},
	desiredCapabilities: {
		browserName: 'chrome'
	}
},
```

I will fix this ASAP.

singles:
Partly-fixes #313.

### DIFF
--- a/tests/src/rules/order.js
+++ b/tests/src/rules/order.js
@@ -206,6 +206,44 @@ ruleTester.run('order', rule, {
         },
       ],
     }),
+    // Option newlines-between: 'never' with multiline imports #1
+    test({
+      code: `
+        import path from 'path';
+
+        import {
+            I,
+            Want,
+            Couple,
+            Imports,
+            Here
+        } from 'bar';
+        import external from 'external'
+      `,
+      options: [{ 'newlines-between': 'always' }]
+    }),
+    // Option newlines-between: 'always' with multiline imports #2
+    test({
+      code: `
+        import path from 'path';
+        import net
+          from 'net';
+
+        import external from 'external'
+      `,
+      options: [{ 'newlines-between': 'always' }]
+    }),
+    // Option newlines-between: 'always' with multiline imports #3
+    test({
+      code: `
+        import foo
+          from '../../../../this/will/be/very/long/path/and/therefore/this/import/has/to/be/in/two/lines';
+
+        import bar
+          from './sibling';
+      `,
+      options: [{ 'newlines-between': 'always' }]
+    }),
   ],
   invalid: [
     // builtin before external module (require)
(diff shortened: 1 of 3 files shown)


singles:
See #313

tests/src/rules/order.js
```diff
@@ -206,6 +206,70 @@ ruleTester.run('order', rule, {
...
+          from '../../../../this/will/be/very/long/path/and/therefore/this/import/has/to/be/in/two/lines';
+        import external from 'external'
+      `,
+      options: [
+        {
+          groups: [
+            ['builtin', 'index'],
+            ['sibling'],
+            ['parent', 'external'],
+          ],
```
jfmengels: Not that it matters all that much, but you could simplify the tests by removing the `groups` option, as you don't need it.

jfmengels:
LGTM.

You say this partially fixes #313, what part of it remains? What I suggested adding tests for in https://github.com/benmosher/eslint-plugin-import/issues/313#issuecomment-217466177?

singles:
@jfmengels Yes, this is pretty tricky, but looks like I'm not far away from finishing that. Also case with `require` in nightwatch config isn't solved yet.

tests/src/rules/order.js
singles: Done.

jfmengels: I meant in all the tests you added and where it's not needed. I think you only removed it from the last one?

jfmengels:
Not sure what you meant with `nightwatch config`..

singles:
@jfmengels That part with requires of `selenium` and `chromedrvier` - it comes from our `nightwatch.js` file. Sorry for not being clear enough.

singles:
I will create new PR which solves all issues mentioned in #313, closing that one.

Issue #328, opened 2016-05-10
Export * and export default


jkimbo:
I get the error `Multiple default exports` when trying to export everything from another file including the default export e.g.

```
export * from './foo';
export { default as default } from './foo';
```

It looks like that this case should throw an error (as shown by this [test case](https://github.com/benmosher/eslint-plugin-import/blob/90fefa3deb0853b4ab286eb07179fba62f4275c6/tests/src/rules/export.js#L32-L35)) however babel compiles the code `export * from './foo';` to:

```
Object.keys(_foo).forEach(function (key) {
  if (key === "default") return;
  Object.defineProperty(exports, key, {
    enumerable: true,
    get: function get() {
      return _foo[key];
    }
  });
});
```
[Reference](https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=export%20*%20from%20'.%2Ffoo'%3B)

so the second export is necessary to ensure the default export also gets exported.

Any suggestions? I'm not sure whether it's actually babel which is at fault here or if the assumption in this plugin is wrong. Happy to contribute a PR if a solution is agreed on.

benmosher:
Interesting. I think the spec dictates that imported/exported namespaces include the `default` name; it's not distinct from other exported names, it just has special extra syntax for exporting/importing.

So I would argue to start that Babel is doing it wrong (maybe in the interest of backwards-compatibility?).

jfmengels:
@jkimbo Are you running ESLint on the transpiled code? If you do, that could explain the error, but I don't see why you would do that.

benmosher:
@jfmengels: if `./foo` contains a default export, it should report `Multiple default exports` on both lines.

@jkimbo: http://www.ecma-international.org/ecma-262/6.0/#sec-getexportednames, 7.d.i. seems to be filtering `default` exports from `export * from ...` statements. As such, looks like this plugin is in error and Babel has it right.

jkimbo:
Thanks for digging into it @benmosher ! I can attempt a PR to fix it (though not until tomorrow) if that would help?

jfmengels:
@jkimbo Yes please, that'd be great :)

benmosher:
You're welcome to take a stab at it, but I think it goes pretty deep into the exported namespace tracking so it may get dicey. You may want to fail fast.

PR #331, opened 2016-05-11
Filter out default export in an ExportAllDeclaration

jkimbo:
Fix for #328 

@benmosher the fix seems suspiciously simple! Are there any potential problems you can see? Any additional tests that would ensure it's working as expected?

### DIFF
--- a/src/rules/export.js
+++ b/src/rules/export.js
@@ -50,7 +50,13 @@ module.exports = function (context) {
         return
       }
       let any = false
-      remoteExports.forEach((v, name) => (any = true) && addNamed(name, node))
+      remoteExports.forEach((v, name) => (
+        (any = true) &&
+        // ignore default export from an export all declaration
+        // Ref: https://github.com/benmosher/eslint-plugin-import/issues/328
+        name !== 'default' &&
+        addNamed(name, node)
+      ))
 
       if (!any) {
         context.report(node.source,
--- a/src/rules/export.js
+++ b/src/rules/export.js
@@ -52,7 +52,7 @@ module.exports = function (context) {
       let any = false
       remoteExports.forEach((v, name) => (
         (any = true) &&
-        // ignore default export from an export all declaration
+        // Ignore default export from an export all declaration
         // Ref: https://github.com/benmosher/eslint-plugin-import/issues/328
         name !== 'default' &&
         addNamed(name, node)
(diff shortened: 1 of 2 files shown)


src/rules/export.js
```diff
@@ -50,7 +50,13 @@ module.exports = function (context) {
         return
       }
       let any = false
-      remoteExports.forEach((v, name) => (any = true) && addNamed(name, node))
+      remoteExports.forEach((v, name) => (
+        (any = true) &&
```
jfmengels: I know it was already in the original code (so this is more like a question to @benmosher), but wouldn't it be simpler & less odd (an assignation in a condition? oô), to do `const any = remoteExports.length > 0`?

I would actually split it up somewhat like this
```js
const namedExports = remoteExports.filter((v, name) => name !== 'default');
const anyNamed = namedExports.length > 0;
namedExports.forEach((v, name) => addNamed(name, node));
```

jkimbo: @jfmengels wouldn't quite work since `any` is used for the `No named exports found in module` check below.

This would work though:
```js
const anyNamed = remoteExports.length > 0;
const namedExports = remoteExports.filter((v, name) => name !== 'default');
namedExports.forEach((v, name) => addNamed(name, node));
```

Happy to change if you think it makes it more readable (I admit I was confused about the assignment in the condition)

benmosher:
@jkimbo I think the reason I was thinking it's more complicated is that it's larger than this. I.E. right now `named` won't report `import SomeDefault from 'foo'` if `foo` has `export * from 'bar'` and `bar` has a default export, which I've now learned is wrong.

So it's a larger issue than just the `export` rule.

I'm going to take a stab at it from inside the export map stuff, let's pause on this for a second.

src/rules/export.js
benmosher: FWIW, `remoteExports` is more of a `Map` than an `Array`, so it has `forEach` but no `filter`. It has a `size`, but this size includes the `default` name.

I agree that the `(any = true)` assignment expression is a little laconic, but it gets the job done.

benmosher: This is what I ended up with, here: https://github.com/benmosher/eslint-plugin-import/pull/332/files#diff-71778249fbd501cbc801fa0ee1cac9a2R53

`default` shouldn't count, so it does trigger the `no exported names` report if you `export * from 'foo'` where `foo` has only a `default` export.

jkimbo: That makes sense! I'll close this PR then.

jkimbo:
Superseded by https://github.com/benmosher/eslint-plugin-import/pull/332/

PR #332, opened 2016-05-11
fix #328: properly suppress `default` when looking at * exports.

### LINKED ISSUE #328: Export * and export default

benmosher:
Supersedes #331, handles `export * from '...'` suppression of `default` in all static analysis rules.

cc @jkimbo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
+## [Unreleased]
+### Fixed
+- `export * from 'foo'` now properly ignores a `default` export from `foo`, if any. ([#328]/[#332], thanks [@jkimbo])
+  This impacts all static analysis of imported names. ([`default`], [`named`], [`namespace`], [`export`])
+
 ## [1.8.0] - 2016-05-11
 ### Added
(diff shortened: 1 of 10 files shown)


jkimbo:
Looks great! Thanks for the quick fix!

benmosher:
@jkimbo no problem! thanks for catching this. also, thanks for taking a stab at it, your PR was a good jumping-off point. I probably wouldn't have found that line in `src/rules/exports.js`.

PR #334, opened 2016-05-11
Check duplicates for normal imports and flow type imports separately

### LINKED ISSUE #225: no-duplicates should ignore flow type imports
I'm using [flow](http://flowtype.org/) and `no-duplicates` triggers on:

```js
import {someFunction} from '../src/foo';
import type {someType} from '../src/foo';
```

But as far as I know there's no way to import both types and other exports with the same line. `no-duplicates` should consider a flow type import as separate from a module export.

sgtpepper43:
Fix #225

If import is a flow `type` import, add to separate map for duplicate checking.

### DIFF
--- a/src/rules/no-duplicates.js
+++ b/src/rules/no-duplicates.js
@@ -4,28 +4,35 @@ import Set from 'es6-set'
 
 import resolve from '../core/resolve'
 
+function checkImports(imported, context) {
+  for (let [module, nodes] of imported.entries()) {
+    if (nodes.size > 1) {
+      for (let node of nodes) {
+        context.report(node, `'${module}' imported multiple times.`)
+      }
+    }
+  }
+}
+
 module.exports = function (context) {
   const imported = new Map()
+  const typesImported = new Map()
   return {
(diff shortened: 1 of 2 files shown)


sgtpepper43:
The travis ci build failed because `nvm install 5` timed out. Probably just needs to be restarted.

benmosher:
Thanks! 😄

dmnd:
Nice, thanks!

mizchi:
I encountered alike problem on `import/named`.

```js
import type { Connector } from 'react-redux'
```

https://github.com/flowtype/flow-typed/blob/master/definitions/npm/react-redux_v5.x.x/flow_v0.30.x-/react-redux_v5.x.x.js#L33-L36

`Connector` is just a type on definition file but there is no instance.

fabiomcosta:
:/
Too bad this wasn't introduced with an option to disable this behavior.
You can import types and other things with the same import with:
```js
import defaultNamed, { named, typed namedType } from './module';
```

This is fine as the default, but it would be nice to provide an option to keep the previous behavior.

PR #335, opened 2016-05-12
Adding handling for multiple Webpack configs

GreenGremlin:
If the Webpack config is an array, then this change will use the first config that includes a resolve section.

### DIFF
--- a/resolvers/webpack/README.md
+++ b/resolvers/webpack/README.md
@@ -25,7 +25,8 @@ or with explicit config file name:
 ---
 settings:
   import/resolver:
-    webpack: { config: 'webpack.dev.config.js' }
+    webpack:
+      config: 'webpack.dev.config.js'
 ```
 
 or with explicit config file name:
@@ -34,6 +35,7 @@ or with explicit config file name:
 ---
 settings:
   import/resolver:
-    webpack: { config: 'webpack.multiple.config.js' }
-    config-index: 1
+    webpack:
+      config: 'webpack.multiple.config.js'
+      config-index: 1   # take the config at index 1
 ```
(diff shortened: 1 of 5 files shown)


GreenGremlin:
I'm have a bear of a time trying to figure out how to configure `eslint-import-resolver-webpack` as a dependency in package.json directly from github. Any ideas?

benmosher:
Code looks great, strategy makes sense. Can you add a note about this new behavior in the resolver README and the root change log?

As for installing from GitHub: not sure, since it's in a folder.

GreenGremlin:
Thanks! I'll add the note.

I ended up getting it to work like this:
``` json
"eslint-import-resolver-webpack": "file:node_modules/eslint-plugin-import/resolvers/webpack",
"eslint-plugin-import": "github:greengremlin/eslint-plugin-import#master-build",
```

GreenGremlin:
Done. I also added support for a `config-index` setting for manually specifying which config to load.

GreenGremlin:
The failure is a timeout in a test that does not cover my changes.

resolvers/webpack/README.md
```diff
@@ -25,3 +27,13 @@ settings:
...
 ```
+
+or with explicit config file name:
+
+```yaml
+---
+settings:
+  import/resolver:
+    webpack: { config: 'webpack.multiple.config.js' }
+    config-index: 1
```
benmosher: This actually needs to be a sibling of the "config" key, I think. The settings that the resolver gets are the contents of the object keyed by "webpack"

benmosher:
Added docs look great, thanks! Just one note on the config-index example. 

And that test timeout thing is a real pain, I thought I had fixed it but apparently not.

resolvers/webpack/README.md
GreenGremlin: Done. I also fixed the JSON piece in the YAML example.

GreenGremlin:
Done. I also fixed the mixed JSON / YAML in the documentation

GreenGremlin:
Is anything else needed to merge this?

benmosher:
Nope, looks good!

GreenGremlin:
Great! Thanks. Any idea when you might push a release to npm?

benmosher:
Likely by Friday evening.

PR #339, opened 2016-05-13
Fix order's `newlines-between` edge cases

singles:
Fixes all issues mentioned in #313:

1. Empty lines between multiline import statements are detected correctly.
2. Not assigned imports are ignored (see comments).
3. `require` used inside objects are ignored.

It also changes requirement - there's no specific limit about 1 empty line between groups. Now it has to be **at least** 1. If one would like to have only one, then `no-multiple-empty-lines` core rule can be used (thx @lo1tuma for hint!).

I'm not 100% sure about implementation for last one - it works for provided test cases, yet I could use another pair of eyes.

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -110,28 +110,33 @@ function convertGroupsToRanks(groups) {
 }
 
 function makeNewlinesBetweenReport (context, imported, newlinesBetweenImports) {
-  const getLineDifference = (currentImport, previousImport) => {
-    return currentImport.node.loc.start.line - previousImport.node.loc.end.line
+  const getNumberOfEmptyLinesBetween = (currentImport, previousImport) => {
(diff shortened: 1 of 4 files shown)


src/rules/order.js
```diff
@@ -110,28 +110,33 @@ function convertGroupsToRanks(groups) {
 }
 
 function makeNewlinesBetweenReport (context, imported, newlinesBetweenImports) {
-  const getLineDifference = (currentImport, previousImport) => {
-    return currentImport.node.loc.start.line - previousImport.node.loc.start.line
+  const getNumberOfEmptyLinesBeetwen = (currentImport, previousImport) => {
```
jfmengels: Between :)

singles: Thx, fixed.

jfmengels:
LGTM

PR #343, opened 2016-05-15
Fix prefer default in multiple named export case

### LINKED ISSUE #342: prefer-default-export doesn't understand destructoring
The `prefer-default-export` rule fires when I do this:

```
// typography.js
export const { rhythm, TypographyStyle } = typography;
```

But I get `import-named` complaining that 'rhythm' cannot be found when I change it back to:

```
// typography.js
export default typography;
```

And continue to reference it like this:

```
// component.js
import { rhythm } from 'utils/typography';
```

I guess I prefer the first option, but that would require that `prefer-default-export` understand destructuring.

On the other hand, do we really want to be that restrictive on the `import` destructuring check? Isn't it reasonable to destructure a default export?

scottnonnenberg:
Fixes #342. Instead of counting the number of `ExportNamedDeclaration` nodes, we count the number of properties in its child `ObjectPattern`.

A couple other style fixes.

### DIFF
--- a/src/rules/newline-after-import.js
+++ b/src/rules/newline-after-import.js
@@ -96,7 +96,9 @@ module.exports = function (context) {
             return
           }
 
-          if (nextStatement && (!nextRequireCall || !containsNodeOrEqual(nextStatement, nextRequireCall))) {
+          if (nextStatement &&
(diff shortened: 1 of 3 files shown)


benmosher:
It's close. Will provide feedback ASAP, next few days or so. 😄

PR #350, opened 2016-05-19
allow multiple named export through variable deconstructing fixes #342

### LINKED ISSUE #342: prefer-default-export doesn't understand destructoring

gavriguy:
This PR fixes #342 by allowing to export more than one variable using variable deconstructing

### DIFF
--- a/src/rules/prefer-default-export.js
+++ b/src/rules/prefer-default-export.js
@@ -1,5 +1,8 @@
 'use strict'
 
+import get from 'lodash.get'
+// import forEach from 'lodash.forEach'
+
 module.exports = function(context) {
   let namedExportCount = 0
   let specifierExportCount = 0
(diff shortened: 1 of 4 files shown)


src/rules/prefer-default-export.js
```diff
@@ -1,5 +1,8 @@
 'use strict'
 
+import get from 'lodash.get'
+// import forEach from 'lodash.forEach'
```
jfmengels: You missed removing a comment here :)

jfmengels:
Not that familiar with this export syntax so I don't know if there are caveats, but LGTM (once the extra comment has been removed :smile:).

benmosher:
Might want to hold off, this duplicates #343 which already solved #342, I think? (If not, can you work from that branch as a start?)

gavriguy:
Oops didn't see the progress that was made - I'll hold off in favor  of #343. @scottnonnenberg feel free to take anything from this PR. and please let me know if you need any help.

PR #351, opened 2016-05-19
Fix prefer default

benmosher:
Builds off of #343, but adds support for

- deep destructuring (`export const { foo: { bar } } = ...`)
- re-export (`export * from './foo'`)
- no exports (threw an error)

CC @scottnonnenberg, @gavriguy


gavriguy:
LGTM nice work

PR #353, opened 2016-05-23
Release v1.8.1, webpack/0.2.5

benmosher:
Releasing bugfixes + minor improvements to core plugin + Webpack resolver.

#332 feels semver-minor because it's large, but in all likelihood it just hasn't impacted anyone before @jkimbo brought it up. So probably valid semver-patch.

Tasks:

- [x] acquire maiden LGTM from @jfmengels 
- [ ] squash + merge (@benmosher)
- [ ] create both tags (@benmosher)
- [ ] publish (@benmosher)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,11 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
-## [Unreleased]
+## resolvers/webpack/0.2.5 - 2016-05-23
 ### Added
 - Added support for multiple webpack configs ([#181], thanks [@GreenGremlin])
 
-## [Unreleased]
+## [1.8.1] - 2016-05-23
 ### Fixed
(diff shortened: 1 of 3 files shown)


jfmengels:
I think we should probably separate submodule changelogs. Either by putting the core module's changelog first, then the resolvers (with a "table of contents" at the top of the page), or in separate files (I would choose option one). And it would be nice to have the same diff links (`[1.8.1`]) for the submodules.

Though we can do this all at a late time, and this LGTM ;)

benmosher:
Yeah, the resolvers have been kindof a pain in the changelog. Esp. because I can't create one of those diff links for a subtree (...right?).

I'd probably put them in their own file, but they have been changing somewhat infrequently so I haven't bothered. 😅

jfmengels:
> I can't create one of those diff links for a subtree (...right?)

You could, if you prepend them with something like
```md
## [webpack-0.2.5] - 2016-05-23
[webpack-0.2.5]: ./foo
```

jfmengels:
:tada:

benmosher:
took long enough 😅

PR #359, opened 2016-05-29
prefer-default-export: Support 'export function foo() {}; export const x = 4;'

### LINKED ISSUE #358: prefer-default-export doesn't count 'export function foo() {}'
This code fires the `prefer-default-export` rule in the recently-released `v1.8.1`:

```javascript
export const foo = 'foo';
export function bar() {};
```

scottnonnenberg:
Fixes #358.

The fixes to support multiple declarations left out a crucial else case.


gavriguy:
LGTM

benmosher:
Great catch!

Can you make a change log note?

scottnonnenberg:
Hopefully that's what you're looking for! :0)

Issue #349, opened 2016-05-18
webpack: Don't throw on externals function


kesne:
Currently, if externals is defined as a function, then the plugin will throw an error. This makes the webpack resolver completely useless if you ever use an externals function. Instead of throwing, maybe we could log out an error, or just list is as a limitation of the module? Just something that still makes the module useable in these cases.

This is the line of code that causes issues:
https://github.com/benmosher/eslint-plugin-import/blob/c15901b2deb70b414a25458ba6ae43b8e3825b82/resolvers/webpack/index.js#L170

I ran into this issue when pointing to a webpack config that uses `webpack-node-externals`, which made it impossible to use this plugin against that config.

benmosher:
Ah, wild, I didn't realize that was an option. Should be easily fixed, I think. Thanks!

benmosher:
Actually could you contribute a PR with a failing test? Looking again, I'm not sure how I would validate that this issue is resolved. I had hoped that the branch using enhanced-resolve would have handled it, but it did not.

kesne:
contributed a PR with a basic fix in #363

PR #363, opened 2016-06-01
[webpack-resolver] Allowing externals to be a function

kesne:
This is a basic fix for #349. It handles externals being a function. I really only tested it against the functionality of `webpack-node-externals`, so it's likely not fully complete, but at least it's a step in the right direction. And this seems to be a relatively less-used feature of Webpack.

One thing that is definitely incompatible is that the callback function must be sync, or else the external function won't resolve in time. Thankfully, the `webpack-node-externals` function is sync, so it handles that usage perfectly.

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -190,7 +190,15 @@ function findExternal(source, externals) {
   }
 
   if (typeof externals === 'function') {
-    throw new Error('unable to handle function externals')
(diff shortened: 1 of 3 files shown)


benmosher:
LGTM, just rebased and merged into master 😎

benmosher:
FYI, just (finally!) published with v0.3.2 of the resolver.

PR #365, opened 2016-06-03
import/core-modules

### LINKED ISSUE #275: Add ability to ignore certain packages from all rules
[Electron](https://github.com/electron/electron) exposes an `const electron = require('electron');` import, but it's not listed anywhere, as it's more of an additional built-in. Same goes for `atom`, when creating Atom plugins.

I think we could just ignore these by default as no one would want it to be linted, but there also might be other cases of this, so would still be useful to have an ignore option.

Here's the errors I get when running this plugin on an Electron project:

```
  browser.js:2
  ✖  2:18  'electron' is not listed in the project's dependencies. Run 'npm i -S electron' to add it  import/no-extraneous-dependencies
  ✖  2:26  Unable to resolve path to module 'electron'.                                               import/no-unresolved

  index.js:4
  ✖  4:18  'electron' is not listed in the project's dependencies. Run 'npm i -S electron' to add it  import/no-extraneous-dependencies
  ✖  4:26  Unable to resolve path to module 'electron'.                                               import/no-unresolved

  menu.js:3
  ✖  3:18  'electron' is not listed in the project's dependencies. Run 'npm i -S electron' to add it  import/no-extraneous-dependencies
  ✖  3:26  Unable to resolve path to module 'electron'.                                               import/no-unresolved

  storage.js:4
  ✖  4:18  'electron' is not listed in the project's dependencies. Run 'npm i -S electron' to add it  import/no-extraneous-dependencies
  ✖  4:26  Unable to resolve path to module 'electron'.                                               import/no-unresolved

  tray.js:2
  ✖  2:18  'electron' is not listed in the project's dependencies. Run 'npm i -S electron' to add it  import/no-extraneous-dependencies
  ✖  2:26  Unable to resolve path to module 'electron'.                                               import/no-unresolved
```

Example import: https://github.com/sindresorhus/caprine/blob/2acdd358eafdc42b45b532c90e87ba5465a30de7/index.js#L4

benmosher:
Fixes #275.

- [x] handle `import/core-modules` in `no-unresolved`
- [x] handle `import/core-modules` in @jfmengels's `importType` support stuff
- [x] create Electron shared config spec-ing `import/core-modules` as `[electron]`

```yaml
settings:
  import/core-modules: [ electron ]
```

### DIFF
--- a/config/electron.js
+++ b/config/electron.js
@@ -1,0 +1,8 @@
+/**
+ * Default settings for Electron applications.
+ */
+module.exports = {
+  settings: {
+    'import/core-modules': ['electron'],
+  },
+}
--- a/src/core/resolve.js
+++ b/src/core/resolve.js
@@ -54,9 +54,13 @@ function fileExistsWithCaseSync(filepath, cacheSettings) {
 }
 
 export function relative(modulePath, sourceFile, settings) {
+  return fullResolve(modulePath, sourceFile, settings).path
+}
+
+function fullResolve(modulePath, sourceFile, settings) {
   // check if this is a bonus core module
   const coreSet = new Set(settings['import/core-modules'])
-  if (coreSet != null && coreSet.has(modulePath)) return { path: null, found: true }
+  if (coreSet != null && coreSet.has(modulePath)) return { found: true, path: null }
 
   const sourceDir = path.dirname(sourceFile)
       , cacheKey = sourceDir + hashObject(settings) + modulePath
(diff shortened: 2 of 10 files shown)


src/core/resolve.js
```diff
@@ -210,3 +214,7 @@ function hashObject(object) {
   settingsShasum.update(JSON.stringify(object))
   return settingsShasum.digest('hex')
 }
+
+const envCores = new Map([
+  ['electron', new Set(['electron'])],
```
sindresorhus: Can you add `atom` to the default too? They too have a builtin require.

config/electron.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Default settings for Electron applications.
+ */
+module.exports = {
+  settings: {
+    'import/core-modules': ['electron'],
```
benmosher: @sindresorhus switched to this model where the core module lists are not baked in, but made a shared config for Electron that can be maintained by the community. 

Another such config can be made for any of Atom, Meteor, etc..

Avoids a constant stream of breaking-change platform-related PRs, let's people solve such problems quickly. I think it's better than baking the sets into the code.

sindresorhus: Works for me.

PR #369, opened 2016-06-04
do not resolve the source directory. Fixes #367

### LINKED ISSUE #367: if relative paths are used, custom resolver cannot be a path relative to node_module root
If I use a relative import, modulePath is relative and then pkg-dir cannot find the package directory and my custom resolver is not found.

lukeapage:
You cannot call resolve without a 2nd argument - but because we now pass the source directory, not the modulePath, it will always be absolute so you don't need to.

Added a test as well, covering the situation this fixes and also the situation I use the custom resolver for.

### DIFF
--- a/src/core/resolve.js
+++ b/src/core/resolve.js
@@ -157,7 +157,7 @@ function requireResolver(name, sourceFile) {
 
   // Try to resolve package with path, relative to closest package.json
   try {
-    const packageDir = pkgDir.sync(resolve(sourceFile))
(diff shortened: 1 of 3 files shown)


benmosher:
👍! Thanks for adding a test, too!

lukeapage:
Thanks. @benmosher Any chance of a release? I'm having to build my own version at the moment for a large project using this.

benmosher:
Yeah, I expect to do a release this week, probably tomorrow or Thursday.

PR #372, opened 2016-06-09
v1.9.0 bump

benmosher:
There are more things I had hoped to complete during this release, but it's been a while so probably best to ship what's done, first.

- [x] get the old once-over and signoff from @jfmengels 
- [x] tag
- [x] publish to `next`
- [ ] ~~give it a day~~
- [ ] ~~move to `latest`~~

officially broken, can't release as-is

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+
+
+## [1.9.0] - 2016-06-10
 ### Added
 - Added support TomDoc comments to [`no-deprecated`]. ([#321], thanks [@josh])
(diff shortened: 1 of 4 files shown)


jfmengels:
Let's do this :)

benmosher:
Once this goes green I'll move to master, but 1.9.0 is officially broken, so it will never move to `next` 😓 .

@lukeapage and @le0nik I need you two to provide feedback on my changes in e044e37, ideally in the form of additional tests.

jfmengels:
What are the problems?

lukeapage:
@benmosher a quick review of it looks good, i can try it out on our repo soon.

lukeapage:
confirmed, it still works nicely for my use case (also the one I added a test for)

benmosher:
@jfmengels the new custom resolver stuff tries to load actual Webpack instead of the Webpack resolver plugin, due to the original precedence order. 

Doesn't show up in tests because Webpack isn't installed, but it happened immediately when I tried it in my actual project. 😅

PR #377, opened 2016-06-14
Webpack resolver plugin

Rogeres:
Hello.

I use `webpack.ResolverPlugin` in my project in order to add my own resolvers:

https://github.com/webpack/docs/wiki/list-of-plugins#resolverplugin

`eslint-import-resolver-webpack` is not supporting such a way adding plugins now. This pull request solves the problem (see the code).

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -86,9 +86,19 @@ exports.resolve = function (source, file, settings) {
   // externals
   if (findExternal(source, webpackConfig.externals)) return { found: true, path: null }
 
+  var otherPlugins = []
+
+  // support webpack.ResolverPlugin
+  if (webpackConfig.plugins) {
+    webpackConfig.plugins.forEach(function (plugin) {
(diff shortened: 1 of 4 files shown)


benmosher:
Ah, cool! Thanks!

Why filter to just ResolverPlugin, though? Do you anticipate other configured plugins may be configured that would be wasteful or wrong?

Rogeres:
I supported only `webpack.Resolver Plugin`, because it seems to be unnecessary to support other plugins.

benmosher:
Fair enough. A good start, for sure. I don't know enough about Webpack plugins to know if it would hurt to just apply all of them.

@le0nik: do you have any thoughts on this? I can't remember if you had custom plugins or were just trying to use this one.

eMaj-Fr:
1.10.4 ?

benmosher:
Probably 0.4.0 of the Webpack resolver. 

Should be out some time next week.

benmosher:
published as v0.4.0: https://www.npmjs.com/package/eslint-import-resolver-webpack

PR #391, opened 2016-06-21
v1.9.2

benmosher:
Patch release, before merging feature PRs.

- [x] once-over from @jfmengels 
- [x] check against test set
- [ ] publish to `latest` on npm


benmosher:
Those wouldn't be part of this anyway, I need to publish the Node resolver separately to fix that. I had marked that as part of milestone 1.10, but really it's just a pre-req (since there are tests that require it now).

jfmengels:
Ok, sounds good to me. Btw, you're on a roll today :smile:

PR #395, opened 2016-06-23
Fix crash with `newline-after-import` related to the use of switch cases (fixes #386)

### LINKED ISSUE #386: `newline-after-import` breaks
```sh
Cannot read property 'type' of undefined
TypeError: Cannot read property 'type' of undefined
    at getScopeBody ($PWD/node_modules/eslint-plugin-import/lib/rules/newline-after-import.js:30:11)
    at $PWD/node_modules/eslint-plugin-import/lib/rules/newline-after-import.js:104:27
    at Array.forEach (native)
    at $PWD/node_modules/eslint-plugin-import/lib/rules/newline-after-import.js:103:22
    at Array.forEach (native)
    at EventEmitter.ProgramExit ($PWD/node_modules/eslint-plugin-import/lib/rules/newline-after-import.js:99:14)
    at emitOne (events.js:82:20)
    at EventEmitter.emit (events.js:169:7)
    at NodeEventGenerator.leaveNode ($PWD/node_modules/eslint/lib/util/node-event-generator.js:49:22)
    at CodePathAnalyzer.leaveNode ($PWD/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:627:23)
    at CommentEventGenerator.leaveNode ($PWD/node_modules/eslint/lib/util/comment-event-generator.js:110:23)
```

No idea what file it breaks on, but it started immediately after enabling this rule.

jfmengels:
Fix crash with `newline-after-import` related to the use of switch cases (fixes #386)

The error came because a switch case has a scope

- I've added a small safe-guard that should prevent the same kind of crash in this part of the code
- I have made it so no errors can come from inside switch cases. I tried returning the switch case `cases` items, but then it reports an error even if I space them correctly. Maybe someone wants to take a further look into it, but I'm out of time right now.

cc @ljharb @singles

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - Added new rule [`no-restricted-paths`]. ([#155]/[#371], thanks [@lo1tuma])
 - [`import/core-modules` setting]: allow configuration of additional module names,
   to be treated as builtin modules (a la `path`, etc. in Node). ([#275] + [#365], thanks [@sindresorhus] for driving)
+### Fixed
+- Fixed crash with `newline-after-import` related to the use of switch cases. (fixes [#386], thanks [@ljharb] for reporting) ([#395])
 
 ## [1.9.2] - 2016-06-21
 ### Fixed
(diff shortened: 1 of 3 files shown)


benmosher:
LGTM. I thought case branches didn't have a scope without wrapping in an explicit block. 

Seems strictly like an improvement to me! Thanks for fixing this!

jfmengels:
:)

Issue #333, opened 2016-05-11
`import/no-unresolved` should find `json` files


jfmengels:
I'm trying to add the plugin to a repo I'm working on, and I'm getting false positives with `no-unresolved` when importing JSON files.

```js
import setEvent from './fixtures/foo'; // <-- `./fixtures/foo.json` is an existing and valid JSON file
```

I'm not aware of it being incorrect to `import` a JSON file, but please let me know if that's the case.

benmosher:
Weird. Assuming resolve supports it, I think it should. Sounds like a bug.

shark0der:
I'm having a similar issue but with jsx files. Webpack doesn't complain at all.

benmosher:
@jfmengels I just added tests but they are passing. Can you take a look? If you can construct a failing test I can patch it, but I'm not sure what the issue is. Maybe something with your config?

jfmengels:
Hmm... Yeah, I'll try to take a look at it tonight.

benmosher:
No rush, speed is up to you. AFAICT it's working. (also, sorry it took so long for me to look at this!)

benmosher:
ah, yeah, so for this, `import/extensions` is the correct solution ATM: https://github.com/benmosher/eslint-plugin-import#importextensions

I think that should resolve it. See note in README about making `['.js']` the default in v2.

benmosher:
Also, I'm confused why you don't have push rights to a branch, I thought you could push anywhere in the repo?

jfmengels:
> Also, I'm confused why you don't have push rights to a branch, I thought you could push anywhere in the repo?

Fixed it, my bad. Didn't configure my git remote stuff well. Just pushed my `issue-333` branch over yours now.

And the `import/extensions` tip works :+1:

benmosher:
Ah, wild. This is actually an "issue" with `node-resolve`. By default it only finds `.js` files.

I just need to add `.json` to the Node resolver and this should work.

jfmengels:
:smile:

benmosher:
Just published with node resolver v0.2.1

PR #400, opened 2016-06-30
v1.10.0

benmosher:
- [x] test locally against real code 
- [x] 300 github stars! 🎉
- [x] the ol' once-over from @jfmengels
- [x] tag in git
- [x] publish to npm

FYI, also just published v0.3.2 of the Webpack resolver this morning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+
+## [1.10.0] - 2016-06-30
 ### Added
 - Added new rule [`no-restricted-paths`]. ([#155]/[#371], thanks [@lo1tuma])
 - [`import/core-modules` setting]: allow configuration of additional module names,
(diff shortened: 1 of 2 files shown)


jfmengels:
> 300 github stars! :tada:

Also reached 100.000 downloads this week!
image
http://www.npmtrends.com/eslint-plugin-import

I notice that #333 was fixed but not added to the changelog (or didn't find it). That might be worth adding.
Looks all good to me otherwise :+1:

benmosher:
Yeah, #333 fix was isolated to the Node resolver, so its changlog was actually born to note it: https://github.com/benmosher/eslint-plugin-import/blob/master/resolvers/node/CHANGELOG.md

Not clear to me whether that should also be mentioned in the main change log. Node resolver is the default, but it's installed independently as a dependency, so upgrading to 1.10+ won't necessarily pick that up.

benmosher:
Re: growth in downloads -- blowing my mind. I don't understand how it's still growing.

jfmengels:
> Not clear to me whether that should also be mentioned in the main change log

It doesn't have to, but since you have to search in the project for the node resolver project and changelog, I think it would be nice to at least have a link to the changelog from the node resolver npm page.

> Re: growth in downloads -- blowing my mind. I don't understand how it's still growing.

I think `XO` just integrated it and released it, so that can account for a few thousand downloads. I guess people are simply starting to spread the word about the benefits and quality of the plugin, so kudos to you :)

That said, I wonder how many people use the project for the static analysis tools vs the stylistic rules (like `order`). I sure hope it's for the former, but considering the popularity of my plugin (finally declining! http://www.npmtrends.com/eslint-plugin-import-order :tada:) and the uptake this project had when I added it here, I'm not sure. Too bad we can't see any data about that...

benmosher:
> That said, I wonder how many people use the project for the static analysis tools vs the stylistic rules (like order).

Me too, man. Though [this reaction](https://github.com/zertosh/eslint-plugin-dependencies/issues/1#issuecomment-219263080) made me wonder if it's getting less obvious that the static analysis stuff is available.

And the style rules are great, people want and love them. I don't use most of them, personally, but I get the desire. My focus is still the analysis, but I'm more than happy to support and distribute the style rules, too.

jfmengels:
Oh, didn't expect that kind of comment. Again, hard to know if that is the case for a lot of developers.

I think we could refresh the plugin description in the README to put an extra emphasis on static analysis and the benefits it brings. It sells itself short now that there are plenty of static analysis tools.

IMO, the analysis rules are the primary value of the plugin, and the stylistic rules are secondary. But they could be used as bait for people to start using the project :smile:

PR #405, opened 2016-07-02
release v1.10.1

benmosher:
- [x] @jfmengels, giving you an opportunity to sign off. looks like a no-brainer to me, let me know if I'm missing something?
- [x] tag
- [x] publish
- [ ] merge back into `master`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## [Unreleased]
 
+
+## [1.10.1] - 2016-07-02
+### Added
+- Officially support ESLint 3.x. (peerDependencies updeted to `2.x - 3.x`)
+
 ## [1.10.0] - 2016-06-30
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## [Unreleased]
 
+
+## [1.10.1] - 2016-07-02
+### Added
+- Officially support ESLint 3.x. (peerDependencies updeted to `2.x - 3.x`)
```
jfmengels: Updeted -> updated

jfmengels:
Dont think this is a breaking change, so it's fine for me.
If you decide to transpile less things to only transpile things not supported in v4, then that'd be a breaking change. In my plugins where I don't transpile code, I'll do a major version to stop supporting 0.10-0.12

benmosher:
@jfmengels yep, working on all that over in #408. I'm actually thinking of playing with TypeScript so 0.12 may come back, but I think 0.10 is done no matter what so I can jettison the `es6-*` polyfills for collections + runtime functions.

PR #408, opened 2016-07-02
v2.0.0

### LINKED ISSUE #233: Expose the resolve function to other plugins

benmosher:
# 2.0!

First real breaking change! tracking progress here. Most discussion should probably occur on the individual issue pages.

## core package (eslint-module-utils)
Pull out src/core into eslint-module-utils, to be published separately for other plugins to leverage.
(copied from #234, closes #233)

- [ ] bikeshed a better name? (if you're interested in this, open an issue and reference this plzzz)
- [ ] document (README, etc)
- [ ] publish as v1
- [ ] add published version to plugin package dependencies


jfmengels:
We should probably drop support for Node v4 (if it's not already done) and update the rules to the latest format (http://eslint.org/blog/2016/07/eslint-new-rule-format). Willing to do the work, but that should not appear on master before we're close to releasing.

benmosher:
I'm up for the new format, but it wasn't obvious to me that it's urgently necessary, since the older format is still supported and updating the rules should not be a noticeable change for users, right?

Also: isn't ESLint still supporting Node v4?

jfmengels:
Sorry, meant dropping support for Node < v4.

The older format is still supported, and the new format is supported by the latest 2.x versions. I consider it a breaking change, even if I haven't encountered any problems with it. Most plugins I've seen that have made the jump released a major version for it, though they almost all also started using ES2015 features supported by Node v4.

benmosher:
Let's knock it out, then. Feel free to push updated rules to this branch. 

And yeah, this branch is v4+. 😎

benmosher:
@jfmengels anything else you can think of that we should address before publishing this?

I'm thinking of publishing as `v2.0.0-beta.0`?

jfmengels:
Ah, good job for the rule format.
Let's publish a beta v2 (don't forget `--tag beta` when publishing).

I think we could take a look at updating the recommended config, so that we can at least give a taste of the useful rules that don't need config. I'm thinking no-unresolved, no-extraneous-dependencies, etc...
But we can do that after the first beta version.

benmosher:
@jfmengels I'll try to merge this over the weekend or early next week.

jfmengels:
Cool! :)

jfmengels:
:tada:

PR #412, opened 2016-07-04
Fixes #411

### LINKED ISSUE #411: In findExternal, context should be a string
(a suggestion I'm not sure about:)

In https://github.com/benmosher/eslint-plugin-import/blob/master/resolvers/webpack/index.js#L174 it might be safer if the second `null` argument corresponding to the 'context' argument is an empty string instead.  Webpack will always provide a string for this argument and no document says it will do otherwise.   Under that circumstance, it is normal to use string methods such as `indexOf` which will make it fail when a `null` is received.  An empty string, I believe,  serves the same purpose and doesn't break anything.

Satyam:
Fix in index.js
Extra configuration on webpack.config.js to trip the error if not fixed

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -84,7 +84,7 @@ exports.resolve = function (source, file, settings) {
   log('Using config: ', webpackConfig)
 
   // externals
-  if (findExternal(source, webpackConfig.externals)) return { found: true, path: null }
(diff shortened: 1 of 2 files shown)


Satyam:
Sorry, I didn't realize the package had separate installs and tests for the resolvers themselves.  I'll fix it and resubmit.  Out of the three errors, only one is mine, the other two are not my fault.

PR #413, opened 2016-07-04
Fixes #411

### LINKED ISSUE #411: In findExternal, context should be a string

Satyam:
index.js contains the fix
webpack.config.js contains code that trips an error if fix not present.
Initial git clone test gave two errors which are not related to this fix.

### DIFF
--- a/resolvers/webpack/test/files/webpack.config.js
+++ b/resolvers/webpack/test/files/webpack.config.js
@@ -14,7 +14,7 @@ module.exports = {
     { 'jquery': 'jQuery' },
     'bootstrap',
     function (context, request, callback) {
-      if (request === 'underscore') {
+      if (request === 'underscore' && context.indexOf('/') !== -1) {
         return callback(null, 'underscore');
       };
       callback();
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -84,7 +84,7 @@ exports.resolve = function (source, file, settings) {
   log('Using config: ', webpackConfig)
 
   // externals
-  if (findExternal(source, webpackConfig.externals)) return { found: true, path: null }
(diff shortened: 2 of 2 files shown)


benmosher:
Thanks for this! 

Looks good at first glance, I just need to check out the test failures and take a closer look at the Webpack spec.

benmosher:
Finally looked at the spec. No good docs on `context` that I could find, but this makes sense to me. Thanks!

resolvers/webpack/test/files/webpack.config.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
     { 'jquery': 'jQuery' },
     'bootstrap',
     function (context, request, callback) {
-      if (request === 'underscore') {
+      if (request === 'underscore' && context.indexOf('/') !== -1) {
```
benmosher: @Satyam: can you explain why this extra predicate? seems to be unnecessary on OSX, and I think it might be why the tests fail on Windows?

Satyam: It is just to force an error when `context` is `null` since `null` has no `indexOf` method.  Whatever produces an error when `context` is null would do just as well, just to make sure the fix is not reversed later on.

Issue #415, opened 2016-07-04
1.10.1 breaks ESLint 2.13.1 and Node < 4


rhettlivingston:
I work with Meteor which is just starting to adopt Node 4 (in Beta). We have many sample apps running Node 0.10.x.

I just downloaded the main Meteor todos app, ran "npm install" and got eslint-plugin-import 1.10.1 installed with ESLint 2.13.1. This produced the following error...

```bash
rhett@dreamshot:~/oob/meteors/todos1.3$ meteor npm run lint

> @ lint /home/rhett/oob/meteors/todos1.3
> eslint .


/home/rhett/oob/meteors/todos1.3/node_modules/eslint-plugin-import/lib/index.js:6
const rules = exports.rules = {
^^^^^

Oops! Something went wrong! :(

ESLint couldn't find the plugin "eslint-plugin-import". This can happen for a couple different reasons:

1. If ESLint is installed globally, then make sure eslint-plugin-import is also installed globally. A globally-installed ESLint cannot find a locally-installed plugin.

2. If ESLint is installed locally, then it's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm i eslint-plugin-import@latest --save-dev

If you still can't figure out the problem, please stop by https://gitter.im/eslint/eslint to chat with the team.


npm ERR! Linux 4.4.0-28-generic
npm ERR! argv "node" "/home/rhett/.meteor/packages/meteor-tool/.1.3.0_3.139i3kk++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/bin/npm" "run" "lint"
npm ERR! node v0.10.43
npm ERR! npm  v2.14.22
npm ERR! code ELIFECYCLE
npm ERR! @ lint: `eslint .`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @ lint script 'eslint .'.
npm ERR! This is most likely a problem with the  package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     eslint .
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs 
npm ERR! Or if that isn't available, you can get their info via:
npm ERR! 
npm ERR!     npm owner ls 
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /home/rhett/oob/meteors/todos1.3/npm-debug.log
```

[...]

PR #426, opened 2016-07-07
Transformed all remaining occurences of for..of to forEach()

rhettlivingston:


benmosher:
fixes #415?

src/core/getExports.js
```diff
@@ -236,18 +236,19 @@ export default class ExportMap {
     if (this.reexports.has(name)) return true
 
     // default exports must be explicitly re-exported (#328)
+    let foundInnerMapName = false
     if (name !== 'default') {
-      for (let dep of this.dependencies.values()) {
-        let innerMap = dep()
+      this.dependencies.forEach((dep) => {
```
benmosher: this would ideally be a `some` so that it short-circuits. not sure how to achieve with a `Set`...

rhettlivingston:
Yes. Didn't put that in the comment because I was merging into the branch
that fixes it.

I have to apologize for what it does to your code. I really wish I knew how
to just make the for..of implementation play well with others. forEach is
very inferior in this case because you can neither 'continue' within it or
'return' from it.

src/core/getExports.js
```diff
@@ -276,24 +277,29 @@ export default class ExportMap {
...
-        if (!innerMap) continue
-
-        // safeguard against cycles
-        if (innerMap.path === this.path) continue
-
-        let innerValue = innerMap.hasDeep(name)
-        if (innerValue.found) {
-          innerValue.path.unshift(this)
-          return innerValue
+      this.dependencies.forEach((dep) => {
```
benmosher: same here, ideally would short-circuit

src/rules/no-unresolved.js
```diff
@@ -53,15 +53,17 @@ module.exports = function (context) {
...
-    for (let element of modules.elements) {
-      if (element.type !== 'Literal') continue
-      if (typeof element.value !== 'string') continue
-
-      if (element.value === 'require' ||
-          element.value === 'exports') continue // magic modules: http://git.io/vByan
-
-      checkSourceValue(element)
-    }
+    modules.elements.forEach((element) => {
```
benmosher: I would prefer that these `continue`s become `returns`, to minimize deep nesting. but that's just code style preference, no impact on semantics/performance 🤓

src/core/getExports.js
rhettlivingston: I learned while doing this that `Maps` and `Sets` are very fundamentally not `Arrays`. So, no `some` and not even a clear way to utilize the basic `for` statement and break out of it (i.e. you can't access them in an indexed fashion).

I suspect the overhead of converting to an array just to use some would be greater than the gain.

src/rules/no-unresolved.js
rhettlivingston: Ahhh. Realized that returns didn't do the same thing inside the foreach, but didn't click on the fact that they did do the same thing as a continue. I agree with you. Will go back and take advantage of that insight. It twists your code a little less.

Frankly, I prefer your original code style and have used it in many similar cases for more than 30 years despite the CompSci perfectionists always criticizing me for not having a single point of return at the end of my functions. I'm a CompEng. We prefer to do what works. :smile:

src/core/getExports.js
rhettlivingston: After thinking about this some,,, I may experiment with directly accessing the iterator in a for loop. This is essentially what is going on with for..of, but I might be able to get away with it if I don't involve the underlying Symbol type and go directly to the Map / Set.

I've also realized I might have changed more than I had too. The instances that definitely caused problems involved using for..of to iterate Maps. In response, I changed all occurrences of for..of. It is possible that using for..of to iterate a Set causes no issue.

benmosher:
**On short-circuits**: I think you're probably right, best to leave it be.

Though notably, `this.dependencies` only ever deals with its values after parse time, so the only use it gets being a `Map` is one-per-path `export * from x` behavior, which could be easily achieved with an ephemeral `Set` of paths during `parse` and just `if (!exportAllPaths.has(remoteMap)) m.dependencies.push(() => ExportMap.for(remoteMap, context)` on line ~163 of `getExports.js`.

FWIW, I'll probably revert to the `for ... of` syntax in v2 anyway since 0.10 will no longer be supported. so no sweat either way. 😁 Best to just get it correct for the short term and not worry too much about the ideal case.

I'm just over-thinking it.

benmosher:
I'll go ahead and merge as-is if it fixes #415, no need to sweat the style stuff ATM.

rhettlivingston:
Thank you very much for accepting this. And I'd absolutely agree with reverting to `for..of` later. It is unquestionably the proper way to traverse Maps and Sets.

PR #444, opened 2016-07-19
consider modules from `bower_components' and 'jspm_modules` as external modules

zloirock:
In one project we wanna use `order` rule with configuration
```json
{
    "groups": [
        ["builtin", "external"],
        ["internal", "parent", "sibling", "index"]
    ],
    "newlines-between": "always"
}
```
but our frontend uses `bower` and `webpack` aliases.

### DIFF
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -66,4 +66,18 @@ describe('importType(name)', function () {
     const electronContext = testContext({ 'import/core-modules': ['electron'] })
     expect(importType('electron', electronContext)).to.equal('builtin')
   })
+
+  it("should return 'external' for module from 'node_modules' with default config", function() {
+    expect(importType('builtin-modules', context)).to.equal('external')
+  })
+
+  it("should return 'internal' for module from 'node_modules' if 'node_modules' missed in 'external-module-folders'", function() {
+    const foldersContext = testContext({ 'import/external-module-folders': [] })
+    expect(importType('builtin-modules', foldersContext)).to.equal('internal')
+  })
+
+  it("should return 'external' for module from 'node_modules' if 'node_modules' contained in 'external-module-folders'", function() {
+    const foldersContext = testContext({ 'import/external-module-folders': ['node_modules'] })
+    expect(importType('builtin-modules', foldersContext)).to.equal('external')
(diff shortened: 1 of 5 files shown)


benmosher:
Interesting, makes sense to me.

What about using a setting (how does `import/external-modules` sound?)--like [`import/core-modules`](https://github.com/zloirock/eslint-plugin-import/blob/ae1fc37e5fb56908d8424aaefbbde253f4fab665/src/core/importType.js#L12) for the `builtin` designation--and defaulting to what you've got defined for `folders` currently?

Then the next person who wants to add `rollup_modules_2017` or whatever is already covered.

Also, [docs, tests, and changelog notes](https://github.com/benmosher/eslint-plugin-import/blob/master/CONTRIBUTING.md#prs) please! 😁

zloirock:
> What about using a setting (how does import/external-module-folders sound?)

Yep, makes sense, I thought about it.

> Also, docs, tests, and changelog notes please!

Sure, I'll add it, the first commit was mainly checking you interest in this feature :)

tests/src/core/importType.js
```diff
@@ -66,4 +66,18 @@ describe('importType(name)', function () {
...
+
+  it("should return 'internal' for module from 'node_modules' if 'node_modules' missed in 'external-module-folders'", function() {
+    const foldersContext = testContext({ 'import/external-module-folders': [] })
+    expect(importType('builtin-modules', foldersContext)).to.equal('internal')
+  })
+
+  it("should return 'external' for module from 'node_modules' if 'node_modules' contained in 'external-module-folders'", function() {
+    const foldersContext = testContext({ 'import/external-module-folders': ['node_modules'] })
+    expect(importType('builtin-modules', foldersContext)).to.equal('external')
+  })
```
zloirock: I don't see a mock for `resolve`, so I used the first module from dependencies for testing.

zloirock:
Updated.

benmosher:
Looks great! Thanks!

zloirock:
@benmosher can you publish a new release?

benmosher:
Just published as v1.12.0! 😎

zloirock:
👍

Issue #447, opened 2016-07-21
Proposal: Disallow importing self


just-paja:
Hi,

I think it would be nice if we had a rule that prevents importing the current file. It might happen when refactoring and moving files across directories. What do you think?

file: foo.jsx

``` js
import foo from './foo';
```

``` js
import foo from './foo.jsx';
```

benmosher:
I would accept a PR for that. 😎 

Call it `no-self-import`?

PR #449, opened 2016-07-21
Added no-self-import rule

just-paja:
New rule checks against importing from the current file. Ref #447.

First time working with eslint rules. Please review.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,6 +8,7 @@ export const rules = {
   'no-mutable-exports': require('./rules/no-mutable-exports'),
   'extensions': require('./rules/extensions'),
   'no-restricted-paths': require('./rules/no-restricted-paths'),
+  'no-self-import': require('./rules/no-self-import'),
 
   'no-named-as-default': require('./rules/no-named-as-default'),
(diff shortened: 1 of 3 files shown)


benmosher:
I think this rule should leverage the [`resolve` infrastructure](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-unresolved.js#L20).

At least, if `resolve` comes back non-null, then the result should be compared with `context.getFilename()`. Then the resolver is responsible for things like missing extensions, aliases, `linklocal`s, etc.

jfmengels:
This would be a cool rule to have!
Are you still up to work on this @just-paja? No worries if you're not :)

just-paja:
Hey, sorry all, I completely forgot about this. @jfmengels, I will happily pass this to you.

jfmengels:
Hey, no problem ;)
Not sure if/when I'll get to this. It's up for the taking if someone wants to have a go at it.

giodamelio:
If it's all right with everyone. I would like to take a crack at this one.

jfmengels:
Please do @giodamelio!

giodamelio:
Awesome, I just opened #727.

PR #459, opened 2016-07-26
add `allowComputed` option to `namespace` (fixes #456)

### LINKED ISSUE #456: Unable to access imported items dynamically
Sorry, don't know the correct jargon for this issue title. Issue I'm hitting:

```js
import * as Module from './Module';

function getItem(name) {
  const item = Module[name];
}
```

error:
```console
error    Unable to validate computed reference to imported namespace 'Module'  import/namespace
```

So I don't know if this is legal by the spec. Babel doesn't object. Perhaps this kind of property access should have its own rule. If it is not permitted by the spec, then it could be an error. If it is legal, then it could be a warning by default? As it is I need to disable `import/namespace` entirely to allow this.

benmosher:
Changed my mind, too much work to make this a separate rule, and it can easily be exposed as an option in v1.x. 😅

I'm not even sure if it should default to `true` in v2...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
-### Modified
+### Added
+- `allowComputed` option for [`namespace`] rule. If set to `true`, won't report
+  computed member references to namespaces. (see [#456])
+
+### Changed
 - Modified [`no-nodejs-modules`] error message to include the module's name ([#453], [#461])
 
 ## [1.12.0] - 2016-07-26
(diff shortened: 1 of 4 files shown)


jfmengels:
LGTM

CHANGELOG.md
```diff
@@ -307,7 +311,11 @@ for info on changes for earlier releases.
 [#157]: https://github.com/benmosher/eslint-plugin-import/pull/157
 [#314]: https://github.com/benmosher/eslint-plugin-import/pull/314
 
+<<<<<<< 30e55b65d6a8a08585aa72c51f8e5871d9a832a6
```
rhys-vdw: Ah, looks like you've commited some merge conflicts, @benmosher.

benmosher: agh! thanks for the tip!

PR #475, opened 2016-08-07
Add webpack 2 support

### LINKED ISSUE #265: Webpack 2 support

taion:
Closes #265

Fixed some lint failures in the plugin along the way.

I don't have test coverage because it seems like it'd be a bunch of additional work to scaffold it into place, because it would require tooling to install dependencies for the test fixtures.

I tested manually and it works on my project, though.

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -128,7 +128,7 @@ function createResolveSync(configPath, webpackConfig) {
 function createWebpack2ResolveSync(webpackRequire, resolveConfig) {
   var EnhancedResolve = webpackRequire('enhanced-resolve')
 
-  return EnhancedResolve.create.sync(assign({}, resolveConfig, webpack2DefaultResolveConfig))
+  return EnhancedResolve.create.sync(assign({}, webpack2DefaultResolveConfig, resolveConfig))
 }
 
 /**
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -180,7 +180,7 @@ function createWebpack1ResolveSync(webpackRequire, resolveConfig, plugins) {
     new ModuleAsFilePlugin('module'),
     new ModuleAsDirectoryPlugin('module'),
     new DirectoryDescriptionFilePlugin(
-      'package.json', ['jsnext:main'].concat(resolveConfig.packageMains || webpack1DefaultMains)
+      'package.json', ['module', 'jsnext:main'].concat(resolveConfig.packageMains || webpack1DefaultMains)
     ),
     new DirectoryDefaultFilePlugin(['index']),
     new FileAppendPlugin(resolveConfig.extensions || ['', '.webpack.js', '.web.js', '.js']),
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -7,6 +7,9 @@ var findRoot = require('find-root')
   , isAbsolute = path.isAbsolute || require('is-absolute')
   , fs = require('fs')
   , coreLibs = require('node-libs-browser')
+  , assign = require('object-assign')
+  , resolve = require('resolve')
+  , semver = require('semver')
 
 var log = require('debug')('eslint-plugin-import:resolver:webpack')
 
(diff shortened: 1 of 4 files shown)


resolvers/webpack/index.js
```diff
@@ -79,75 +82,137 @@ exports.resolve = function (source, file, settings) {
...
+    var webpackFilename = resolve.sync('webpack', { basedir: path.dirname(configPath) })
+    var webpackResolveOpts = { basedir: path.dirname(webpackFilename) }
+
+    webpackRequire = function (id) {
+      return require(resolve.sync(id, webpackResolveOpts))
+    }
+  } catch (e) {
+    // Something has gone wrong (or we're in a test). Use our own bundled
+    // enhanced-resolve.
+    webpackRequire = require
```
taion: Should never hit this in real use given the peer dependency. It's really just for tests.

resolvers/webpack/index.js
```diff
@@ -79,75 +82,137 @@ exports.resolve = function (source, file, settings) {
...
+
+  return createWebpack1ResolveSync(webpackRequire, resolveConfig, webpackConfig.plugins)
+}
+
+function createWebpack2ResolveSync(webpackRequire, resolveConfig) {
+  var EnhancedResolve = webpackRequire('enhanced-resolve')
 
-var SyncNodeJsInputFileSystem = require('enhanced-resolve/lib/SyncNodeJsInputFileSystem')
-var syncFS = new SyncNodeJsInputFileSystem()
+  return EnhancedResolve.create.sync(assign({}, resolveConfig, webpack2DefaultResolveConfig))
```
taion: Yeah... this plus the config defaults below are (I think) all it takes. Neat.

resolvers/webpack/package.json
```diff
@@ -32,10 +32,14 @@
...
-    "node-libs-browser": "^1.0.0"
+    "node-libs-browser": "^1.0.0",
+    "object-assign": "^4.1.0",
+    "resolve": "^1.1.7",
+    "semver": "^5.3.0"
   },
   "peerDependencies": {
-    "eslint-plugin-import": ">=1.4.0"
+    "eslint-plugin-import": ">=1.4.0",
+    "webpack": "^1.11.0 || ^2.1.0-beta || ^2.1.0"
```
taion: 1.11.0 is the earliest webpack version that uses enhanced-resolver 0.9.x, so it seems like a reasonable minimum version.

resolvers/webpack/index.js
```diff
@@ -79,75 +82,137 @@ exports.resolve = function (source, file, settings) {
...
+  mainFields: ['browser', 'module', 'main'],
+}
 
 // adapted from tests &
 // https://github.com/webpack/webpack/blob/v1.13.0/lib/WebpackOptionsApply.js#L322
-function createResolver(resolve, otherPlugins) {
-  var resolver = new Resolver(syncFS)
+function createWebpack1ResolveSync(webpackRequire, resolveConfig, plugins) {
+  var Resolver = webpackRequire('enhanced-resolve/lib/Resolver')
+  var SyncNodeJsInputFileSystem = webpackRequire('enhanced-resolve/lib/SyncNodeJsInputFileSystem')
```
taion: https://github.com/webpack/enhanced-resolve/blob/v0.9.1/lib/SyncNodeJsInputFileSystem.js

This file system object has no caching attached to it AFAICT, so the penalty for constructing it on-the-fly should be trivial.

resolvers/webpack/index.js
guzart: I think the order of `resolveConfig` and `webpack2DefaultResolveConfig` should be reversed, so that the custom resolveConfig overrides the values on webpack2DefaultResolveConfig. Great job!

taion: Oops, you're right.

taion:
This diff looks worse than it is. It really just does:

1. Use the `resolve` trick to make a `webpackRequire` that requires on behalf of webpack
2. Check the version of `enhanced-resolve` to see which branch to take
3. Move the webpack 1 imports into the function for creating the webpack 1 resolver, and use `webpackRequire` there
4. Add the webpack 2 resolver, which is trivial because of the new `ResolverFactory`, plus the default config
5. Shim `resolveSync` to have the same signature between both versions

benmosher:
@taion thanks for the detailed breakdown!

Kovensky:
Trying taion's branch on a small webpack2 project, after removing the `root` and `fallback` compatibility keys from my webpack config's `resolve`:

```shell
$ git clone https://github.com/taion/eslint-plugin-import
...
$ rm -rf node_modules/eslint-import-resolver-webpack
$ ln -s ../eslint-plugin-import/resolvers/webpack node_modules/eslint-import-resolver-webpack
$ $(npm bin)/eslint src | grep no-unresolved | wc -l
      15
$ cd eslint-plugin-import && git checkout webpack-2 && cd ..
Branch webpack-2 set up to track remote branch webpack-2 from origin.
Switched to a new branch 'webpack-2'
$ $(npm bin)/eslint src | grep no-unresolved | wc -l
       0
```

benmosher:
@Kovensky if I'm reading this right, you lost 15 `no-unresolved` reports. 

is that.. good? i.e. were there 15 wrong reports that were fixed, or are 15 correct reports now missing?

Kovensky:
@benmosher there were 15 wrong reports that were fixed. The wrong reports were happening even for relative imports (!) or things that could be found inside a `node_modules` a few dirs up. Using the `webpack-2` branch resolves everything correctly.

benmosher:
Merged and published as v0.5.0 to npm.

@taion, thanks again!

PR #476, opened 2016-08-07
Make options more readable

karlhorky:


jfmengels:
Thanks @karlhorky! I noticed it too, just didn't get to it.
I actually thought about using `-` to make a list, but this is much better already, so LGTM.

karlhorky:
No worries, thanks for the project!

PR #485, opened 2016-08-12
Implement new rule: no-internal-modules

spalger:
[See doc page](https://github.com/spalger/eslint-plugin-import/blob/8a96faf4ee5b66a464e5fe0e411366607c7372fc/docs/rules/no-reaching-inside.md) for details


benmosher:
Ah, sounds cool. Normally I'd like to see an issue with a proposal first, but a PR makes for a very explicit proposal. 😄

Adding `minimatch` is a non-trivial dependency, as other settings/options may leverage some sort of glob evaluator, and they should all use the _same_ one. It's worth discussing what (if any) glob parsers are used for other plugins / ESLint core (i.e. the `.eslintignore` file must use something).

Need to get @jfmengels' thoughts on this, too.

@sindresorhus, I think you may have thoughts an feelings on something like this, too, if you're willing to take the time to evaluate and share. (no worries if not!)


Also, I'm not sure I love `no-reaching-inside` as a name, but I don't immediately have a better one in mind. I'd probably go toward something positive, i.e. `import/shallow` or something. But that is a more half-baked thought. I think it is important the name be as clear as possible, though, as it will live as long as the rule does.

jfmengels:
Thanks @spalger!

> Normally I'd like to see an issue with a proposal first

Same here, but this sounds interesting anyway :)

Not sure about the rule name either. I like the idea of shallow, but starting with no- is more ESLint like.

I do think that ../foo/index should be allowed though.

Not very constructive, but gotta go sorry ^^'

spalger:
> Normally I'd like to see an issue with a proposal first

Sorry :) Happy to see it go in, but not married to the idea if y'all would prefer a different direction entirely.

> I'd probably go toward something positive, i.e. import/shallow or something.

I like the idea, but one thing I don't want to imply is that the rule is missing a `depth` parameter, which is a feeling I get with the name "shallow".  In my opinion the rule is sort of useless if you can reach into a module at all.

> Adding minimatch is a non-trivial dependency

I would love to replace minimatch with another solution. Especially if it allow removing the path-separator-normalization dance. The initial implementation followed the lead of `no-restricted-paths` but it's dependency on `process.cwd()` meant it didn't work with `linter-eslint` in atom.

benmosher:
Could you refactor to use `glob` instead of `minimatch`? Looks like that's [ESLint's globber of choice](https://github.com/eslint/eslint/blob/master/package.json#L46), I'd like to be consistent with them (and then, ideally, the dependency can be shared).

spalger:
minimatch is the matching library used by glob. glob though is about finding files on the file system that match the pattern, not about checking if a known path (the resolved path of an import statement) matches a pattern.

benmosher:
Ah, got it. As you can tell, I'm not very familiar with either. Just reminded me of an old ESLint issue about gitignore vs eslintignore. That makes sense, now.

docs/rules/no-reaching-inside.md
```diff
@@ -0,0 +1,64 @@
+# no-reaching-inside
+
+Use this rule to prevent importing the submodules of other modules.
+
+## Rule Details
+
+This rule has one option, `allow` which is an array of minimatch patterns to identify directories whose children can be imported explicitly.
```
jfmengels: Can you make `minimatch patterns` a link that links to documentation about minimatch patterns?

jfmengels:
This looks pretty good to me. I still think that `../foo/index` and `../foo/index.js` should be allowed though, as that is what is happening most of the time under the hood when you require `../foo`, and is then only a coding style preference.

spalger:
@jfmengels In my eyes, the spirit of the rule is that the consumer of a module shouldn't be concerned with the implementation of the module. This includes where the entry point is.

If the module is using an `index.js` one day, moved to a single file the next, and ended up with a directory containing `package.json` and a custom `main` entry, the consumer shouldn't need to know.

Conveniently, you can add `allow: ["**/index.js","**/index"]` to your eslint config to enable this behavior.

spalger:
@jfmengels I updated the rule so that `allow` now whitelists files that can be imported with reaching. It will first check for an allow pattern that matches the import statement, then an allow pattern that matches the resolved file path, and if either succeeds then the import will pass.

Because of this you could use `allow: ["*/index{.js,}"]` to support importing `../foo/index` and `../foo/index.js` but not `../foo/internal/index.js`.

src/rules/no-reaching-inside.js
```diff
@@ -0,0 +1,100 @@
...
+module.exports = function noReachingInside(context) {
+  const options = context.options[0] || {}
+  const allowRegexps = (options.allow || []).map(p => minimatch.makeRe(p))
+
+  // test if reaching to this destination is allowed
+  function reachingAllowed(importPath) {
+    return !!find(allowRegexps, re => re.test(importPath))
+  }
+
+  // minimatch patterns are expected to use / path seperators, like import
```
jfmengels: nitpick: seperators --> separators :sweat_smile:

src/rules/no-reaching-inside.js
```diff
@@ -0,0 +1,100 @@
...
+        }
+      }, [])
+
+    if (steps.length <= 1) return false
+
+    // before trying to resolve, see if the raw import (with relative
+    // segments resolved) matches an allowed pattern
+    const justSteps = steps.join('/')
+    if (reachingAllowed(justSteps)) return false
+    if (reachingAllowed(`/${justSteps}`)) return false
```
jfmengels: more nitpicking
`if (reachingAllowed(justSteps) || reachingAllowed(`/${justSteps}`)) return false`

src/rules/no-reaching-inside.js
```diff
@@ -0,0 +1,100 @@
...
+
+    // if the import statement doesn't match directly, try to match the
+    // resolved path if the import is resolvable
+    const resolved = resolve(importPath, context)
+    if (!resolved) return false
+    if (reachingAllowed(normalizeSep(resolved))) return false
```
jfmengels: even more nitpicking
`if (!resolved || reachingAllowed(normalizeSep(resolved))) return false`

src/rules/no-reaching-inside.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  function checkImportForReaching(importPath, node) {
+    switch (importType(importPath, context)) {
+      case 'parent':
+      case 'index':
+      case 'sibling':
+      case 'external':
+      case 'internal': {
+        if (isReachViolation(importPath)) {
```
jfmengels: An other way to do this, which I prefer:
```js
const potentialViolationTypes = ['parent', 'index', 'sibling', 'external', 'internal']

function checkImportForReaching(importPath, node) {
  if (potentialViolationTypes.indexOf(importType(importPath, context)) !== -1 &&
    isReachViolation(importPath)
  ) {
    // report
  }
}
```

jfmengels:
Some nitpicks/code style improvements from me, but looks good to me otherwise :)

src/rules/no-reaching-inside.js
spalger: I spell that word wrong everywhere, all the time, I think I need to setup autohotkey just for it.

spalger:
Hey, any updates here?

benmosher:
Sorry, been extra busy lately. 

I'm still stuck on the name. I don't feel like it quickly conveys the purpose and effect of the rule. 

Unfortunately, I also have no ideas what to call it. 😬

jfmengels:
I was thinking of something like `no-module-internals-import`, as that conveys the message of what we want to avoid. Not sure it's a good end name, but maybe a good direction to go in.

spalger:
No worries @benmosher. I like the start @jfmengels. Because of the plugin id being part of the user-facing rule name I'd be happy with `import/no-internal-modules`

jfmengels:
I get the reasoning of not wanting to duplicate the plugin name, but I find `import/no-internal-modules` confusing as it makes me think that you're not allowed to have internal modules.

(PS : Bikeshedding is hard :D)

benmosher:
@jfmengels I think that's a reasonable concern in general, but in this case it's not obvious to me what incorrect behavior that name would imply. 

i.e. what would you expect not to be allowed, given `no-internal-modules` as a name, that _is_ allowed? or vice versa?

This rule is pretty restrictive in that way. Seems like a fair characterization.

jfmengels:
> what would you expect not to be allowed, given no-internal-modules as a name, that is allowed

Well, I wouldn't expect much to be allowed I guess if this rule forbid internal modules altogether :p
But in my mind, that's what the rule name says, even if it doesn't make much sense, and it's a bit hard for me to understand that it forbids `importing` an internal module, even though I know what it does.

You're right that naming and functionality-wise it's very close to `no-nodejs-modules`, but that is less confusing because I can't create core nodejs modules (now that I say it, I hope that it's clear that core modules are meant).

Let's put it this way: When I read the rule name, I read `no internal modules`, not `import no internal modules` (I don't read the plugin name when I read the rule). If you think that most people read it in the second way, then the name is fine by me.

benmosher:
I guess I want to err on the side of being concise, because absolutely _no_ internal modules ever would be a wild world in which to try to code (and lint!).

jfmengels:
Ok, let's go with that then.
@spalger Just want to make sure of something. what happens if I want to import `lodash/fp` or `lodash/fp/get` for instance? Do I need to put it in the allowed imports, or is that allowed by default?

spalger:
You would need to add `lodash/**` or `lodash/*` to `allow:`

spalger:
You would need to add `lodash/**` or `lodash/*` to the allow patterns

benmosher:
Thanks for merging from master! Could you go ahead and rename it to `no-internal-modules`?

spalger:
Whoops, got distracted previously I guess 😄 . Updated!

test-build.js
```diff
@@ -0,0 +1,2 @@
+require('babel-register')
+require('./tests/src/rules/no-internal-modules.js')
```
benmosher: what is this file?

jfmengels:
Apart from @benmosher's [comment](https://github.com/benmosher/eslint-plugin-import/pull/485#discussion_r78439354), this looks good to me. Thanks a lot @spalger! :smile:

test-build.js
spalger: Whoops :)

benmosher:
lol @jfmengels, me too. @spalger: looks great! thanks for working with us!

spalger:
🎉 🎉 🎉 🎉 🎉

jfmengels:
Hey, once again: Thanks a lot @spalger, this is awesome!

PR #489, opened 2016-08-13
Feature/max dependencies

### LINKED ISSUE #486: Rule: max-imports
Curious what you would think about a max-imports rule. Usually if a module has, say, more than 10 imports or so, that is code smell and the module should be split up into smaller pieces. 

Would you accept such a rule as a PR, or should such a rule live somewhere separately? Or does it already exist and I'm not aware? :)

tizmagik:
Closes #486 

I decided on naming it `imports/max-dependencies` since it's a bit more generic (also supports `require()` calls) and less redundant sounding than `imports/max-imports`.

### DIFF
--- a/docs/rules/max-dependencies.md
+++ b/docs/rules/max-dependencies.md
@@ -2,7 +2,7 @@
 
 Forbid modules to have too many dependencies (`import` or `require` statements).
 
-This is a useful rule because a module with too many dependencies is code smell, and usually indicates the module is doing too much and/or should be broken up into smaller modules.
+This is a useful rule because a module with too many dependencies is a code smell, and usually indicates the module is doing too much and/or should be broken up into smaller modules.
 
 Importing multiple named exports from a single module will only count once (e.g. `import {x, y, z} from './foo'` will only count as a single dependency).
 
--- a/docs/rules/max-dependencies.md
+++ b/docs/rules/max-dependencies.md
@@ -1,0 +1,44 @@
+# max-dependencies
+
+Forbid modules to have too many dependencies (`import` or `require` statements).
+
+This is a useful rule because a module with too many dependencies is code smell, and usually indicates the module is doing too much and/or should be broken up into smaller modules.
+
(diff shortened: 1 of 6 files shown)


src/rules/max-dependencies.js
```diff
@@ -0,0 +1,49 @@
+import Set from 'es6-set'
+import isStaticRequire from '../core/staticRequire'
+
+const DEFAULT_MAX = 10
```
tizmagik: I'm not quite sure howe defaults are handled. I wanted it so that if you enable the rule, but don't specify `{max}` option (is that possible?) it would at least default to a sensible 10.

Not sure if that was the right way to go about it... thoughts?

tizmagik:
Any feedback here @benmosher @jfmengels ?

benmosher:
Sorry for the slow feedback, I think we've both been pretty busy this week and last week. I will do my best to make time Tuesday or Thursday of next week, if @jfmengels hasn't beaten me to it (and @jfmengels: don't worry about it if you're busy too).

benmosher:
Need to add a note in the change log, but beyond that, looks good to me! Thanks! Great docs and tests!

@jfmengels, any thoughts/concerns?

docs/rules/max-dependencies.md
```diff
@@ -0,0 +1,44 @@
+# max-dependencies
+
+Forbid modules to have too many dependencies (`import` or `require` statements).
+
+This is a useful rule because a module with too many dependencies is code smell, and usually indicates the module is doing too much and/or should be broken up into smaller modules.
```
jfmengels: is A code smell

jfmengels:
Sorry for blocking this!

LGTM. Just a doc typo to fix, and the PR needs a rebase.

Thanks for this @tizmagik!

tizmagik:
@jfmengels debased and ready

jfmengels:
LGTM! :smile:

PR #500, opened 2016-08-18
Allow no-restricted-paths to restrict from packages

lencioni:
I am working in a project where we have a package installed that has
a "private" directory with some files in it that are not intended to be
part of the public API. It would be nice to be able to restrict these
files from being imported with configuration like:

```json
"import/no-restricted-paths": ["error", {
  "zones": [
    { "target": ".", "from": "my-package/private" },
  ],
}],
```

This commit takes a swing at making this work. I am not entirely sure if
this is the right approach and I was unable to get the specs running the
way I wanted, but I hope this can at least get the ball rolling.

I'm happy to finish this up with a little direction, or feel free to take what
I've started and run with it.

### DIFF
--- a/src/rules/no-restricted-paths.js
+++ b/src/rules/no-restricted-paths.js
@@ -16,22 +16,22 @@ module.exports = function noRestrictedPaths(context) {
   })
 
   function checkForRestrictedImportPath(importPath, node) {
-      const absoluteImportPath = resolve(importPath, context)
+    const absoluteImportPath = resolve(importPath, context)
 
-      if (!absoluteImportPath) {
-        return
+    if (!absoluteImportPath) {
+      return
+    }
+
+    matchingZones.forEach((zone) => {
+      const absoluteFrom = path.resolve(basePath, zone.from)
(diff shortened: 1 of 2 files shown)


ljharb:
I wonder if this might be more appropriate as an option or enhancement to the core rule [no-restricted-imports](http://eslint.org/docs/rules/no-restricted-imports)?

benmosher:
I think I agree with @ljharb, feels like [this line](https://github.com/eslint/eslint/blob/master/lib/rules/no-restricted-imports.js#L42) in the rule he referenced could be enhanced from 

```js
if (restrictedImports.indexOf(value) !== -1) {
```

to roughly

```js
if (restrictedImports.some(ri => ri.startsWith(value))) {
```

without being considered a breaking change.

I don't feel particularly strongly either way, though.

benmosher:
Closing, pending ESLint team's reaction to https://github.com/eslint/eslint/issues/6963. Will reopen (at least for discussion) if they reject it.

PR #503, opened 2016-08-20
support alt-JS dependencies

### LINKED ISSUE #407: parse typescript?
Would be cool if `*.ts` were magically parsed by a (_the_?) TypeScript parser, such that JavaScript files that import from TypeScript could be properly linted.

Use case: transitioning from one to the other in a Webpack project, where loaders allow both to exist.

Possible general config option: `import/parsers` that maps from regexes to parser modules. Non-matches would continue to use the ESLint parser.

Could theoretically lint CoffeeScript or whatever else, assuming some module spit out a roughly ESTree-compatible AST (i.e. it has compatible export nodes in a `Program` `body`).

benmosher:
Closes #407. Allows TypeScript or other alt-JS dependencies to be configured via [`import/parsers` setting](https://github.com/benmosher/eslint-plugin-import/tree/typescript#importparsers).

cc @santiagoaguiar, @denneulin for feedback since you two 👍'd the original issue I created. 😎

### DIFF
--- a/tests/src/core/getExports.js
+++ b/tests/src/core/getExports.js
@@ -83,11 +83,12 @@ describe('getExports', function () {
     var imports = ExportMap.parse(
       path,
       contents,
-      { parserPath: 'babel-eslint' }
+      { parserPath: 'babel-eslint', settings: {} }
     )
 
-    expect(imports).to.exist
-    expect(imports.get('default')).to.exist
+    expect(imports, 'imports').to.exist
+    expect(imports.errors).to.be.empty
+    expect(imports.get('default'), 'default export').to.exist
     expect(imports.has('Bar')).to.be.true
   })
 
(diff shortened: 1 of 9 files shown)


tests/src/core/getExports.js
```diff
@@ -307,4 +310,35 @@ describe('getExports', function () {
 
   })
 
+  context('alternate parsers', function () {
+    const configs = [
+      // ['string form', { 'typescript-eslint-parser': '.ts' }],
+      ['array form', { 'typescript-eslint-parser': ['.ts', '.tsx'] }],
```
jfmengels: What is the reason that the `string form` one is commented out?

benmosher: I started with just one string, then made it support both a string and an array.

Then I decided I was making it too complicated.

Do you think it would be better to support both?

PR #499, opened 2016-08-16
Fix `namespace` rule runtime crash

wKich:

PR #504, opened 2016-08-20
release v1.14.0

### LINKED ISSUE #416: Cannot read property 'namespace' of undefined
TypeError: Cannot read property 'namespace' of undefined from `eslint-plugin-import/lib/rules/namespace.js`

Getting this error and not exactly sure why. Thanks.

benmosher:
cc @jfmengels. I want to get `import/parsers` published so I can use it at work.

Also I clearly forgot to do this for 1.13.0 😳, only changes are the bugfix for #416 (PR #499) and `import/parsers` via #503. I have QA'd.

**PR summary:**
- #499: fix #416 
- #503: add `import/parsers` to support ES-module-compliant alt-JS interop

It'd probably be good if you took a look at the `import/parsers` config schema. Though v2 is on the horizon so we could break it if needed/desired as part of that release... only thing I'm thinking is whether it should be glob- or regex-based instead of array-of-extensions. I like AoE mostly for consistency with `import/extensions`, and it's easy to grok. Con: less powerful, but could be enhanced to support glob/regex later if desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,7 +91,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - Added an `optionalDependencies` option to [`no-extraneous-dependencies`] to allow/forbid optional dependencies ([#266], thanks [@jfmengels]).
 - Added `newlines-between` option to [`order`] rule ([#298], thanks [@singles])
 - add [`no-mutable-exports`] rule ([#290], thanks [@josh])
-- [`import/extensions` setting]: a whitelist of file extensions to parse as modules
+- [`import/extensions` setting]: a list of file extensions to parse as modules
   and search for `export`s. If unspecified, all extensions are considered valid (for now).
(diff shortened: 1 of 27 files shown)


jfmengels:
> whether it should be glob- or regex-based instead of array-of-extensions

You should probably create an issue to discuss this, so that people interested by this participate and give their meaning.
I like the glob idea, as long as you allow an array of globs (`["src/**/*.ts", "!src/foo/**/*.ts"]`), as it gives you more control over how to control this setting. But before doing that, I would like to know of at least one use case that would benefit from a regex/glob configuration, because I can't seem to think of one.

I've left you a comment here https://github.com/benmosher/eslint-plugin-import/pull/503/files#r75605034, but LGTM otherwise.

benmosher:
> But before doing that, I would like to know of at least one use case that would benefit from a regex/glob configuration, because I can't seem to think of one.

Yeah, this is where I ended up as well, thinking through it. Globs are great, and very expressive, but I feel like >95% of the time, one or two file extensions will be totally sufficient.

PR #508, opened 2016-08-22
Fix false positive with default import having a "default" property

### LINKED ISSUE #507: no-named-as-default-member false positive, with default export with 'default' property
The following produces the incorrect error;

```js
//bar.js
export default { 
  default: 'baz'
};
```

```js
//foo.js
import bar from './bar';

const qux = bar.default;
```
It's incorrectly noting that there is a named export named "default", following the suggestion would lead to this: `import bar, { default } from './bar` which is clearly not right, and I'd suggest not ever what you want to do in this situation. Perhaps that if the property is from the default export, special case allowing 'default' as a property of that export? 

**edit**: actually that would hold true for any named export, with a property of the same name as the export on it.

jquense:
Fixes #507

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## [Unreleased]
 
+### Fixed
+- [`no-named-as-default-member`] Allow default import to have a property named "default" ([#507])
+
 
 ## [1.14.0] - 2016-08-22
 ### Added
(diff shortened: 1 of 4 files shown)


jfmengels:
LGTM, thanks @jquense :)

benmosher:
Indeed, thanks much for a high-quality PR! 😎

jfmengels:
Only thing missing was adding thanks to yourself in the changelog, but we'll do that next time we update it ;)
(and also link to this PR in the changelog)
Thanks again!

benmosher:
@jfmengels I got it in the rebase: https://github.com/benmosher/eslint-plugin-import/commit/0d98253ea3bc1c5998223cab7b1a095ba943b210#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR11 😁

Can't really blame folks for not wanting to thank themselves, I suppose... 😄

jfmengels:
Truthfylly, I think it's odd to ask someone to thank themselves or add themselves to a contributors list. It's better done by the maintainers to show that your work is really appreciated :)

Btw, do you want to add a link to this PR in the changelog?

jquense:
thanks y'all!

benmosher:
@jfmengels maybe we should start doing the changelog update in the merge commit instead of the PR. that would be less merge-conflict-prone, too.

PR #509, opened 2016-08-22
Add `allow` option to `no-nodejs-modules` rule (fixes #452)

### LINKED ISSUE #452: Add `allow` option to `no-nodejs-modules`
The `no-nodejs-modules` rule is very useful, but some modules are fine to import - like `events` - due to a good browserify shim.

I'd like the ability to add exceptions to this rule.

jfmengels:
Add `allow` option to `no-nodejs-modules` rule (fixes #452)

cc @ljharb

### DIFF
--- a/src/rules/no-nodejs-modules.js
+++ b/src/rules/no-nodejs-modules.js
@@ -1,20 +1,23 @@
 import importType from '../core/importType'
 import isStaticRequire from '../core/staticRequire'
 
-function reportIfMissing(context, node, name) {
-  if (importType(name, context) === 'builtin') {
+function reportIfMissing(context, node, allowed, name) {
+  if (!allowed.includes(name) && importType(name, context) === 'builtin') {
     context.report(node, 'Do not import Node.js builtin module "' + name + '"')
   }
 }
 
 module.exports = function (context) {
+  const options = context.options[0] || {}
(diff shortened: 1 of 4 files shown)


src/rules/no-nodejs-modules.js
```diff
@@ -2,7 +2,7 @@ import importType from '../core/importType'
 import isStaticRequire from '../core/staticRequire'
 
 function reportIfMissing(context, node, allowed, name) {
-  if (!allowed.includes(name) && importType(name, context) === 'builtin') {
+  if (allowed.indexOf(name) === -1 && importType(name, context) === 'builtin') {
```
benmosher: ah, bummer, I also thought Node v4 had `includes`.

jfmengels: Not that surprising actually, since it's not even ES2015, but an ES2016 feature. I thought Babel would actually take care of this, but I guess was too hopeful (or it's just something to add). Anyway `indexOf` is fine.

benmosher:
I was thinking presets (like @graingert suggested) would be a good idea, but that can be a responsibility of shared configs. LGTM. 😎

jfmengels:
2pm-wooyoung-relevant-11LMswao05npcs

PR #527, opened 2016-08-27
Allow filename patterns for rule no-extraneous-dependencies.

### LINKED ISSUE #470: could `import/no-extraneous-dependencies` get a path regex for dev deps?
Currently, the airbnb config only allows deps, and that means that everyone has to manually do something like https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/test/.eslintrc#L8-L10 for their tests.

It would be ideal if in the shared config, we could specify that folders named "test" or "spec", for example, could have dev deps, but nothing else. I think that would cover the majority use case and remove the need for most devs to have to override this rule.

knpwrs:
Following the suggestion in https://github.com/benmosher/eslint-plugin-import/issues/470#issuecomment-237035169.

Fixes #470.

### DIFF
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -25,25 +25,13 @@ You can also use globs instead of literal booleans:
 "import/no-extraneous-dependencies": ["error", {"devDependencies": "*.test.js"}]
 ```
 
-You can also use regular expressions instead of literal booleans or globs (note: the string must begin and end with `/`):
+When using a glob, the setting will be activated if the name of the file being linted matches the given glob. In addition, you can use an array to specify multiple globs:
 
 ```js
-"import/no-extraneous-dependencies": ["error", {"devDependencies": "/test|spec/"}]
+"import/no-extraneous-dependencies": ["error", {"devDependencies": ['*.test.js', '*.spec.js']}]
 ```
 
-If you are using JavaScript configuration (e.g., `.eslintrc.js`), then you can use a RegExp literal instead of a string:
-
-```js
-"import/no-extraneous-dependencies": ["error", {"devDependencies": /test|spec/}]
-```
-
-When using a glob or regular expression, the setting will be activated if the name of the file being linted matches the given glob or regular expression. In addition, you can use an array to specify multiple globs or regular expressions:
-
-```js
-"import/no-extraneous-dependencies": ["error", {"devDependencies": [/test/, '/spec/', '*.test.js', '*.spec.js']}]
-```
-
-When using an array of globs or regular expressions, the setting will be activated if the name of the file being linted matches a single glob or regular expression in the array.
+When using an array of globs, the setting will be activated if the name of the file being linted matches a single glob in the array.
 
 ## Rule Details
 
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -19,13 +19,7 @@ You can set the options like this:
 "import/no-extraneous-dependencies": ["error", {"devDependencies": false, "optionalDependencies": false, "peerDependencies": false}]
 ```
 
-You can also use globs instead of literal booleans:
-
-```js
-"import/no-extraneous-dependencies": ["error", {"devDependencies": "*.test.js"}]
-```
-
-When using a glob, the setting will be activated if the name of the file being linted matches the given glob. In addition, you can use an array to specify multiple globs:
+You can also use an array of globs instead of literal booleans:
 
 ```js
 "import/no-extraneous-dependencies": ["error", {"devDependencies": ['*.test.js', '*.spec.js']}]
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -31,7 +31,7 @@ If you are using JavaScript configuration (e.g., `.eslintrc.js`), then you can u
 "import/no-extraneous-dependencies": ["error", {"devDependencies": /test|spec/}]
 ```
 
-When using a regular expression the result of running [`test`] against the name of the file being linted is used as the boolean value. For example, the above configurations will allow the import of `devDependencies` in files whose names include `test` or `spec`.
+When using a regular expression, the setting will be activated if the name of the file being linted matches the given regular expression. For example, the above configurations will allow the import of `devDependencies` in files whose names include `test` or `spec`.
 
 ## Rule Details
 
@@ -99,5 +99,3 @@ import react from 'react';
 ## When Not To Use It
 
 If you do not have a `package.json` file in your project.
-
-[`test`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test "RegExp.prototype.test"
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -19,10 +19,16 @@ You can set the options like this:
 "import/no-extraneous-dependencies": ["error", {"devDependencies": false, "optionalDependencies": false, "peerDependencies": false}]
 ```
 
-You can also use regular expressions instead of literal booleans:
+You can also use globs instead of literal booleans:
 
 ```js
-"import/no-extraneous-dependencies": ["error", {"devDependencies": "test|spec"}]
+"import/no-extraneous-dependencies": ["error", {"devDependencies": "*.test.js"}]
+```
+
+You can also use regular expressions instead of literal booleans or globs (note: the string must begin and end with `/`):
+
+```js
+"import/no-extraneous-dependencies": ["error", {"devDependencies": "/test|spec/"}]
 ```
 
 If you are using JavaScript configuration (e.g., `.eslintrc.js`), then you can use a RegExp literal instead of a string:
@@ -31,7 +37,13 @@ If you are using JavaScript configuration (e.g., `.eslintrc.js`), then you can u
 "import/no-extraneous-dependencies": ["error", {"devDependencies": /test|spec/}]
 ```
 
-When using a regular expression, the setting will be activated if the name of the file being linted matches the given regular expression. For example, the above configurations will allow the import of `devDependencies` in files whose names include `test` or `spec`.
+When using a glob or regular expression, the setting will be activated if the name of the file being linted matches the given glob or regular expression. In addition, you can use an array to specify multiple globs or regular expressions:
+
+```js
+"import/no-extraneous-dependencies": ["error", {"devDependencies": [/test/, '/spec/', '*.test.js', '*.spec.js']}]
+```
+
+When using an array of globs or regular expressions, the setting will be activated if the name of the file being linted matches a single glob or regular expression in the array.
 
 ## Rule Details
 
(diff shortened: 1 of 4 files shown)


docs/rules/no-extraneous-dependencies.md
```diff
@@ -19,6 +19,19 @@ You can set the options like this:
...
+"import/no-extraneous-dependencies": ["error", {"devDependencies": "test|spec"}]
+```
+
+If you are using JavaScript configuration (e.g., `.eslintrc.js`), then you can use a RegExp literal instead of a string:
+
+```js
+"import/no-extraneous-dependencies": ["error", {"devDependencies": /test|spec/}]
+```
+
+When using a regular expression the result of running [`test`] against the name of the file being linted is used as the boolean value. For example, the above configurations will allow the import of `devDependencies` in files whose names include `test` or `spec`.
```
jfmengels: `When using a regular expression` --> `When using a regular expression,` (+ comma)

`the result of running [test] against the name of the file being linted is used as the boolean value.` --> `the setting will be activated if the name of the file being linted matches the given regular expression.`

jfmengels:
Hi @knpwrs! Thanks a lot for this, it looks great!

I'm asking myself now whether a glob configuration would not more appropriate (as we're adding `minimatch` in some other PR anyway), as I think it would give more control, and be easier to write than regex when things get a bit complicated. What do you think @benmosher @ljharb? If you're fine with only a regex (or waiting until requests come in and want to eventually support both regex and glob), then that's fine with me.

knpwrs:
Yeah, I saw the discussion about glob vs regular expression but didn't see if it went anywhere. I could see it going either way. I've used globs for everything filename related until webpack came along.

Might be overkill, but maybe it would be worth supporting both? Would it be safe to assume that a string bookended by `/` and `/` is a regular expression? Essentially:

* `'foo'`: glob (via string)
* `'/foo/'`: regex (via string)
* `/foo/`: regex (literal, only from `js` configuration)
* `true` or `false`: boolean

src/rules/no-extraneous-dependencies.js
```diff
@@ -102,9 +111,9 @@ module.exports.schema = [
   {
     'type': 'object',
     'properties': {
-      'devDependencies': { 'type': 'boolean' },
-      'optionalDependencies': { 'type': 'boolean' },
-      'peerDependencies': { 'type': 'boolean' },
+      'devDependencies': { 'type': ['boolean', 'string', 'object'] },
+      'optionalDependencies': { 'type': ['boolean', 'string', 'object'] },
+      'peerDependencies': { 'type': ['boolean', 'string', 'object'] },
```
jfmengels: You're allowing `object`, because a regular expression value is an object, right? Just want confirmation.

knpwrs: Yeah, AFAIK the schema is checked with `typeof`, and `typeof /foo/ === 'object'`.

jfmengels:
I like globs as you can decompose your allowance rules `['**/test/**.js', '*/*.test.js', '!**/fixtures/**']`. Since you can do anything a regex does using this, I'd prefer this option. And having both options sounds like overkill, but why not, doesn't cost anything. But you're right that there may be a conflict in figuring out whether `"test"` is a glob or a regex...

knpwrs:
For two of those examples (`**/test/**.js` and `!**/fixtures/**`) I believe the current solution of a directory-level `.eslintrc{,.yml,.js}` is sufficient, no? The purpose this serves is for when test files live next to implementation files.

jfmengels:
True, didn't think them through much ^^
But maybe you have a lot of directories in test/, each with their fixtures (or something else) subdirectory, in which the devdependency rule should be deactivated.
```js
test/
  A/
    a.js
    a2.js
    fixtures/
      no-dev-deps.js
  B/
    b.js
    b2.js
    fixtures/
      no-dev-deps.js
```
If you have a lot of them, it gets tedious to add a `.eslintrc` file in each `fixtures/` folder. You'd probably rather do that with `['**/test/**.js', '*/*.test.js', '!**/test/*/fixtures/**']`.

Anyway, you'd make a clearer selection using globs than regexes IMO.

knpwrs:
Kaboom. No kill like overkill! [Looks like everything new is covered.](https://coveralls.io/builds/7644730/source?filename=src%2Frules%2Fno-extraneous-dependencies.js)

docs/rules/no-extraneous-dependencies.md
```diff
@@ -31,7 +37,13 @@ If you are using JavaScript configuration (e.g., `.eslintrc.js`), then you can u
...
 ```
 
-When using a regular expression, the setting will be activated if the name of the file being linted matches the given regular expression. For example, the above configurations will allow the import of `devDependencies` in files whose names include `test` or `spec`.
+When using a glob or regular expression, the setting will be activated if the name of the file being linted matches the given glob or regular expression. In addition, you can use an array to specify multiple globs or regular expressions:
+
+```js
+"import/no-extraneous-dependencies": ["error", {"devDependencies": [/test/, '/spec/', '*.test.js', '*.spec.js']}]
+```
+
+When using an array of globs or regular expressions, the setting will be activated if the name of the file being linted matches a single glob or regular expression in the array.
```
jfmengels: > matches a single glob or regular expression in the array

Hmm... Not sure this part works. Globs as an array allows you to specify multiple things `['test/**/*.js', 'src/**/*.js']`, but also to forbid elements allowed by previous "expressions": `['test/**/*.js', 'src/**/*.js', '!fixtures']`. Here `<root>/test/fixtures/a.js` should not work, whereas from your docs (and guessing implementation), it would.

(This is a glob right? Or am I confusing terms?)

knpwrs: Dang it. Back to the drawing board.

Maybe we should only support arrays for globs? But then what is excluded? Just things that contain an exclamation mark? In the case of `['*.test.js', '*.spec.js']`, I'd want that to be additive. In the case of `['*.test.js', '!**/fixtures/**']`, I'd want that to be subtractive.

This is approaching a point of complexity that may not be best for this plugin to handle. In your multiple test and fixture directory example we're already approaching a status of, "really, now?" At least IMHO.

A thought occurs. Maybe allow a function for configuration? Pass the name to the function and expect a boolean back. This plugin could support the most basic use cases and then people who need more complex functionality can implement it themselves.

Re: your closing question... Is what a glob? The string containing `<root>`? Not sure.

knpwrs: Also worth noting that passing a function obviously requires a JavaScript configuration file. I think that's a reasonable requirement if somebody requires super-specific rules.

jfmengels: > ['*.test.js', '*.spec.js'] I'd want that to be additive. In the case of ['*.test.js', '!**/fixtures/**'], I'd want that to be subtractive.

That is the case. Imagine

```js
[
  '**/*.test.js', // All the file ending with '.test.js'
  'test/**/*.js', // + All the files in test/
  '!**/fixtures/**' // Take all the files selected by the previous two selectors,
                    // and remove those in a fixtures (sub-)folder
]
```

> This is approaching a point of complexity that may not be best for this plugin to handle.

I'm not sure I agree, but that's a subjective point a view ;)

> Maybe allow a function for configuration

The function is doable, but then we're pushing the responsibility of doing the glob/regex logic to the user. I'd use functions as a last last resort.

ljharb: These aren't ignore files - I don't think they need to be subtractive. There's nothing wrong with requiring people to nest `.eslintrc` files for things that can easily be achieved that way, but that would otherwise complicate the config.

benmosher:
> having both options sounds like overkill, but why not, doesn't cost anything

I think we've found the cost of both is in the design discussion. 😅

I think it should be one or the other. There is precedent for both at this point (pending the PR @jfmengels referenced).

My gut says it should be globs, going forward, whenever we're defining configuration for sets of files. My suspicion is that one or two globs will be sufficient in 99% of the cases where this is used, i.e. `[ "**/*.spec.js" ]`.

So, I propose that it be an array of globs, handled by `minimatch`, until such time as it can be shown that there is a prevalent use case that is solved well and simply by regex and not globs.

knpwrs:
So let's say we drop regex, how do we want to handle the additive / subtractive problem with globs?

jfmengels:
@knpwrs Sorry, what problem?

knpwrs:
I'm just currently thinking in the mindset of how I implemented the array functionality using `Array.prototype.some`. If a single pattern matches, then the whole test passes. Simple. With globs it has to match any given pattern in the array but then if there are any patterns with exclamation marks then it has to be checked against those as well and negated. We can't use `Array.prototype.every` because a given file shouldn't be expected to match every pattern. Right now my initial thought is to separate the patterns into positive and negative patterns and then test `positive.some(/* filename test */) && !negative.some(/* filename test */)`. But maybe I'm overthinking it.

ljharb:
That all seems overcomplicated. If you want to exempt a path, either use a nested `.eslintrc`, or write your "include" glob more precisely.

knpwrs:
Unless there's something about glob composition I'm missing.

knpwrs:
Right, okay, so that's going to a point I was making at https://github.com/benmosher/eslint-plugin-import/pull/527#discussion_r76541951. So is the current idea just to roll this back to nothing but matching `some` glob pattern?

ljharb:
That's what makes the most sense to me.

knpwrs:
Cool, I'll knock it out when I get home. :)

knpwrs:
ɯooqɐʞ

docs/rules/no-extraneous-dependencies.md
```diff
@@ -19,6 +19,19 @@ You can set the options like this:
 "import/no-extraneous-dependencies": ["error", {"devDependencies": false, "optionalDependencies": false, "peerDependencies": false}]
 ```
 
+You can also use globs instead of literal booleans:
+
+```js
+"import/no-extraneous-dependencies": ["error", {"devDependencies": "*.test.js"}]
```
ljharb: it seems like it'd be simpler to just require this to always be an array - eslint can enforce it, and that way the user would get an immediate indication that multiple globs were possible.

src/rules/no-extraneous-dependencies.js
```diff
@@ -71,18 +72,28 @@ function reportIfMissing(context, deps, depsOptions, node, name) {
   context.report(node, missingErrorMessage(packageName))
 }
 
+function testConfig(config, filename) {
+  // Simplest configuration first
+  if (typeof config === 'boolean' || typeof config === 'undefined') {
+    return config
+  }
+  // Account for the possibility of an array for multiple configuration
+  return [].concat(config).some(c =>  minimatch(filename, c))
```
ljharb: two spaces before `minimatch`?

knpwrs: Still getting used to vim. 😵

docs/rules/no-extraneous-dependencies.md
knpwrs: Sure, I'll go ahead and do that.

knpwrs:
How's this looking now?

jfmengels:
LGTM :smile:

benmosher:
:shipit:

tristanbbq:
@benmosher 
My settings:
```"import/no-extraneous-dependencies": ["error", {"devDependencies": ['*.test.js', '*.spec.js']}]```

Filename: `SOMETHING.test.js` - still get this 'should be listed in the projects dependencies, not devDependencies' - something I'm missing here??

benmosher:
if it's deep, might need to be `['**/*.test.js']`, otherwise it might only match if the file is in the CWD. not positive, though

benmosher:
if that works, let me know--we should update the example in the docs to that

tristanbbq:
@benmosher works 👍  - can confirm that blob `'**/*.test.js'` works but not `'*.text.js'`

PR #533, opened 2016-08-29
Support webpack configs that export a function

grahamb:
Webpack 2 supports webpack configuration files that export a function (see https://webpack.github.io/docs/roadmap.html). This commit adds support for this by testing if the `require`ed config is a function, and if so, executing it.


benmosher:
Published with Webpack resolver v0.6.0 😎

thanks for this!

Issue #238, opened 2016-04-04
webpack doesn't resolve relative to `resolve.root`


seekshiva:
webpack lets us define a value for `resolve.root`, and we can import modules relative to the root directly.

Eg:

webpack.config.js
```
resolve: {
  root: path.resolve('./jsx')
}
```

/jsx/components/Button.jsx
```jsx
import React, { Component } from 'react'
export default class Button extends Component {
  render() {
    return <button {...this.props} />
  }
}
```

/jsx/Some/Deep/Nested/Dir/Component.jsx
```jsx
import '../../../../components/Button'
```

can be written as

/jsx/Some/Deep/Nested/Dir/Component.jsx
```jsx
import 'components/Button'
```
-----

Since webpack config has defined a value for `resolve.root`, I can directly import from 'components/Button' rather than from '../../../../components/Button'. This syntax is super helpful when dealing with deep nested directory structures. It would be great if the webpack resolver package could support it.

Thanks!

benmosher:
This situation is actually why I wrote the Webpack resolver. It works for me. 

Can you provide more details about what exactly isn't working? Or a toy test case that fails? Some info that might be helpful:

- Webpack config file location
- your eslintrc
- the CWD where you ran ESLint from
- the exact error you got that you didn't expect (or the error you didn't get that you did expect)

(BTW: if you're using Sublime, see the end of the README.)

benmosher:
For reference, here are the existing resolve.root tests: https://github.com/benmosher/eslint-plugin-import/blob/master/resolvers/webpack/test/root.js

If you could write a failing test, that would be fantastic. 😄

benmosher:
closing for now, if you can provide a failing test case we can reopen?

Chyzwar:
ok, I have similar problem, 

First I tried to add in .eslintrc
```
    "settings": {
      "import/resolver": {
        "webpack": {
          "config": "./scripts/configs/serverDev.js"
        }
      }
    },
```
and in my serverDev.js I have:
```  
resolve: {
    extensions: ['', '.js', '.jsx'],
    root: path.join(__dirname, '../../src'),
  },
```


Since first approach was still producing lint errors I tried:
```
    "settings": {
      "resolve.root": "src"
    },
```

benmosher:
Thanks for the test repo, will make identifying the issue a lot simpler. I'll take a look!

ravasthi:
[...]

PR #535, opened 2016-08-29
Suggested change to readme

danielchangNYC:
Took a while to hunt down why my config wasn't working and figured out that I needed to add this dependency only after reading through [#238](https://github.com/benmosher/eslint-plugin-import/issues/238) and noticing [this line in a linked example](https://gist.github.com/ravasthi/abcfee465411fc45a8bc28decb9d8e5e#file-package-json-L24).

I think adding this line may add some clarity and save time for several other webpack / eslint noobs (like me 😄 ).


benmosher:
Yeah, this has come up a few times in different issues. I like the way you've addressed it, thanks! 😄

PR #538, opened 2016-08-30
Add `no-absolute-path` rule (fixes #530)

### LINKED ISSUE #530: New rule: no-absolute-path
Based off [this comment](https://github.com/benmosher/eslint-plugin-import/issues/512#issuecomment-241761338)

This rule should forbid the use of imports with an absolute path. This is valid in Node.js, but should not appear in code in npm packages or in production, or better: never.

### Invalid

```js
import Foo from '/some/absolute/path.js';
```

### Valid

```js
import Foo from './some/relative/path.js';
import Foo from 'foo';
```

jfmengels:
Add `no-absolute-path` rule (fixes #530)

### DIFF
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -8,7 +8,7 @@ import { testContext } from '../utils'
 describe('importType(name)', function () {
   const context = testContext()
 
-  it("should return 'absolute' for node.js modules", function() {
+  it("should return 'absolute' for paths starting with a /", function() {
     expect(importType('/', context)).to.equal('absolute')
     expect(importType('/path', context)).to.equal('absolute')
     expect(importType('/some/path', context)).to.equal('absolute')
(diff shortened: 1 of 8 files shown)


tests/src/core/importType.js
```diff
@@ -8,6 +8,12 @@ import { testContext } from '../utils'
 describe('importType(name)', function () {
   const context = testContext()
 
+  it("should return 'absolute' for node.js modules", function() {
```
benmosher: What does 'node.js' mean here?

benmosher:
LGTM! :shipit:

tests/src/core/importType.js
jfmengels: That means typo :sweat_smile:

jfmengels: Fixed

PR #543, opened 2016-09-05
no-extraneous-dependencies allow array of globs option

ramasilveyra:
`no-extraneous-dependencies` rule options `devDependencies`, `peerDependencies` and `optionalDependencies` should allow boolean and array values.

See docs: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md

### DIFF
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -113,9 +113,15 @@ module.exports.schema = [
   {
     'type': 'object',
     'properties': {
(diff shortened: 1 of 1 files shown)


ramasilveyra:
Just saw that was added under the unreleased version :)

juliusosokinas:
Great, will it be released with the next version? Exactly what I wanted to use today.

benmosher:
@juliusosokinas should be, just haven't had a chance to publish

PR #555, opened 2016-09-12
[New] `extensions`: accept both a string, and an object to override it.

### LINKED ISSUE #390: `extensions`: how to specify a default while specifying granular extensions?
I'd like the default to be "always" but i'd like `js` and `jsx` to be "never". How do I specify this? The schema only allows a string, or an object, but not both.

ljharb:
Fixes #390.

### DIFF
--- a/docs/rules/extensions.md
+++ b/docs/rules/extensions.md
@@ -6,9 +6,11 @@ In order to provide a consistent use of file extensions across your code base, t
 
 ## Rule Details
 
-This rule has one option which could be either a string or an object. If it is `"never"` (the default value) the rule forbids the use for any extension. If `"always"` then the rule enforces the use of extensions for all import statements.
+This rule either takes one string option, one object option, or a string and an object option. If it is the string `"never"` (the default value), then the rule forbids the use for any extension. If it is the string `"always"`, then the rule enforces the use of extensions for all import statements.
 
 By providing an object you can configure each extension separately, so for example `{ "js": "always", "json": "never" }` would always enforce the use of the `.js` extension but never allow the use of the `.json` extension.
+
+By providing both a string and an object, the string will set the default setting for all extensions, and the object can be used to set granular overrides for specific extensions. For example, `[<enabled>, "never", { "svg": "always" }]` would require that all extensions are omitted, except for "svg".
 
 ### Exception
 
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -10,6 +10,7 @@ var findRoot = require('find-root')
   , assign = require('object-assign')
   , resolve = require('resolve')
   , semver = require('semver')
+  , has = require('has')
 
 var log = require('debug')('eslint-plugin-import:resolver:webpack')
 
@@ -265,7 +266,7 @@ function findExternal(source, externals, context) {
 
   // else, vanilla object
   for (var key in externals) {
-    if (!externals.hasOwnProperty(key)) continue
+    if (!has(externals, key)) continue
     if (source === key) return true
   }
   return false
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -1,22 +1,27 @@
 import path from 'path'
 import endsWith from 'lodash.endswith'
+import has from 'has'
+import assign from 'object-assign'
 
 import resolve from '../core/resolve'
 import { isBuiltIn } from '../core/importType'
 
-const has = Object.prototype.hasOwnProperty;
-
 module.exports = function (context) {
   const configuration = context.options[0] || 'never'
   const defaultConfig = typeof configuration === 'string' ? configuration : null
-  const modifiers = typeof configuration === 'object' ? configuration : context.options[1] || {}
+  const modifiers = assign(
+    {},
+    typeof configuration === 'object' ? configuration : context.options[1]
+  )
 
   function isUseOfExtensionRequired(extension) {
-    return (has.call(modifiers, extension) || defaultConfig) === 'always'
+    if (!has(modifiers, extension)) { modifiers[extension] = defaultConfig }
+    return modifiers[extension] === 'always'
   }
 
   function isUseOfExtensionForbidden(extension) {
-    return (has.call(modifiers, extension) || defaultConfig) === 'never'
+    if (!has(modifiers, extension)) { modifiers[extension] = defaultConfig }
+    return modifiers[extension] === 'never'
   }
 
   function isResolvableWithoutExtension(file) {
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -4,15 +4,19 @@ import endsWith from 'lodash.endswith'
 import resolve from '../core/resolve'
 import { isBuiltIn } from '../core/importType'
 
+const has = Object.prototype.hasOwnProperty;
+
 module.exports = function (context) {
   const configuration = context.options[0] || 'never'
+  const defaultConfig = typeof configuration === 'string' ? configuration : null
+  const modifiers = typeof configuration === 'object' ? configuration : context.options[1] || {}
 
-  function isUseOfExtensionEnforced(extension) {
-    if (typeof configuration === 'object') {
-      return configuration[extension] === 'always'
-    }
+  function isUseOfExtensionRequired(extension) {
+    return (has.call(modifiers, extension) || defaultConfig) === 'always'
+  }
 
-    return configuration === 'always'
+  function isUseOfExtensionForbidden(extension) {
+    return (has.call(modifiers, extension) || defaultConfig) === 'never'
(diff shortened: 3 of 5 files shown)


ljharb:
@benmosher Now that 1.15 is out, is this ready to be merged?

benmosher:
will review ASAP, likely tomorrow morning 😅

docs/rules/extensions.md
```diff
@@ -6,10 +6,12 @@ In order to provide a consistent use of file extensions across your code base, t
 
 ## Rule Details
 
-This rule has one option which could be either a string or an object. If it is `"never"` (the default value) the rule forbids the use for any extension. If `"always"` then the rule enforces the use of extensions for all import statements.
+This rule has one option which could be either a string, an object, or both. If it is `"never"` (the default value) the rule forbids the use for any extension. If `"always"` then the rule enforces the use of extensions for all import statements.
```
jfmengels: It's not `one option` anymore, is it? It's either one, which is either a string or an object, or two, with first a string then an object. Or am I misunderstanding something?

tests/src/rules/extensions.js
```diff
@@ -33,6 +33,16 @@ ruleTester.run('extensions', rule, {
...
+    test({
+      code: [
+        'import lib from "./bar"',
+        'import lib from "./bar.json"',
+        'import lib from "./bar.hbs"',
+      ].join('\n'),
+      options: [ 'always', { js: 'never', jsx: 'never' } ],
+      settings: { 'import/resolve': { 'extensions': [ '.js', '.jsx', '.json', '.hbs' ] } },
+    }),
+
```
jfmengels: Maybe add a test case with never, for completeness' sake.
I'd also like to see both valid and invalid test cases with this change, if it's not too much to ask.

benmosher:
Ah, fair point @jfmengels. Could probably use the added test case as a starting point.

jfmengels:
That said, this looks pretty good and I like this way of configuring the rule :+1:

ljharb:
@benmosher friendly ping

ljharb:
@jfmengels thanks for the feedback; it helped me find some bugs. Updated!

docs/rules/extensions.md
```diff
@@ -6,10 +6,12 @@ In order to provide a consistent use of file extensions across your code base, t
 
 ## Rule Details
 
-This rule has one option which could be either a string or an object. If it is `"never"` (the default value) the rule forbids the use for any extension. If `"always"` then the rule enforces the use of extensions for all import statements.
+This rule either takes one string option, one object option, or a string and an object option. If it is the string `"never"` (the default value) the rule forbids the use for any extension. If the string `"always"` then the rule enforces the use of extensions for all import statements.
```
jfmengels: If it is the string `"never"` (the default value)
-->
`If it is the string `"never"` (the default value), then
(for consistency)

If the string `"always"` then the rule
-->
If it is the string `"always"` then the rule

src/rules/extensions.js
```diff
@@ -37,15 +39,15 @@ module.exports = function (context) {
     const extension = path.extname(resolvedPath || importPath).substring(1)
 
     if (!extension || !endsWith(importPath, extension)) {
-      if (isUseOfExtensionEnforced(extension)) {
+      if (isUseOfExtensionRequired(extension) && !isUseOfExtensionForbidden(extension)) {
```
jfmengels: You could compute the value for the extension `modifiers[extension] || defaultConfig` once, instead of recomputing it several times (nitpick)

docs/rules/extensions.md
ljharb: alrighty

src/rules/extensions.js
ljharb: Do you just mean, cache it per-extension?

ljharb: Sure, can-do.

jfmengels: Hmm, not cache per-se, though I don't see a reason why not. I was thinking of computing the value of it before line 41, as you will use the result later

ljharb: oh, that's true. let me do that.

ljharb: actually no, i think this is better, since these functions take `extension` as an argument.

ljharb: I'm confused by that - dependencies are always better. In this case it's ensuring that `hasOwnProperty` can't be broken.

Removing the dependency will effectively mean inlining the code - what's the benefit except reduced deducing in people's `node_modules`?

ljharb: This change may not even be worth it overall tho - the tests are failing in Travis because the object is not extensible.

jfmengels: What are the reasons that `externals.hasOwnProperty(key)` could break? Is that just when externals is nil?
If you feel like it's better to have it, keep it, I'm not that against it, just didn't see the value if it was just to check if `externals[key]` exists.

ljharb: so, the reason bracket access and `in` won't work is because someone could define an extension named "toString" or "hasOwnProperty" or similar. The reason to avoid `Object.prototype.hasOwnProperty.call` is because that relies on both `Function#call` and `Object#hasOwnProperty` being undisturbed in the environment. Using `has` is robust against both.

ljharb: k, tests should be passing now

jfmengels:
LGTM! (if the tests pass obviously)
I'm letting @benmosher take a last look

ljharb:
hm, now I'm getting failures that are unrelated to my branch/this rule, that I can't reproduce locally.

ljharb:
@benmosher any idea why the tests are getting "duplicate export" errors for a rule that I'm not changing?

ljharb:
@benmosher actually i just reprod it locally on master when i `rm -rf node_modules && npm install`.

jfmengels:
I agree it sounds like an Espree error. I made a fix in #565

src/rules/extensions.js
benmosher: So, I'm not anti-dependencies in general, but I don't understand why 

- someone would have file extensions of `toString` or `hasOwnProperty`
- someone would be surprised if things are broken when they monkeypatch JS core functions

Am I correct that these are the cases where `has` matters over just a plain ol' `hasOwnProperty` call or `in` operator? Because if that's true, I'd still like to revert to `hasOwnProperty` or really, just `extension in modifiers` because these are well-understood JS core idioms, and in this specific case, not practically distinguishable.

resolvers/webpack/index.js
```diff
@@ -265,7 +266,7 @@ function findExternal(source, externals, context) {
 
   // else, vanilla object
   for (var key in externals) {
-    if (!externals.hasOwnProperty(key)) continue
+    if (!has(externals, key)) continue
```
benmosher: (restarted as an explicit line review, mostly out of curiosity + to gain familiarity with the new feature)

So, I'm not anti-dependencies in general, but I don't understand why 

- someone would have file extensions of `toString` or `hasOwnProperty`
- someone would be surprised if things are broken when they monkeypatch JS core functions

Am I correct that these are the cases where `has` matters over just a plain ol' `hasOwnProperty` call or `in` operator? Because if that's true, I'd still like to revert to `hasOwnProperty` or really, just `extension in modifiers` because these are well-understood JS core idioms, and in this specific case, not practically distinguishable.

benmosher: Looks ready to :shipit:, other than this, though. 😄 Thanks to both of you for working through @jfmengels' feedback, too!

ljharb: The first one, I also don't know why someone *would* do it, but given that it's a possibility, that would be a really silly reason for this package to break, don't you think?

As to the second, for a tool like an eslint plugin, that's a stronger case, since it won't be running in browser environments, or environments with unknown third-party code.

I'll remove the dependency if you insist, but I think this is still strictly better because you don't have to think about all these edge cases. I do not think `in` or `hasOwnProperty` are well-understood idioms, which is why this problem keeps coming up in libs all over the place, and why `has` exists.

ljharb: (let me know your final call, and i'll rebase out the last commit)

benmosher: Fair points. It's not a huge deal either way, or it wouldn't be so easy to have this back and forth. 😄

I hear what you're saying, and maybe one day all our key checks turn into `has` calls, but I don't think today is that day. 

I didn't realize the `has` part was a separate commit -- I'll just go ahead and merge the first one.

benmosher: Agh, nevermind, it's bundled with some other stuff. I'll just keep it for now. 😁

jfmengels:
Yay! Thanks @ljharb! :tada:

PR #561, opened 2016-09-14
docs: Update obsoleted reference

preco21:
Since [Additional export-from statements in ES7](https://github.com/leebyron/ecmascript-more-export-from) has been separated repo in [export-ns-from](https://github.com/leebyron/ecmascript-export-ns-from) and [export-default-from](https://github.com/leebyron/ecmascript-export-default-from), reference link need to update.


benmosher:
Thanks for the heads up! 😄

PR #568, opened 2016-09-20
Add `no-dynamic-require` rule (fixes #567)

### LINKED ISSUE #567: Rule proposal: no-dynamic-require
I have just found `require` calls of the following form in a codebase
```js
const type = '...';
const foo = require('../../bar/' + type + '/foo');
```

It can be useful when doing real dynamic package imports (with a plugin system for instance), but in most cases, it's not a good practice, for all the reasons that pushed the ES6 import/export system to be static.

I'm proposing a rule to forbid dynamic paths in `require()`, i.e. using something else than a string literal as a require argument. For the name, I think `no-dynamic-require` fits well.

I'm wondering whether template literals without expressions should be allowed.

### Invalid

```js
// Any non-static argument to require
var foo = require('../' + f + '/foo');
var foo = require(`../${f}/foo`);
var foo = require(f());
var foo = require(f + g);
```

### Valid
```js
var foo = require('../bar/foo');
import foo from '../bar/foo';

var foo = require(`../bar/foo`); // Allow this?
```

jfmengels:
Add `no-dynamic-require` rule (fixes #567)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## [Unreleased]
 ### Added
+- New rule [`no-dynamic-require`]: restrict deep package imports to specific folders. ([#567])
 - New rule [`no-internal-modules`]: restrict deep package imports to specific folders. ([#485], thanks [@spalger]!)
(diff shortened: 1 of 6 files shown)


jfmengels:
Wow, no comments? Am I getting better at this? :D

benmosher:
:shipit: 😎

PR #569, opened 2016-09-21
resolve configs for webpack2

toshafed:
Added configs for webpack2 (since version 2.1.0-beta.23)
`modulesDirectories` now [change](https://github.com/webpack/webpack.js.org/blob/develop/content/how-to/upgrade-from-webpack-1.md#resolveroot-resolvefallback-resolvemodulesdirectories) to `modules`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
   ([#566], thanks [@preco21])
 - [`default`]: allow re-export of values from ignored files as default
   ([#545], thanks [@skyrpex])
+- Added `resolve.modules` to configs for webpack2 support ([#569])
 
 ## [1.15.0] - 2016-09-12
 ### Added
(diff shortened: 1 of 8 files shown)


benmosher:
It's not clear to me what this PR accomplishes. Changes appear to be formatting and config for tests, only.

Can you elaborate on what the intended outcome is? I'm familiar with the `modules` config change for Webpack 2, but it's not clear to me why the tests needed to be updated.

cc @charlieschwabacher, who is apparently pretty enthused about this 😄

charlieschwabacher:
@benmosher - including `modulesPaths` in a webpack config w/ webpack 2 leads to an error:

```
WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.resolve has an unknown property 'modulesPaths'. These properties are valid:
   object { alias?, aliasFields?, cachePredicate?, descriptionFiles?, enforceExtension?, enforceModuleExtension?, extensions?, fileSystem?, mainFields?, mainFiles?, moduleExtensions?, modules?, plugins?, resolver?, symlinks?, unsafeCache? }
```

So you end up needing to remove the `resolve.modulesPaths` option to get webpack to build, but doing so breaks linting if you have added any paths.  Adding `resolve.modules` as a fallback in index.js here will fix the issue I am having.

benmosher:
@charlieschwabacher ah, got it, thanks for the detail. I looked right at those lines but just didn't see it the first time.

benmosher:
merged (had to move the changelog note, but kept your changes)

thanks!

PR #572, opened 2016-09-22
Update webpack resolver to allow passing a config object rather than filename

jameslnewell:
...because my Webpack config is generated by an API (and not written to disk) and `.eslintrc.js` allows us to use complex JavaScript objects.


benmosher:
ooo, cool! makes sense to me. will review in detail when I get a chance. thanks!

valentin-nemcev:
This PR also adds support for es6 module syntax (`export default`) in webpack config file, which is very useful

benmosher:
👍 :shipit: 

sorry about the wait, will publish asap

PR #574, opened 2016-09-23
Add web.js extension in react-native config

MoOx:
Make sense when you use react-native and react-native-web.
Not sure if that should go in another config tho.


benmosher:
thanks! makes sense to me. shouldn't hurt to have it here.

PR #578, opened 2016-09-23
[no-internal-modules] allow importing @-scoped modules

### LINKED ISSUE #577: no-internal-modules treats `@org/package` as a violation
The `no-internal-modules` rule is incorrectly flagging `import async from '@org/async'` as a violation, but should probably be allowed.

spalger:
Fixes #577 by counting the steps that do not start with `@`

### DIFF
--- a/src/rules/no-internal-modules.js
+++ b/src/rules/no-internal-modules.js
@@ -54,7 +54,8 @@ module.exports = {
           }
         }, [])
 
-      if (steps.length <= 1) return false
+      const nonScopeSteps = steps.filter(step => step.indexOf('@') !== 0)
+      if (nonScopeSteps.length <= 1) return false
 
       // before trying to resolve, see if the raw import (with relative
       // segments resolved) matches an allowed pattern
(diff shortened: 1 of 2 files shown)


jfmengels:
Thanks a lot for the fix @spalger!

jfmengels:
@benmosher I think there are failing tests for the `exports` rule on master. Do you know of anything changing recently?

spalger:
@jfmengels updated

jfmengels:
Awesome, LGTM :)

@benmosher For the errors, seems like a dependency (a parser?) now considers some of the invalid test cases from `export` as invalid syntax. So that's all good, we just need to update the expected error message / change the parser for those tests to keep supporting other parsers that allow the syntax.

Anyway, nothing for you to worry for this PR @spalger ;)

jfmengels:
We just merged the v2 branch, this needs to be rebased :s

spalger:
I'm tethering, so I'd rather not pull all the new deps, but I think that should do it

jfmengels:
Still LGTM, thanks for your patience on this!

PR #579, opened 2016-09-23
Fix #570: No newline after inline require

### LINKED ISSUE #570: `newline-after-import` should not apply to inline `require`'s
Only ones at the top. Or at least provide an option to ignore them.

When I have inline requires, it doesn't really help much to have a newline and it just get's in the way. I've been hit by this many times.

Example: https://github.com/zeit/hyperterm/commit/8a8fdacffe370ca0e267a7cffbff6734bef43036#diff-2460efa50e29d32731d41893921a7739R9

jfmengels:
Breaking change - newline-after-import: Do not require empty line after inline `require` (fixes #570).

cc @singles @sindresorhus

I made two commits:
- One applying the change and making the new tests pass
- One simplifying the implementation by removing the whole usage of scopes (redundant with the level system I picked from the order rule, and a bit more complex IMO)


jfmengels:
Rebased due to the merge of the v2 branch.

benmosher:
:shipit:

PR #584, opened 2016-09-24
Fix #519: Add option `newlines-between: "ignore"` to `order`

### LINKED ISSUE #519: Question: how to configure `import/order` to not enforce any policy on empty lines?
My rule is configured as:
```
"import/order": [ "warn" ],
```

So I assume this clause should be triggering, as I omit `newlines-between`:
> newlines-between: [always|never]:
> ...
> If omitted, assertion messages will be neither enforced nor forbidden.

But I still get a warning about empty lines:
```
import/order There should be no empty line between import groups at line 5 col 1
```
for the following code:
```
import foo from 'foo-core';

import {
	FooError,
} from 'foo-utils';
```

I expect no warnings about empty lines if `newlines-between` is not configured.

Probably it would be nice to have an explicit `"ignore"` value along with `"always"` and `"never"` which will be the default behavior if `newslines-between` isn't configured.

jfmengels:
Fix #519: Add option `newlines-between: "ignore"` to `order`

I haven't yet changed the default option, which I think we should, as I have not yet see any comments from others. That can be done in a second time, but should it be done, it should be done before the v2 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
   with some `parserOptions` in the mix. ([#402])
 - `react` shared config: added `jsx: true` to `parserOptions.ecmaFeatures`.
 - Added [`no-webpack-loader-syntax`] rule: forbid custom Webpack loader syntax in imports. ([#586], thanks [@fson]!)
+- Add option `newlines-between: "ignore"` to [`order`] ([#519])
 
 ### Breaking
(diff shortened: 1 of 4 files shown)


jfmengels:
I'm putting this in the v2 milestone. At least to not forget evaluating whether to include the breaking change or not.

benmosher:
`ignore` as default makes sense to me. Do you want to add that to this PR?

benmosher:
sweet, once that's in I'll publish `2.0.0-beta.0`?

jfmengels:
Updated. Yeah, v2 sounds good :)

PR #586, opened 2016-09-25
Add new rule `no-webpack-loader-syntax`

fson:
[Webpack allows specifying loaders](https://webpack.github.io/docs/loaders.html#introduction) and their configuration inline in imports using a non-standard syntax in the path. This rule allows disabling imports like this.

It can be useful for people using Webpack who don't want their code to rely on this Webpack-specific feature and we can possibly use it in Create React App to fix https://github.com/facebookincubator/create-react-app/issues/733.

### DIFF
--- a/docs/rules/no-webpack-loader-syntax.md
+++ b/docs/rules/no-webpack-loader-syntax.md
@@ -2,12 +2,12 @@
 
 Forbid Webpack loader syntax in imports.
 
-[Webpack](http://webpack.github.io) allows specifying [loaders](http://webpack.github.io/docs/loaders.html) and their configuration inline in imports using a special syntax like this:
+[Webpack](http://webpack.github.io) allows specifying the [loaders](http://webpack.github.io/docs/loaders.html) to use in the import source string using a special syntax like this:
 ```js
 var moduleWithOneLoader = require("my-loader!./my-awesome-module");
 ```
 
-This syntax is non-standard, so it couples the code using to Webpack. The recommended way to specify Webpack loader configuration is in a [Webpack configuration file](http://webpack.github.io/docs/loaders.html#loaders-by-config).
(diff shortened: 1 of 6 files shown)


docs/rules/no-webpack-loader-syntax.md
```diff
@@ -0,0 +1,36 @@
+# no-webpack-loader-syntax
+
+Forbid Webpack loader syntax in imports.
+
+[Webpack](http://webpack.github.io) allows specifying [loaders](http://webpack.github.io/docs/loaders.html) and their configuration inline in imports using a special syntax like this:
+```js
+var moduleWithOneLoader = require("my-loader!./my-awesome-module");
+```
+
+This syntax is non-standard, so it couples the code using to Webpack. The recommended way to specify Webpack loader configuration is in a [Webpack configuration file](http://webpack.github.io/docs/loaders.html#loaders-by-config).
```
jfmengels: `it couples the code to Webpack`

docs/rules/no-webpack-loader-syntax.md
```diff
@@ -0,0 +1,36 @@
+# no-webpack-loader-syntax
+
+Forbid Webpack loader syntax in imports.
+
+[Webpack](http://webpack.github.io) allows specifying [loaders](http://webpack.github.io/docs/loaders.html) and their configuration inline in imports using a special syntax like this:
```
jfmengels: Suggestion:
`Webpack allows specifying the loaders to use in the imports source string using ...`

fson:
Thanks for the review, I made the changes you suggested.

I also understand why it's a good idea to open an issue for new features like this first, and I'll do it the next time I want something added :)

jfmengels:
LGTM, thanks :)

benmosher:
LGTM too, thanks @fson for the submission (I think I'll use this, too!) and as always, thanks @jfmengels for reviewing!

PR #588, opened 2016-09-27
Fix #529: Add rule `no-unassigned-import`

### LINKED ISSUE #529: Rule proposal: Forbid imports with explicit side-effects
IMO, modules that have side-effect just from requiring them are really bad. I would like to see a rule that forbids importing/requiring modules that are not assigned, which implicitly means there is a side-effect or simply not used and therefore useless.

### Invalid

```js
import 'dosideeffect';
import './dosideeffect';
require('./dosideeffect');
```

### Valid

```js
import foo from 'bar';

// Explicit effect
import doEffect from './dosideeffect';
doEffect();
```

As far as rule name goes: `no-unnassigned-import`, `no-side-effect-import`, ...? I'm leaning toward the former at the moment, though it doesn't capture the rationale of it.

jfmengels:
Fix #529: Add rule `no-unassigned-import`

Just wanted to work on this rule. If you feel like adding it to v2.0, feel free to do so, or we can wait a bit and include it then. Either way is fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Added
+- Added [`no-unassigned-import`] rule ([#529])
 
 ## 2.0.0 - WIP
 ### Added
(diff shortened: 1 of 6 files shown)


benmosher:
👍 :shipit:

jfmengels:
Sounds good @benmosher.

cc @sindresorhus who wanted to know when we'd publish it. If you feel like reviewing the beta version :)

Issue #592, opened 2016-09-29
Cannot read property 'loc' of undefined crash in newline-after-import


lukekarrys:
Seeing this when greenkeeper updated a project to version 2. It's using decorators, which I know might not be supported, but I don't think it should be crashing. Here's a reduced test case that I came up with (I put the whole thing in a [gist too](https://gist.github.com/lukekarrys/403897c78dc5b0b7387c15851fe48b5a)):

**.eslintrc**
```
{
  "parser": "babel-eslint",
  "plugins": ["import"],
  "rules": {
    "import/newline-after-import": 2
  }
}
```

**.babelrc**
```
{
  "presets": ["es2015"]
}
```

**app.js**
```js
@SomeDecorator(require('./some-file'))
export default class App {}
```

**output**
```
> eslint-plugin-import-v2-test@1.0.0 lint /Users/lukekarrys/Desktop/403897c78dc5b0b7387c15851fe48b5a
> DEBUG=eslint-plugin-import:rules:newline-after-import eslint .

  eslint-plugin-import:rules:newline-after-import exit processing for +0ms /Users/lukekarrys/Desktop/403897c78dc5b0b7387c15851fe48b5a/app.js
  eslint-plugin-import:rules:newline-after-import got scope: +2ms [ Node {
    type: 'ExportDefaultDeclaration',
    start: 39,
    end: 66,
    loc: SourceLocation { start: [Object], end: [Object] },
    declaration:
     Node {
       type: 'ClassDeclaration',
       start: 54,
       end: 66,
       loc: [Object],
       id: [Object],
       superClass: null,
       body: [Object],
       decorators: [Object],
       range: [Object],
       _babelType: 'ClassDeclaration',
       parent: [Circular] },
    range: [ 39, 66 ],
    _babelType: 'ExportDefaultDeclaration',
    parent:
     Node {
       type: 'Program',
       start: 0,
       end: 67,
       loc: [Object],
       comments: [],
       tokens: [Object],
       range: [Object],
       sourceType: 'module',
       directives: undefined,
       body: [Circular],
       parent: null } } ]
  eslint-plugin-import:rules:newline-after-import node position in scope: +13ms -1
Cannot read property 'loc' of undefined
TypeError: Cannot read property 'loc' of undefined
[...]

PR #595, opened 2016-09-30
newline-after-import with decorators

lukekarrys:
This fixes the crash outlined in #592, when using a `require` statement inside a decorator.

While I was in there I also noticed that `newline-after-import` wasn't properly reporting if the next line was a decorator. I added this so it would be properly reported, but I can remove that if decorator support isn't something you want to add (I know it's still experimental).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 ### Fixed
 - [`prefer-default-export`] handles re-exported default exports ([#609])
+- Fix crash when using `newline-after-import` with decorators ([#592])
+- Properly report `newline-after-import` when next line is a decorator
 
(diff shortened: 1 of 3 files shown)


jfmengels:
@lukekarrys Awesome! Thanks a lot for this! Also thanks for the great report, which I did not have time to look at until now.

I'm not familiar with the decorators syntax, but this looks good to me.

As for supporting decorators... It's currently at stage 2, which is not the safest stage but means it will *likely* make it in the spec. I think the way this is implemented it should be fine even if it doesn't make it into the spec, or if it changes a bit. I'm fine with this at the moment. @benmosher?

lukekarrys:
I rebased this against `master` and fixed a conflict in `CHANGELOG.md`. I can remove the changelog commit too if that's something that you normally do before releases.

benmosher:
Nope, I'm cool with decorators. Maybe not always all of stage 2, but definitely decorators. 😎

Issue #481, opened 2016-08-10
Disallow importing "default" as a named import


ljharb:
Babel allows `import { default as Foo } from 'foo'` - I'm not sure if the spec does too.

I'd like a rule that requires `import { default as Foo, Bar }` be written as `import Foo, { Bar }`.

benmosher:
Last I checked, importing a default export as a name is valid ES6, but this seems like a reasonable style-guide rule. 

Name: `no-named-default`?

giodamelio:
This should be closed now that #596 is merged right?

PR #596, opened 2016-10-01
Add `no-named-default` rule

ntdb:
👋

This is an attempt at building the rule requested in #481. I'm not new to ES2015 or ESLint but I may have misstepped on some terminology and certainly on some logic. Please take a look and let me know what I can improve!


src/rules/no-named-default.js
```diff
@@ -0,0 +1,32 @@
+module.exports = {
+  meta: {
+    docs: {},
+  },
+
+  create: function (context) {
+    function checkSpecifiers(type, node) {
```
jfmengels: There's no reason to specify the type IMO, as I don't think this code will be extended to account for other types. You should remove the `type` parameter and the call to `bind` below. To make things a little bit clearer, you could simply assign this function to ImportDeclaration directly (like `{ ImportDeclaration: function (node) { ... } }`

src/rules/no-named-default.js
```diff
@@ -0,0 +1,32 @@
...
+      })
+
+      if (!hasImportSpecifier) {
+        return
+      }
+
+      node.specifiers.forEach(function (im) {
+        if (im.type !== type) return
+
+        if (im.imported.name === 'default') {
```
jfmengels:
```js
if (im.type === 'ImportSpecifier' && im.imported.name === 'default') {
}
```

src/rules/no-named-default.js
```diff
@@ -0,0 +1,32 @@
...
+          context.report(im.local,
+            'Using name \'' + im.local.name +
+            '\' as identifier for default export.')
```
jfmengels:
```js
context.report({
  node: im.local,
  message: 'Using name...'
})
```

Both syntaxes are valid but I think the first one was deprecated. Anyhow, this would be more consistent with the other rules in this repo AFAIK.

tests/src/rules/no-named-default.js
```diff
@@ -0,0 +1,31 @@
...
+    test({
+      code: 'import { default as bar } from "./bar";',
+      errors: [ {
+        message: 'Using name \'bar\' as identifier for default export.'
+      , type: 'Identifier' } ] }),
+    test({
+      code: 'import { foo, default as bar } from "./bar";',
+      errors: [ {
+        message: 'Using name \'bar\' as identifier for default export.'
+      , type: 'Identifier' } ] }),
```
jfmengels: Could you add the following invalid test case? From what I can see, that looks like a syntax error using the default parser, but still valid with `babel-eslint`.

```js
import {default} from './bar';
```

tests/src/rules/no-named-default.js
```diff
@@ -0,0 +1,31 @@
...
+    , rule = require('rules/no-named-default')
+
+ruleTester.run('no-named-default', rule, {
+  valid: [
+    test({code: 'import bar from "./bar";'}),
+    test({code: 'import bar, { foo } from "./bar";'}),
+
+    // es7
+    test({ code: 'import bar from "./bar";', parser: 'babel-eslint' }),
+    test({ code: 'import bar, { foo } from "./bar";', parser: 'babel-eslint' }),
```
jfmengels: What are the differences with the previous non-es7 test cases? Are you just testing the parsers?

docs/rules/no-named-default.md
```diff
@@ -0,0 +1,30 @@
...
+```js
+// foo.js
+export default 'foo';
+export const bar = 'baz';
+```
+
+...these would be valid:
+```js
+import foo from './foo.js';
+import foo { bar } from './foo.js';
```
jfmengels: `foo, { bar }`

src/rules/no-named-default.js
```diff
@@ -0,0 +1,32 @@
...
+      if (!hasImportSpecifier) {
+        return
+      }
+
+      node.specifiers.forEach(function (im) {
+        if (im.type !== type) return
+
+        if (im.imported.name === 'default') {
+          context.report(im.local,
+            'Using name \'' + im.local.name +
```
jfmengels: The error is currently: "Using name `bar` as identifier for default export".
Technically, if you do `import bar from './foo'`, `bar` is also the name of the default export, but that's valid.

I'd change the error message to "Use default import syntax to import `bar`".

tests/src/rules/no-named-default.js
ntdb: Yup these are here because I was unsure about whether the parser required testing for each rule. Are they safe to remove?

src/rules/no-named-default.js
ntdb: I also went ahead and removed the preceding checks as I don't believe they were helping us anymore.

tests/src/rules/no-named-default.js
jfmengels: I'd say it's safe yes. At this point and AFAIK, the only difference is that one parser might consider some syntax invalid while another accepts it (and it goes both ways now with the latest version of Espree).

src/rules/no-named-default.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  meta: {
+    docs: {},
+  },
+
+  create: function (context) {
+    return {
+      'ImportDeclaration': function (node) {
+        if (node.source == null) return
```
jfmengels: I'm curious, is there ever a case where the source is not defined? :question:
We can leave it in case you're not sure, but I don't when this would happen.

src/rules/no-named-default.js
```diff
@@ -0,0 +1,22 @@
...
+
+        node.specifiers.forEach(function (im) {
+          if (im.type === 'ImportSpecifier' && im.imported.name === 'default') {
+            context.report({
+              node: im.local,
+              message: 'Use default import syntax to ' +
```
jfmengels: You could use template literal syntax here by the way, makes it a bit simpler (and I think you're allowed to put this all on one line, but let me know if you aren't)
```js
`Use default import syntax to import '${im.local.name}'`
```

src/rules/no-named-default.js
ntdb: TBH, this is here because other rules included it (`named`, `no-deprecated`). It now seems likely that it's not needed here, will remove.

src/rules/no-named-default.js
ntdb: Good thought! I'm used to an 80-char line length, this is definitely under the project's max of 99. :)

ntdb:
@jfmengels I've just pushed changes based on your most recent review. Cheers!

jfmengels:
Awesome, thanks @ntdb! :)

benmosher:
👌🏻 Thanks for this, @ntdb! Cool to see you contributing!

And thanks for all the back and forth working together from both you and @jfmengels!

ntdb:
@benmosher Sure thing, thanks for maintaining a great plugin! I've just rebased, should be good to go.

PR #597, opened 2016-10-01
Fix single quotes in JSON example

ybiquitous:


jfmengels:
Oops, that's my mistake. Thanks for having my back there @ybiquitous! :)

PR #599, opened 2016-10-02
utils: [Fix] report the error stack on a resolution error

### LINKED ISSUE #536: eslint-import-resolver-webpack swallows errors in webpack.config.js
I was seeing the following error for each of my checked files:
`1:0   error  Resolve error: Unexpected token import`

I was trying to understand why (correct plugin configuration, babel and webpack setup correctly) for about 1h until I realized, that I forgot to start `eslint` with `babel-node` and my `webpack.config.js` was written in babel ES6+ code. Obviosly, it was starting with
`import path from 'path';`
which seems to have produced said errors.

I think instead `eslint-import-resolver-webpack` should error when trying to read the webpack configuration, similar to as it does, when `settings.import/resolver.webpack` cannot be resolved.

sompylasar:
Resolve errors are most likely caused by invalid configuration, and the reason is easier to determine with the full details rather than just `err.message`.

See https://github.com/benmosher/eslint-plugin-import/issues/536

With this change, it reports something like:

```
import/no-unresolved: Resolve error: SyntaxError: Unexpected token import
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:387:25)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at module.exports (/__censored__/webpack/configFactory.js:216:3)
    at configProdClient (/__censored__/webpack/configProdClient.js:5:36)
    at Object.<anonymous> (/__censored__/webpack/configForEslintImportResolver.js:1:126)
```

### DIFF
--- a/tests/files/load-error-resolver.js
+++ b/tests/files/load-error-resolver.js
@@ -1,1 +1,1 @@
-throw new Error('TEST ERROR')
+throw new SyntaxError('TEST SYNTAX ERROR')
--- a/tests/src/core/resolve.js
+++ b/tests/src/core/resolve.js
@@ -8,6 +8,11 @@ import * as fs from 'fs'
 import * as utils from '../utils'
 
 describe('resolve', function () {
+  // We don't want to test for a specific stack, just that it was there in the error message.
+  function replaceErrorStackForTest(str) {
+    return typeof str === 'string' ? str.replace(/(\n\s+at .+:\d+\)?)+$/, '\n<stack-was-here>') : str
+  }
+
   it('throws on bad parameters', function () {
     expect(resolve.bind(null, null, null)).to.throw(Error)
   })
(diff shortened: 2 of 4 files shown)


benmosher:
Ah, cool. Makes sense, though in retrospect I'm thinking maybe the stacktrace should be emitted by a `debug` logger instead of the error message, since it's so verbose. 

What do you think?

sompylasar:
I think the stack for this specific error is meaningful at the exact place where the error has occurred, to be able yo quickly fix it without switching context.

I use Atom Linter, so I expect all the error details in the linter error message which editor shows me.

We could **also** log the error, but the core benefit is being able to see its cause quickly.

st-sloth:
Is this PR moving? It would really be helpful.

sompylasar:
@st-sloth Looks like the tests need to be updated to reflect the change. I'll have a look if time allows.

sompylasar:
Changed the logic slightly to only add stack to unknown errors (that do not originate from `eslint-plugin-import` itself, which is designated by the error name of `'EslintPluginImportResolveError'`). Updated the tests to reflect the change.

PR #607, opened 2016-10-04
#604: Fixed code that relied on removed dependencies

### LINKED ISSUE #604: [2.0.0] Missing dependencies fail on npm@2.x.x
`eslint-plugin-import` seems to be missing a dependency on `es6-set` and `object-assign`.

This goes unnoticed in `npm@3.x.x` since they are included with `eslint-module-utils`, but fails on `npm@2.x.x`.

jfmengels:
Fixes #604.
I replaced `object-assign` and `es6-set` by native `Objct.assign` and `Set`. Since we dropped support for Node < v4, this should not break anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
-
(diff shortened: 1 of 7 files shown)


benmosher:
Agh, thanks. I could've sworn I did a full text search for these before removing. 

Also thought CI would catch this. Need to turn on `no-extraneous-dependencies`! 😅

Issue #479, opened 2016-08-08
expose semi-core modules via resolvers


benmosher:
https://github.com/clayne11/eslint-import-resolver-meteor/issues/11

TL;DR: semi-magic modules (in this case, `meteor/*`) that exist on disk (and thus can be linted) but are baked into the environment (and thus, are not explicitly dependencies) ideally should not blow up `no-extraneous-dependencies`.

Hand-wavy possible solution: allow resolvers to define import types somehow. (maybe move the `importType` functionality down into the resolver spec?)

Or could keep the existing behavior when/if chosen resolver does not explicitly provide the designation.

merlinpatt:
Any progress on this?

jfmengels:
@merlinpatt No, no progress. If you feel like helping out, there could be some :)

merlinpatt:
@jfmengels do you have suggestions on where to look to get started? I have 2 years dev experience but I've never written or contributed to a resolver before

jfmengels:
Thanks for considering contributing! Unfortunately, that's a part I haven't touched either, and I don't want to send you on the wrong path.

@benmosher or @tleunen could probably advise you.

benmosher:
@merlinpratt I appreciate your enthusiasm! I am not sure this is a good place for a first contribution, though, as it's mostly an API change. 

This is top of mind for me now that v2 is released, though, so I will be on it ASAP. 😁

SrimalPriyanga:
Any progress to solve this issue?

PR #608, opened 2016-10-04
#601: Fixed default order for the order rule

### LINKED ISSUE #601: Documented defaults for `import/order` groups are inaccurate

``` js
```

jfmengels:
Fixes #601.
This fixes the default order to be the one described by the documentation, and they have been inconsistent ever since the rule was introduced.

Though it fixes the implementation to correspond to what the documentation says, this is a fix that can create new linting errors for users that use internal modules and have not overridden the order. In this case, I think that can be accepted as a fix/patch, and not a breaking change/major, but wanted to have some confirmation.

Fortunately, I think that the use of internal modules is relatively low, so it should not break anything for most people.

cc @ljharb who I know is very passionnate about respecting semver and might have more experience with this kind of situation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 ### Fixed
 - Fixed code that relied on removed dependencies. ([#604])
+- Fixed default order for the [`order`] rule ([#601])
 
 ## [2.0.0]! - 2016-09-30
 ### Added
@@ -397,6 +398,7 @@ for info on changes for earlier releases.
 [#314]: https://github.com/benmosher/eslint-plugin-import/pull/314
 
 [#604]: https://github.com/benmosher/eslint-plugin-import/issues/604
+[#601]: https://github.com/benmosher/eslint-plugin-import/issues/601
 [#577]: https://github.com/benmosher/eslint-plugin-import/issues/577
 [#570]: https://github.com/benmosher/eslint-plugin-import/issues/570
 [#567]: https://github.com/benmosher/eslint-plugin-import/issues/567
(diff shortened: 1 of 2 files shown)


ljharb:
I'm actually kind of confused how this plugin even determines "internal" - does it read the webpack config?

Based on the code in https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#enforce-a-convention-in-module-import-order , the "internal" module there should be indistinguishable from an npm-installed module.

jfmengels:
@ljharb It uses the resolver system (that @benmosher knows way better than me), which AFAIK is either configured by the user or reads your webpack config (or Babel config, as you can do that too using Babel). When nothing is configured, you'll never get `internal` but always `external` (== npm package).

Obviously, I also forgot to mention that we have the option of simply fixing the docs to reflect the implementation.

ljharb:
While I think that its likely that very few users will be affected by this change, I do think it constitutes a breaking change, since it's been the case the entire time, across majors.

jfmengels:
Right. So maybe the way to go would be to simply change the documentation. I'm fine with that.

ljharb:
Let's update the docs now; and then open a separate PR for the breaking change that can be queued up for v3.

benmosher:
@jfmengels is the originally-documented order more ideal than the way that it has been implemented?

If not, does it make sense to just update the docs and never make the breaking change?

jfmengels:
> is the originally-documented order more ideal than the way that it has been implemented?

A bit, yes. They're pretty much the same, but there is support for `internal` modules, which only occur when you use Babel/Webpack to change the resolver logic.

I'm fine with changing the docs instead of the implementation, and I think that's what I'll do. Let's close this PR and I'll make another one to fix the docs instead. Sounds good?

benmosher:
Makes sense to me. 

I think once all the other issues stemming from #479 are fixed, this might actually be necessary, but if it hasn't been much of a pain point so far, probably makes sense to leave alone for now.

benmosher:
reopening for v3 milestone tracking, though feel free to close and open a new issue/PR or hack around on this one

PR #612, opened 2016-10-05
Remove `es6-set` dependency from max-depdendencies.js

### LINKED ISSUE #611: Fails on "Error: Cannot find module 'es6-set'"
I'm seeing an issue trying to use this plugin: it fails on "cannot find module 'es6-set'".  It appears that [9880ae7](https://github.com/benmosher/eslint-plugin-import/commit/9880ae7db99865a6e4c887151d59dd4de97debfa) removed `es6-set` from the package.json, but there's still a reference to `es6-set` in [max-dependencies.js](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/max-dependencies.js)

Retsam:
`es6-set` was already removed from the package.json

Fixes #611.

### DIFF
--- a/src/rules/max-dependencies.js
+++ b/src/rules/max-dependencies.js
@@ -1,4 +1,3 @@
-import Set from 'es6-set'
 import isStaticRequire from '../core/staticRequire'
 
 const DEFAULT_MAX = 10


benmosher:
duplicate of #607, thanks though! 😄

Issue #609, opened 2016-10-05
prefer-default-export false negative with named exports


FredyC:
I have a simple file `Memory.js` which contains:

```js
import t from 'tcomb'
import MemoryKey from './MemoryKey'

const Memory = t.declare('Memory')

export const MemoryValue = t.union([ Memory, t.String, t.Boolean, t.Number,], 'MemoryValue')
export default Memory.define(t.dict(MemoryKey, MemoryValue))
```

Then in `index.js` which exports all the types, I have it like this...

```js
export Memory, { MemoryValue } from './Memory' // eslint-disable-line import/prefer-default-export
export MemoryDictionary from './MemoryDictionary'
export MemoryKey from './MemoryKey'
```

The rule `prefer-default-export` is failing on that line for some reason. Babel transpiles that code just fine.

> eslint: 3.7.0
> eslint-plugin-import: 1.16.0
> eslint-config-airbnb-base: 8.0.0

benmosher:
Are you using babel-eslint?

FredyC:
Yes, of course. Not sure why I omitted that one from the list :)

> babel-eslint: 7.0.0

benmosher:
Sounds like a bug, then. Will take a look 😄

jfmengels:
Fixed in #622 :)

FredyC:
Awesome, thanks ! :)

PR #622, opened 2016-10-11
(#609) Make prefer-default-export count re-exported default exports

jfmengels:
(#609) Make prefer-default-export count re-exported default exports

The rule didn't count re-exporting default exports, which led to it reporting where it shouldn't have.

I also moved out the `captureDeclaration` function declaration because I easily could and so that it wouldn't get re-declared for every encountered `ExportNamedDeclaration`.


benmosher:
Oh cool, already fixed! Thanks!

jfmengels:
ineedthisforreactions-youre-welcome-gif-wl8OeD1gqUe4M

PR #623, opened 2016-10-12
 Fixed documentation for the default values for the `order` rule (fixes #601)

### LINKED ISSUE #601: Documented defaults for `import/order` groups are inaccurate

jfmengels:
Fixes #601, and replaces #608.

cc @ljharb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 ### Fixed
 - [`prefer-default-export`] handles re-exported default exports ([#609])
+- Fixed documentation for the default values for the [`order`] rule ([#601])
 
 ## [2.0.1] - 2016-10-06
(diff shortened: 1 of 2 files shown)


jfmengels:
Hmm, looks like another dependency update breaking... Looking into it a bit.

Anyway, thanks @ljharb for your advice :)

jfmengels:
Builds on master are broken, and the culprit is Babylon. I fixed the error and have made a PR over there https://github.com/babel/babylon/pull/170.

This is safe to merge, as it's just md files, but we'll need a new Babylon release before being able to have a working test suite. Tip: Do not remove your node_modules folder @benmosher :)

hzoo:
@jfmengels released babylon@6.11.6

jfmengels:
@hzoo Thanks! Tests are passing now :)

benmosher:
rebased and merged via 024a8a7

jfmengels:
Cool, thanks!

PR #625, opened 2016-10-14
Add deprecated flags to imports-first metadata

randycoulman:
Over on [eslint-find-rules](https://github.com/sarbbottam/eslint-find-rules), there has been some discussion about being able to detect deprecated rules in order to [not report them as unused](https://github.com/sarbbottam/eslint-find-rules/issues/172) and to [report when they are still being used](https://github.com/sarbbottam/eslint-find-rules/issues/188).

In order to implement those features, there needs to be a way to detect that a rule has been deprecated.  Core ESLint rules that are deprecated are marked as such in their metadata.

This PR adds the same `deprecated` flag to the metadata of the deprecated rule in this plugin.

I added a very specific spec for this rule to make sure that `imports-first` is marked as deprecated, but `first` (which is its new name) is not.  There might be a better way to write a more generic test, but when there’s only a single deprecated rule, it’s hard to know what the right direction is.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -33,7 +33,7 @@ export const rules = {
   'no-deprecated': require('./rules/no-deprecated'),
 
   // deprecated aliases to rules
-  'imports-first': require('./rules/first'),
+  'imports-first': require('./rules/imports-first'),
 }
 
 export const configs = {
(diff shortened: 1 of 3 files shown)


jfmengels:
@benmosher Any reason you didn't merge this? I'm :+1: with this

PR #630, opened 2016-10-18
[New] `no-extraneous-dependencies`: check both the absolute and relative filename

### LINKED ISSUE #602: no-extraneous-deps globs don't also check against `process.cwd() + glob`
I want to set my glob to `tests/**`, so that only the top-level "tests" dir can contain things that require dev deps.

However, this rule compares against the full absolute path of the file, which includes the path to the project root - which isn't really useful, and forces people to start 100% of their globs with "**/".

I'd like the rule to check both against the full filename, and also against `path.join(process.cwd(), glob)`, to remove the need for this boilerplate in airbnb's shared config.

ljharb:
Fixes #602.

This should be semver-minor.

### DIFF
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -1,4 +1,5 @@
 import fs from 'fs'
+import path from 'path'
 import pkgUp from 'pkg-up'
 import minimatch from 'minimatch'
 import importType from '../core/importType'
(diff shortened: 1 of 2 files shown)


ljharb:
@benmosher i'd love to see this merged and released! :-D

ljharb:
@jfmengels any chance you could do the honors? :-)

jfmengels:
I still don't know the release process, sorry :/
Try DMing @benmosher on Twitter maybe.

Would gladly do it next time though ^^'

Issue #620, opened 2016-10-11
Rule Proposal: exports-last


k15a:
I would like to propose a new rule which enforces that all export statements are at the end of the file.

#### Why?

I really like to keep all my exports at the end of the file so I can quickly scan what a file exports.

jfmengels:
Sounds good to me.
With commonjs, I have the same preference. I don't necessarily have it for ES6 modules at the moment, but this makes sense to me to have this rule.

@k15a Would you be interested in working on this? :)

k15a:
I am interested but I tried working on another issue 2 times and failed hard so it's not a guarantee for success :laughing:

benmosher:
Sounds like a reasonable style guide rule to me. 👍

sevor005:
Good afternoon! The question is: can I cancel the rule (put it in ignor) for classes that are exported at the beginning of the file **export default class SomeClass**, followed by **defaultProps** and **PropTypes**?

RodrigoNovais:
> Good afternoon! The question is: can I cancel the rule (put it in ignor) for classes that are exported at the beginning of the file **export default class SomeClass**, followed by **defaultProps** and **PropTypes**?

This would be reasonable for storybook csf files, since templating would require exports to be mixed within the file
It's not a big deal to disable the rule on every stories file but would be nice to be able to handle it in the eslint config

ljharb:
@RodrigoNovais you can use `overrides` to disable them on every story file at once, in the config.

RodrigoNovais:
Didn't know that's possible, thank you @ljharb

PR #632, opened 2016-10-19
exports-last

k15a:
Hey,

I am currently working on #620 and would like some feedback.

Currently my implementation only checks for ES Modules export declaration but I would also like to support CommonJS modules as well. Does anybody has an idea how to check if an ExpressionStatement is an CommonJS export?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 ### Added
 - [`no-anonymous-default-export`] rule: report anonymous default exports ([#712], thanks [@duncanbeevers]).
-- Add new value to [`order`]'s `newlines-between` option to allow newlines inside import groups ([#627], [#628], thanks [@giodamelio])
-- Add `count` option to the [`newline-after-import`] rule to allow configuration of number of newlines expected ([#742], thanks [@ntdb])
+- Add new value to `order`'s `newlines-between` option to allow newlines inside import groups ([#627], [#628], thanks [@giodamelio])
+- [`exports-last`] lints that export statements are at the end of the file ([#620] + [#632])
 
 ### Changed
 - [`no-extraneous-dependencies`]: use `read-pkg-up` to simplify finding + loading `package.json` ([#680], thanks [@wtgtybhertgeghgtwtg])
@@ -35,9 +35,6 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - Fix crash when using [`newline-after-import`] with decorators ([#592])
 - Properly report [`newline-after-import`] when next line is a decorator
 - Fixed documentation for the default values for the [`order`] rule ([#601])
-
-### Added
-- [`exports-last`] lints that export statements are at the end of the file ([#620] + [#632])
 
 ## [2.0.1] - 2016-10-06
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`no-anonymous-default-export`] rule: report anonymous default exports ([#712], thanks [@duncanbeevers]).
 - Add new value to [`order`]'s `newlines-between` option to allow newlines inside import groups ([#627], [#628], thanks [@giodamelio])
 - Add `count` option to the [`newline-after-import`] rule to allow configuration of number of newlines expected ([#742], thanks [@ntdb])
-- [`exports-last`] lints that export statements are at the end of the file ([#620] + [#632])
 
 ### Changed
 - [`no-extraneous-dependencies`]: use `read-pkg-up` to simplify finding + loading `package.json` ([#680], thanks [@wtgtybhertgeghgtwtg])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - Properly report [`newline-after-import`] when next line is a decorator
 - Fixed documentation for the default values for the [`order`] rule ([#601])
 
+### Added
+- [`exports-last`] lints that export statements are at the end of the file ([#620] + [#632])
+
 ## [2.0.1] - 2016-10-06
 ### Fixed
 - Fixed code that relied on removed dependencies. ([#604])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 ### Added
 - [`no-anonymous-default-export`] rule: report anonymous default exports ([#712], thanks [@duncanbeevers]).
-- Add new value to `order`'s `newlines-between` option to allow newlines inside import groups ([#627], [#628], thanks [@giodamelio])
+- Add new value to [`order`]'s `newlines-between` option to allow newlines inside import groups ([#627], [#628], thanks [@giodamelio])
+- Add `count` option to the [`newline-after-import`] rule to allow configuration of number of newlines expected ([#742], thanks [@ntdb])
 - [`exports-last`] lints that export statements are at the end of the file ([#620] + [#632])
 
 ### Changed
--- a/docs/rules/exports-last.md
+++ b/docs/rules/exports-last.md
@@ -1,0 +1,41 @@
+# exports-last
+
+This rule reports all export declaration which come before any non-export statements.
+
+## This will be reported
+
+```JS
+
+const bool = true
+
+export default bool
+
+const str = 'foo'
+
+```
+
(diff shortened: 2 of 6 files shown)


benmosher:
The ES module part looks good to me. 👍 

I'd leave out CommonJS for now and add it later behind a flag (unless it's a use case you're looking forward to linting).

k15a:
I've just added documentation for this rule. 📖

k15a:
I've resolved the merge conflicts but now travis ci is failing and I couldn't figure out why.

k15a:
I've removed the CommonJS part. Let me know if I can do something else to get this merged.

benmosher:
LGTM, but would like to have one of @jfmengels or @ljharb take a quick look and sign off as well before merging 👍🏻

src/rules/exports-last.js
```diff
@@ -0,0 +1,35 @@
+function isExportStatement({ type }) {
+  // ES Module export statements
+  if (type === 'ExportDefaultDeclaration' || type === 'ExportNamedDeclaration') {
```
jfmengels: Simpler:
```js
function isExportStatement({ type }) {
  return type === 'ExportDefaultDeclaration' || type === 'ExportNamedDeclaration';
}
```

And you should probably also test for ExportAllDeclaration (and add tests for that case)

src/rules/exports-last.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  return false
+}
+
+const rule = {
+  create(context) {
+    return {
+      Program({ body }) {
+        const lastNonExportStatement = body.reduce((acc, node, index) => {
```
jfmengels: rename to `lastNonExportStatementIndex`

src/rules/exports-last.js
```diff
@@ -0,0 +1,35 @@
...
+          if (isExportStatement(node)) {
+            return acc
+          }
+          return index
+        }, 0)
+
+        body.forEach((node, index) => {
+          if (isExportStatement(node) && index < lastNonExportStatement) {
+
```
jfmengels: remove this blank line (sorry about the nitpicking :sweat_smile:)

src/rules/exports-last.js
```diff
@@ -0,0 +1,35 @@
...
+      Program({ body }) {
+        const lastNonExportStatement = body.reduce((acc, node, index) => {
+          if (isExportStatement(node)) {
+            return acc
+          }
+          return index
+        }, 0)
+
+        body.forEach((node, index) => {
+          if (isExportStatement(node) && index < lastNonExportStatement) {
```
jfmengels: It might be a bit faster to do the `index < lastNonExportStatement` before the `isExportStatement`.

Or even better (?): only run the `forEach` on the statements after `lastNonExportStatement`.

```js
// maybe add +/-1 to lastNonExportStatement to make it work right
body.slice(lastNonExportStatement).forEach(node => {
  if (isExportStatement(node)) {
    // report
  }
})
```

docs/rules/exports-last.md
```diff
@@ -0,0 +1,41 @@
...
+export function func() {
+  console.log('Hello World 🌍')
+}
+
+export const str = 'foo'
+```
+
+## When Not To Use It
+
+If you don't mind exports being sprinkled throughout a file, you may not want to enable this rule.
```
jfmengels: It would be nice to indicate that this is only for the export statements using the `export` keyword, and not for CommonJS `module.exports/exports`

k15a:
Hey @jfmengels, I've added some more tests unfortunately one test is breaking but I think that's not an issue with my rule. I've also refactored the code. Let me know if I can do something else.

tests/src/rules/exports-last.js
```diff
@@ -0,0 +1,117 @@
...
+
+const ruleTester = new RuleTester()
+
+const errors = ['Export statements should appear at the end of the file']
+
+ruleTester.run('exports-last', rule, {
+  valid: [
+    // Empty file
+    test({
+      code: '',
```
k15a: This line is breaking the tests.

docs/rules/exports-last.md
```diff
@@ -0,0 +1,48 @@
+# exports-last
+
+This rule reports all export declaration which come before any non-export statements.
```
jfmengels: `This rule enforces that all exports are declared at the bottom of the file. This rule will report any export declarations that comes before any non-export statements.`

tests/src/rules/exports-last.js
```diff
@@ -0,0 +1,117 @@
...
+      code: `
+        export default 'such foo many bar'
+        export const so = 'many'
+        const foo = 'bar'
+        export const exports = ':)'
+        export const i = 'cant'
+        export const even = 'count'
+        export const how = 'many'
+      `,
+      errors,
```
jfmengels: Shouldn't there be 2 errors here? One for the `export default` and one for ` export const so = 'many'`?

tests/src/rules/exports-last.js
```diff
@@ -0,0 +1,117 @@
+import { test } from '../utils'
+
+import { RuleTester } from 'eslint'
+import rule from 'rules/exports-last'
+
+const ruleTester = new RuleTester()
+
+const errors = ['Export statements should appear at the end of the file']
```
jfmengels:
```js
const error = {
  ruleId: 'exports-last',
  message: 'Export statements should appear at the end of the file',
};
```
You can make this into an array by default but I think you'll need to make an array of multiple errors for some tests.

Would be nice if you could add:
- either the location of the reported node
- the type of the Node : `type: 'ImportDefaultSpecifier`. This might be easier.

At the moment, you are not reporting the export statements, but non-export statements. Any would be fine (though I' prefer reporting the export statements), but the message mentions the export statement, so let's report those. That's why I'd like to add this additional verification.

It could look something like:
```js
const error = type => {
  ruleId: 'exports-last',
  message: 'Export statements should appear at the end of the file',
  type
};

// ------

     test({
        code: `
          export * from './foo'
          const bar = true
        `,
        errors: [error('ExportAllDeclaration')],
      }),
```

src/rules/exports-last.js
```diff
@@ -0,0 +1,27 @@
...
+              })
+            }
+          })
+        }
+      },
+    }
+  },
+}
+
+export default rule
```
jfmengels: Would you mind making the file structure (exports and stuff) look a bit more like the others, for consistency? Here's an example: https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-dynamic-require.js

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - Properly report [`newline-after-import`] when next line is a decorator
 - Fixed documentation for the default values for the [`order`] rule ([#601])
 
+### Added
+- [`exports-last`] lints that export statements are at the end of the file ([#620] + [#632])
```
jfmengels: You'll need to rebase and move this back into the `Unreleased` section. Things have changed ever since you mad your PR :sweat_smile: 

If you don't know how to rebase, don't mind it, I'll do it later for you.

tests/src/rules/exports-last.js
jfmengels: Ok, try replacing it with a comment
```js
test({
        code: '// comment',
})
```
and if that doesn't work out, just remove the test case.

docs/rules/exports-last.md
```diff
@@ -0,0 +1,48 @@
...
+export const bool = true
+
+const str = 'foo'
+
+```
+
+## This will not be reported
+
+```JS
+export const bool = true
```
jfmengels: Add a simple non-exported variable declaration up top, to show that that is allowed

k15a:
Hey @jfmengels,

I think I know solved all the requested changes. Let me know if I can do something else to get this merged.

ljharb:
@k15a please rebase latest master, so as to remove all merge commits

k15a:
I tried my best. Did this removed the merge commits?

CHANGELOG.md
```diff
@@ -6,8 +6,8 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 ### Added
 - [`no-anonymous-default-export`] rule: report anonymous default exports ([#712], thanks [@duncanbeevers]).
-- Add new value to [`order`]'s `newlines-between` option to allow newlines inside import groups ([#627], [#628], thanks [@giodamelio])
-- Add `count` option to the [`newline-after-import`] rule to allow configuration of number of newlines expected ([#742], thanks [@ntdb])
```
ljharb: we lost one of the changelog entries - let's fix that :-)

k15a:
Fixed that as well. Hope everything is ok now.

alexkrolick:
I'd like to use this rule - anyone want to take a look at it again?

sevor005:
Good afternoon! The question is: can I cancel the rule (put it in ignor) for classes that are exported at the beginning of the file export default class SomeClass, followed by defaultProps and PropTypes?

Issue #626, opened 2016-10-15
React native absolute path imports


willlma:
Having the same issue as this [Stackoverflow Question](http://stackoverflow.com/questions/40025214/absolute-paths-with-react-native-raise-eslint-errors)

> in my React Native app, within my `package.json` I've set the name to `rpms`. This allows me to do:

```
import Component from 'rpms/App/common/Component'
```

> […] eslint-plugin-import raises […] `import/no-unresolved`.

sindresorhus:
You need to create a [resolver](https://github.com/benmosher/eslint-plugin-import#resolvers) for React Native.

guzart:
Just publish a package that piggy backs on eslint-import-resolver-node, hope this helps

https://www.npmjs.com/package/eslint-import-resolver-reactnative

benmosher:
oo, thanks @guzart!

Gapur:
You can use absolute paths like in this article

https://medium.com/better-programming/using-absolute-paths-in-react-native-3be369244fb1

I hope you found this piece useful.

PR #638, opened 2016-10-24
Add React Native Absolute import support to node resolver

maraujop:
Hi there,

First, thanks for eslint-plugin-import it really helps development.

I was having some issues using it with React Native and opened a StackOverflow question that turned into an issue of the project #626. Instead of creating a new resolver for React Native as suggested in this issue I've added a new setting to node resolver that is used to resolve absolute imports. If you prefer it to be a separate resolver, I can refactor it.

This is working in my RN project as expected.

Thanks, cheers
Miguel

### DIFF
--- a/resolvers/node/README.md
+++ b/resolvers/node/README.md
@@ -20,6 +20,9 @@ settings:
         - .es6
         - .coffee
 
+      # This allows to resolve absolute imports in React Native
+      projectName: 'YourAppNameInPackage.json'
+
       paths:
         # an array of absolute paths which will also be searched
         # think NODE_PATH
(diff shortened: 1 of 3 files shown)


maraujop:
BTW, just a suggestion. It would be nice to include how to run tests in the contribution guidelines and also how to trigger log debugging `DEBUG=eslint-plugin-import:resolver:node`

benmosher:
Hmm, tough call. I think as part of this v3 push I'd like to make resolvers more composable, so you could wrap a light React Native resolver around the normal Node resolver.

adding to v3 milestone to decide during planning

benmosher:
actually, see #626, looks like @guzart beat you to it? just catching up on issues/PRs so I just saw that.

closing this to defer to his resolver. 😎

PR #639, opened 2016-10-24
Ignoring flow type exports

### LINKED ISSUE #484: import/prefer-default-export should ignore type for flowtype
I have a file with this:

``` js
// @flow
export type CommentType = {
  mobile_image_thumbnail: string,
  author_name: string,
  preview_text: string,
  updated_at: string,
}
```

I get this error:

app/bundles/comments/types/index.js
  2:1  error  Prefer default export  import/prefer-default-export

[See doc note](https://flowtype.org/blog/2015/02/18/Import-Types.html):

```
Note that we only support the explicit named-export statements for now (i.e. export type UserID = number;). In a future version we can add support for latent named-export statements (i.e. type UserID = number; export {UserID};) and default type exports (i.e. export default type MyType = ... ;)…but for now these forms aren’t yet supported for type exports.
```

I just checked and `export default type` is not supported in 0.27.

Any chance of ignoring `export type` as an option to this rule?

jakubsta:
resolve #484

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## [Unreleased]
 ### Fixed
+- [`prefer-default-export`] handles flow `export type` ([#484])
 - [`prefer-default-export`] handles re-exported default exports ([#609])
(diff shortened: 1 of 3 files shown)


jakubsta:
@jfmengels Could you have a look once again?

PR #644, opened 2016-10-31
Add some core-module tests for no-extraneous-dependencies

gnarf:
Added some test cases (one of which fails) for core-modules in no-extraneous-dependencies


tests/src/rules/no-extraneous-dependencies.js
```diff
@@ -128,3 +128,30 @@ ruleTester.run('no-extraneous-dependencies', rule, {
...
+    test({
+      code: 'import "core-module"',
+      settings: { 'import/core-modules': ['core-module'] },
+    }),
+    test({
+      code: 'import "@namespaced/module"',
+      settings: { 'import/core-modules': ['@namespaced/module'] },
+    }),
+    test({
+      code: 'import "@namespaced/module/and/sub/path"',
```
gnarf: I'm not sure if this is supposed to work, but I was trying to use `core-modules` to fix the `no-extraneous-dependencies` for my webpack `resolve.alias` that points `@namespaced/module/lib/foo` to `./src/foo`

Maybe related to #479 ?

benmosher: Ah, weird. Yeah, #479 is the ideal path to resolving this issue without explicit config (Webpack resolver would report as a `project` or `internal`) but what you've done should theoretically work.

Must be either the `@` or `/` killing it. I suspect the latter...

ljharb:
Rebased this. Part of the issue was that modules have to *resolve* in order to be reported, and "core-module", for example, didn't exist on disk. I removed one of the namespace tests because `@generated/bar` is already tested.

ljharb (approved):

PR #654, opened 2016-11-05
fix unambiguous regex

### LINKED ISSUE #615: v2 is SLOW
Upgrading eslint-plugin-import led to significant change in speed of eslint scan on several codebases I support.

For example, https://github.com/FreeFeed/freefeed-server:
- 1.x: 16 seconds
- 2.x: ~110 seconds

Is this expected?

### LINKED ISSUE #634: Parser error on files in node_modules
I'm not sure where this issue belongs, I thought I'd start here as it's related to the rules in this package.

I'm using this plugin as it's included with [XO](https://github.com/sindresorhus/xo). I'm getting an error as a result of the following rules: `import/no-named-as-default`, `import/no-named-as-default-member`, `import/default`, `import/namespace` and `import/named`

``` bash
Parse errors in imported module browser-sync: Unexpected character # (1:1)
```

This happens when I'm importing `browser-sync`. It looks as though BrowserSync has a `#! /usr/bin/env node` at the top of their [main file](https://github.com/BrowserSync/browser-sync/blob/master/index.js), which I assume is causing the issue.

So appears there are a few things going on here:
- `eslint-plugin-import` is parsing files in my `node_modules` folder. I haven't spent much time understanding how this plugin works, I just consume it through XO.
- `browser-sync` has a `#! /usr/bin/env node` line at the top of their main file. I'm only aware of using this in node shell scripts, and I don't think the main BrowserSync file is a script
- it looks like the eslint parser falls over when it encounters a `#` as the first character in a javascript module file

There's a few things here and I'm happy to follow up with any changes to the other projects, I just wanted to check with others that the above is correct. Thanks! 😀

### LINKED ISSUE #649: Large performance regression in v2.0
I'm observing a very large performance hit between v1.x and v2.0.1.  Linting my project used to take approximately 13 seconds with v1.x, and now takes about 25 minutes with v2.0.1. 

I'm using ESLint v3.9.1, `eslint-plugin-import` v2.0.1, `eslint-import-resolver-webpack` v0.6.0, and `babel-eslint` v7.1.0.  

I've tried reverting to earlier versions each of those packages, and I only see a noticeable performance regression with `eslint-plugin-import` v2.0.x.  

---

My `.eslintrc.js` is as follows:

```js
module.exports = {
  rules: {
    "import/default": 0,
  },
  parser: "babel-eslint",
  env: {
    browser: true,
    es6: true
  },
  globals: {
    console: false,
    require: false
  },
  extends: [
      "../.eslintrc.js",
      "plugin:import/errors",
      "plugin:import/warnings"
  ],
  plugins: [
      "import"
  ],
  settings: {
      "import/resolver": "webpack",
      "import/extensions": ['.js']
  }
}
```

I can't release the source code of my project, but would be happy to help track down this bug by some other means.  I've already thrown a profiler on the process, and unfortunately didn't see anything particularly interesting...

benmosher:
Key notes:
- used a `&&` instead of `||` for the auto-ignore (fail) so it wasn't even used unless the file is ignored anyway, and then it would ignore the ignore if the regex passed
- did not set regex to `multiline` mode, so it would fail on files unless `import`/`export` is the first line (double fail)

I also had to comment out a parser error test for CoffeeScript that should never have passed in v2 if I had implemented this correctly to begin with.

To publish this, need to bump+publish `eslint-module-utils`, then bump the dep in the plugin and publish it too.

I'm thinking this should be semver-minor since it's a large enough fix that it's nearly a feature.

Technically a breaking change to `eslint-module-utils`, but that's [not a big deal](https://www.npmjs.com/browse/depended/eslint-module-utils).

fixes #615
fixes #634
fixes #649

### DIFF
--- a/tests/src/core/getExports.js
+++ b/tests/src/core/getExports.js
@@ -4,6 +4,7 @@ import ExportMap from '../../../src/ExportMap'
 import * as fs from 'fs'
 
 import { getFilename } from '../utils'
+import * as unambiguous from 'eslint-module-utils/unambiguous'
 
 describe('ExportMap', function () {
   const fakeContext = {
@@ -341,6 +342,25 @@ describe('ExportMap', function () {
         })
       })
     })
+
+  })
+
+  // todo: move to utils
+  describe('unambigous regex', function () {
+
+    const testFiles = [
+      ['deep/b.js', true],
+      ['bar.js', true],
+      ['deep-es7/b.js', true],
+      ['common.js', false],
+    ]
+
+    for (let [testFile, expectedRegexResult] of testFiles) {
+      it(`works for ${testFile} (${expectedRegexResult})`, function () {
+        const content = fs.readFileSync('./tests/files/' + testFile, 'utf8')
+        expect(unambiguous.test(content)).to.equal(expectedRegexResult)
+      })
+    }
 
   })
 
--- a/tests/src/core/getExports.js
+++ b/tests/src/core/getExports.js
@@ -346,7 +346,7 @@ describe('ExportMap', function () {
   })
 
   // todo: move to utils
-  describe('unambigous regex', function () {
+  describe('unambiguous regex', function () {
 
     const testFiles = [
       ['deep/b.js', true],
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -286,7 +286,8 @@ ExportMap.for = function (path, context) {
   const content = fs.readFileSync(path, { encoding: 'utf8' })
 
   // check for and cache ignore
-  if (isIgnored(path, context) && !unambiguous.potentialModulePattern.test(content)) {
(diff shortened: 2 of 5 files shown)


tests/src/core/getExports.js
```diff
@@ -344,4 +345,23 @@ describe('ExportMap', function () {
 
   })
 
+  // todo: move to utils
+  describe('unambigous regex', function () {
```
jfmengels: unambiguous

benmosher:
@jfmengels I agree about bumping `utils` to v2 👍🏻 I'm going to go stage the release right now

schmod:
Just tested v2.2.0 out on my codebase, and performance now appears to be comparable to v1.16.0.  Thanks so much!  (Details in #649)

Issue #414, opened 2016-07-04
import/extensions should have a ignorePackages option


laurentgoudet:
Currently setting `import/extensions` to `always` reports the following as an error:

```
import jQuery from 'jquery';
```

However, packages default entry point (and extension) should be resolved by the loader, so it doesn't really make sense to do:

```
import jQuery from 'jquery'/index.js;
```

benmosher:
This makes sense to me.

Feels like another mode between `always` and `never`. A flag wouldn't be meaningful in `never` mode.

vvo:
Great! I too want to force extensions only for everything that is not a main entry in package.json

`"extensions" : ["error", "always", {allowMainEntry: true}]`?

`"extensions": ["error", "as-needed"]`?

Naming is hard :D

benmosher:
maybe `[ "error", "always", "!packages" ]`?

vvo:
Do you have other example of eslint rules with two different string parameters?

I am not comfortable with "!packages", again do you have example of an eslint rule having that sort of string based configuration with a negation?

Why not provide an option object an drop the string based one? Major ftw :D

benmosher:
I think a string and then a flag object is probably ideal. No one is surprised by booleans on an options object. 

Or maybe `["error", "always", { packages: "never" }]`, so it's obvious the object is overriding. 

I suppose this new behavior would ignore index.js files too. Not sure how to denote that via the flag override name.

vvo:
I would not ignore index.js files, only packages.

benmosher:
Sure, but if `./foo/index.js` exists,

```js
import { x } from './foo' // acceptable
// vs
import { x } from './foo/index' // definitely still reported
```

I think of `index.js` being the default value of `package.json/main`, but maybe that's just me.

vvo:
In my project I specifically want to always enforce file extensions on anything that is not the main export of an external package.

Again that's also maybe just me, but this is what I am looking for.

vvo:
My goal being: when looking at local imports, I never want to ask myself: "Is this a folder? Is this the index.js file?" => always be specific on local

benmosher:
Ah, got it. I'm not sure that the two are distinguishable, ATM (explicit package main vs. implicit folder index file).

Would it be workable if it ignored both? or are you depending on this rule to suss out implicit `index` references?

vvo:
[...]

PR #663, opened 2016-11-11
Ignore root external modules when checking for extensions

kileras:
When checking for extensions we got a bunch of error dues to the linter checking the name of packages like ‘import Decimal from “decimal.js”’ because we are using ‘never’. The opposite is true too, as seen in #414, when using ‘always’ will trigger linting is you do ‘import query form “query”’.

So what this does is change the ‘checkExtension’ method to check if a file is an external module and if it is then check if its a root one, in which case it is ignored. In that ticket there was a conversation about adding different options but I actually think that this behaviour makes more sense. Naming of packages is a bit weird and will always be so its probably safer just ignore them completely in this case.

A ton of things may be wrong as I’m not familiar with eslint and there may be a better way of checking if the package is root or not that I’m don’t know.

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -84,6 +84,10 @@ function typeTest(name, settings, path) {
   return 'unknown'
 }
 
+export function isScopedModule(name) {
+  return name.indexOf('@') === 0
+}
+
 export default function resolveImportType(name, context) {
   return typeTest(name, context.settings, resolve(name, context))
 }
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -1,7 +1,7 @@
 import path from 'path'
 
 import resolve from 'eslint-module-utils/resolve'
-import { isBuiltIn, isExternalModule, isScoped } from '../core/importType'
+import { isBuiltIn, isExternalModule, isScoped, isScopedModule } from '../core/importType'
 import docsUrl from '../docsUrl'
 
 const enumValues = { enum: [ 'always', 'ignorePackages', 'never' ] }
@@ -126,6 +126,14 @@ module.exports = {
       return resolvedFileWithoutExtension === resolve(file, context)
     }
 
+    function isExternalRootModule(file) {
+      const slashCount = file.split('/').length - 1
+
+      if (isScopedModule(file) && slashCount <= 1) return true
+      if (isExternalModule(file, context, resolve(file, context)) && !slashCount) return true
+      return false
+    }
+
     function checkFileExtension(node) {
       const { source } = node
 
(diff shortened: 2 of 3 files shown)


src/rules/extensions.js
```diff
@@ -64,13 +64,23 @@ module.exports = {
       return resolvedFileWithoutExtension === resolve(file, context)
     }
 
+    function isExternalRootModule(file) {
+      if (!isExternalModule(file, context,resolve(file, context))) return false
```
ljharb: Space after `context`

src/rules/extensions.js
```diff
@@ -64,13 +64,23 @@ module.exports = {
...
+      if (file.indexOf('/') >= 0) return false
```
ljharb: `@foo/bar` is a valid external module that is importing the root.

src/rules/extensions.js
kileras: Ups

src/rules/extensions.js
kileras: Nice catch, totally forgot and I actually using private modules! Should be handled properly now.

kileras:
Looks like some network error, the build agent got a 400 when doing `npm install` but I have no clue on how to re-trigger it, apart form adding a random line in the pr :P

src/core/importType.js
```diff
@@ -49,6 +49,10 @@ function isRelativeToSibling(name) {
   return name.indexOf('./') === 0
 }
 
+export function isPrivateModule(name) {
+  return name.indexOf('@') === 0
```
ljharb: note that this might not actually be true - I could make a folder called "@foo" and put a package.json in it, and `require('@foo')` would work, but it wouldn't be a private module.

ljharb: also, the name isn't "private", it's "scoped" - scoped modules can be either public or private.

k15a: But wouldn't the folder have to be required like `./@foo`?

ljharb: not if it's inside `node_modules` - anyone can `mkdir` on disk inside `node_modules` and `require` will respect that.

k15a: Yeah that makes sence. 👍

kileras: Ups, renamed to scoped 😄 

Regarding `require('@foo')` as k15a pointed out if you have a folder with that name you'll do `./@foo` and if you have a folder that starts with '@' in your `node_modules` folder well that is the definition a scoped module. If someone create a random folder that follows the same pattern as a scoped package is not unreasonable to expect it to behave like a scoped module. Myself did in the past use scoped modules for local development of multiple packages and nowadays have webpack aliases that use the '@' and load modules from arbitrary locations.

ljharb:
@kileras would you mind checking "allow edits" on the right hand side of the PR?

kileras:
Nice option, didn’t knew about it! It is done, you can mess with my master now 😜

ljharb:
Only until the PR is closed or merged, at least :-p

PR #672, opened 2016-11-21
Check importPath extension with 'dot'

dplusic:
Assume that `importPath` is 'exceljs'.
The path does not have an extension.
But it ends with 'js' and will be handled as having an extension.
This misleads the module.
Fixed that by checking an extension with 'dot'.


dplusic:
I added the test case.

dplusic:
`tests/files/node_modules/exceljs/excel.js` is added for resolving. To resolve `importPath`, there should be js file. `tests/files/node_modules/exceljs/index.js` is also possible, but I want to follow the structure of 'exceljs'. The content is empty because it is not executed as same as other test files(e.g. `tests/files/node_modules/@org/package/index.js`).

`tests/files/node_modules/exceljs/package.json` is added to reference `excel.js`. Without this file, resolver finds only `index.js`. This file is not empty, github just suppressed the diff. You can show the diff by clicking it.

`tests/files/internal-modules/package.json` is added to simulate [Two package.json Structure](https://github.com/electron-userland/electron-builder/wiki/Two-package.json-Structure). With this structure and problem of misjudging extension(the first problem of this issue, that is, without my change) together, resolver reports error message. This makes test fail. This file is empty because only the existence affects the problem. The content does not matter.

ljharb:
Took care of the rebase.

PR #680, opened 2016-11-28
Use `read-pkg-up` over `pkg-up`.

### LINKED ISSUE #675: `pkg-up` vs `read-pkg-up`.
`pkg-up` is used to find the filepath of the nearest `package.json`.  That filepath is then [used to read that `package.json`](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-extraneous-dependencies.js#L15).  Could `pkg-up` be replaced with [`read-pkg-up`](https://github.com/sindresorhus/read-pkg-up)?  Both packages are written by the same author, but `read-pkg-up` reads the `package.json` and seems to be more maintained.
I can submit a Pull Request if you would like.

wtgtybhertgeghgtwtg:
Closes #675.


wtgtybhertgeghgtwtg:
I can't seem to find why this is failing.  Does anyone with more experience with `rule-tester` have any insight?

src/rules/no-extraneous-dependencies.js
```diff
@@ -1,18 +1,14 @@
 import fs from 'fs'
```
jfmengels: Can you remove this import, now that it is not used?

src/rules/no-extraneous-dependencies.js
```diff
@@ -1,18 +1,14 @@
...
 function getDependencies(context) {
-  const filepath = pkgUp.sync(context.getFilename())
-  if (!filepath) {
-    return null
-  }
-
   try {
-    const packageContent = JSON.parse(fs.readFileSync(filepath, 'utf8'))
+    const pkg = readPkgUp.sync({cwd: context.getFilename(), normalize: false})
+    const packageContent = JSON.parse(pkg.pkg)
```
jfmengels: The problem that makes the build fail is that `getDependencies` is now returning something else. `pkg.pkg` is now an object, and not a string that remains to be parsed. By attempting to parse it you are causing an error, which will be cached and make the function return `null`.

This works better:
```js
    const pkg = readPkgUp.sync({cwd: context.getFilename(), normalize: false})
    if (!pkg || !pkg.pkg) {
      return null
    }

    const packageContent = pkg.pkg
```

wtgtybhertgeghgtwtg: Thank you, I don't know how I missed that.

jfmengels:
LGTM, but do you know why the whole file is considered a having been changed in that last commit, instead of only the modified lines? That's pretty odd :thinking:
If possible, I would love to have only the lines that have really changed be marked as changed, for debugging purposes later on. If you feel like looking at that, that'd be great. Otherwise, it's not that big of a deal, but surprising still.

wtgtybhertgeghgtwtg:
I accidently had it as CRLF instead of LF.  That should be fixed now.

wtgtybhertgeghgtwtg:
Thank you for helping me with this Pull Request.

benmosher:
thanks @wtgtybhertgeghgtwtg!

wtgtybhertgeghgtwtg:
Thank you for merging.

Issue #458, opened 2016-07-26
no-extraneous-dependencies doesn't support nested package.json


nkt:
Example of my project structure:

```
├── package.json
├── src
│   ├── components
│   │   ├── Avatar
│   │   │   ├── Avatar.css
│   │   │   ├── Avatar.js
│   │   │   ├── README.md
│   │   │   └── package.json
│   │   ├── Button
│   │   │   ├── Button.css
│   │   │   ├── Button.js
│   │   │   ├── ButtonSpinner.css
│   │   │   ├── ButtonSpinner.js
│   │   │   ├── README.md
│   │   │   └── package.json
```

Root `package.json` contains all dependencies, nested `package.json` looks like this:

``` json
{
  "name": "Avatar",
  "main": "Avatar.js"
}
```

That's why I got wrong warnings:

```
src/components/Avatar/Avatar.js
  1:1  error  'react' should be listed in the project's dependencies. Run 'npm i -S react' to add it            import/no-extraneous-dependencies
  2:1  error  'classnames' should be listed in the project's dependencies. Run 'npm i -S classnames' to add it  import/no-extraneous-dependencies
```

Is this fixable?

jfmengels:
@sindresorhus Any chance you have already written a tool to list all package.json files (like `pkg-up` but multiple? I might create one otherwise.

@nkt Out of curiosity, what is the use of those `package.json` files? Do you distribute plenty of packages like this?

jfmengels:
@benmosher I propose to have an option (name TBD) to use all `package.json` files, and keep the current behavior by default. I'm not used to monorepos, but I guess that looking at all previous `packagge.json` files is not always wanted, depending on how you declare your dependencies for each sub-package.

nkt:
@jfmengels no, I'm not. I use `package.json` instead of `index.js`.

sindresorhus:
I haven't. Go ahead.

benmosher:
@jfmengels I'm thinking maybe this just shouldn't be supported. Autofix would be impossible in this case, plus `npm i src/components/Avatar` would fail if it did not declare _all_ of its dependencies.

I occasionally run into problems with the resolvers because they are descendants of the folder with the root `package.json` for this project. CI passes, but installing from `npm` fails.

Up to you, though, if it's easy and sane to do it behind an option.

jfmengels:
> I'm thinking maybe this just shouldn't be supported

[...]

PR #685, opened 2016-12-09
Add support to specify the package.json for no-extraneous-dependencies rule

ramasilveyra:
Add support for nested package.json, projects with a folder architecture with internals package.json files (see https://github.com/benmosher/eslint-plugin-import/issues/458).

### Discussion

Actually change the behavior of use the closest package.json to use the root package.json breaks the rule when you _want_ to check the dependencies from an internal package.json, so I have this two paths in mind:

1) Per default get the dependencies from the closest package.json and if the option `getPkgFromRoot` is true get the dependencies from the root package.json file.

2) Add an option like `pkgPath` to the rule to be able to specify the package.json path and if is null just find the closest package.json like the last version. IMHO this looks better, it's more flexible.

**What do you think about this?**

## 12/24/2016 update

I choose the _2._ path, the option is called `packagePath`.

## 01/29/2017 update

Commit name changed to `Add support to specify the package.json` as this solution doesnt fully solve all the cases of #458 Look: https://github.com/benmosher/eslint-plugin-import/pull/685#pullrequestreview-17807295

### DIFF
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -1,7 +1,7 @@
 # Forbid the use of extraneous packages
 
 Forbid the import of external modules that are not declared in the `package.json`'s `dependencies`, `devDependencies`, `optionalDependencies` or `peerDependencies`.
-The closest parent `package.json` will be used. If no `package.json` is found, the rule will not lint anything.
+The closest parent `package.json` will be used. If no `package.json` is found, the rule will not lint anything. This behaviour can be changed with the rule option `packageDir`.
 
 ### Options
 
@@ -26,6 +26,12 @@ You can also use an array of globs instead of literal booleans:
 ```
 
 When using an array of globs, the setting will be activated if the name of the file being linted matches a single glob in the array.
+
+Also there is one more option called `packageDir`, this option is to specify the path to the folder containing package.json and is relative to the current working directory.
+
+```js
+"import/no-extraneous-dependencies": ["error", {"packageDir": './some-dir/'}]
+```
 
 ## Rule Details
 
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -1,16 +1,20 @@
 import path from 'path'
+import fs from 'fs'
 import readPkgUp from 'read-pkg-up'
 import minimatch from 'minimatch'
 import importType from '../core/importType'
 import isStaticRequire from '../core/staticRequire'
 
-function getDependencies(context) {
+function getDependencies(context, packageDir) {
   try {
-    const pkg = readPkgUp.sync({cwd: context.getFilename(), normalize: false})
-    if (!pkg || !pkg.pkg) {
+    const packageContent = packageDir
+      ? JSON.parse(fs.readFileSync(path.join(packageDir, 'package.json'), 'utf8'))
+      : readPkgUp.sync({cwd: context.getFilename(), normalize: false}).pkg
+
+    if (!packageContent) {
       return null
     }
-    const packageContent = pkg.pkg
+
     return {
       dependencies: packageContent.dependencies || {},
       devDependencies: packageContent.devDependencies || {},
@@ -18,6 +22,19 @@ function getDependencies(context) {
       peerDependencies: packageContent.peerDependencies || {},
     }
   } catch (e) {
+    if (packageDir && e.code === 'ENOENT') {
+      context.report({
+        message: 'The package.json file could not be found.',
+        loc: { line: 0, column: 0 },
+      })
+    }
+    if (e.name === 'JSONError' || e instanceof SyntaxError) {
+      context.report({
+        message: 'The package.json file could not be parsed: ' + e.message,
+        loc: { line: 0, column: 0 },
+      })
+    }
+
     return null
   }
 }
@@ -93,6 +110,7 @@ module.exports = {
           'devDependencies': { 'type': ['boolean', 'array'] },
           'optionalDependencies': { 'type': ['boolean', 'array'] },
           'peerDependencies': { 'type': ['boolean', 'array'] },
+          'packageDir': { 'type': 'string' },
         },
         'additionalProperties': false,
       },
@@ -102,7 +120,7 @@ module.exports = {
   create: function (context) {
     const options = context.options[0] || {}
     const filename = context.getFilename()
-    const deps = getDependencies(context)
+    const deps = getDependencies(context, options.packageDir)
 
     if (!deps) {
       return {}
--- a/tests/src/rules/no-extraneous-dependencies.js
+++ b/tests/src/rules/no-extraneous-dependencies.js
@@ -1,10 +1,19 @@
 import { test } from '../utils'
 import * as path from 'path'
+import * as fs from 'fs'
 
 import { RuleTester } from 'eslint'
-
 const ruleTester = new RuleTester()
     , rule = require('rules/no-extraneous-dependencies')
+
+const packageDirWithSyntaxError = path.join(__dirname, '../../files/with-syntax-error')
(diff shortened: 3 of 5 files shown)


ramasilveyra:
I found time to finish this pr 😀, It's ready for a review.

I choose the `2.` solution. The option that allows you to specify the `package.json` path is `packagePath`.

src/rules/no-extraneous-dependencies.js
```diff
@@ -1,16 +1,22 @@
...
 import importType from '../core/importType'
 import isStaticRequire from '../core/staticRequire'
 
-function getDependencies(context) {
+function getDependencies(context, packagePath) {
   try {
-    const pkg = readPkgUp.sync({cwd: context.getFilename(), normalize: false})
-    if (!pkg || !pkg.pkg) {
+    const pkg = packagePath
+      ? JSON.parse(fs.readFileSync(packagePath, 'utf8'))
```
jfmengels: If the file is not found, the rule should report an error indicating that the file could not be found.
If the file could not be parsed, the rule should report an error indicating that too.
Please add tests for those cases :)

src/rules/no-extraneous-dependencies.js
```diff
@@ -1,16 +1,22 @@
...
 import minimatch from 'minimatch'
 import importType from '../core/importType'
 import isStaticRequire from '../core/staticRequire'
 
-function getDependencies(context) {
+function getDependencies(context, packagePath) {
   try {
-    const pkg = readPkgUp.sync({cwd: context.getFilename(), normalize: false})
-    if (!pkg || !pkg.pkg) {
+    const pkg = packagePath
```
jfmengels: Rename `pkg` to `packageContent`

src/rules/no-extraneous-dependencies.js
ramasilveyra:
```
...
  } catch (e) {
    if (packagePath && e.code === 'ENOENT') {
      context.report({ message: 'The package.json file could not be find.' })
    }
    if (packagePath && e instanceof SyntaxError) {
      context.report({ message: 'The package.json file could not be parsed due to syntax errors.' })
    }

    return null
  }
...
```

doesn't work because `context.report(` needs a `loc` or `node`, cc: http://eslint.org/docs/developer-guide/working-with-rules#contextreport

But maybe I can throw the error in those cases?

src/rules/no-extraneous-dependencies.js
ramasilveyra: Ready 😄

docs/rules/no-extraneous-dependencies.md
```diff
@@ -27,6 +27,12 @@ You can also use an array of globs instead of literal booleans:
 
 When using an array of globs, the setting will be activated if the name of the file being linted matches a single glob in the array.
 
+Also there is one more option called `packagePath`, this option is to specify the path of the package.json.
```
ramasilveyra: Ready 😄

src/rules/no-extraneous-dependencies.js
jfmengels: You could report the error at the beginning of the file, like [here](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/no-abusive-eslint-disable.js#L14-L20).

I would advise against throwing an error, as ESLint doesn't properly handle them, the process will crash and the error will be hard to understand. By reporting it this way, the user will get an error for pretty much every file, but it will be much clearer and visible than a wondering `console.error`.

```js
 if (packagePath && err.code === 'ENOENT') {
    // find --> found
      context.report({ message: 'The package.json file could not be found.' })
    }
    if (packagePath && err instanceof SyntaxError) {
      // Change the error message to this and append the error message
      context.report({ message: 'The package.json file could not be parsed: ' + err.message})
    }
```

ramasilveyra: Good hack! You are right, thanks for your comments and explanation!

Now this change request is fixed.

ramasilveyra:
PR title and commit title changed to `Add support to specify the package.json` since what @jfmengels says it's right and this doesn't fully solve #685 .

ramasilveyra:
@jfmengels test case for when the package.json is not found added

tests/src/rules/no-extraneous-dependencies.js
```diff
@@ -154,5 +158,29 @@ ruleTester.run('no-extraneous-dependencies', rule, {
...
+      code: 'import "not-a-dependency"',
+      options: [{packagePath: path.join(__dirname, '../../../package.json')}],
+      errors: [{
+        ruleId: 'no-extraneous-dependencies',
+        message: '\'not-a-dependency\' should be listed in the project\'s dependencies. Run \'npm i -S not-a-dependency\' to add it',
+      }],
+    }),
+    test({
+      code: 'import "not-a-dependency"',
+      options: [{packagePath: path.join(__dirname, '../../../doesnt-exist.json')}],
```
jfmengels: This kind of highlights a weirdness of this rule: you can specify a file that is not named `package.json`, which I don't think makes sense. I'm wondering whether `packagePath` should not rather be `packageDir` and be the path to the folder containing package.json.
i.e. `[{packagePath: path.join(__dirname, '../../../package.json')}]` --> `[{packageDir: path.join(__dirname, '../../../')}]`

What do you think? :thinking:

ramasilveyra: Brainstorming: The name of this property could be named also as `manifestFilePath`. https://en.wikipedia.org/wiki/Manifest_file

ramasilveyra: But yes, `packageDir ` seems better! I will do that.

jfmengels: Yes, packageDir seems better, less confusing for npm users :)

ramasilveyra: Ready!

src/rules/no-extraneous-dependencies.js
```diff
@@ -1,23 +1,40 @@
...
       peerDependencies: packageContent.peerDependencies || {},
     }
   } catch (e) {
+    if (packageDir && e.code === 'ENOENT') {
+      context.report({
+        message: 'The package.json file could not be found.',
+        loc: { line: 0, column: 0 },
+      })
+    }
+    if (packageDir && e instanceof SyntaxError) {
```
ljharb: This feels like it would be a useful error in all cases - even though currently an invalid package.json is skipped silently.

ramasilveyra: You are right!

ramasilveyra: Ready, `read-pkg-up` outputs a different kind of error for json syntax errors https://github.com/sindresorhus/parse-json/blob/master/index.js#L5

johnsardine:
Can we specify if it should look also in `bower_components`?

ljharb:
No, bower should not be used by anyone, and I would be very against supporting it.

johnsardine:
@ljharb I understand. Can you share some information about why you think that? I'm just curious and looking to learn more. Thanks

ljharb:
@johnsardine to start: 1) there never was a reason for bower to exist, npm has always been also for frontend code. 2) bower isn't a package manager, it's a CLI over curl-to-github, and one should use a proper package manager (that allows for nested dependencies when it is needed). 3) anything on bower that isn't on npm isn't worth using, and there'll be a plethora of better alternatives that are on npm.

ramasilveyra:
FYI @ljharb @jfmengels I just resolved the conflict on the CHANGELOG.md (rebase).

0xR:
It seems this PR is waiting on additional review which has not been performed for several months. Can we just merge this? If any major issues show up it will reported here and we can fix them or revert.

benmosher:
:shipit: 👍🏻 thanks for review @jfmengels

PR #689, opened 2016-12-15
Fix incompatibility with webpack 2.2.0-rc.0

Kovensky:
enhanced-resolve has been bumped to 3.0.


benmosher:
thanks! 

were you able to test against a project using Webpack 2.2.0? I don't have one handy.

I'm going to go ahead and merge this though, I figure it can only help

PR #693, opened 2016-12-20
Update no-dynamic-require.md

yooungt13:
require('../name' + name);  
it should not be passed

### DIFF
--- a/docs/rules/no-dynamic-require.md
+++ b/docs/rules/no-dynamic-require.md
@@ -18,5 +18,6 @@ require(name());
 ### Pass
 
 ```js
+require('../name');
 require(`../name`);
 ```
--- a/docs/rules/no-dynamic-require.md
+++ b/docs/rules/no-dynamic-require.md
@@ -18,7 +18,5 @@ require(name());
 ### Pass
 
 ```js
-require('../name');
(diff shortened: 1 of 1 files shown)


docs/rules/no-dynamic-require.md
```diff
@@ -18,7 +18,5 @@ require(name());
 ### Pass
 
 ```js
-require('../name');
```
ljharb: this isn't a dynamic require, and it should pass.

jfmengels: True, but it seems like it is duplicated a bit below.
@yooungt13 If you could, instead of removing it, replace it by a library import (like `lodash`), that might be better for the purpose of the example :)

jfmengels:
Thanks @yooungt13, good catch :)

I'm fine with this, but I left a comment to improve this if you don't mind :)

docs/rules/no-dynamic-require.md
ljharb: I thought the purpose was to illustrate that both a string literal, and a template literal with no interpolations, would pass?

jfmengels:
I fixed the issue. Thanks again for reporting this @yooungt13!

PR #697, opened 2016-12-21
[Tests] make package tests more robust by only checking .js files

ljharb:
From https://github.com/benmosher/eslint-plugin-import/pull/696#issuecomment-268664237


ljharb:
Not sure why appveyor failed but it looks unrelated?

jfmengels:
LGTM. Were you bit by this problem?

And yes, the appveyor error seems unrelated :thinking:

PR #701, opened 2016-12-25
Added order rule to support regexs

HenriBeck:
This will make the order rule more configurable especially when you want to have exact control over the order of the imports.

I have already added some tests if you need more I will be glad to add them.

I will add the documentation and the changelog probably tomorrow when I will have time.

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -7,6 +7,10 @@ const defaultGroups = ['builtin', 'external', 'parent', 'sibling', 'index']
 
 // REPORTING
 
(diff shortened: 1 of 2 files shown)


HenriBeck:
What do you mean with wild cards?
Do you mean globs?

ljharb:
@HenriBeck not globs as in `**`, wildcards as in `*` standing in for any string of length 0 or more.

HenriBeck:
I would be fine with that too

PR #712, opened 2017-01-05
Add `no-anonymous-default-export` rule

duncanbeevers:
Forbids exporting anonymous functions, literals, class expressions, or arrow functions as module default.

This is intended to improve the grepability of codebases by encouraging the re-use of identifiers at their declaration site and at their import sites.

### Fail
```js
export default []

export default () => {}

export default class {}

export default function () {}

export default 123

export default {}
```

### Pass
```js
const foo = 123
export default foo

export default class MyClass() {}

export default function foo() {}

/* eslint import/no-anonymous-default-export: [2, {"allowArray": true}] */
export default []

/* eslint import/no-anonymous-default-export: [2, {"allowArrowFunction": true}] */
export default () => {}

/* eslint import/no-anonymous-default-export: [2, {"allowAnonymousClass": true}] */
export default class {}

/* eslint import/no-anonymous-default-export: [2, {"allowAnonymousFunction": true}] */
export default function () {}

/* eslint import/no-anonymous-default-export: [2, {"allowLiteral": true}] */
export default 123

/* eslint import/no-anonymous-default-export: [2, {"allowObject": true}] */
export default {}
```

### DIFF
--- a/docs/rules/no-anonymous-default-export.md
+++ b/docs/rules/no-anonymous-default-export.md
@@ -1,0 +1,67 @@
+# no-anonymous-default-export
+
+Reports if a module's default export is unnamed. This includes several types of unnamed data types; literals, object expressions, arrays, anonymous functions, arrow functions, and anonymous class declarations.
+
+Ensuring that default exports are named helps improve the grepability of the codebase by encouraging the re-use of the same identifier for the module's default export at its declaration site and at its import sites.
+
+## Options
+
+By default, all types of anonymous default exports are forbidden, but any types can be selectively allowed by toggling them on in the options.
+
+The complete default configuration looks like this.
+
+```js
+"import/no-anonymous-default-export": ["error", {
+  "allowArray": false,
+  "allowArrowFunction": false,
+  "allowAnonymousClass": false,
+  "allowAnonymousFunction": false,
+  "allowLiteral": false,
+  "allowObject": false
+}]
+```
+
+## Rule Details
+
+### Fail
+```js
+export default []
+
+export default () => {}
+
+export default class {}
+
+export default function () {}
+
+export default 123
+
+export default {}
+```
+
+### Pass
+```js
+const foo = 123
+export default foo
+
+export default class MyClass() {}
+
+export default function foo() {}
+
+/* eslint import/no-anonymous-default-export: [2, {"allowArray": true}] */
+export default []
+
+/* eslint import/no-anonymous-default-export: [2, {"allowArrowFunction": true}] */
+export default () => {}
+
+/* eslint import/no-anonymous-default-export: [2, {"allowAnonymousClass": true}] */
+export default class {}
+
+/* eslint import/no-anonymous-default-export: [2, {"allowAnonymousFunction": true}] */
+export default function () {}
+
+/* eslint import/no-anonymous-default-export: [2, {"allowLiteral": true}] */
+export default 123
+
+/* eslint import/no-anonymous-default-export: [2, {"allowObject": true}] */
+export default {}
+```
--- a/src/rules/no-anonymous-default-export.js
+++ b/src/rules/no-anonymous-default-export.js
@@ -1,0 +1,84 @@
+/**
+ * @fileoverview Rule to disallow anonymous default exports.
+ * @author Duncan Beevers
+ */
+
+const defs = {
+  ArrayExpression: {
+    option: 'allowArray',
+    description: 'If `false`, will report default export of an array',
+    message: 'Assign array to a variable before exporting as module default',
+  },
+  ArrowFunctionExpression: {
+    option: 'allowArrowFunction',
+    description: 'If `false`, will report default export of an arrow function',
+    message: 'Assign arrow function to a variable before exporting as module default',
+  },
+  ClassDeclaration: {
+    option: 'allowAnonymousClass',
+    description: 'If `false`, will report default export of an anonymous class',
+    message: 'Unexpected default export of anonymous class',
+    forbid: (node) => !node.declaration.id,
+  },
+  FunctionDeclaration: {
+    option: 'allowAnonymousFunction',
+    description: 'If `false`, will report default export of an anonymous function',
+    message: 'Unexpected default export of anonymous function',
+    forbid: (node) => !node.declaration.id,
+  },
+  Literal: {
+    option: 'allowLiteral',
+    description: 'If `false`, will report default export of a literal',
+    message: 'Assign literal to a variable before exporting as module default',
+  },
+  ObjectExpression: {
+    option: 'allowObject',
+    description: 'If `false`, will report default export of an object expression',
+    message: 'Assign object to a variable before exporting as module default',
+  },
+  TemplateLiteral: {
+    option: 'allowLiteral',
+    description: 'If `false`, will report default export of a literal',
+    message: 'Assign literal to a variable before exporting as module default',
+  },
+}
+
+const schemaProperties = Object.keys(defs).
+  map((key) => defs[key]).
+  reduce((acc, def) => {
+    acc[def.option] = {
+      description: def.description,
+      type: 'boolean',
+      default: false,
+    }
+
+    return acc
+  }, {})
+
+module.exports = {
+  meta: {
+    schema: [
+      {
+        type: 'object',
+        properties: schemaProperties,
+        'additionalProperties': false,
+      },
+    ],
+  },
+
+  create: function (context) {
+    const options = Object.assign({}, context.options[0])
+
+    return {
+      'ExportDefaultDeclaration': (node) => {
+        const def = defs[node.declaration.type]
+
+        // Recognized node type and allowed by configuration,
+        //   and has no forbid check, or forbid check return value is truthy
+        if (def && !options[def.option] && (!def.forbid || def.forbid(node))) {
+          context.report({ node, message: def.message })
+        }
+      },
+    }
+  },
+}
--- a/tests/src/rules/no-anonymous-default-export.js
+++ b/tests/src/rules/no-anonymous-default-export.js
@@ -1,0 +1,50 @@
+import { test, SYNTAX_CASES } from '../utils'
+
+import { RuleTester } from 'eslint'
+
+const ruleTester = new RuleTester()
+const rule = require('rules/no-anonymous-default-export')
+
+ruleTester.run('no-anonymous-default-export', rule, {
+    valid: [
+      // Exports with identifiers are valid
+      test({ code: 'const foo = 123\nexport default foo' }),
+      test({ code: 'export default function foo() {}'}),
+      test({ code: 'export default class MyClass {}'}),
+
+      // Allow each forbidden type with appropriate option
+      test({ code: 'export default []', options: [{ allowArray: true }] }),
+      test({ code: 'export default () => {}', options: [{ allowArrowFunction: true }] }),
+      test({ code: 'export default class {}', options: [{ allowAnonymousClass: true }] }),
+      test({ code: 'export default function() {}', options: [{ allowAnonymousFunction: true }] }),
+      test({ code: 'export default 123', options: [{ allowLiteral: true }] }),
+      test({ code: 'export default \'foo\'', options: [{ allowLiteral: true }] }),
+      test({ code: 'export default `foo`', options: [{ allowLiteral: true }] }),
+      test({ code: 'export default {}', options: [{ allowObject: true }] }),
+
+      // Allow forbidden types with multiple options
+      test({ code: 'export default 123', options: [{ allowLiteral: true, allowObject: true }] }),
+      test({ code: 'export default {}', options: [{ allowLiteral: true, allowObject: true }] }),
+
+      // Sanity check unrelated export syntaxes
+      test({ code: 'export * from \'foo\'' }),
+      test({ code: 'const foo = 123\nexport { foo }' }),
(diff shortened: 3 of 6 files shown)


ljharb:
Should this then warn if you default-export an arrow function? what about an anonymous `class` expression?

Why are literals forbidden? The identifier name of their import would usually match the filename, by convention.

duncanbeevers:
The idea is to improve grepability of the codebase. When grepping around for an identifier, filenames aren't matched on, only the contents of the file.

I agree that it should warn if you default-export an arrow function or an anonymous `class` expression. I'll add specs for those cases.

duncanbeevers:
@ljharb Now forbids default exports of anonymous class expressions and arrow functions. Added functionality, specs and documentation to cover these cases.

duncanbeevers:
@ljharb Each forbidden unnamed default export type can now be selectively allowed with an option.

src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,61 @@
...
+ * @fileoverview Rule to disallow anonymous default exports.
+ * @author Duncan Beevers
+ */
+
+module.exports = {
+  meta: {},
+
+  create: function (context) {
+
+    const forbid = (option) => !~context.options.indexOf('allow-' + option)
```
ljharb: let's use `> -1` here, `!~` is a bit too clever ;-)

src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * @fileoverview Rule to disallow anonymous default exports.
+ * @author Duncan Beevers
+ */
+
+module.exports = {
```
ljharb: please add a schema to the rule, so that invalid options are rejected.

duncanbeevers:
@ljharb 
* Added specs demonstrating use of multiple options
* Added schema
* Removed `~`+`indexOf` gymnastics

src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,79 @@
...
+            node: node,
+            message: 'Unexpected default export of object expression',
+          })
+          return
+        }
+
+        if (type === 'FunctionDeclaration' && noID && forbid(MAP[type])) {
+          context.report({
+            node: node,
+            message: 'Unexpected default export of anonymous function',
```
jfmengels: I'm not a native speaker, but doesn't `of an anonymous function` sound better?

src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,79 @@
...
+
+        if (type === 'Literal' && forbid(MAP[type])) {
+          context.report({
+            node: node,
+            message: 'Unexpected default export of literal',
+          })
+          return
+        }
+
+        if (type === 'ObjectExpression' && forbid(MAP[type])) {
```
jfmengels: I think you can make all this less repetitive:

```js
const types = { // or whatever you wish to name it
  ObjectExpression: {
    validation: () => true,
    description: 'object expression',
  },
  FunctionDeclaration: {
    validation: id => !id,
    description: 'anonymous function',
  },
  // ...
}

// ....


'ExportDefaultDeclaration': (node) => {
  if (forbid(MAP[type]) && types[type] && types[type].validation(node.declaration.id)) {
    context.report({
      node: node,
      message: 'Unexpected default export of ' + types[type].description,
    })
  }
}
```

docs/rules/no-anonymous-default-export.md
```diff
@@ -0,0 +1,51 @@
+# no-anonymous-default-export
+
+Reports if a module's default export is unnamed. This includes several types of unnamed data types; literals, object expressions, anonymous functions, arrow functions, and anonymous class declarations.
```
jfmengels: I would also enforce this on arrays, which I for instance sometimes use to create a fixtures file.
Also, I think it would make sense to also handle template literals if we intend to handle pretty much any value ?

src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,79 @@
...
+
+const FLAGS = []
+for (const nodeType in MAP) {
+  FLAGS.push(MAP[nodeType])
+}
+
+module.exports = {
+  meta: {
+    schema: {
+      type: 'array',
```
jfmengels: I think the options should better be passed as an object, with a `true`/`false` values to each key. I found that having one object with settings instead of multiple values in a given order makes it much more extendable.

My gut feeling is to make it so that a field set to `true` would make it forbidden, and `false` would make it ignored, and that you'd have it `true` by default.

```js
/*eslint import/no-anonymous-default-export: [2, "allow-arrow-function-expression"]*/
-->
/*eslint import/no-anonymous-default-export: [2, {ArrowFunctionExpression: false}]*/
```

The schema would look something like this (IIRC) (adapted from what is done in https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-extraneous-dependencies.js)

```js
 schema: [
      {
        'type': 'object',
        'properties': {
          'Literal': { 'type': 'boolean' },
          'ObjectExpression': { 'type': 'boolean' },
          'FunctionDeclaration': { 'type': 'boolean' },
          'ArrowFunctionExpression': { 'type': 'boolean' },
          'ClassDeclaration': { 'type': 'boolean' },
        },
        'additionalProperties': false,
      },
    ],
```

And you'd get options like
```js

  create: function (context) {
  }
```

src/rules/no-anonymous-default-export.js
ljharb: it does.

docs/rules/no-anonymous-default-export.md
ljharb: Agreed - it should handle arrays, but template literals i'd expect to be covered under "literals".

jfmengels: Yes, I was thinking the same thing, no real reason to handle them separately from normal strings and numbers.

src/rules/no-anonymous-default-export.js
jfmengels: I would be fine with simplifying the name of the properties (like ArrowFunctionExpression --> ArrowFunction), you'd just need to account for the difference for that in the code.

tests/src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,36 @@
+import { test, SYNTAX_CASES } from '../utils'
+
+import { RuleTester } from 'eslint'
+
+var ruleTester = new RuleTester()
+var rule = require('rules/no-anonymous-default-export')
+
+ruleTester.run('no-anonymous-default-export', rule, {
+    valid: [
+      // Named exports are valid
```
duncanbeevers: Updated this comment to
```js
// Exports with identifiers are valid
```

tests/src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,36 @@
...
+      test({ code: 'export default 123', options: ['allow-literal'] }),
+      test({ code: 'export default {}', options: ['allow-object-expression'] }),
+      test({ code: 'export default class {}', options: ['allow-anonymous-class-declaration'] }),
+      test({ code: 'export default function() {}', options: ['allow-anonymous-function-declaration'] }),
+      test({ code: 'export default () => {}', options: ['allow-arrow-function-expression'] }),
+
+      // Allow forbidden types with multiple options
+      test({ code: 'export default 123', options: ['allow-literal', 'allow-object-expression'] }),
+      test({ code: 'export default {}', options: ['allow-literal', 'allow-object-expression'] }),
+
```
duncanbeevers: Added a few specs for a couple of flavors of exports and labeled the section
```js
// Sanity check unrelated export syntaxes
```

src/rules/no-anonymous-default-export.js
duncanbeevers: Options are now handled as a single object. Details are spelled out in the rule description file.

src/rules/no-anonymous-default-export.js
duncanbeevers: I've added articles are all the default export types for consistency.

```
a literal
an array
an object
```
etc...

docs/rules/no-anonymous-default-export.md
duncanbeevers: Added enforcement for arrays, and a spec for the handling of template literals.

src/rules/no-anonymous-default-export.js
duncanbeevers: The architecture of the rule now roughly matches this suggested layout.

tests/src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,47 @@
+import { test, SYNTAX_CASES } from '../utils'
+
+import { RuleTester } from 'eslint'
+
+var ruleTester = new RuleTester()
+var rule = require('rules/no-anonymous-default-export')
```
jfmengels: var -> const

tests/src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,47 @@
...
+      // Sanity check unrelated export syntaxes
+      test({ code: 'export * from \'foo\'' }),
+      test({ code: 'const foo = 123\nexport { foo }' }),
+      test({ code: 'const foo = 123\nexport { foo as default }' }),
+
+      ...SYNTAX_CASES,
+    ],
+
+    invalid: [
+      test({ code: 'export default []', errors: [{ message: 'Unexpected default export of an array' }] }),
```
jfmengels: I think the error messages are confusing.
`Unexpected default export of an anonymous class/function` are fine, because they mention that what is wrong is the anonymousness.

For the others, it feels like this rule says that arrays/literals/objects are not allowed to be the default export for some reason.
I think it would be clearer if we recommended giving it a name or assigning it to a variable.

tests/src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,47 @@
...
+      test({ code: 'export default () => {}', errors: [{ message: 'Unexpected default export of an arrow function' }] }),
+      test({ code: 'export default class {}', errors: [{ message: 'Unexpected default export of an anonymous class' }] }),
+      test({ code: 'export default 123', errors: [{ message: 'Unexpected default export of a literal' }] }),
```
jfmengels: Could you add a test with a string and one with a template literal ?

src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,84 @@
...
+    acc[def.options] = false
+
+    return acc
+  })
+
+const schemaProperties = Object.keys(defs).
+  map((key) => defs[key]).
+  reduce((acc, def) => {
+    acc[def.option] = {
+      description: 'If `false`, will not report default export of ' + def.message,
```
jfmengels: This sounds like the opposite of what the name suggests.
From what I understand, if `allowLiteral` is set to true, then Literals will be reported. :thinking: 

This should be reversed and the default value should be false.

src/rules/no-anonymous-default-export.js
```diff
@@ -0,0 +1,84 @@
...
+  ObjectExpression: {
+    option: 'allowObject',
+    message: 'an object expression',
+  },
+}
+
+const defaultOptions = Object.keys(defs).
+  map((key) => defs[key]).
+  reduce((acc, def) => {
+    acc[def.options] = false
```
jfmengels: `def.option` ? :thinking:

src/rules/no-anonymous-default-export.js
jfmengels: The names are explicit, I like it :+1:

src/rules/no-anonymous-default-export.js
duncanbeevers: This `defaultOptions` object doesn't strictly need to be constructed. I created it (poorly) for parity with the documentation. Since it doesn't affect behavior, I'm rewinding my decision here and removing this.

src/rules/no-anonymous-default-export.js
duncanbeevers: Yes, this fell out of date as I reworked the implementation. Corrected.

tests/src/rules/no-anonymous-default-export.js
duncanbeevers: Updated

tests/src/rules/no-anonymous-default-export.js
duncanbeevers: The error messages and schema description strings have all been made explicit. The messages about anonymous functions and classes remain the same, while the rest of the messages are updated to use a more active voice.

```
'Assign literal to a variable before exporting as module default'
```

tests/src/rules/no-anonymous-default-export.js
duncanbeevers: I added tests with string and template literal cases to both the success and failure corpuses, and found that template literals were not yet forbidden.

I corrected this by adding a `TemplateLiteral` clause to the `defs` structure whose value is intentionally identical to `Literal`'s

duncanbeevers:
I added an example case to the documentation of exporting an identified rather than anonymous class.

duncanbeevers:
@benmosher Let me know if there's anything I can do to shepherd this along.

jfmengels:
Thanks a lot for this @duncanbeevers! (and sorry for the delay in merging this)

PR #719, opened 2017-01-11
fix resolvers link

stevemao:


benmosher:
ah, nice catch, thanks for the PR!

PR #721, opened 2017-01-16
New: `group-exports` rule

robertrossmann:
Adds a new rule which prefers a particular style in exporting named members from modules.

Looking forward to hearing your feedback. Let me know if I have forgotten anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## [Unreleased]
 
+- Add [`group-exports`] rule: style-guide rule to report use of multiple named exports ([#721], thanks [@robertrossmann])
 
 ## [2.8.0] - 2017-10-18
 ### Added
@@ -431,6 +432,7 @@ for info on changes for earlier releases.
 [`unambiguous`]: ./docs/rules/unambiguous.md
 [`no-anonymous-default-export`]: ./docs/rules/no-anonymous-default-export.md
 [`exports-last`]: ./docs/rules/exports-last.md
+[`group-exports`]: ./docs/rules/group-exports.md
 
 [`memo-parser`]: ./memo-parser/README.md
 
@@ -442,6 +444,7 @@ for info on changes for earlier releases.
 [#744]: https://github.com/benmosher/eslint-plugin-import/pull/744
 [#742]: https://github.com/benmosher/eslint-plugin-import/pull/742
 [#737]: https://github.com/benmosher/eslint-plugin-import/pull/737
+[#721]: https://github.com/benmosher/eslint-plugin-import/pull/721
 [#712]: https://github.com/benmosher/eslint-plugin-import/pull/712
 [#696]: https://github.com/benmosher/eslint-plugin-import/pull/696
 [#685]: https://github.com/benmosher/eslint-plugin-import/pull/685
@@ -661,3 +664,4 @@ for info on changes for earlier releases.
 [@mplewis]: https://github.com/mplewis
 [@rosswarren]: https://github.com/rosswarren
 [@alexgorbatchev]: https://github.com/alexgorbatchev
+[@robertrossmann]: https://github.com/robertrossmann
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 * Forbid unassigned imports ([`no-unassigned-import`])
 * Forbid named default exports ([`no-named-default`])
 * Forbid anonymous values as default exports ([`no-anonymous-default-export`])
+* Prefer named exports to be grouped together in a single export declaration ([`group-exports`])
 
 [`first`]: ./docs/rules/first.md
 [`exports-last`]: ./docs/rules/exports-last.md
@@ -91,6 +92,7 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 [`no-unassigned-import`]: ./docs/rules/no-unassigned-import.md
 [`no-named-default`]: ./docs/rules/no-named-default.md
 [`no-anonymous-default-export`]: ./docs/rules/no-anonymous-default-export.md
+[`group-exports`]: ./docs/rules/group-exports.md
 
 ## Installation
 
--- a/docs/rules/group-exports.md
+++ b/docs/rules/group-exports.md
@@ -1,0 +1,87 @@
+# group-exports
+
+Reports when named exports are not grouped together in a single `export` declaration or when multiple assignments to CommonJS `module.exports` or `exports` object are present in a single file.
+
+**Rationale:** An `export` declaration or `module.exports` assignment can appear anywhere in the code. By requiring a single export declaration all your exports will remain at one place, making it easier to see what exports a module provides.
+
+## Rule Details
+
+This rule warns whenever a single file contains multiple named export declarations or multiple assignments to `module.exports` (or `exports`).
+
+### Valid
+
+```js
+// A single named export declaration -> ok
+export const valid = true
+```
+
+```js
+const first = true
+const second = true
+
+// A single named export declaration -> ok
+export {
+  first,
+  second,
+}
+```
+
+```js
+// A single exports assignment -> ok
+module.exports = {
+  first: true,
+  second: true
+}
+```
+
+```js
+const first = true
+const second = true
+
+// A single exports assignment -> ok
+module.exports = {
+  first,
+  second,
+}
+```
+
+```js
+function test() {}
+test.property = true
+test.another = true
+
+// A single exports assignment -> ok
+module.exports = test
+```
+
+
+### Invalid
+
+```js
+// Multiple named export statements -> not ok!
+export const first = true
+export const second = true
+```
+
+```js
+// Multiple exports assignments -> not ok!
+exports.first = true
+exports.second = true
+```
+
+```js
+// Multiple exports assignments -> not ok!
+module.exports = {}
+module.exports.first = true
+```
+
+```js
+// Multiple exports assignments -> not ok!
+module.exports = () => {}
+module.exports.first = true
+module.exports.second = true
+```
+
+## When Not To Use It
+
+If you do not mind having your exports spread across the file, you can safely turn this rule off.
--- a/src/rules/group-exports.js
+++ b/src/rules/group-exports.js
@@ -1,0 +1,98 @@
+const meta = {}
+/* eslint-disable max-len */
+const errors = {
+  ExportNamedDeclaration: 'Multiple named export declarations; consolidate all named exports into a single export declaration',
+  AssignmentExpression: 'Multiple CommonJS exports; consolidate all exports into a single assignment to `module.exports`',
+}
+/* eslint-enable max-len */
+
+/**
+ * Returns an array with names of the properties in the accessor chain for MemberExpression nodes
+ *
+ * Example:
+ *
+ * `module.exports = {}` => ['module', 'exports']
+ * `module.exports.property = true` => ['module', 'exports', 'property']
+ *
+ * @param     {Node}    node    AST Node (MemberExpression)
+ * @return    {Array}           Array with the property names in the chain
+ * @private
+ */
+function accessorChain(node) {
+  const chain = []
+
+  do {
+    chain.unshift(node.property.name)
+
+    if (node.object.type === 'Identifier') {
+      chain.unshift(node.object.name)
+      break
+    }
+
+    node = node.object
+  } while (node.type === 'MemberExpression')
+
+  return chain
+}
+
+function create(context) {
+  const nodes = {
+    modules: new Set(),
+    commonjs: new Set(),
+  }
+
+  return {
+    ExportNamedDeclaration(node) {
+      nodes.modules.add(node)
+    },
+
+    AssignmentExpression(node) {
+      if (node.left.type !== 'MemberExpression') {
+        return
+      }
+
+      const chain = accessorChain(node.left)
+
+      // Assignments to module.exports
+      // Deeper assignments are ignored since they just modify what's already being exported
+      // (ie. module.exports.exported.prop = true is ignored)
+      if (chain[0] === 'module' && chain[1] === 'exports' && chain.length <= 3) {
+        nodes.commonjs.add(node)
+        return
+      }
+
+      // Assignments to exports (exports.* = *)
+      if (chain[0] === 'exports' && chain.length === 2) {
+        nodes.commonjs.add(node)
+        return
+      }
+    },
+
+    'Program:exit': function onExit() {
+      // Report multiple `export` declarations (ES2015 modules)
+      if (nodes.modules.size > 1) {
+        nodes.modules.forEach(node => {
+          context.report({
+            node,
+            message: errors[node.type],
+          })
+        })
+      }
+
+      // Report multiple `module.exports` assignments (CommonJS)
+      if (nodes.commonjs.size > 1) {
+        nodes.commonjs.forEach(node => {
+          context.report({
+            node,
+            message: errors[node.type],
+          })
+        })
+      }
+    },
+  }
+}
+
+export default {
+  meta,
+  create,
+}
--- a/tests/src/rules/group-exports.js
+++ b/tests/src/rules/group-exports.js
@@ -1,0 +1,221 @@
+import { test } from '../utils'
+import { RuleTester } from 'eslint'
+import rule from 'rules/group-exports'
+
+/* eslint-disable max-len */
+const errors = {
+  named: 'Multiple named export declarations; consolidate all named exports into a single export declaration',
+  commonjs: 'Multiple CommonJS exports; consolidate all exports into a single assignment to `module.exports`',
+}
+/* eslint-enable max-len */
+const ruleTester = new RuleTester()
+
+ruleTester.run('group-exports', rule, {
+  valid: [
+    test({ code: 'export const test = true' }),
+    test({ code: `
+      export default {}
+      export const test = true
+    ` }),
+    test({ code: `
+      const first = true
+      const second = true
+      export {
+        first,
+        second
+      }
+    ` }),
+    test({ code: `
+      export default {}
+      /* test */
+      export const test = true
+    ` }),
+    test({ code: `
+      export default {}
+      // test
+      export const test = true
+    ` }),
+    test({ code: `
+      export const test = true
+      /* test */
+      export default {}
+    ` }),
+    test({ code: `
+      export const test = true
+      // test
+      export default {}
+    ` }),
+    test({ code: 'module.exports = {} '}),
+    test({ code: `
+      module.exports = { test: true,
+        another: false }
+    ` }),
+    test({ code: 'exports.test = true' }),
+
+    test({ code: `
+      module.exports = {}
+      const test = module.exports
+    ` }),
+    test({ code: `
+      exports.test = true
+      const test = exports.test
+    ` }),
+    test({ code: `
+      module.exports = {}
+      module.exports.too.deep = true
+    ` }),
+    test({ code: `
+      module.exports.deep.first = true
+      module.exports.deep.second = true
+    ` }),
+    test({ code: `
+      module.exports = {}
+      exports.too.deep = true
+    ` }),
+    test({ code: `
+      export default {}
+      const test = true
+      export { test }
+    ` }),
+    test({ code: `
+      const test = true
+      export { test }
+      const another = true
+      export default {}
+    ` }),
+    test({ code: `
+      module.something.else = true
+      module.something.different = true
+    ` }),
+    test({ code: `
+      module.exports.test = true
+      module.something.different = true
+    ` }),
+    test({ code: `
+      exports.test = true
+      module.something.different = true
+    ` }),
+    test({ code: `
+      unrelated = 'assignment'
+      module.exports.test = true
+    ` }),
+  ],
+  invalid: [
+    test({
+      code: `
+        export const test = true
+        export const another = true
+      `,
+      errors: [
+        errors.named,
+        errors.named,
+      ],
+    }),
+    test({
+      code: `
+        module.exports = {}
+        module.exports.test = true
+        module.exports.another = true
+      `,
+      errors: [
+        errors.commonjs,
+        errors.commonjs,
+        errors.commonjs,
+      ],
+    }),
+    test({
+      code: `
+        module.exports = {}
+        module.exports.test = true
+      `,
+      errors: [
+        errors.commonjs,
+        errors.commonjs,
+      ],
+    }),
+    test({
+      code: `
+        module.exports = { test: true }
+        module.exports.another = true
+      `,
+      errors: [
+        errors.commonjs,
+        errors.commonjs,
+      ],
+    }),
+    test({
+      code: `
+        module.exports.test = true
+        module.exports.another = true
+      `,
+      errors: [
+        errors.commonjs,
+        errors.commonjs,
+      ],
+    }),
+    test({
+      code: `
+        exports.test = true
+        module.exports.another = true
+      `,
+      errors: [
+        errors.commonjs,
+        errors.commonjs,
+      ],
+    }),
+    test({
+      code: `
+        module.exports = () => {}
+        module.exports.attached = true
+      `,
+      errors: [
+        errors.commonjs,
+        errors.commonjs,
+      ],
+    }),
+    test({
+      code: `
+        module.exports = function test() {}
+        module.exports.attached = true
+      `,
+      errors: [
+        errors.commonjs,
+        errors.commonjs,
+      ],
+    }),
+    test({
+      code: `
+        module.exports = () => {}
+        exports.test = true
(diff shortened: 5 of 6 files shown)


ljharb:
Why would this be preferred over `prefer-default-export`?

Specifically, why would exporting an object be better than named exports?

Alaneor:
Well, first things first:

```js
const first = true
export {
  first,
}
```

Here, I am not "exporting an object" - I am simply declaring all named exports at one place. The above is equivalent to

```js
export const first = true
```

The reason to use this rule primarily comes from some developers exporting at various places in a file. This makes it a bit hard to find out what exactly is exported from a particular module (because you have to scan through the whole file to find all the named exports). Preferring a singular `export {}` declaration simplifies this.

This rule is not preferred over `prefer-default-export`. Rather, it seems to be a complement to that rule. There are situations where you do not want to provide a single, default export because the things you export do not necessarily relate, ie. a `utils` module full of various standalone functions. Using named exports for such functions seems more appropriate.

> Also, for commonjs modules exporting on a single place also increases performance because you declare the exported object at once, therefore v8 does not have to create new hidden classes for the exported object every time you assign new property to it. Though I have to admit this perf gain is probably marginal.

ljharb:
Negligent performance theorizing shouldn't be used to inform code decisions, especially since v8 isn't the only place this code will be run.

I could certainly understand a rule that requires that all export statements in a file appear grouped - that seems to achieve your goal without requiring that people export an object (a bag of things), ie, without requiring that people forfeit the benefit of using named exports.

Alaneor:
I would like to stress out that you are not forfeiting the benefit of named exports by doing `export { thing1, thing2 }` - these **are** named exports. [See MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#Examples).

ljharb:
Ah, I misunderstood. Your rule isn't suggesting that people switch from multiple named exports to an export of an object - it's suggesting that people consolidate all named exports into a single statement?

Alaneor:
Yes, that's exactly the aim - to have all the exports in one place so that it's absolutely obvious what is being exported. 🎉

As per your other comment, it could be worthwhile considering implementing a rule which requires all `export` statements to be grouped. That would be a nice addition to this rule since it would require both named and default exports to be grouped together, further improving readability. But that's for another issue. 😄

ljharb:
Then in that case, I think this is an excellent rule, but that the rule name needs a bit of work :-)

Maybe `group-named-exports`?

If we wanted to combine this, with the ability to require grouping all exports, then perhaps `group-exports` taking an object with `named` and `default` booleans?

Alaneor:
Sounds like a good idea!

Do you think it would be useful to give users option to specify if the default export should be before or after the named exports? E.g. `default: "before-named|after-named"`. Alternatively, we can also allow a `true` value in which case the default export may appear before or after the named exports.

Also, what should happen if `named` is set to `false` (no grouping of named exports) and `default` is enabled? Should the default export appear near the first, last or any named export?

Alaneor:
I have implemented a simplified version of the default export placement check - for now, it requires the  default export to be adjacent to the named export (either before or after it). To make it simple, it does not check the placement if there are still multiple named exports. We may decide how this should behave in such situation (maybe it should then be adjacent to the last named export?).

Let me know what you think so far. We can discuss the options we would like to provide for this rule, then I will extend the current implementation with those options.

Thanks!

tests/src/rules/group-exports.js
```diff
@@ -0,0 +1,70 @@
...
+    test({ code: 'exports.test = true' }),
+  ],
+  invalid: [
+    test({
+      code: [
+        'export const test = true',
+        'export const another = true',
+      ].join('\n'),
+      errors: [
+        'Multiple named export declarations',
```
ljharb: let's have this error message contain something actionable - maybe "Multiple named export declarations; consolidate all named exports into a single statement"

Also, could this be autofixable?

tests/src/rules/group-exports.js
```diff
@@ -0,0 +1,70 @@
...
+    test({
+      code: [
+        'module.exports = {}',
+        'module.exports.test = true',
+        'module.exports.another = true',
+      ].join('\n'),
+      errors: [
+        'Multiple CommonJS exports',
+        'Multiple CommonJS exports',
+        'Multiple CommonJS exports',
```
ljharb: similarly here, "Multiple CommonJS exports; do a single assignment to `module.exports` for all of your exports."

Also, what about assigning to `exports` without the `module` prefix?

tests/src/rules/group-exports.js
Alaneor: Messages updated as requested.

Autofixing would be quite difficult, it seems - we have to
- remove `export` from in front of the things being exported
- add one extra `export` declaration somewhere into the file with all the exported names

The first step seems easy, but I am unsure how to do the second, given you can do only one fix per rule violation.

I guess we could use the last found export and do a full replacement of the source where we append the export at the end...Not sure if this is a good idea, though - if the code we replace contains other issues, I have no idea how ESLint would solve that.

tests/src/rules/group-exports.js
Alaneor: Assignment to the `exports` object is covered [here](https://github.com/benmosher/eslint-plugin-import/pull/721/files#diff-7c6599b72765e69ef4f27ae9f27ff175R91).

Alaneor:
I just pushed another update to this PR, incorporating remaining suggestions and fixing some issues:

- Assigning to `module.exports.exported.*` or deeper members is no longer considered as an export
- Accessing properties on `module.exports` is no longer mistakenly considered as an export declaration 😱

docs/rules/group-exports.md
```diff
@@ -0,0 +1,62 @@
...
+export default function test() {}
+// A single named export -> ok
+export const valid = true
+```
+
+```js
+const first = true
+const second = true
+
+// A single named export -> ok
```
ljharb: technically this is a single named export *statement* but it's two named exports - can we add "statement"?

docs/rules/group-exports.md
```diff
@@ -0,0 +1,62 @@
...
+module.exports = {
+  first: true,
+  second: true
+}
+```
+
+### Invalid
+
+```js
+// Multiple named exports -> not ok!
```
ljharb: "Multiple named export statements"

src/rules/group-exports.js
```diff
@@ -0,0 +1,133 @@
+const meta = {}
+const errors = {
+  ExportNamedDeclaration:
+    'Multiple named export declarations; consolidate all named exports into a single statement',
+  ExportDefaultDeclaration:
+    'Default export declaration should be adjacent to named export',
+  MemberExpression:
+    'Multiple CommonJS exports; consolidate all exports into a single assignment to ' +
```
ljharb: any reason this isn't on a single line? if `max-len` is complaining, let's enable the option that ignores long strings.

src/rules/group-exports.js
```diff
@@ -0,0 +1,133 @@
...
+        return void exports.named.add(node)
+      }
+
+      if (node.object.name === 'exports') {
+        // exports.exported.*: assignments this deep should not be considered as exports
+        if (accessorDepth(node) > 1) {
+          return
+        }
+
+        return void exports.named.add(node)
```
ljharb: Why are these using `return void` instead of doing two things in two statements? Namely, `exports.named.add(node); return;`

tests/src/rules/group-exports.js
```diff
@@ -0,0 +1,80 @@
...
+import { RuleTester } from 'eslint'
+import rule from 'rules/group-exports'
+
+const errors = {
+  named:
+    'Multiple named export declarations; consolidate all named exports into a single statement',
+  default:
+    'Default export declaration should be adjacent to named export',
+  commonjs:
+    'Multiple CommonJS exports; consolidate all exports into a single assignment to ' +
```
ljharb: same here re, let's let this be a single line

src/rules/group-exports.js
Alaneor: I put those in a `/* eslint-disable max-len */` block.

src/rules/group-exports.js
Alaneor: Personal preference, I guess. I like to return together with the last statement in the function. Now they are split into separate lines.

Alaneor:
Updated with recent comments.

I pushed it as a separate commit, figured it would be easier to review incrementally than going through the whole thing again and again. Once you are ready to merge, I can squash them all into a single commit.

tests/src/rules/group-exports.js
jfmengels: I think this would be pretty hard to auto-fix yes :/

docs/rules/group-exports.md
```diff
@@ -0,0 +1,62 @@
...
+// Default export is not adjacent to the named export statement -> not ok!
+export default {}
+const first = true
+export { first }
+```
+
+```js
+// Multiple exports assignments -> not ok!
+exports.first = true
+exports.second = true
```
jfmengels: This is a it more complicated for CommonJS. If you wish for your module to export a function, but also some other things under "named" exports

```js
module.exports = function foo() {};
module.exports.first = true;
module.exports.second = true;
```

You could potentially do this like in the following example, but that's pretty odd to satisfy this rule.
```js
module.exports = Object.assign(function foo {}, {first: true, second: true});
```

IMO, this rule should either:
- only reports errors for ES modules
- only report the use of several instructions like `module.exports.foo = true` if there is no other direct assignment (`module.exports = foo`). <-- I'd go for this as I myself prefer grouping `module.exports = { foo, bar }` when possible.

src/rules/group-exports.js
```diff
@@ -0,0 +1,133 @@
...
+  while (node.type === 'MemberExpression') {
+    depth++
+    node = node.parent
+  }
+
+  return depth
+}
+
+function create(context) {
+  const exports = {
```
jfmengels: Can you rename this to something else? `exports` is what I consider a "semi-reserved" keyword in Node, better to name it something else to avoid confusion.

src/rules/group-exports.js
```diff
@@ -0,0 +1,133 @@
...
+      }
+
+      // Member expressions on the right side of the assignment do not interest us
+      if (node.parent.type === 'AssignmentExpression' && node.parent.left !== node) {
+        return
+      }
+
+      if (node.object.name === 'module' && node.property.name === 'exports') {
+        // module.exports.exported.*: assignments this deep should not be considered as exports
+        if (accessorDepth(node) > 2) {
```
jfmengels: Could you add tests for this? Multiple instructions like `module.exports.foo.bar = 1` for instance.

README.md
```diff
@@ -76,6 +76,7 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 * Limit the maximum number of dependencies a module can have ([`max-dependencies`])
 * Forbid unassigned imports ([`no-unassigned-import`])
 * Forbid named default exports ([`no-named-default`])
+* Prefer single named export declaration ([`group-exports`])
```
robertrossmann: Changed to _Prefer named exports to be grouped together in a single export declaration_

docs/rules/group-exports.md
```diff
@@ -0,0 +1,62 @@
+# group-exports
+
+Reports when multiple named exports or CommonJS assignments are present in a single file and when the default export is not adjacent to the named export.
```
robertrossmann: Checks about grouping named and default exports next to each other have been removed.

robertrossmann: Reworded: _Reports when named exports are not grouped together in a single `export` statement or when multiple assignments to CommonJS `module.exports` or `exports` object are present in a single file._

docs/rules/group-exports.md
robertrossmann: I have refactored the rule to only do anything if the thing assigned to `module.exports` is an object literal or there is nothing being assigned to that property. If it's anything else (function, string, what have you) it will not report any errors.

So, this is still an error:

```js
module.exports = {}
module.exports.prop = true
```

But this is now ok:

```js
module.exports = () => {}
module.exports.first = true
module.exports.second = true
```

I guess this is the best of both approaches.

tests/src/rules/group-exports.js
```diff
@@ -0,0 +1,78 @@
...
+
+ruleTester.run('group-exports', rule, {
+  valid: [
+    test({ code: 'export const test = true' }),
+    test({ code: 'export default {}\nexport const test = true' }),
+    test({ code: [
+      'const first = true',
+      'const second = true',
+      'export { first,\nsecond }',
+    ].join('\n') }),
```
robertrossmann: Updated, thanks!

src/rules/group-exports.js
robertrossmann: Removed a bunch of now-obsolete code and ultimately renamed the Set to `named`, since we are now only tracking named exports.

src/rules/group-exports.js
robertrossmann: Added!

robertrossmann:
@jfmengels Thanks for the feedback! I have incorporated all the changes you requested, please have a look.

Since this rule now only does anything useful with named exports, I suggest we rename the rule to `group-named-exports`.

docs/rules/group-exports.md
ljharb: What about `exports = {}` or `exports.foo = {}`?

robertrossmann: @ljharb `exports = {}` will not export anything nor will it modify whatever value is in `module.exports`, the `exports` variable is but a local reference to the default `module.exports` object.

As for `exports.foo = {}`, that's basically a named export of an object.

ljharb: @robertrossmann yes, i'm aware of that; it could still be invalid, but doesn't necessarily have to be covered by this rule.

imo `exports.foo = {}` should be treated the same as `module.exports.foo = {}`.

If someone wants to export a function with extra properties, they should do something like `const foo = () => {}; foo.first = true; foo.second = true; module.exports = foo;` - I don't think the rule should allow multiple module.exports mutations under any circumstances.

robertrossmann: @ljharb assignments to `exports` and `module.exports` are indeed treated the same. 😄 I have added more tests to validate this.

So

```js
exports.first = true
module.exports.second = true
```

will cause this rule to issue a warning for both lines.

Currently the rule will ignore situations in which you assign something else than an object literal to `module.exports` (the rule will ignore all CommonJS module operations), so

```js
module.exports = function test() {}
module.exports.extra = true
module.exports.field = true
```

will not cause any warnings. Do you think it should warn even in situations like these?

robertrossmann:
I just pushed a major refactor of the rule's implementation as I discovered some bugs in how CommonJS modules were handled. See 22e103c.

Thank you for your review!

ljharb:
I think "group exports" is still the proper name; since named exports are the only way you could ever have > 1 to group, that was always the case, but `exports.foo` is not a named export - it's a property on the default export.

robertrossmann:
@ljharb In the future, you may want to have a rule which requires all `export` statements, whether they are named or default, to be grouped together/adjacent. That's why it would seem to be more appropriate to name this rule `group-named-exports` and the other just `group-exports`. Of course if you would prefer different name for the latter it's ok to keep it as it is now.

From your second comment it would seem you would prefer that this rule does not do anything from CommonJS modules (because everything you assign to `module.exports` is but a default export) ? Or how would you like to handle CommonJS modules?

Currently the assumption is that if you assign an object to `module.exports`, that object's properties are treated as "named" exports, and assigning anything else to `module.exports` is treated as a kind of "default" export and therefore ignored by this rule.

Here is an example CommonJS syntax that I was trying to cover. Of course in CommonJS there exist no technical difference between default and named exports since you can only export one thing, whether that is a function or an object or something else, but you can "emulate" named exports by exporting an object with its properties being the actual things you want to export.

```js
// "named exports" in CommonJS
function test() {}
function another() {}
module.exports = {
  test,
  another,
}

// "default export" in CommonJS
module.exports = function test() {}
```

ljharb:
You can emulate it, but those should be required to be grouped just like named exports would.

My expectations for this rule are:
 1. in ES modules, all `export` statements are grouped together.
 1. in CJS modules, all modifications to `module` or `exports` are grouped together.

robertrossmann:
@ljharb Thanks for feedback, I have updated the implementation as you specified. Please have a look and let me know if you have any further suggestions. Thanks!

CHANGELOG.md
```diff
@@ -382,10 +383,16 @@ for info on changes for earlier releases.
 [`no-webpack-loader-syntax`]: ./docs/rules/no-webpack-loader-syntax.md
 [`no-unassigned-import`]: ./docs/rules/no-unassigned-import.md
 [`unambiguous`]: ./docs/rules/unambiguous.md
+<<<<<<< HEAD
```
ljharb: this looks like a bad merge?

CHANGELOG.md
```diff
@@ -572,6 +579,10 @@ for info on changes for earlier releases.
 [@ntdb]: https://github.com/ntdb
 [@jakubsta]: https://github.com/jakubsta
 [@wtgtybhertgeghgtwtg]: https://github.com/wtgtybhertgeghgtwtg
+<<<<<<< HEAD
 [@duncanbeevers]: https://github.com/duncanbeevers
 [@giodamelio]: https://github.com/giodamelio
 [@ntdb]: https://github.com/ntdb
+=======
+[@Alaneor]: https://github.com/Alaneor
+>>>>>>> New: `group-exports` rule
```
ljharb: here too

docs/rules/group-exports.md
```diff
@@ -0,0 +1,87 @@
+# group-exports
+
+Reports when named exports are not grouped together in a single `export` declaration or when multiple assignments to CommonJS `module.exports` or `exports` object are present in a single file.
+
+**Rationale:** An `export` declaration or `module.exports` assignment can appear anywhere in the code. By requiring a single export declaration all your exports will remain at one place, making it easier to see what exports a module provides.
+
+## Rule Details
+
+This rule warns whenever a single file contains multiple named export declarations or multiple assignments to `module.exports` (or `exports`).
```
ljharb: doesn't it only warn when this is true *and* when they're not grouped together lexically?

src/rules/group-exports.js
```diff
@@ -0,0 +1,98 @@
...
+        nodes.commonjs.add(node)
+        return
+      }
+    },
+
+    'Program:exit': function onExit() {
+      // Report multiple `export` declarations (ES2015 modules)
+      if (nodes.modules.size > 1) {
+        for (const node of nodes.modules) {
+          context.report({
```
ljharb: could we include the source code location here?

src/rules/group-exports.js
```diff
@@ -0,0 +1,98 @@
...
+            node,
+            message: errors[node.type],
+          })
+        }
+      }
+
+      // Report multiple `module.exports` assignments (CommonJS)
+      if (nodes.commonjs.size > 1) {
+        for (const node of nodes.commonjs) {
+          context.report({
```
ljharb: also here?

CHANGELOG.md
robertrossmann: Oops, fixed everywhere... Sorry!

docs/rules/group-exports.md
robertrossmann: They do not need to be grouped lexically (I assume that means that the statements are consecutive), there must really be only a single assignment or export declaration.

ljharb: ah ok, thanks for clarifying.

In that case, is there a reason this rule couldn't be autofixable? (ie, delete all but the last export statements, and replace the last one with the combined group)

src/rules/group-exports.js
robertrossmann: Source code location should already be embedded in the `node` at `node.loc`, see [Eslint docs](http://eslint.org/docs/developer-guide/working-with-rules#contextreport).

ljharb: sweet, awesome

docs/rules/group-exports.md
robertrossmann: I do not currently see any issues with auto-fixing, either, but I have not included a fixer in this PR for several reasons:

- I am kind of new to ESLint rule development
- I have no prior experience manipulating or generally working with AST
- I wanted to keep this small and push it out fast (even though I then got lost in other tasks and forgot about this PR for several months 😄 )

So I'd rather not add fixer at this moment, even though it should be possible to implement it, eventually.

robertrossmann:
@jfmengels Could you please have a look at this PR once again and let me know if you have any additional notes? Thank you!

If you like, I can squash the commits into one before merging.

robertrossmann:
@jfmengels @ljharb Any chance merging this? I believe all the review notes have been addressed.

I have also squashed the changes into a single commit. Thanks!

robertrossmann:
Rebased on current master again. I would love to have this merged, let me know if there are any outstanding issues. Thanks!

_cc @ljharb @jfmengels_

ljharb:
@robertrossmann the appveyor tests are failing, i'm not sure why.

@jfmengels would you mind reviewing this again? it looks like your issues might be resolved.

robertrossmann:
@ljharb the test suite failed due to coveralls error, looks like something to do with me rebasing the branch. Restarting the suite should fix it. Thanks!

robertrossmann:
Closing due to constantly failing tests and an unnecessarily long discussion record and outdated review notes which I fear have been preventing merge for almost 7 months now.

Opened a new issue.

ljharb:
@robertrossmann please don't close PRs and open duplicates; in no way are any of the things you mentioned preventing merges.

Please reopen this and close the duplicate - duplicate PRs clutter the repo forever, since github creates undeleteable refs to them.

robertrossmann:
@ljharb Reopened. Just let me know what needs to be done in order to have this merged, please. Or tell me you don't like it so I don't waste time coming back to this. Thanks!

ljharb:
I do like it; @jfmengels has an outstanding review, and I'd prefer them to either approve or dismiss it before this is merged. If @benmosher approves first, we may be able to bypass that, but it's not ideal.

robertrossmann:
Any chance to have this merged or reviewed? Thanks.

robertrossmann:
Rebased again. Could we get this reviewed and merged in?

Regarding tests: They seem to be failing due to an unrelated error in `exports-last valid` test. See the logs.

Thank you.

CHANGELOG.md
```diff
@@ -4,20 +4,13 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
-
-
-## [2.8.0] - 2017-10-18
```
ljharb: I think this needs rebasing, and some fixing - it appears some info has been lost

docs/rules/group-exports.md
ljharb: Has this been addressed?

src/rules/group-exports.js
```diff
@@ -0,0 +1,98 @@
...
+      if (chain[0] === 'exports' && chain.length === 2) {
+        nodes.commonjs.add(node)
+        return
+      }
+    },
+
+    'Program:exit': function onExit() {
+      // Report multiple `export` declarations (ES2015 modules)
+      if (nodes.modules.size > 1) {
+        for (const node of nodes.modules) {
```
ljharb: could we use `nodes.modules.forEach` here, instead of a loop?

src/rules/group-exports.js
```diff
@@ -0,0 +1,98 @@
...
+          context.report({
+            node,
+            message: errors[node.type],
+          })
+        }
+      }
+
+      // Report multiple `module.exports` assignments (CommonJS)
+      if (nodes.commonjs.size > 1) {
+        for (const node of nodes.commonjs) {
```
ljharb: also here

src/rules/group-exports.js
robertrossmann: Fixed!

src/rules/group-exports.js
robertrossmann: Fixed!

CHANGELOG.md
robertrossmann: Ooops, sorry - fixed! If you'd like, I can drop the changes to CHANGELOG completely. Not sure if it's supposed to be part of a PR.

docs/rules/group-exports.md
robertrossmann: Yes - the rule now does not care what is being assigned to `module.exports`. Some tests for that behaviour can be [found here](https://github.com/benmosher/eslint-plugin-import/pull/721/files#diff-e516480e5c5622695cf87cc6d15e177fR176-212).

robertrossmann:
@ljharb thanks for feedback - I think I have made all the requested changes.

robertrossmann:
Looks like AppVeyor tests pass but uploading coverage data failed on all jobs and marked the build as failed.

CHANGELOG.md
ljharb: nah it's good to add it to "unreleased"

ljharb:
Let's give this a few more days, but then I'm going to merge it.

PR #727, opened 2017-01-22
Add rule 'no-self-import'

### LINKED ISSUE #447: Proposal: Disallow importing self

giodamelio:
Stops a file from importing itself. See #449 for details.

Progress:

- [x] Create rule
- [x] Write tests
- [x] Update docs
- [x] Add note in changelog
- [ ] Squash the above into one nice commit(No sure if I should do this?)

A few questions

- Is there any special way the error message should be written?
- Should I include the name of the file in it?
- Is it best practices to dynamically generate error messages based on the context?
  

Closes #449. Fixes #447.


src/rules/no-self-import.js
```diff
@@ -0,0 +1,42 @@
...
+import resolve from 'eslint-module-utils/resolve'
+import isStaticRequire from '../core/staticRequire'
+
+function isImportingSelf(context, node, requireName) {
+  const filePath = context.getFilename()
+  const resolvedPath = resolve(requireName, context)
+  if (filePath === resolvedPath) {
+    context.report({
+        node,
+        // TODO: check in about message, should it print the name of the module itself
```
ljharb: yes, i think it should - it's much more helpful that way. Ideally, it'd include the path to the file, and, *if different*, the path used in the `import` statement.

src/rules/no-self-import.js
```diff
@@ -0,0 +1,42 @@
...
+        message: 'Module imports itself',
```
jfmengels: > should it print the name of the module itself

I think there is no need. If they see the error in the editor, they will already have the file open and see which file the error is for. If they see it in the terminal or a CI environment's terminal, ESLint will display the location of the file it is in.

src/rules/no-self-import.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+
+module.exports = {
+  meta: {
+    doc: {
+      description: 'Prohibits a file from importing itself',
+      recommended: true,
+    },
+    fixable: 'code',
```
jfmengels: Remove this line as the rule is not fixable.

tests/src/rules/no-self-import.js
```diff
@@ -0,0 +1,42 @@
...
+    , rule = require('rules/no-self-import')
+
+const error = {
+  ruleId: 'no-self-import',
+  message: 'Module imports itself',
+}
+
+ruleTester.run('no-self-import', rule, {
+  valid: [
+    test({ code: 'import _ from "lodash"'}),
```
ljharb: how can any of these valid examples be deemed such without a filename to test against?

It's probably good to keep all of these, to ensure that when there's no filename (like stdin input), the rule bails out, but these should include a filename explicitly.

src/rules/no-self-import.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * @fileOverview Prohibits a file from importing itself
+ * @author Gio d'Amelio
+ */
+
+import resolve from 'eslint-module-utils/resolve'
+import isStaticRequire from '../core/staticRequire'
+
+function isImportingSelf(context, node, requireName) {
+  const filePath = context.getFilename()
```
ljharb: what happens when the input is from stdin, and thus there's no filename?

jfmengels: Good catch! When that happens, `filePath` will equal `'<text>`. In some other plugin, we've handled it like: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/filename-case.js#L71-L73

tests/src/rules/no-self-import.js
giodamelio: Thanks, that makes sense, I don't totally grok the testing setup yet. I'll add that.

src/rules/no-self-import.js
ljharb: That seems totally reasonable - it might also be useful to create a generic way to tag rules as "does not apply when the file is from stdin"

src/rules/no-self-import.js
ljharb: That's true - but because resolution is being applied, it might not be immediately apparent which import path is pointing back to the file.

jfmengels: ESLint will tell you on what line the import is, and in the editor, it will put red squiggly lines under the import statement / require call. That sounds sufficient to me, especially since (IMO) having the whole path to the file in the error message would probably make the error harder to read.

src/rules/no-self-import.js
jfmengels: I have actually never used the `--stdin` option and don't know the uses for it :thinking:

giodamelio:
That should take care of the changes you guys requested.

src/rules/no-self-import.js
giodamelio: I didn't even know that was a thing you could do.

ljharb: I use it all the time when providing reproduction steps for eslint bugs :-p

giodamelio:
That should cover the changelog and docs.

giodamelio:
I think the AppVeyor build might have failed because I pushed a second commit while the first one was still running, I saw that happen earlier. I didn't change any code in these last two commits, so it shouldn't have broken.

giodamelio:
Yep those files are needed. I just added a comment in them explaining.

src/rules/no-self-import.js
benmosher: last I knew, Sublime uses `--stdin`, but with `--stdin-filename` (or something like that) to fake in the filename

giodamelio:
Sounds good to me.

ljharb:
(heads up tho, if you squash directly on github you actually lose the link to the git log for the PR - it's far better to rebase/squash on the command line before clicking the button on github)

giodamelio:
Oh my god, @coveralls bot is getting a bit annoying.

jfmengels:
> (heads up tho, if you squash directly on github you actually lose the link to the git log for the PR - it's far better to rebase/squash on the command line before clicking the button on github)

@ljharb I'm sorry, what git log are you talking about?
I was about to merge this, but I don't know how to squash this locally and then re-push to GitHub, as I don't have control over @giodamelio's fork. What do you recommend otherwise, squashing with the button anyway, or squashing on the command line and pushing to master?

Sorry about the delay @giodamelio!

giodamelio:
I think he meant when you squash on the cli you can include the commit messages from all the commits in the body of the squashed commit. I can squash it for you if you want, but I don't really mind losing the messages.

jfmengels:
Hmm, this is what I'd get if I were to squash it from GitHub (I have the option to edit if I want to):

```
* Add rule itself

* Add test

* Add requested changes

* Change the word prohibit to forbid to match other rules

* Update docs

* Update changelog

* Add comment in test files

* Fix typos

* Simplify implementation a bit

* Add thanks :)

* extra comma
```

So I'm not sure what logs would be lost.
But yeah sure, if you feel like squashing @giodamelio, might as well :)

giodamelio:
Oh, I didn't think Github could do that. You might as well do it on your end then, the squash was giving me a bit of trouble and I would rather Github take care of it then me borking something.

ljharb:
@jfmengels i'm talking about when you fetch the `origin/pr/727` ref from the command line - that "ref" (a git concept) will be orphaned when using squashmerge or rebasemerge on the github UI.

In other words, it's not just the commit messages. I'm content to do the squash for you, if that's necessary?

ljharb:
@jfmengels also, if you note the text "Add more commits by pushing to the feature-no-self-import branch on giodamelio/eslint-plugin-import." - PRs now by default have the box checked that allows maintainers to push branches *directly to the fork* - so in fact you can directly update this PR's branch, which then reruns the tests, before actually doing the merge.

ljharb:
I've just done the rebase and force pushed it to the fork. Once tests have passed, please feel free to do a normal merge with a merge commit.

giodamelio:
Not sure exactly why the AppVeyor build failed, I don't think it handled the rebase well.

giodamelio:
@benmosher @jfmengels I just wanted to ping you guys and see about getting this merged and inquire about a new release, that could hopefully include this.

ljharb:
@giodamelio it'll need another rebase to remove the merge commit (the "update branch" button doesn't help)

giodamelio:
Yep, I shouldn't have used the "fix conflicts" button. Only causes me more issues :expressionless:. Fix incoming.

giodamelio:
Should be good to go.

giodamelio:
Hey guys, I totally forgot about this pull. I think it should be ready to go still.

ljharb:
@giodamelio would you rebase it on the command line, and we do need to figure out why the tests are failing

ljharb:
(it's possible that tests are failing on master, tho)

ljharb:
Took care of the rebase; tests are passing now.

giodamelio:
Thanks, I totally forgot about this again :). Merged after almost a year. :)

PR #731, opened 2017-01-27
Fix path to README.

eventualbuddha:


jfmengels:
Thanks a lot @eventualbuddha! :smile:

PR #732, opened 2017-01-28
Flow types support in the `order` rule

### LINKED ISSUE #645: Add type import support to import/order
Example:

```js
import React from 'react';
import reactCssModules from 'react-css-modules';
import styles from './static/styles.css';
import type {
  TopicType
} from '../../types';
```

Refer to https://flowtype.org/blog/2015/02/18/Import-Types.html

avaly:
Closes #645

~~PS: I also included a `yarn.lock` file: https://yarnpkg.com/en/docs/version-control~~

Please autosquash my fixup commits before merging. :smile:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ node_modules
 # generated output
 lib/
 **/.nyc_output/
+yarn.lock
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -20,6 +20,8 @@ import bar from './bar';
 import baz from './bar/baz';
 // 6. "index" of the current directory
 import main from './';
+// 7. Flow types
+import type { Element } from 'react';
 ```
 
 Unassigned imports are ignored, as the order they are imported in may be important.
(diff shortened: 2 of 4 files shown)


avaly:
I removed the `yarn.lock` file.

jfmengels:
Thanks a lot for your work on this @avaly!

Since the issue was created, Flow v0.38 came out and allowed importing both types and values in a single import statement, which sounds like a way better syntax to me and would make this new type less useful.

I described my concerns and propositions on the original issue, let me know what you think https://github.com/benmosher/eslint-plugin-import/issues/645#issuecomment-276196243

avaly:
@jfmengels I was aware of the new syntax in Flow 0.38, however I still feel this PR is relevant. Since the old syntax to import Flow types is still available, this rule should at least offer the opportunity to configure the code style some may want for their projects. Personally, I still prefer having all the Flow types imports separate from the normal imports.

avaly:
Any more thoughts on this? I would really like to have this available.

jfmengels:
The `order` rule is about the imported module, not what is being imported. I'm pretty sure that if we add the `type`/`flow` import type, then we'd soon have a request to order the type imports with the same logic (so you'd need subtypes like `flow parent`, `flow external`, ...), which would make sense because that is the purpose of this rule to start with, and things will start to get messy.
Having this new feature in Flow 0.38 is a fortunate and welcome solution to this problem IMO.

I don't want to tell you and other users how to import your types, but I really don't think this is the best way to handle this... I'm open to comments though.

Please note that if you really feel like going this path anyway, you always have the option to fork this plugin and create a new package with the forked order rule.

(cc @gajus who opened the original issue, if you feel like commenting)

avaly:
I'll close this PR and fork this rule into a separate package.

PR #737, opened 2017-02-05
Support `allow` option in `no-unassigned-import`, fix #671

### LINKED ISSUE #671: Add option to `no-unassigned-import` for allowed glob
When using the style and css webpack loaders, sometimes I want to be able to import a CSS file which will result in injecting the styles in the DOM (a side-effect) but this is acceptable in some use cases (I realize there are CSS modules, but importing things like bootstrap etc.)

So I suggest:

```javascript
"import/no-unassigned-import": [2, {
  "allowed": ["*.css"]
}]
```

Thoughts?

kevin940726:
Add `allow` option to `no-unassigned-import` to fix #671.

I basically follow the same logic from `no-extraneous-dependencies`. With the glob matching start from `process.cwd()`. I'm not sure that this is the best solution (maybe use regexp like webpack do?), but it works. Please let me know if there is something need to be improved. Big thanks!

- [x] write tests
- [x] update docs
- [x] make a note in change log

### DIFF
--- a/docs/rules/no-unassigned-import.md
+++ b/docs/rules/no-unassigned-import.md
@@ -10,7 +10,7 @@ This rule aims to remove modules with side-effects by reporting when a module is
 
 This rule supports the following option:
 
-`allow`: An Array of globs. The files that match any of these patterns would be ignored/allowed by the linter. This can be usefull for some build environment (e.g. css-loader in webpack).
+`allow`: An Array of globs. The files that match any of these patterns would be ignored/allowed by the linter. This can be useful for some build environments (e.g. css-loader in webpack).
 
 Note that the globs start from the where the linter is executed (usually project root), but not from each file that includes the source. Learn more in both the pass and fail examples below.
 
--- a/docs/rules/no-unassigned-import.md
+++ b/docs/rules/no-unassigned-import.md
@@ -6,11 +6,24 @@ With both CommonJS' `require` and the ES6 modules' `import` syntax, it is possib
 
 This rule aims to remove modules with side-effects by reporting when a module is imported but not assigned.
 
+### Options
+
+This rule supports the following option:
+
+`allow`: An Array of globs. The files that match any of these patterns would be ignored/allowed by the linter. This can be usefull for some build environment (e.g. css-loader in webpack).
+
+Note that the globs start from the where the linter is executed (usually project root), but not from each file that includes the source. Learn more in both the pass and fail examples below.
+
+
 ## Fail
 
 ```js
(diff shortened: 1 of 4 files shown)


docs/rules/no-unassigned-import.md
```diff
@@ -6,11 +6,24 @@ With both CommonJS' `require` and the ES6 modules' `import` syntax, it is possib
 
 This rule aims to remove modules with side-effects by reporting when a module is imported but not assigned.
 
+### Options
+
+This rule supports the following option:
+
+`allow`: An Array of globs. The files that match any of these patterns would be ignored/allowed by the linter. This can be usefull for some build environment (e.g. css-loader in webpack).
```
ljharb: `s/usefull/useful`, `s/environment/environments`

src/rules/no-unassigned-import.js
```diff
@@ -7,15 +9,40 @@ function report(context, node) {
   })
 }
 
+function testIsAllow(globs, filename, source) {
+  if (!Array.isArray(globs)) {
+    return false // default doens't allow any pattern
```
ljharb: `s/doens't/doesn't`, `s/pattern/patterns`

src/rules/no-unassigned-import.js
```diff
@@ -33,6 +60,7 @@ module.exports = {
           'devDependencies': { 'type': ['boolean', 'array'] },
           'optionalDependencies': { 'type': ['boolean', 'array'] },
           'peerDependencies': { 'type': ['boolean', 'array'] },
+          'allow': { 'type': 'array' },
```
ljharb: should this require that it be an array of strings specifically?

kevin940726:
Sorry for the typos, will change it next commit!

Yes, I think it should.

thiagodebastos:
Hey guys! Are we close to this being resolved? It looks like the only PR conflict is the `CHANGELOG.md`

ljharb:
@kevin940726 please rebase on the command line so as to avoid merge commits; if you have trouble, i can do it for you.

ljharb:
(also; tests are failing)

kevin940726:
My bad, didn't notice that. I'll try my best. :pensive:

kevin940726:
Is the tests failing on `master`? Or what have I done wrong 😢

ljharb:
@kevin940726 sorry for the delay here. I've rebased your branch for you and resolved the conflicts. Tests seem to pass locally, so let's see if this will get them to pass.

PR #744, opened 2017-02-13
Support scoped modules containing hyphens

rosswarren:
I was having problems with the order rule and scoped modules. It was returning NaN for the rank. I discovered the regex was not matching our scope name which contained hyphens.

This regex should match the full range of scoped module names I think. Please check it though.

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -27,7 +27,7 @@ function isExternalModule(name, settings, path) {
   return externalModuleRegExp.test(name) && isExternalPath(path, name, settings)
 }
 
-const scopedRegExp = /^@\w+\/\w+/
+const scopedRegExp = /^@[^\/]+\/[^\/]+/
 function isScoped(name) {
   return scopedRegExp.test(name)
 }
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -31,6 +31,9 @@ describe('importType(name)', function () {
   it("should return 'external' for scopes packages", function() {
     expect(importType('@cycle/core', context)).to.equal('external')
     expect(importType('@cycle/dom', context)).to.equal('external')
+    expect(importType('@some-thing/something', context)).to.equal('external')
+    expect(importType('@some-thing/something/some-module', context)).to.equal('external')
+    expect(importType('@some-thing/something/some-directory/someModule.js', context)).to.equal('external')
   })
 
   it("should return 'internal' for non-builtins resolved outside of node_modules", function () {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Yanked due to critical issue in eslint-module-utils with cache key resulting fro
 ### Fixed
 - attempt to fix crash in [`no-mutable-exports`]. ([#660])
 - "default is a reserved keyword" in no-maned-default tests by locking down babylon to 6.15.0 (#756, thanks @gmathieu)
+- support scoped modules containing non word characters
 
 
 ## [2.2.0] - 2016-11-07


src/core/importType.js
```diff
@@ -27,7 +27,7 @@ function isExternalModule(name, settings, path) {
   return externalModuleRegExp.test(name) && isExternalPath(path, name, settings)
 }
 
-const scopedRegExp = /^@\w+\/\w+/
```
ljharb: would this be simpler as `/^@[^\/]+\/[^\/]+`?

rosswarren: Yes, just tested this on our codebase and that works 👍  I have updated the pull request.

ljharb:
Instead of regexes, should we use the official https://github.com/npm/validate-npm-package-name ? It'll be a scoped package if it's an invalid old one but a valid new one, and we won't have to maintain our own logic.

rosswarren:
@ljharb I am not sure if that will help us with detection of scoped modules 🤔 . It will validate but not tell us what type it is.

ljharb:
It returns two booleans, "old" and "new" - new meaning, including scoped. If they're both true it's scoped, if just "old" is true it's unscoped, if both are false, it's invalid.

rosswarren:
@ljharb unfortunately this isn't how it works

capture

ljharb:
@rosswarren what about checking "new", and if it starts with an "@"?

That's an official npm package, and it'd really be nice to use that instead of messing around with regexps.

rosswarren:
@ljharb OK, good idea. I will get it done!

rosswarren:
@ljharb hmm, I tried to implement this but it seems that the `isScoped` function is passed the whole import for example `@scope/donthaveit/lib/foo` rather than just the package which would be `@scope/donthaveit` so you would still need to try and parse out the package name with some code or a regex.

rosswarren:
@ljharb would it be a problem to only check if the first character is an "@"?. From NPM docs "If a package's name begins with @, then it is a scoped package". Not sure we really need to check anything else.

ljharb:
@rosswarren Probably - but `@foo` isn't a scoped package, and could be a webpack alias - same with `@`.

rosswarren:
@ljharb do you have any further suggestion given my previous comment? Not sure how to proceed here.

tests/src/core/importType.js
```diff
@@ -31,6 +31,7 @@ describe('importType(name)', function () {
   it("should return 'external' for scopes packages", function() {
     expect(importType('@cycle/core', context)).to.equal('external')
     expect(importType('@cycle/dom', context)).to.equal('external')
+    expect(importType('@some-thing/something', context)).to.equal('external')
```
jfmengels: Could you add additional tests for the following cases? (and more if you think of some, better safe than sorry)
- `@something/some-thing`
- `@something/something/something`

rosswarren: have added 👍

ljharb:
You could use `fs` methods to split up the path, without using a regex, but that's probably fine to leave for a future PR.

@rosswarren would you mind rebasing, so as to remove all merge commits?

rosswarren:
@ljharb yeah that would require some more thinking, there is probably a better solution for this. However, right now of our code importing scoped modules is full of extra line breaks so I am anxious to see this fixed 😄 

And yeah no problem I rebased the branch 👍

afairb:
Hi. I see this PR is approved but not merged quite yet. Is there any chance it could get in soon as it's causing issues in my project when importing scoped modules? Thanks!

ljharb:
The Windows build is still failing.

danny-andrews:
Can we get this merged? I'm having this same issue.

danny-andrews:
@ljharb 
> You could use fs methods to split up the path, without using a regex, but that's probably fine to leave for a future PR.

Just out of curiosity, how would you do this?

ljharb:
@danny-andrews a combo of `path.parse`, splitting on `path.delimiter`, etc.

danny-andrews:
Siiiiick. Thanks for merging!

danny-andrews:
This looks like a good solution to this: https://github.com/mattdesl/require-package-name.

ljharb:
It doesn't seem to work for relative paths tho.

danny-andrews:
Brutal. Could use it in combination with https://www.npmjs.com/package/relative-require-regex?
 Maybe not worth it for solving such a simple problem doe.

Issue #756, opened 2017-02-28
no-named-default test failure


gmathieu:
@ntdb looks like a lot of PRs are failing because of the following error:
> default is a reserved word

Is it okay to disable [this test](https://github.com/gmathieu/eslint-plugin-import/blame/cjs-default/tests/src/rules/no-named-default.js#L17)?

ntdb:
Interesting... @jfmengels any thoughts on this? I'm hesitant to just disable a spec but obviously we need builds to run.

ljharb:
Can we pin down what dep update is causing the error?

gmathieu:
I've created a diff between the latest passing PR build [1849](https://ci.appveyor.com/project/benmosher/eslint-plugin-import/build/1849/job/113boykxinv96jfy) and my branch's build [1858](https://ci.appveyor.com/project/benmosher/eslint-plugin-import/build/1858).

```diff
--- passing.txt    Tue Feb 28 16:12:34 2017
+++ failing.txt    Tue Feb 28 16:12:37 2017
@@ -5,7 +5,7 @@
 | | `-- invariant@2.2.2
 | +-- babel-types@6.23.0
 | | `-- to-fast-properties@1.0.2
-| +-- babylon@6.15.0
+| +-- babylon@6.16.1
 | +-- lodash.assign@3.2.0
 | | +-- lodash._baseassign@3.2.0
 | | | `-- lodash._basecopy@3.0.1
@@ -62,8 +62,8 @@
 |   | | `-- is-buffer@1.1.4
 |   | +-- normalize-path@2.0.1
 |   | +-- object.omit@2.0.1
-|   | | +-- for-own@0.1.4
-|   | | | `-- for-in@0.1.6
+|   | | +-- for-own@0.1.5
+|   | | | `-- for-in@1.0.2
 |   | | `-- is-extendable@0.1.1
 |   | +-- parse-glob@3.0.4
 |   | | +-- glob-base@0.3.0
@@ -186,7 +186,7 @@
 +-- doctrine@1.5.0
 | +-- esutils@2.0.2
 | `-- isarray@1.0.0
-+-- eslint@3.15.0
++-- eslint@3.16.1
 | +-- babel-code-frame@6.22.0
 | | `-- js-tokens@3.0.1
 | +-- chalk@1.1.3
@@ -196,7 +196,7 @@
 | | `-- supports-color@2.0.0
 | +-- concat-stream@1.6.0
[...]

PR #757, opened 2017-03-01
Fix "default is a reserved keyword" error no-named-default tests

gmathieu:
#756

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Fixed
 - attempt to fix crash in [`no-mutable-exports`]. ([#660])
+- "default is a reserved keyword" in no-maned-default tests by locking down babylon to 6.15.0 (#756, thanks @gmathieu)
 
 
 ## [2.2.0] - 2016-11-07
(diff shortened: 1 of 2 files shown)


gmathieu:
Anyone know how to restart travis-ci? It's clearly hung.

ljharb:
I'm glad we figured this out - but I think if the parsing error is per-spec, that we should instead fix the code.

gmathieu:
I would recommend removing [this test](https://github.com/gmathieu/eslint-plugin-import/blame/cjs-default/tests/src/rules/no-named-default.js#L17) (possibly the rule?) since most environments throw an exception when assign variables to `default`.

#### Safari 10.0.3 (12602.4.8)
```
var { default } = { default: 'foo' }
SyntaxError: Cannot use abbreviated destructuring syntax for keyword 'default'.
```

#### Chrome 56.0.2924.87
```
var { default } = { default: 'foo' }
VM11560:1 Uncaught SyntaxError: Unexpected token }
```
```
var default = 'foo'
VM11582:1 Uncaught SyntaxError: Unexpected token default
```
#### Firefox 51.0.1
```
var { default } = { default: 'foo' }
SyntaxError: default is a reserved identifier
```

#### Node 6.9.1
```
var { default } = { default: 'foo' }
              ^
SyntaxError: Unexpected token }
```
```
var default = 'foo'
    ^^^^^^^
SyntaxError: Unexpected token default
```

@ljharb I'm following @ntdb's recommendation to leave the test in and downgrade the library. My only goal is to get tests to pass so I can unblock quite a few PRs (selfishly #754 😄)

PR #768, opened 2017-03-14
log a useful error in a module bug arises

### LINKED ISSUE #767: webpack resolver fails on CircleCI
Everything works great locally, but I kick it up to CircleCI & I get a few thousand errors saying:

`Resolve error: Cannot convert undefined or null to object`

example (under npm run lint)
https://circleci.com/gh/ParabolInc/action/2142

my settings: 
```
  "settings": {
    "import/parser": "babel-eslint",
    "import/resolver": {
      "webpack": {
        "config": "./webpack/prod.babel.js" // points to a file that babelifies my webpack config
      }
    },
  },
```

mattkrick:
is not the root cause of #767, but it at least makes it possible for folks to figure out real fast what is causing the error (use log because eslint swallows thrown errors)
fix #767

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -65,7 +65,12 @@ exports.resolve = function (source, file, settings) {
 
       log('Config path resolved to:', configPath)
       if (configPath) {
-        webpackConfig = require(configPath)
+        try {
+          webpackConfig = require(configPath)
+        } catch(e) {
+          console.log('Error resolving webpackConfig', e)
+          throw e
+        }
       } else {
         log("No config path found relative to", file, "; using {}")
         webpackConfig = {}


resolvers/webpack/index.js
```diff
@@ -65,7 +65,11 @@ exports.resolve = function (source, file, settings) {
 
       log('Config path resolved to:', configPath)
       if (configPath) {
-        webpackConfig = require(configPath)
+        try {
+          webpackConfig = require(configPath)
+        } catch(e) {
+          console.log('Error resolving webpackConfig', e)
```
ljharb: I think we should also rethrow the error; that way code execution continues to stop.

ljharb:
@mattkrick why close this? This is a good change.

mattkrick:
just cleaning up some PRs. I no longer use the package

ljharb:
I can't understand no longer linting imports, but I'll try to take this over the finish line if you're not interested in doing so.

PR #775, opened 2017-03-23
loosen the version of `rimraf` module

shinnn:
I think there is no reason to pin the `rimraf` version to `2.5.2` strictly.


shinnn:
Rebased from the latest master branch.

Issue #788, opened 2017-04-04
Eslint is slow when used with webpack


mieszko4:
I am wondering if it is possible to boost up performance when using eslint-plugin-import with webpack.

In my project:
* I have 4500 modules (`find node_modules/ -type f -name "package.json" | wc -l`)
* I have 500 files that I am eslinting
* I use 8 aliases in webpack config `resolve.alias`
* I use 8 loader rules in webpack config `module.rules`.

Eslinting lasts ~30s, and I can see that `import/no-unresolved` takes most of the time:
```
Rule                              | Time (ms) | Relative
:---------------------------------|----------:|--------:
import/no-unresolved              |  4983.197 |    28.2%
react/prop-types                  |  1694.653 |     9.6%
react/sort-comp                   |  1321.233 |     7.5%
react/no-multi-comp               |  1264.966 |     7.2%
react/require-render-return       |  1221.941 |     6.9%
import/no-named-as-default        |   665.776 |     3.8%
react/no-deprecated               |   524.655 |     3.0%
indent                            |   261.081 |     1.5%
import/no-named-as-default-member |   248.505 |     1.4%
react/jsx-curly-spacing           |   198.665 |     1.1%

```

Any suggestions on improvement?

ljharb:
If you disable the webpack aliases and the resolver config for them, does it go faster?

mieszko4:
I found the problem!
I am using `webpack.config.babel.js` and the problem is that it reads and transpiles the entire webpack config each time it resolves the rule. And there are lots of calls, in my case over 150 of them.

I suggest the config is read once and then cached. I will make a pr.

ljharb:
Glad you figured it out; seems like a webpack issue.

ljharb:
Wait, or are you saying that eslint-plugin-import reads your webpack config?

mieszko4:
Yes, `eslint-import-resolver-webpack` reads the config each time `resolve()` is called.
The improvement could be that it reads it once and `resolve()` calles the cached version.

ljharb:
Sounds like a great improvement, thanks!

mieszko4:
After some further investigation I concluded that caching it would not speed it up (too much) since `require()` that is used caches the call already. It is simply slow because of babel the first time it is called.
Each subsequent time it is quick as it uses the cached version already.

I've created a pr anyway as it moves out some common processing.

mieszko4:
[...]

PR #789, opened 2017-04-04
import/resolver move out webpack reading and processing (#788)

mieszko4:
This pr is related to https://github.com/benmosher/eslint-plugin-import/issues/788

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -49,7 +49,6 @@ exports.resolve = function (source, file, settings) {
     , configIndex = get(settings, 'config-index')
     , webpackConfig
     , resolveSync
-    , packageDir
 
   log('Config path from settings:', configPath)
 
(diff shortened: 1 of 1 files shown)


mieszko4:
Not sure why tests are failing. `master` is failing for me, too.

ljharb:
nodejs.org was down earlier today. I'll rerun.

mieszko4:
Not a problem of nodejs.org.
Locally my tests also fail for my commit and even for master commit.
I even tried to follow the commands from this successful build (https://ci.appveyor.com/project/benmosher/eslint-plugin-import/build/1696/job/5gwjncjqkyj0srii) that I've found using the same npm/node versions and it locally fails with 2 pending and 6 errors. I am on Linux.

benmosher:
as noted in #788, I don't know if this is a good idea as some things hold the plugin in memory for an extended period. would need to periodically reload or at least check file modified time / hash of contents for changes. the latter would still be a major perf gain.

also: looks like babel is reporting some things as syntax errors that weren't before, thus the test fails

ljharb:
@mieszko4 not sure if this is still relevant; but it needs a rebase. Please mark it as ready for review when it's so.

mieszko4:
@ljharb I will close it, It's been 4 years so I do not think it is relevant anymore.

PR #804, opened 2017-04-14
[New] Add `no-import-module-exports` rule

### LINKED ISSUE #760: add no-import-module-exports
Due to [this issue](https://github.com/webpack/webpack/issues/4039) with webpack tree-shaking, and the fact that some people cannot adopt `no-commonjs` quite yet, I think this would be really handy. I've written the rule for my own stuff and would be happy to add it here (or retrofit `no-commonjs` with options to cover this use case).

Here's my rule:

```javascript
'use strict'

module.exports = {
	meta: {
		docs: {
			description: 'Disallow import statements with module.exports',
			category: 'Best Practices',
			recommended: true,
		},
		fixable: 'code',
		schema: [],
	},
	create(context) {
		const importDeclarations = []
		let hasReported = false
		function report() {
			if (hasReported) {
				return
			}
			hasReported = true
			importDeclarations.forEach(node => {
				context.report({
					node,
					message: 'Cannot use import declarations in modules that export using CommonJS (`module.exports = \'foo\'` or `exports.bar = \'hi\'`)',
				})
			})
		}
		return {
			ImportDeclaration(node) {
				importDeclarations.push(node)
			},
			MemberExpression(node) {
				if (isModuleExports(node)) {
					report()
				}

				function isModuleExports(n) {
					return n.object.type === 'Identifier' && (/^(module|exports)$/).test(n.object.name)
				}
			},
		}
	},
}
```

And the associated tests:

```javascript
'use strict'

const { RuleTester } = require('eslint')
const rule = require('./no-import-module-exports')

const parserOptions = { ecmaVersion: 6, sourceType: 'module' }
const error = {
	message: 'Cannot use import declarations in modules that export using CommonJS (`module.exports = \'foo\'` or `exports.bar = \'hi\'`)',
	type: 'ImportDeclaration',
}

ruleTester.run('no-blockless-if', rule, {
	valid: [
		{
			code: `
				const thing = require('thing')
				module.exports = thing
			`,
			parserOptions,
		},
		{
			code: `
				import thing from 'otherthing'
				console.log(thing.module.exports)
			`,
			parserOptions,
		},
		{
			code: `
				import thing from 'other-thing'
				export default thing
			`,
			parserOptions,
		},
	],
	invalid: [
		{
			code: `
				import { stuff } from 'starwars'
				module.exports = thing
			`,
			errors: [error],
			parserOptions,
		},
		{
			code: `
				import thing from 'starwars'
				const baz = module.exports = thing
				console.log(baz)
			`,
			errors: [error],
			parserOptions,
		},
		{
			code: `
				import * as allThings from 'starwars'
				exports.bar = thing
			`,
			errors: [error],
			parserOptions,
		},
	],
})
```

Let me know what you all think :)

ttmarek:
Fixes #760 
Still a work in progress. 
Updates to the changelog and docs to follow.

### DIFF
--- a/docs/rules/no-import-module-exports.md
+++ b/docs/rules/no-import-module-exports.md
@@ -1,0 +1,74 @@
+# no-import-module-exports
+
+Reports the use of import declarations with CommonJS exports in any module
+except for the [main module](https://docs.npmjs.com/files/package.json#main).
+
+If you have multiple entry points or are using `js:next` this rule includes an
+`exceptions` option which you can use to exclude those files from the rule.
+
+## Options
+
+#### `exceptions`
+ - An array of globs. The rule will be omitted from any file that matches a glob
+   in the options array. For example, the following setting will omit the rule
+   in the `some-file.js` file.
+
+```json
+"import/no-import-module-exports": ["error", {
+    "exceptions": ["**/*/some-file.js"]
+}]
+```
+
+## Rule Details
+
+### Fail
+
+```js
+import { stuff } from 'starwars'
+module.exports = thing
+
+import * as allThings from 'starwars'
+exports.bar = thing
+
+import thing from 'other-thing'
+exports.foo = bar
+
+import thing from 'starwars'
+const baz = module.exports = thing
+console.log(baz)
+```
+
+### Pass
+Given the following package.json:
+
+```json
+{
+  "main": "lib/index.js",
+}
+```
+
+```js
+import thing from 'other-thing'
+export default thing
+
+const thing = require('thing')
+module.exports = thing
+
+const thing = require('thing')
+exports.foo = bar
+
+import thing from 'otherthing'
+console.log(thing.module.exports)
+
+// in lib/index.js
+import foo from 'path';
+module.exports = foo;
+
+// in some-file.js
+// eslint import/no-import-module-exports: ["error", {"exceptions": ["**/*/some-file.js"]}]
+import foo from 'path';
+module.exports = foo;
+```
+
+### Further Reading
+ - [webpack issue #4039](https://github.com/webpack/webpack/issues/4039)
--- a/src/rules/no-import-module-exports.js
+++ b/src/rules/no-import-module-exports.js
@@ -1,0 +1,66 @@
+import minimatch from 'minimatch';
+import path from 'path';
+import pkgUp from 'pkg-up';
+
+function getEntryPoint(context) {
+  const pkgPath = pkgUp.sync(context.getFilename());
+  return require.resolve(path.dirname(pkgPath));
+}
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow import statements with module.exports',
+      category: 'Best Practices',
+      recommended: true,
+    },
+    fixable: 'code',
+    schema: [
+      {
+        'type': 'object',
+        'properties': {
+          'exceptions': { 'type': 'array' },
+        },
+        'additionalProperties': false,
+      },
+    ],
+  },
+  create(context) {
+    const importDeclarations = [];
+    const entryPoint = getEntryPoint(context);
+    const options = context.options[0] || {};
+    let alreadyReported = false;
+
+    function report(node) {
+      const fileName = context.getFilename();
+      const isEntryPoint = entryPoint === fileName;
+      const isIdentifier = node.object.type === 'Identifier';
+      const hasKeywords = (/^(module|exports)$/).test(node.object.name);
+      const isException = options.exceptions &&
+        options.exceptions.some(glob => minimatch(fileName, glob));
+
+      if (isIdentifier && hasKeywords && !isEntryPoint && !isException) {
+        importDeclarations.forEach(importDeclaration => {
+          context.report({
+            node: importDeclaration,
+            message: `Cannot use import declarations in modules that export using ` +
+              `CommonJS (module.exports = 'foo' or exports.bar = 'hi')`,
+          });
+        });
+        alreadyReported = true;
+      }
+    }
+
+    return {
+      ImportDeclaration(node) {
+        importDeclarations.push(node);
+      },
+      MemberExpression(node) {
+        if (!alreadyReported) {
+          report(node);
+        }
+      },
+    };
+  },
+};
--- a/tests/src/rules/no-import-module-exports.js
+++ b/tests/src/rules/no-import-module-exports.js
@@ -1,0 +1,101 @@
+import path from 'path';
+import { RuleTester } from 'eslint';
+
+import { test } from '../utils';
+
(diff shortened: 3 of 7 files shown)


ttmarek:
@ljharb 
sure...just to be clear, you're thinking we should ensure that the following is invalid:

```js
    test({
      code: `
        import thing from 'other-thing'
        exports.foo = bar
      `,
      errors: [error],
    })
```

and the following would be valid:

```js
    test({
      code: `
        const thing = require('thing')
        exports.foo = bar
      `,
    }),
```

ljharb:
Yes, exactly

src/rules/no-import-module-exports.js
```diff
@@ -0,0 +1,72 @@
+import minimatch from 'minimatch'
+import path from 'path'
+import readPkgUp from 'read-pkg-up'
+
+function getEntryPoint(context) {
+  try {
```
ljharb: I'm wondering if it'd be easier to just use `require.resolve(pathToPackageJSON)` here, which reads "main" for you?

ttmarek: hmm...I'm not sure I follow.
From what I understand `require.resolve` gives you the absolute path to a given module.
So if I wanted the path to `minimatch`'s package.json I could do:
```js
require.resolve('minimatch/package.json')
```
But I can't use it to get an absolute path to the current module's package.json...Or can I? 

By the way, I borrowed this approach from [no-extraneous-dependencies](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-extraneous-dependencies.js). If we come up with a better solution, it might not be a bad idea to implement it there as well.

ljharb: If you require.resolve a dir path that contains a package.json, it gives you the path to the "main"

ljharb: I'm suggesting finding the package.json first, and then resolving that dir. Not sure if it will work or not here.

ttmarek: I see what you mean, let me take a looksee

ttmarek: The `require.resolve` approach seems to work well. The only thing to take into consideration is the use of `read-pkg-up` for finding the path to the nearest `package.json`. Using `read-pkg-up` the implementation would look something like:
```js
import path from 'path'
import readPkgUp from 'read-pkg-up'

function getEntryPoint(context) {
    const pkg = readPkgUp.sync({ cwd: context.getFilename() })
    return require.resolve(path.dirname(pkg.path))
}
```

But since all we need is the path to the nearest `package.json` (we don't actually need to parse it) we could use the [`pkg-up`](https://www.npmjs.com/package/pkg-up) lib instead. In which case the implementation would look like:

```js
import path from 'path'
import pkgUp from 'pkg-up'

function getEntryPoint(context) {
  const pkgPath = pkgUp.sync(context.getFilename())
  return require.resolve(path.dirname(pkgPath))
}
```

So, we can:
A) Stick to the current implementation where we parse the `package.json`, and create the path to "main" ourselves.
B) Use `read-pkg-up` to find the path to the nearest `package.json`, then use `require.resolve` to get the path to `main`.
C) Install and use `pkg-up` to find the path to the nearest `package.json`, then use `require.resolve` to get the path to `main`.

I prefer B and C over A....and lean towards C, but I'm not sure what the cost would be of adding another dependency. What do you think?

ljharb: pkg-up is already a dep of read-pkg-up, and the cost of adding dependencies to a dev dependency is virtually zero. I'd say C.

However, I just realized my comment on the original issue does also add that it should probably support `jsnext:main`, and for that, you'd need A or B.

I'd still lean towards C for the primary case.

ttmarek: C it is. Here's the [diff](https://github.com/benmosher/eslint-plugin-import/pull/804/commits/e2a8ccb337a373e71823dcdaec3f76fdc085f7ae).

Those using `jsnext:main` can always add their second entry to the exceptions in the rule options. I will add something to that effect in the docs.

src/rules/no-import-module-exports.js
```diff
@@ -0,0 +1,66 @@
...
+    const options = context.options[0] || {}
+    let alreadyReported = false
+
+    function report(node) {
+      const fileName = context.getFilename()
+      const isEntryPoint = entryPoint === fileName
+      const isIdentifier = node.object.type === 'Identifier'
+      const hasKeywords = (/^(module|exports)$/).test(node.object.name)
+      const isException = options.exceptions
+        ? options.exceptions.some(glob => minimatch(fileName, glob))
```
ljharb: this could be `options.exceptions && options.exceptions.some(…)`, but this is fine too :-)

docs/rules/no-import-module-exports.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### `exceptions`
+An array of globs. The rule will be omitted from any file that matches a glob in
+the options array.
+
+For example, the following setting will omit the rule in the `some-file.js` file.
+
+```json
+"import/no-import-module-exports": ["error", {
+    "exceptions": ['**/*/some-file.js']
```
ljharb: json needs double quotes

docs/rules/no-import-module-exports.md
```diff
@@ -0,0 +1,72 @@
...
+
+import thing from 'otherthing'
+console.log(thing.module.exports)
+
+/* in lib/index.js */
+import foo from 'path';
+module.exports = foo;
+
+/* in some-file.js */
+/* eslint import/no-import-module-exports: ["error", {"exceptions": ['**/*/some-file.js']}] */
```
ljharb: these should also probably be double quotes, but also, the `*/` actually does end the comment, so this should probably be a `//` comment :-)

tests/src/rules/no-import-module-exports.js
```diff
@@ -93,16 +93,6 @@ ruleTester.run('no-import-module-exports', rule, {
         import foo from 'path';
         module.exports = foo;
       `,
-      // When the file does not match the entry point defined in package.json
-      // See tests/files/package.json
```
ljharb: why was this test removed?

ttmarek: Ahh....the four tests above this one are already testing that the rule applies for files that aren't "main". By default the `test` util sets the default file to `eslint-plugin-import/tests/files/foo.js`. Which isn't equal to the "main" I set up in `eslint-plugin-import/tests/files/package.json`. It seemed a bit redundant to me...but I can add it back if you prefer.

ttmarek:
@benmosher @jfmengels any thoughts?

ttmarek:
🎵 all I want for christmaaas 🎵...  is for this to be merged.  🎅

ljharb (approved):

PR #827, opened 2017-05-10
import/extensions should have a ignorePackages option.  Fixes #414

### LINKED ISSUE #414: import/extensions should have a ignorePackages option

collinsauve:
For issue #414

Unfortunately I cannot run the tests from this repo on my machine (lots of unexplained failures) so I will have trouble writing additional tests for this new behavior.

### DIFF
--- a/docs/rules/extensions.md
+++ b/docs/rules/extensions.md
@@ -6,7 +6,7 @@ In order to provide a consistent use of file extensions across your code base, t
 
 ## Rule Details
 
-This rule either takes one string option, one object option, or a string and an object option. If it is the string `"never"` (the default value), then the rule forbids the use for any extension. If it is the string `"always"`, then the rule enforces the use of extensions for all import statements.
+This rule either takes one string option, one object option, or a string and an object option. If it is the string `"never"` (the default value), then the rule forbids the use for any extension. If it is the string `"always"`, then the rule enforces the use of extensions for all import statements. If it is the string `"ignorePackages"`, then the rule enforces the use of extensions for all import statements except package imports.
 
 By providing an object you can configure each extension separately, so for example `{ "js": "always", "json": "never" }` would always enforce the use of the `.js` extension but never allow the use of the `.json` extension.
 
@@ -84,6 +84,30 @@ import Component from './Component.jsx'
 import express from 'express/index.js';
 
 import * as path from 'path';
+```
+
+The following patterns are considered problems when configuration set to "ignorePackages":
+
+```js
+import foo from './foo';
+
+import bar from './bar';
+
+import Component from './Component'
+
+```
+
+The following patterns are not considered problems when configuration set to "ignorePackages":
+
+```js
+import foo from './foo.js';
+
+import bar from './bar.json';
+
+import Component from './Component.jsx'
+
+import express from 'express';
+
 ```
 
 ## When Not To Use It
--- a/docs/rules/extensions.md
+++ b/docs/rules/extensions.md
@@ -41,7 +41,7 @@ import foo from './foo.js';
 
 import bar from './bar.json';
 
-import Component from './Component.jsx'
+import Component from './Component.jsx';
 
 import express from 'express/index.js';
 ```
@@ -53,7 +53,7 @@ import foo from './foo';
 
 import bar from './bar';
 
-import Component from './Component'
+import Component from './Component';
 
 import express from 'express/index';
 
@@ -67,7 +67,7 @@ import foo from './foo';
 
 import bar from './bar';
 
-import Component from './Component'
+import Component from './Component';
 
 import express from 'express';
 ```
@@ -79,7 +79,7 @@ import foo from './foo.js';
 
 import bar from './bar.json';
 
-import Component from './Component.jsx'
+import Component from './Component.jsx';
 
 import express from 'express/index.js';
 
@@ -93,7 +93,7 @@ import foo from './foo';
 
 import bar from './bar';
 
-import Component from './Component'
+import Component from './Component';
 
 ```
 
@@ -104,7 +104,7 @@ import foo from './foo.js';
 
 import bar from './bar.json';
 
-import Component from './Component.jsx'
+import Component from './Component.jsx';
 
 import express from 'express';
 
--- a/docs/rules/extensions.md
+++ b/docs/rules/extensions.md
@@ -110,6 +110,17 @@ import express from 'express';
 
 ```
 
+The following patterns are not considered problems when configuration set to `[ 'always', {ignorePackages: true} ]`:
+
+```js
+import Component from './Component.jsx';
+
+import baz from 'foo/baz.js';
+
+import express from 'express';
+
+```
+
 ## When Not To Use It
 
 If you are not concerned about a consistent usage of file extension.
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -1,13 +1,55 @@
 import path from 'path'
-import has from 'has'
 
 import resolve from 'eslint-module-utils/resolve'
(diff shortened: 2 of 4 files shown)


docs/rules/extensions.md
```diff
@@ -86,6 +86,30 @@ import express from 'express/index.js';
 import * as path from 'path';
 ```
 
+The following patterns are considered problems when configuration set to "ignorePackages":
```
ljharb: let's add an example of `require('foo/bar.js')` still being required with both "always" and "ignorePackages" set.

src/rules/extensions.js
```diff
@@ -46,9 +46,10 @@ module.exports = {
       typeof configuration === 'object' ? configuration : context.options[1]
     )
 
-    function isUseOfExtensionRequired(extension) {
+    function isUseOfExtensionRequired(extension, isModule) {
       if (!has(modifiers, extension)) { modifiers[extension] = defaultConfig }
-      return modifiers[extension] === 'always'
+      const modifier = modifiers[extension]
+      return (!isModule && modifier === 'ignorePackages') || modifier === 'always'
```
ljharb: it should not be ignored just because it's an external package, rather only when it's the "main" of an external package.

In other words, `require('foo')` and `require('@foo/bar')` should be ignored, but subpaths under these should not.

collinsauve:
@ljharb object options are already taken by the extensions.  Adding anything to the object options will create ambiguity.

ljharb:
That's true. We could allow for a separate object, though, in either order, and later, migrate the extensions options to an "extensions" key in that new object, and deprecate/remove the top-level extensions config object.

src/rules/extensions.js
collinsauve: Good catch!  I'm not well versed in all of the possible import permutations, but here's a regex I just made up.  Can you tell me if I'm missing something?

```
^(([\w]((?!\/).)*)|(@[^\/]+\/?[^\/]+))$
```

This matches the following:

```
foo
foo2
foo.js
@foo
@foo2
@foo.js
@foo/bar
@foo2/bar
@foo2.js/bar
@foo2.js/bar.js
```

And will not match the following:

```
foo/bar
foo2/bar
foo/bar2
foo/bar.js
foo/bar/hello
@foo2/bar/hello
@foo2/bar/hello.js
```

collinsauve:
Say we have a new object type (for now called `newProperties` ) with props like `extensions`, and `ignorePackages`.  We now need to add a new item to `meta.schema.anyOf`.  

As far as I understand this schema system, since `patternProperties` catches all objects we can't use args `[newProperties]`.  For the same reason we also cant use `[string, newProperties]`.  We need to use `[string, patternProperties, newProperties]` or `[patternProperties, newProperties]`.  

Then what's the migration path?  Perhaps deprecate `[patternProperties]` and `[string, patternProperties]` in verion N and add support for `[newProperties]` and `[string, newProperties]` in version N+1?

src/rules/extensions.js
ljharb: I'd really prefer to avoid complex regexes if we can, but your matches look good.

ljharb:
ok, so the current schema allows any of:
1. one item, "always" or "never"
1. one item, an object of "pattern properties"
1. two items, "always/never" and an object of "pattern properties".

Indeed, this is problematic, and "patternProperties" should never have had top-level configurable keys.

It seems like the migration path is the following schema:

1. one item, "always" or "never"
1. **new** two items, "always/never" and an object with a "pattern" key (containing pattern properties)
1. **new** two items, "always/never" and an object with an "ignorePackages" key
1. **new** two items, "always/never" and an object with an "ignorePackages" key and a "pattern" key
1. **new** one item, an object with an "ignorePackages" key
1. **new** one item, an object with an "ignorePackages" key and a "pattern" key
1. one item, an object of "pattern properties"
1. two items, "always/never" and an object of "pattern properties".

And then, we'd deprecate and eventually remove the "object of pattern properties" schemas (the last two above), and when they were removed, we'd be able to collapse the 5 new schema types into one or two.

src/rules/extensions.js
collinsauve: This regex is probably better understood if I split it:

```
function isMainImportOrScopedMainImport(importPath) {
  const mainImportRegExp = /^[\w]((?!\/).)*$/
  const scopedMainImportRegExp = /^@[^\/]+\/?[^\/]+$/
  return import.test(mainImportRegExp) || import.test(scopedMainImportRegExp)
}
```

This probably belongs in `importType.js`, and I might split it into 2 functions if I can think of the names for them :)

collinsauve:
Thanks for that @ljharb, makes sense now.  I will try to implement your suggestions tonight or tomorrow.

collinsauve:
I can't really work on building tests for this this (to fix coverage issue) due to #838

millerized:
Hi, @collinsauve @ljharb – great work here. I am not a formal contributor yet, but I am curious if there is anything I can do help get this moving again?

🍻

ljharb:
@millerized if you can contribute a commit to improve coverage on top of this PR, and link it here, I can push it to the branch - I think that's the main thing that's missing.

jdalton:
Awesome PR! Was just looking for a way to configure this.

millerized:
Hi, @ljharb @collinsauve –  I’ve rebased this branch with upstream to fix conflicts and added some test coverage around the `ingnorePackages` option. Please see https://github.com/millerized/eslint-plugin-import/pull/1. Please let me know if there is still missing coverage and I will get it in asap. 🍻

ljharb:
Thanks! I've updated the PR with those commits, and will review later today.

millerized:
>Thanks! I've updated the PR with those commits, and will review later today.

Thanks, @ljharb!

docs/rules/extensions.md
ljharb: this still needs an example, and a matching test

millerized: Thanks for the reminder, @ljharb. I beleive [these commits](https://github.com/millerized/eslint-plugin-import/pull/1/files/00062fbdcf6e3eaf492fcbb4020e9b35239fdf05..59b9dfd2f52cec10b802d2d8f355552cb4002afb) should suffice. lmk what you think.

millerized:
>[this still needs an example, and a matching test](https://github.com/benmosher/eslint-plugin-import/pull/827#discussion_r148113887)

Thanks for the reminder, @ljharb. I beleive [these commits](https://github.com/millerized/eslint-plugin-import/pull/1/files/00062fbdcf6e3eaf492fcbb4020e9b35239fdf05..59b9dfd2f52cec10b802d2d8f355552cb4002afb) should suffice. lmk what you think.

ljharb:
I'll merge this tomorrow if there's no objections.

millerized:
>continuous-integration/appveyor/pr — AppVeyor build failed

Unsure about the AV CI failure. It appears to be a false negative. Perhaps it needs a re-run, @ljharb. 

>Coverage decreased (-0.01%) to 96.082% when pulling 81d3b36 on collinsauve:issue414 into eca7b4d on benmosher:master.

Coverage decreasing seems odd.

ljharb:
appveyor is passing, but coveralls is throwing errors; you can ignore it.

millerized:
>appveyor is passing, but coveralls is throwing errors; you can ignore it.

Gtk. Thanks, @ljharb. Any last items to address or are we ready to merge?

PR #840, opened 2017-05-19
eslint-module-utils: filePath in parserOptions

sompylasar:
Refs https://github.com/benmosher/eslint-plugin-import/issues/839

### DIFF
--- a/utils/parse.js
+++ b/utils/parse.js
@@ -21,6 +21,10 @@ exports.default = function parse(path, content, context) {
 
   // always attach comments
   parserOptions.attachComment = true
+  
+  // provide the `filePath` like eslint itself does, in `parserOptions`
+  // https://github.com/eslint/eslint/blob/3ec436eeed0b0271e2ed0d0cb22e4246eb15f137/lib/linter.js#L637
+  parserOptions.filePath = path
 
   // require the parser relative to the main module (i.e., ESLint)
   const parser = moduleRequire(parserPath)


sompylasar:
@ljharb OK, will add, what specifically? Assert the path to be actually passed?

ljharb:
@sompylasar that's ideal, yes

sompylasar:
@ljharb 
1. I need a spy to intercept parser call in the test. I've implemented a naive one in 6 lines, but if you'd like to import a library like `sinon` or `chai-spies` instead, let me know.
2. I've bumped the `eslint-module-utils` version to `2.1.0`, added lines to both CHANGELOGs, and updated the `eslint-plugin-import` dependency version of `eslint-module-utils`.

sompylasar:
Now CI fail with `No compatible version found: eslint-module-utils@^2.1.0`, how do you update these versions? @ljharb

utils/CHANGELOG.md
```diff
@@ -3,9 +3,13 @@ All notable changes to this module will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
-## v2 - 2016-11-07
+## v2.1.0 - 2017-05-20
```
ljharb: The version number and date aren't decided yet; If you want to call this section "Unreleased" that'd be helpful :-)

utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-module-utils",
-  "version": "2.0.0",
+  "version": "2.1.0",
```
ljharb: please revert this; package.json should never be bumped in a PR - only in a commit directly made to master.

ljharb:
It'd have to be `npm publish`ed - but that won't happen until after it's merged.

sompylasar:
@ljharb Reverted: https://github.com/benmosher/eslint-plugin-import/pull/840/commits/7712ce132ec38e38e0c756695efd10df7c2c3eeb

sompylasar:
> I'd probably bring in sinon (more deps is a good thing), but this is fine for now.

👍 I'm for sinon, too. I'd also use chai assert instead of expect (too much magic), but it was already there, so consistency...

sompylasar:
@benmosher Updated, PTAL. Replaced hand-made spy with `sinon.spy`.

tests/src/core/parse.js
```diff
@@ -21,4 +22,37 @@ describe('parse(content, { settings, ecmaFeatures })', function () {
       .not.to.throw(Error)
   })
 
+  it('passes expected parserOptions to custom parser', function () {
+    const parseSpy = sinon.spy()
+    const parserOptions = { ecmaFeatures: { jsx: true } }
+    require('./parseStubParser').parse = parseSpy
```
ljharb: require has a cache, so you get the same value out of it every time - any reason this isn't required at the top level?

sompylasar: Right, can pull to top. Both this and resolve. Just a habit of having self-contained tests.

sompylasar: Done: https://github.com/benmosher/eslint-plugin-import/commit/57327424f0a3fb3c37e7e8c8199f6f8558d8e837

tests/src/core/parse.js
```diff
@@ -1,11 +1,14 @@
...
 import { expect } from 'chai'
+import sinon from 'sinon'
 import parse from 'eslint-module-utils/parse'
 
 import { getFilename } from '../utils'
 
 describe('parse(content, { settings, ecmaFeatures })', function () {
   const path = getFilename('jsx.js')
+  const parseStubParser = require('./parseStubParser')
+  const parseStubParserPath = require.resolve('./parseStubParser')
```
ljharb: requires should generally all only be static and at the top level of the file - can these be moved one level higher?

sompylasar: @ljharb This `require` was left here like that on purpose, this is not just a dependency like any other, this is a stub which must be in a file that can be literally `require`d because of the way `eslint`, and following it, `eslint-import-resolver`, resolves the parser function via `require` function (and we can do nothing about that).

sompylasar: @ljharb For the same reason, the line above: `const path = getFilename('jsx.js')` could be moved up, but it's not.

ljharb: I'm not sure I understand, but this can stay where it is.

sompylasar: In other words, `require('./parseStubParser')` and `require.resolve('./parseStubParser')` should sit next to each other, they are closely related in this test.

utils/parse.js
```diff
@@ -22,6 +22,10 @@ exports.default = function parse(path, content, context) {
   // always attach comments
   parserOptions.attachComment = true
 
+  // provide the `filePath` like eslint itself does, in `parserOptions`
+  // https://github.com/eslint/eslint/blob/3ec436ee/lib/linter.js#L637
+  parserOptions.filePath = path
```
ljharb: Perhaps the path needs to be path.normalized?

sompylasar: Maybe. What is the error case? The serialized parserOptions participates in some cache key, right? Where is it?

benmosher: The internal parser has a cache that is a hash of the parse settings and the module path. So I'm guessing the path must be relative. 

An absolute path would work, or the cache key could drop the file path key. 

I am not at a computer right now but I think the cache access is in utils/resolve.js

benmosher: Sorry, not thinking -- relative would be fine as long as it's all against cwd. So maybe path.normalize would be fine

sompylasar: Thanks for the pointers, I'll look into that and suggest a change and tests when I get to my computer (afk, too).

PR #843, opened 2017-05-24
Performance increase for no-absolute-path

jseminck:
See my reply here for a detailed analysis on why I made this change: https://github.com/benmosher/eslint-plugin-import/issues/803#issuecomment-303754437

I remove the call to `resolve()` in `no-absolute-path` rule. That should make this rule faster when it is used as a standalone rule or with other rules that are not dependent on having to resolve file paths.

As I currently see no reason to have to call `resolve()` for this rule at all. However, I am new to the code, so I may be wrong. :)

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -8,10 +8,6 @@ function constant(value) {
   return () => value
 }
 
-function isAbsolute(name) {
-  return name.indexOf('/') === 0
-}
-
 export function isBuiltIn(name, settings) {
   const extras = (settings && settings['import/core-modules']) || []
   return builtinModules.indexOf(name) !== -1 || extras.indexOf(name) > -1
@@ -50,7 +46,6 @@ function isRelativeToSibling(name) {
 }
 
 const typeTest = cond([
-  [isAbsolute, constant('absolute')],
   [isBuiltIn, constant('builtin')],
   [isExternalModule, constant('external')],
   [isScoped, constant('external')],
--- a/src/rules/no-absolute-path.js
+++ b/src/rules/no-absolute-path.js
@@ -1,10 +1,13 @@
-import importType from '../core/importType'
 import isStaticRequire from '../core/staticRequire'
 
 function reportIfMissing(context, node, name) {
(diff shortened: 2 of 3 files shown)


jseminck:
@benmosher, what do you think of this change? If it's not useful/needed, then I will close the PR. Thanks!

src/core/importType.js
```diff
@@ -8,10 +8,6 @@ function constant(value) {
   return () => value
 }
 
-function isAbsolute(name) {
```
benmosher: instead of removing this, could you export it and then import it in `no-absolute-path`?

src/core/importType.js
```diff
@@ -50,7 +46,6 @@ function isRelativeToSibling(name) {
 }
 
 const typeTest = cond([
-  [isAbsolute, constant('absolute')],
```
benmosher: also keep this here, needed for `import/order` AFAIK

src/rules/no-absolute-path.js
```diff
@@ -1,12 +1,15 @@
...
 
-function reportIfMissing(context, node, name) {
-  if (importType(name, context) === 'absolute') {
+function reportIfAbsolute(context, node, name) {
+  if (isAbsolute(name)) {
     context.report(node, 'Do not import modules using an absolute path')
   }
 }
 
+function isAbsolute(name) {
```
benmosher: so this is replaced with `import { isAbsolute } from '../core/importType'`

tests/src/core/importType.js
```diff
@@ -8,12 +8,6 @@ import { testContext } from '../utils'
 describe('importType(name)', function () {
   const context = testContext()
 
-  it("should return 'absolute' for paths starting with a /", function() {
```
benmosher: and then put these tests back 😁

jseminck:
@benmosher I missed your comments. Do you still want to make the suggested refactoring or was it done in another PR/commit? 

Sorry - been a bit busy due to holidays and other priorities - and somehow I completely missed your messages.

PR #845, opened 2017-05-25
[docs] Fix `newline-after-import` example

daltones:
Fix missing error level in “Example options usage” and add JS highlighting.

### DIFF
--- a/docs/rules/newline-after-import.md
+++ b/docs/rules/newline-after-import.md
@@ -73,9 +73,8 @@ const FOO = 'BAR'
 
 
 ## Example options usage
-```
+```json
 {
-  ...
   "rules": {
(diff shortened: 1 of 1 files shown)


docs/rules/newline-after-import.md
```diff
@@ -72,11 +72,11 @@ const FOO = 'BAR'
 
 
 ## Example options usage
-```
+```js
```
ljharb: eslint config should really be json

docs/rules/newline-after-import.md
```diff
@@ -72,11 +72,11 @@ const FOO = 'BAR'
...
 {
-  ...
+  // ...
```
ljharb: and we can just remove this line then

Issue #838, opened 2017-05-18
Tests failing


collinsauve:
When running `npm run test` from master I get 50 errors that look like this:

```
45) no-unresolved syntax verification valid import json from "./data.json":

      AssertionError: Should have no errors but had 1: [ { ruleId: 'no-unresolved syntax verification',
    severity: 1,
    message: 'Casing of ./data.json does not match the underlying filesystem.',
    line: 1,
    column: 18,
    nodeType: 'Literal',
    source: 'import json from "./data.json"' } ]
      + expected - actual

      -1
      +0

      at testValidTemplate (node_modules\eslint\lib\testers\rule-tester.js:423:20)
      at Context.RuleTester.it (node_modules\eslint\lib\testers\rule-tester.js:548:25)
```

I am running:

* Windows 10 (v1607 build 14393.447)
* NPM v4.1.2
* Node v7.7.4
* GNU bash, version 4.3.46(2)-release (x86_64-pc-msys)

benmosher:
can you confirm that `data.json` case does actually match?

collinsauve:
Is it supposed to be all lower-case?

```
$ find . -iname data.json
./resolvers/node/test/data.json
./tests/files/data.json
```

benmosher:
yep. super weird. I wonder if it's a Node 7 issue? I have not tested past Node v6

benmosher:
oh actually: does your shell `pwd` path match the filesystem? there was an old Windows issue where if you navigated case-insensitive to your code folder it would fail because the absolute bath (based on `pwd`) didn't match FS.

benmosher:
If #846 passes I will close, assuming it's that.

collinsauve:
> does your shell pwd path match the filesystem? 

That was it!  All tests are passing now - thank you so much!

PR #846, opened 2017-05-25
update CI to build on Node 6+7

benmosher:
investigating #838 + probably a good idea anyway

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - 4
   - 6
+  - 7
 
 os:
   - linux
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - 4
```
ljharb: um, why is node 4 dropped? eslint still supports node 4; this makes it a breaking change.

ljharb: i strongly recommend continuing to test on node 4 for a very long time.

benmosher: fair enough. I don't expect to explicitly drop it, it just adds a bunch of build time.

would be fine to add it back

PR #847, opened 2017-05-25
Update no-nodejs-modules rule to allow ignored files

deecewan:
I have a primarily web-based project, however in my config files, etc, I have access to the node runtime and sometimes need it (to read files, etc).

I want to be able to use node modules in these files, but not anywhere else in the project.

This extension to the rule does that.

### DIFF
--- a/src/rules/no-nodejs-modules.js
+++ b/src/rules/no-nodejs-modules.js
@@ -1,3 +1,5 @@
+import path from 'path'
+import minimatch from 'minimatch'
 import importType from '../core/importType'
 import isStaticRequire from '../core/staticRequire'
 
@@ -5,6 +7,23 @@ function reportIfMissing(context, node, allowed, name) {
   if (allowed.indexOf(name) === -1 && importType(name, context) === 'builtin') {
     context.report(node, 'Do not import Node.js builtin module "' + name + '"')
   }
+}
+
+function isIgnored(ignored, filename) {
+  return ignored.some(c => (
+    minimatch(filename, c) ||
+    minimatch(filename, path.join(process.cwd(), c))
+  ))
+}
+
+function guaranteeArray(item) {
+  if (item instanceof Array) {
+    return item
+  }
+  if (item) {
+    return [item]
+  }
+  return []
 }
 
 module.exports = {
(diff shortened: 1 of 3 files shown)


ljharb:
Wouldn't it be preferable to handle this by disabling the rule in a `.eslintrc` or eslint override comment?

deecewan:
Is it possible to disable for certain files in a `.eslintrc`?

It could be done with a directive. I try and discourage directives in our codebase. There are other rules that allow an 'ignore' globbing syntax, like `no-extraneous-dependencies`.

ljharb:
My suggestion would be to make the "node files" be the only thing in their directory; it'd be confusing to have a node file and a browser file right next to each other.

I agree that there's precedent for this option, as you say.

src/rules/no-nodejs-modules.js
```diff
@@ -7,6 +9,23 @@ function reportIfMissing(context, node, allowed, name) {
...
 
+function isIgnored(ignored, filename) {
+  return ignored.some(c => (
+    minimatch(filename, c) ||
+    minimatch(filename, path.join(process.cwd(), c))
+  ))
+}
+
+function guaranteeArray(item) {
+  if (item instanceof Array) {
```
ljharb: `intsanceof` is never reliable; `Array.isArray(item)` is the proper way to check.

However, this entire function can be replaced with `[].concat(item || [])`, i believe

deecewan:
The only problem with that is for things like `webpack.config.js` which most of the time sit at the top level right next to the `.eslintrc`. None of the files outside of `src/`, for example, should complain about node dependencies.

Maybe I've never seen it before, but if I put another `.eslintrc` inside of `src/`, will it have to completely duplicate the existing one, sans rule? or will it inherit the parent?

src/rules/no-nodejs-modules.js
deecewan: that's brilliant. thanks for this.

ljharb:
`.eslintrc` merges upwards, unless it has `"root": true` - but you're correct that things like webpack.config.js are tricky without globs or override comments.

src/rules/no-nodejs-modules.js
```diff
@@ -7,21 +9,53 @@ function reportIfMissing(context, node, allowed, name) {
...
+    minimatch(filename, path.join(process.cwd(), c))
+  ))
+}
+
 module.exports = {
   meta: {
     docs: {},
+    schema: [{
+      type: 'object',
+      properties: {
```
ljharb: Awesome! Could we also set "allow additional properties" or whatever it's called to false?

deecewan: for you? anything.

deecewan:
can I get an update on this please? it'd be great to send it live.

deecewan:
you're right, that is confusing.

Also, if we're honest, I did not realise that ESLint 4 had shipped glob support. That'll solve my issue, I think.

PR #855, opened 2017-05-31
build on node v4, again

benmosher:


benmosher:
ah sweet, I haven't been tracking the release schedule but that makes sense to me

PR #858, opened 2017-06-01
Fix issue #720

### LINKED ISSUE #720: False positives in import/no-unresolved when command is run from path not correctly cased
I was getting a lot of false positives like this:
```
   1:34  error  Casing of react does not match the underlying filesystem                                            import/no-unresolved
   2:25  error  Casing of react-redux does not match the underlying filesystem                                      import/no-unresolved
   3:27  error  Casing of redux-form does not match the underlying filesystem                                       import/no-unresolved
```

In Windows, you have a case-insensitive file system which enables you to enter directories regardless of their casing. I.e. the path: `C:\Git\projects\myproject\src` can be entered like this:

```
cd C:\
cd git <-- this is the culprit
cd projects
cd myproject
cd src
npm run eslint
```

When using your lint rule `import/no-unresolved`, this will crash with the above false positives. They will disappear of you go into the directory `Git` instead of `git`.

I think the lint rule should not go further back than the level where the `package.json` is. This problem will appear when you're using either PowerShell or Bash. And it was a huge pain to figure out...

laysent:
Simply prevent eslint to check any parent folder of current working directory.

Fixes #720

### DIFF
--- a/tests/src/core/resolve.js
+++ b/tests/src/core/resolve.js
@@ -3,6 +3,7 @@ import { expect } from 'chai'
 import resolve, { CASE_SENSITIVE_FS, fileExistsWithCaseSync } from 'eslint-module-utils/resolve'
 import ModuleCache from 'eslint-module-utils/ModuleCache'
 
+import * as path from 'path'
 import * as fs from 'fs'
 import * as utils from '../utils'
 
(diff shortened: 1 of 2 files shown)


laysent:
@ljharb Hi, I have added a unit test for this. Please kindly review.

laysent:
@ljharb Hi, any chance this commit could get merged? I believe the unit test error is not caused by changed code. Re-run the test might fix the issue.
Let me know if there is anything I can improve

ljharb:
@laysent I'm unable to rerun appveyor tests; if you rebase and force push, that will trigger a rerun.

ljharb:
@laysent please rebase from the command line; the UI "update" button creates a merge commit.

benmosher:
Ah, great idea! thanks!

PR #860, opened 2017-06-02
Release 2.4.0

benmosher:
need to move eslint-module-utils@v2.1.0 to the `latest` tag before publishing


benmosher:
oops, need to base against `release`, will re-open

PR #861, opened 2017-06-02
Release 2.4.0

benmosher:
need to move eslint-module-utils@v2.1.0 to `latest` tag before publishing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - 4
   - 6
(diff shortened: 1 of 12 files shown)


benmosher:
@ljharb should I unpublish this as well? I deprecated but it's literally unusable since I unpublished the broken `utils` module.

benmosher:
yeah, unpublished in terror at the idea of Greenkeeper, etc., asking folks to upgrade to a dead version.

PR #878, opened 2017-06-18
fix: Written wrong

ali-master:
replace *find* with *isArray* in loadash import

### DIFF
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -89,7 +89,7 @@ var foo = require('./foo');
 
 import test from 'ava';
 import find from 'lodash.find';
-import find from 'lodash.isarray';
+import isArray from 'lodash.isarray';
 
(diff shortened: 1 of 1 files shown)


docs/rules/no-extraneous-dependencies.md
```diff
@@ -89,7 +89,7 @@ var foo = require('./foo');
 
 import test from 'ava';
 import find from 'lodash.find';
-import find from 'lodash.isarray';
+import isArray from 'lodash.isarray';
```
ljharb: Instead tho, maybe we should not even bother importing this, since `Array.isArray` is in the language :-)

can we replace this with `omit` or something, since that's actually useful?

benmosher: 🤷🏼‍♂️😁

ali-master:
Hi @benmosher

You'r wellcome,
I hope success for You.

PR #880, opened 2017-06-23
Add `allow-require` option to `no-commonjs` rule

### LINKED ISSUE #548: Allow require() but forbid module.exports
I'm porting a codebase to ES6. In some cases we explicitly want to use `require()`, because `import` must be at the top of the file.

For example, we sometimes write:

``` js
if (process.env.NODE_ENV == "production") {
  require('../foo.css');
}
```

or in tests we sometimes want to only require things at certain points. We also conditionally require polyfills.

Could we add a configuration option to allow `require()` inside `no-commonjs`? Alternatively, would it make sense to split the rule into `no-commonjs-require` and `no-commonjs-exports`?

futpib:
This fixes #548

### DIFF
--- a/docs/rules/no-commonjs.md
+++ b/docs/rules/no-commonjs.md
@@ -21,6 +21,24 @@ module.exports = { a: "b" }
 exports.c = "d"
 ```
 
+### Allow require
+
+If `allow-require` is provided as an option, `require` calls are valid:
+
+```js
+/*eslint no-commonjs: [2, "allow-require"]*/
+
+if (typeof window !== "undefined") {
+  require('that-ugly-thing');
+}
+```
+
+but `module.exports` is reported as usual.
+
+This is useful for conditional requires.
+
+### Allow primitive modules
+
 If `allow-primitive-modules` is provided as an option, the following is valid:
 
 ```js
--- a/docs/rules/no-commonjs.md
+++ b/docs/rules/no-commonjs.md
@@ -23,10 +23,10 @@ exports.c = "d"
 
 ### Allow require
 
-If `allow-require` is provided as an option, `require` calls are valid:
+If `allowRequire` option is set to `true`, `require` calls are valid:
 
 ```js
-/*eslint no-commonjs: [2, "allow-require"]*/
+/*eslint no-commonjs: [2, { allowRequire: true }]*/
 
 if (typeof window !== "undefined") {
   require('that-ugly-thing');
@@ -39,10 +39,10 @@ This is useful for conditional requires.
 
 ### Allow primitive modules
 
-If `allow-primitive-modules` is provided as an option, the following is valid:
+If `allowPrimitiveModules` option is set to `true`, the following is valid:
 
 ```js
-/*eslint no-commonjs: [2, "allow-primitive-modules"]*/
+/*eslint no-commonjs: [2, { allowPrimitiveModules: true }]*/
 
 module.exports = "foo"
 module.exports = function rule(context) { return { /* ... */ } }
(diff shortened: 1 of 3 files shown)


docs/rules/no-commonjs.md
```diff
@@ -21,6 +21,24 @@ module.exports = { a: "b" }
 exports.c = "d"
 ```
 
+### Allow require
+
+If `allow-require` is provided as an option, `require` calls are valid:
+
+```js
+/*eslint no-commonjs: [2, "allow-require"]*/
```
ljharb: Let's use an object schema instead, so it's extensible - like `{ allowRequire: true, allowPrimitiveModules: true }`.

We should move away from string schemas (but continue to support the existing ones, ofc)

docs/rules/no-commonjs.md
```diff
@@ -21,6 +21,24 @@ module.exports = { a: "b" }
...
+
+if (typeof window !== "undefined") {
+  require('that-ugly-thing');
+}
+```
+
+but `module.exports` is reported as usual.
+
+This is useful for conditional requires.
```
ljharb: When you have this rule enabled, wouldn't you use `import()` for conditionals?

docs/rules/no-commonjs.md
futpib: Keeping or dropping compatibility with string options?

docs/rules/no-commonjs.md
futpib: Well, you could. I guess it's for compatibility only then.

futpib: Actually, I just looked though some [babel-plugin-dynamic-import-node test](https://github.com/airbnb/babel-plugin-dynamic-import-node/blob/master/test/fixtures/basic-import/expected.es2015.js), the new dynamic `import` is async and thus is not a drop-in replacement for `require`. So I guess adding a hint like this would suffice:
> This is useful for conditional requires. If you don't rely on synchronous module loading, check out [dynamic import](https://github.com/airbnb/babel-plugin-dynamic-import-node).

ljharb: True enough; your suggestion is a good addition.

src/rules/no-commonjs.js
```diff
@@ -27,10 +27,34 @@ function allowRequire(node, options) {
...
+  additionalProperties: false,
+}
 
 module.exports = {
   meta: {
     docs: {},
+
+    schema: {
+      anyOf: [
+        {
```
ljharb: i think this first anyOf item can just be `schemaString` by itself?

docs/rules/no-commonjs.md
benmosher: This was my knee jerk reaction as well, but I think the sync argument tracks.

src/rules/no-commonjs.js
futpib: I don't think so, options are always [an array of things](http://eslint.org/docs/developer-guide/working-with-rules#contextoptions).

futpib:
I'm so waiting for this to release. Is there anything else to do for me?

ljharb:
@futpib i'd ilke @benmosher or @jfmengels to review it prior to merging.

benmosher:
Hi @futpib, sorry it has been literally forever since you submitted this and I'm finally getting back to it. 😅 

I just upgraded this to semver-major because it changes the options on this rule. If you'd like to get it merged into the next 2.x release, can you make the options match the existing rule?

i.e. `'allow-require'` instead of `allowRequire: true`. This is not my first choice; I like the flags-based options you've implemented better, but again, they are a breaking change to the existing rule.

futpib:
@benmosher still got semver-major label though

PR #881, opened 2017-06-23
Fix flow interface imports giving false-negative with `named` rule

### LINKED ISSUE #708: False positives when importing from flow-typed module declarations
Explicit imports from [flow-typed](https://github.com/flowtype/flow-typed) module declarations seem to yield false positives.

Example:

```js
import type { ActionCreators } from "redux"
// error  ActionCreators not found in 'redux'  import/named
```

The error should not appear because [flow can attach declared types to named modules via flow-typed or other library files](https://flowtype.org/docs/third-party.html#using-flow-typed), even if the modules themselves do not export those types. In the example above, redux does not export any types. Instead, a [flow-typed library](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/redux_v3.x.x/flow_v0.33.x-/redux_v3.x.x.js) handles the type declarations.

futpib:
This fixes #708 without ignoring all type imports.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -155,7 +155,7 @@ export default class ExportMap {
       const d = dep()
       // CJS / ignored dependencies won't exist (#717)
       if (d == null) return
-      
+
       d.forEach((v, n) =>
         n !== 'default' && callback.call(thisArg, v, n, this))
     })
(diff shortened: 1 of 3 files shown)


futpib:
No idea on that travis-ci error, seems unrelated to the changes in this PR.

julienw:
It just needs a merge now, right ? ;)

benmosher:
published with 2.6.1 👍🏻 thanks!

athomann:
Looks like I'm still experiencing this issue in 2.7.0.

SpainTrain:
@athomann we are seeing it also, but it may be a different issue (possibly https://github.com/benmosher/eslint-plugin-import/issues/616 ?) since `redux-form` >= 7 exports flow defs from the module itself.

To expound on that - for us, this works with flow but results in `named` error from eslint-plugin-import

```
import type {FormProps} from 'redux-form'
```

but this works and results in no eslint error

```
import type {FormProps} from 'redux-form/es/index.js.flow'
```

PR #891, opened 2017-07-09
Allow core modules to resolve files inside declared modules

### LINKED ISSUE #886: core-modules does not resolve files inside declared modules
**tl;dr** I want to tell eslint-plugin-import that I have a package named `assets` and it shouldn't worry about resolving this package in my project dependencies. 

I am using Rails 5.1 with Webpacker. The default app config allows Webpack to import assets from the Rails asset pipeline transparently as follows:

```js
import myLogo from 'assets/images/my_logo.svg';
```

...where `my_logo.svg` lives in `/app/assets/images/my_logo.svg`.

I searched and found #433, which is similar to my issue. I added the following to my `.eslintrc.json`:

```json
"settings": {
  "import/core-modules": ["assets"]
},
```

However, I'm running into another problem. In `importType.js`, when `isBuiltIn()` evaluates my imports based on the `import/core-modules` setting, it's called with the following arguments and **returns false**:

```js
{
  name: 'assets/images/my_logo.svg',
  settings: {
    'import/ignore': ['node_modules', '\\.(coffee|scss|css|less|hbs|svg|json)$'],
    ... 
  },
  extras: ['assets']
}
```

The error message reads as follows:

```
'assets' should be listed in the project's dependencies. Run 'npm i -S assets to add it (import/no-extraneous-dependencies)
```

I believe I have marked `assets` as a package that is present. No errors are thrown if I write something such as `import foo from 'assets'`.

But the import is being evaluated as `assets/images/my_logo.svg` and the plugin believes the package is not present.

How can I tell the plugin that `assets` and **everything under it** are present?

mplewis:
Closes #886.

Lets users add packages to `core-modules` such as `electron`, then import resources from paths such as `electron/some/path/to/resource.json` without encountering lint errors.

You have edit access to this branch – let me know if there is anything else you're looking for. Thanks!

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -8,14 +8,23 @@ function constant(value) {
   return () => value
 }
 
+function baseModule(name) {
+  if (isScoped(name)) {
+    const [scope, pkg] = name.split('/')
+    return `${scope}/${pkg}`
+  }
+  const [pkg] = name.split('/')
+  return pkg
+}
+
 export function isAbsolute(name) {
   return name.indexOf('/') === 0
 }
 
 export function isBuiltIn(name, settings) {
-  const baseModule = name.split('/')[0]
+  const base = baseModule(name)
   const extras = (settings && settings['import/core-modules']) || []
-  return builtinModules.indexOf(baseModule) !== -1 || extras.indexOf(baseModule) > -1
+  return builtinModules.indexOf(base) !== -1 || extras.indexOf(base) > -1
 }
 
 function isExternalPath(path, name, settings) {
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -13,8 +13,9 @@ export function isAbsolute(name) {
 }
 
 export function isBuiltIn(name, settings) {
+  const baseModule = name.split('/')[0]
   const extras = (settings && settings['import/core-modules']) || []
-  return builtinModules.indexOf(name) !== -1 || extras.indexOf(name) > -1
+  return builtinModules.indexOf(baseModule) !== -1 || extras.indexOf(baseModule) > -1
 }
 
 function isExternalPath(path, name, settings) {
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -5,7 +5,7 @@ import importType from 'core/importType'
 
 import { testContext } from '../utils'
 
-describe.only('importType(name)', function () {
+describe('importType(name)', function () {
   const context = testContext()
 
   it("should return 'absolute' for paths starting with a /", function() {
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -5,7 +5,7 @@ import importType from 'core/importType'
 
 import { testContext } from '../utils'
 
-describe('importType(name)', function () {
+describe.only('importType(name)', function () {
   const context = testContext()
 
   it("should return 'absolute' for paths starting with a /", function() {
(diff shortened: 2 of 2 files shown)


src/core/importType.js
```diff
@@ -13,8 +13,9 @@ export function isAbsolute(name) {
 }
 
 export function isBuiltIn(name, settings) {
+  const baseModule = name.split('/')[0]
```
ljharb: what if the core module name is `@foo/bar`?

mplewis: That's a good point, thanks. Maybe we need something better than a simple split[0]. Let me try.

mplewis: How would you feel about replacing the base module name logic with the following?

```js
function baseModule(name) {
  const parts = name.split('/')
  if (name.startsWith('@')) { return parts.slice(0, 2).join('/') }
  return parts[0];
}
baseModule('foo')              //=> "foo"
baseModule('foo/bar/baz')      //=> "foo"
baseModule('@bat/foo/bar/baz') //=> "@bat/foo"
```

ljharb: This logic already exists in the project: https://github.com/benmosher/eslint-plugin-import/blob/ee5a986cfd3ed9d97857889655a969031f65f66d/src/core/importType.js#L31-L33 (and it's best to avoid directly indexing into arrays)

Something like this could be good:
```js
function baseModule(name) {
  if (isScoped(name)) {
    const [scope, pkg] = name.split('/');
    return `${scope}/${pkg}`;
  }
  const [pkg] = name.split('/');
  return pkg;
}
```

mplewis: Thanks for pointing that out. I'll see if I can integrate that.

mplewis:
Just integrated your example logic, along with some tests for scoped packages.

tests/src/core/importType.js
```diff
@@ -69,9 +69,21 @@ describe('importType(name)', function () {
   it("should return 'builtin' for additional core modules", function() {
     // without extra config, should be marked external
     expect(importType('electron', context)).to.equal('external')
+    expect(importType('@org/foobar', context)).to.equal('external')
 
     const electronContext = testContext({ 'import/core-modules': ['electron'] })
     expect(importType('electron', electronContext)).to.equal('builtin')
+
+    const scopedContext = testContext({ 'import/core-modules': ['@org/foobar'] })
```
ljharb: we should probably also support a value of `@org` here - that way you can mark an entire package scope as builtin

mplewis: I'm not sure if that's needed. I work at an org that uses private packages a lot, and I'd fully expect to have to include each imported package one by one. It seems like it should be a bug to accidentally import `@myorg` and whitelist all `@myorg/*` packages. What do you think?

ljharb: I think that it'd be a fine addition later, so there's no need to block this PR on it :-)

mplewis:
I'm happy with this PR if you are! But it looks like the Appveyor build is failing even though all tests are passing. It looks like this error has happened earlier in the project history too, so I don't think it's part of this PR.

Anything else I can help with before we merge this?

ljharb:
Don't worry about the appveyor build for now. Let's get some reviews from the other collabs.

benmosher:
@jfmengels you're off the hook unless you'd like to comment. I'll merge when I can, with changelog notes. thanks @mplewis!

mplewis:
Hey @benmosher! I was wondering if there is any way to merge this into eslint-plugin-import before it gets stale against master. Thanks.

benmosher:
Makes sense, would merge it now but I'm on my phone and the Appveyor noise is blocking me 😓

PR #902, opened 2017-07-25
Add no-commonjs-exports-with-import rule

lencioni:
We want to start using Webpack to resolve imports instead of Babel. One
key difference is that Webpack will break the bundle if you use `import`
in the same file that `module.exports` is used, but it will appear to
build pretty alright. When this happens, webpack outputs a warning like:

```
WARNING in ./foo.js
4:12-13 "export 'default' (imported as 'bar') was not found in './bar'
```

And when the JS is run in the browser, an error like the following is
logged:

```
Uncaught TypeError: Cannot assign to read only property 'exports' of
object '#<Object>'
```

To make this safer to turn on, I am adding this
no-commonjs-exports-with-import rule which reports when it encounters an
`import` in the same file as a CommonJS `module.exports` or `exports.*`.


ljharb:
This might be a duplicate of #804.

ljharb:
#804 is now merged.

PR #908, opened 2017-08-02
Add autofixer for order rule

### LINKED ISSUE #711: Add autofixer for order
It would be nice if order/groups and order/newline had fix options.

I have done a little preliminary digging on what would be involved; order/newline looks like it will be easy to resolve and rather harmless as a whitespace only change. The difficulty seems to be in reordering imports with fix, since as far as I can tell its a little beyond the scope of a normal fix.

For food for thought, I have a [plugin ](https://github.com/AlexJuarez/eslint-plugin-import-order-autofix) that might make a good basis for conversation.

tihonove:
I really want this feature :-)
Fixes #711


tihonove:
Hi, i dont have ideas how to fix [appveyor build](https://ci.appveyor.com/project/benmosher/eslint-plugin-import/build/2136). Could you give me some a hint?
Another point: travis builds seems to unstable. Here is [failed build](https://travis-ci.org/benmosher/eslint-plugin-import/builds/260213433) and after fix coverage [success build](https://travis-ci.org/benmosher/eslint-plugin-import/builds/260230798)

tihonove:
@ljharb, currently autofixer just fix errors reported by original rule. Code
```
var async = require('async');
var a = 1;
var fs = require('fs');
```
transforms to
```
var fs = require('fs');
var async = require('async');
var a = 1;
```
due to rule requires 'fs' to occurs before 'async'. 
In my opinion fixer should fix errors reported by rule. 
Should i modify rule to avoid this behavior?

ljharb:
What about:
```js
var b = require('b');
require('a')(b);
```
?

In other words, I think this rule is only safely fixable in a subset of the circumstances that the rule warns on - eslint core's policy (which I think everyone should follow) is that autofixes must always be "safe", and not change the meaning of the code, even if that means not autofixing some things.

tihonove:
:see_no_evil: Currently autofixer brokes the code above. Im going to update autofixer to avoid unsafe fixes.

ljharb:
Specifically: I think that to start, autofixing can only be said to be safe only within a chunk of requires and/or imports, where the requires are *purely* assignments - and any code that's not a variable-assigned-require or an import-with-a-binding has to be considered a boundary across which autofixing can't cross.

For example, side-effecting requires/imports (like just `import 'foo';` or `require('foo');`, eg polyfills/shims/etc), are likely in place intentionally, which means that every require/import that happens before it must *remain* before it, and everything after it must *remain* after it.

tihonove:
Hi. Currently reorder fixes is very safe and only strict var-assignement-require and imports can be reordered and autofixes would not applied to the following code
```
var foo = require('./foo').bar.bar.bar;
var fs = require('fs');
```
Im not sure about this. Maybe such reorderings should be allowed

ljharb:
@tihonove using `fs` would make it safe; but if you imported something with side effects it wouldn't be :-/

tihonove:
Ok, are there anything i can do to complete pr?

tihonove:
> We'll need to get the tests passing before it can be merged tho.

I dont have ideas how to fix [appveyor build](https://ci.appveyor.com/project/benmosher/eslint-plugin-import/build/2153). Could you give me a hint? Seems it fails after all test successfully passed

ljharb:
I believe the appveyor build might be broken on master; @benmosher, can you confirm?

lukeapage:
Would be great to see this merged..

lukeapage:
@benmosher any chance of merging and releasing this?

tihonove:
Sync and rebase pr

lukeapage:
@tihonove after your rebase there is one test failure..

lukeapage:
ah I see. :(

lukeapage:
@tihonove - master build is now passing again
@benmosher - any chance you can take a look at merging this?

mrm007:
It sure would be great to have this merged 🙏

lukeapage:
@ljharb I see you have merge rights now. Given you approved this and Ben does not seem to be about, would you consider merging it? Thanks.

ljharb:
@lukeapage i've had them this whole time; I don't think an autofixer - especially one as risky as an ordering autofixer - should be merged with only one collaborator having reviewed it.

MoOx:
@benmosher anything we can do to make this review easy? It could be really nice to get this in a  release :)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [2.8.0] - 2017-10-18
 ### Added
 - [`exports-last`] rule ([#620] + [#632], thanks [@k15a])
+- Autofixer for [`order`] rule ([#711], thanks [@tihonove])
```
MoOx: Might need to be adjusted

benmosher:
will merge once linux builds pass 👍

MoOx:
macOS build seems to be struggling. Are those really necessary btw (as this lib doesn't involve any "macos" specific code?)

benmosher:
@MoOx I agree that it _shouldn't_ matter but the tests are there for the times when it mysteriously _does_. 😅 LGTM now, though 👍

ljharb:
See #1086 - this PR caused a performance regression. It’d be great if anyone interested in keeping this autofixer in the plugin could look into that issue.

ajhyndman:
You're my hero, @tihonove!  I just needed to apply an import/order rule to 395 files; this saved me a full day of hair pulling.

PR #910, opened 2017-08-07
Don’t check the comments if they’re not provided

j-f1:
This caused problems for me when linting TypeScript.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -330,17 +330,19 @@ ExportMap.parse = function (path, content, context) {
   })
 
   // attempt to collect module doc
-  ast.comments.some(c => {
-    if (c.type !== 'Block') return false
-    try {
-      const doc = doctrine.parse(c.value, { unwrap: true })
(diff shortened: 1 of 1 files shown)


j-f1:
Closing in favor of https://github.com/eslint/typescript-eslint-parser/issues/346.

j-f1:
From the issue I opened:

> You need to specify that you want comments to be collected within the parser options:
> ```js
> parser.parse('// foo', { comment: true })
> ```
>
> […]
>
> These are the relevant options within espree:
>
> ```js
> // create a top-level comments array containing all comments
> comment: false,
> 
> // attach comments to the closest relevant node as leadingComments and trailingComments
> attachComment: false
> ```
> 
> `comment` is the option which addresses this issue, `attachComment` is no longer relevant for this > parser as we made the decision a while ago to move comment attribution into ESLint itself.

It looks like the `comments` option should at least be included by default [here](https://github.com/benmosher/eslint-plugin-import/blob/dd281306089ba15394aeb2b85241509286442e34/utils/parse.js#L23).

PR #912, opened 2017-08-08
add no-useless-path-segments rule Fixes #471

### LINKED ISSUE #471: detect up a directory then down the same directory
I have a structure like:

```
/ham/egg/chips/index.js
/ham/egg/foo.js
```

then in /ham/egg/chips/index.js I sometimes have code that accidental ends up like:

``` js
import x from '../../egg/foo.js'
```

but I'd much rather it be:

``` js
import x from '../foo.js'
```

graingert:
add no-useless-path-segments rule Fixes #471

### DIFF
--- a/src/rules/no-useless-path-segments.js
+++ b/src/rules/no-useless-path-segments.js
@@ -37,6 +37,13 @@ module.exports = {
 
   create: function (context) {
     const currentDir = path.dirname(context.getFilename())
+    const resolvesToSameModule = (path1, path2) => {
+      try {
+        return require.resolve((path1)) === require.resolve((path2))
+      } catch(e) {
+        return true
+      }
+    }
 
     function checkSourceValue(source) {
       const { value } = source
@@ -54,7 +61,7 @@ module.exports = {
       }
 
       const normed = normalize(value)
-      if (normed !== value) {
+      if (normed !== value && resolvesToSameModule(value, normed)) {
         return report(normed)
       }
 
--- a/src/rules/no-useless-path-segments.js
+++ b/src/rules/no-useless-path-segments.js
@@ -1,0 +1,46 @@
+/**
+ * @fileOverview Ensures that there are no useless path segments
+ * @author Thomas Grainger
+ */
+
+import path from 'path'
+import resolve from 'eslint-module-utils/resolve'
+import moduleVisitor from 'eslint-module-utils/moduleVisitor'
+
+function relative(from, to) {
+  const rel = path.relative(from, to)
+  return rel.startsWith('.') ? rel : `.${path.sep}${rel}`
+}
+
+module.exports = {
+  meta: {},
+
+  create: function (context) {
+    const currentDir = path.dirname(context.getFilename())
+
+    function checkSourceValue(source) {
+      const { value } = source
+      if (!value.startsWith('.')) {
+        return
+      }
+
+      const resolvedPath = resolve(value, context)
+      if (resolvedPath === undefined) {
+        return
+      }
+
+      const expected = path.parse(relative(currentDir, resolvedPath))
+      const valueParsed = path.parse(value)
+
+      if (valueParsed.dir !== expected.dir) {
+        const proposed = path.format({ dir: expected.dir, base: valueParsed.base })
+        context.report({
+          node: source,
+          message: `Useless path segments for "${value}", should be "${proposed}"`,
+        })
+      }
+    }
+
+    return moduleVisitor(checkSourceValue, context.options[0])
+  },
+}
--- a/src/rules/no-useless-path-segments.js
+++ b/src/rules/no-useless-path-segments.js
@@ -9,7 +9,7 @@ import resolve from 'eslint-module-utils/resolve'
 import moduleVisitor from 'eslint-module-utils/moduleVisitor'
 
 /**
- * convert a potentialy relative path from node utils into a true
+ * convert a potentially relative path from node utils into a true
  * relative path.
  *
  * ../ -> ..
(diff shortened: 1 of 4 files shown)


graingert:
@benmosher not sure why it's still failing on appveyor

src/rules/no-useless-path-segments.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+import path from 'path'
+import sumBy from 'lodash.sumby'
+import escapeRegExp from 'lodash.escaperegexp'
+import resolve from 'eslint-module-utils/resolve'
+import moduleVisitor from 'eslint-module-utils/moduleVisitor'
+
+function toRel(rel, sep) {
+  const rSep = escapeRegExp(sep)
```
ljharb: is there a way this could be done without having to escape regexes, or use them at all?

graingert: Ah it can be removed, this is only used on the posix side now.

graingert:
> It'd also be good to warn on /absolute/path/to/foo when it in fact should be './foo and similar.

Maybe on the no absolute rule?

ljharb:
If I want to permit absolute paths, but not ones that could instead be relative, this one seems the appropriate place.

graingert:
Would need to know how far up is too far for absolute paths.

'../../../../users/ljharb/Downloads/bower/jquery-ui'

As that escapes out of the project

ljharb:
Right, you'd locate the package.json, and consider that the root - anything within that should be relative instead of absolute.

When phrased like that, however, it might make sense as an option to `no-absolute-path` :-D

src/rules/no-useless-path-segments.js
```diff
@@ -0,0 +1,93 @@
...
+ * .foo/bar -> ./.foo/bar
+ * ..foo/bar -> ./..foo/bar
+ * foo/bar -> ./foo/bar
+ *
+ * @param rel {string} relative posix path potentially missing leading './'
+ * @returns {string} relative posix path that always starts with a ./
+ **/
+function toRel(rel) {
+  const stripped = rel.replace(/\/$/g, '')
+  return stripped.match(/^((\.\.)|(\.))($|\/)/) ?
```
ljharb: since we're not using the match object here, let's use `regex.test(str)` instead.

tests/src/rules/no-useless-path-segments.js
```diff
@@ -0,0 +1,50 @@
...
+      test({
+        code: 'import "../files/malformed.js"',
+        errors: [ 'Useless path segments for "../files/malformed.js", should be "./malformed.js"'],
+      }),
+      test({
+        code: 'import "../files/malformed"',
+        errors: [ 'Useless path segments for "../files/malformed", should be "./malformed"'],
+      }),
+      test({
+        code: 'import "./test-module/"',
```
ljharb: We should test all of these cases:
1. `foo/index.something` exists → path should be "foo"
1. `foo.something` exists → path should be "foo"
1. both `foo/index.something` and `foo.something` exist → `foo/` and `foo` are different, and the difference is important, and this rule should not warn on it.

graingert: > foo.something exists → path should be "foo"

I think that should be handled by https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md

ljharb: yes, that should handle whether there's an extension or not - i'm saying that "whether this rule can warn on a path or not" actually depends on what's on disk, because the trailing slash is meaningful.

danny-andrews: Can we add a test case for catching multiple slashes, e.g. `foo//something`? See https://github.com/benmosher/eslint-plugin-import/issues/915#issuecomment-322898231.

ljharb: That's a good idea.

graingert:
I've not got a lot of time to fix this up, so please consider this PR abandoned.

To those looking to finish this off, you need to handle the case of:

```
.
├── bar
│   └── index.js
├── bar.js
└── index.js
```

```js
import './bar';  // --> ./bar.js
// vs
import './bar/'; // --> ./bar/index.js
```

danny-andrews:
@ljharb I took this over and handled the edge-case in https://github.com/benmosher/eslint-plugin-import/pull/946.

src/rules/no-useless-path-segments.js
```diff
@@ -0,0 +1,91 @@
...
+ * @author Thomas Grainger
+ */
+
+import path from 'path'
+import sumBy from 'lodash.sumby'
+import resolve from 'eslint-module-utils/resolve'
+import moduleVisitor from 'eslint-module-utils/moduleVisitor'
+
+/**
+ * convert a potentialy relative path from node utils into a true
```
ljharb: potentially

danny-andrews: Weird, I thought I had a test case for that. That's what I added https://github.com/benmosher/eslint-plugin-import/pull/912/commits/9f3c3a7d91aa3b3428bbe392f95be2239edcdf57#diff-8b276518a9cbd15a6895feee0de942b1 for. I'll add it back in and fix the typo.

danny-andrews:
Weird, I thought I had a test case for that. That's what I added 9f3c3a7#diff-8b276518a9cbd15a6895feee0de942b1 for. I'll add it back in and fix the typo.

danny-andrews:
Can't get tests to run locally, but they seem to run in CI...strange.

graingert:
@danny-andrews @ljharb is this ready to go?

danny-andrews:
Ready on my end!

ljharb:
I'm going to rebase this soon and merge it.

graingert:
Hype

ljharb:
@benmosher do you have the release process documented anywhere? If so, I'll cut a release; if not, would you mind? :-D

PR #923, opened 2017-08-31
Handle unresolved JS import when type is exported with the same name

Coobaha:
This was not giving any error

```
// fileA.js
export type MyType = {};

// fileB.js
import { MyType } from './fileA.js' 
```

P.S. Some specs are failing in master?
```
1) ExportMap does not return a cached copy after modification:
2) exports-last valid
        const foo = 'bar'
        export default function foo () {
          const very = 'multiline'
        }
        export const bar = true
      :
```

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -533,9 +533,12 @@ ExportMap.parse = function (path, content, context) {
         case 'TSTypeAliasDeclaration':
         case 'TSInterfaceDeclaration':
         case 'TSAbstractClassDeclaration':
-        case 'TSModuleDeclaration':
-          m.namespace.set(n.declaration.id.name, captureDoc(source, docStyleParsers, n));
+        case 'TSModuleDeclaration': {
+          const meta = captureDoc(docStyleParsers, n);
+          meta.exportKind = n.exportKind;
+          m.namespace.set(n.declaration.id.name, meta);
           break;
+        }
         case 'VariableDeclaration':
           n.declaration.declarations.forEach((d) =>
             recursivePatternCapture(d.id,
--- a/src/rules/named.js
+++ b/src/rules/named.js
@@ -50,7 +50,14 @@ module.exports = {
               `${im[key].name} not found via ${deepPath}`);
           } else {
             context.report(im[key],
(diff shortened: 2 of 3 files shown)


src/rules/named.js
```diff
@@ -40,6 +40,13 @@ module.exports = {
             context.report(im[key],
               im[key].name + ' not found in \'' + node.source.value + '\'')
           }
+        } else if (node.importKind === 'value') {
+          const meta = deepLookup.path[deepLookup.path.length - 1].namespace.get(im[key].name)
+          const wrongType = meta && meta.exportKind !== undefined && meta.exportKind !== 'value'
+          if (wrongType) {
+            context.report(im[key],
+              im[key].name + ' not found in \'' + node.source.value + '\'')
```
ljharb: This should probably use a template literal

src/ExportMap.js
```diff
@@ -406,9 +406,12 @@ ExportMap.parse = function (path, content, context) {
           case 'FunctionDeclaration':
           case 'ClassDeclaration':
           case 'TypeAlias': // flowtype with babel-eslint parser
-          case 'InterfaceDeclaration':
-            m.namespace.set(n.declaration.id.name, captureDoc(docStyleParsers, n))
+          case 'InterfaceDeclaration': {
+            const meta = captureDoc(docStyleParsers, n)
```
ljharb: Maybe captureDoc should set the exportKind?

benmosher:
There have been enough Flow issues lately that I'm tempted to have `ExportMap` just ignore them completely, especially since Flow will complain if things are wrong... but if the merge conflicts resolve without test issues, I'll probably merge as-is for now. Thanks!

ljharb:
@Coobaha i've rebased this, and around 10 tests are failing. When i comment out your implementation changes, only 2 tests are failing.

I know it's been awhile, but it'd be great to get this updated and landed :-) please mark it as ready for review once it's so.

ljharb:
@Coobaha im not sure why you deleted the branch; it’d be great to get this in. Can you restore and reopen it?

Coobaha:
@ljharb Sorry was doing github cleanup and pruning stale branches. Branch is now restored

PR #926, opened 2017-09-13
Point download count link to npmcharts.com?

cheapsteak:
The api requests on npmtrends.com seems to have broken and no longer displays a chart?

changes link from this
http://www.npmtrends.com/eslint-plugin-import

to this:
https://npmcharts.com/compare/eslint-plugin-import?minimal=true


benmosher:
Closing because npmtrends seems to work fine, but maybe I am misunderstanding what you meant to do?

PR #925, opened 2017-09-10
Always include comments when parsing

j-f1:
See #910 and eslint/typescript-eslint-parser#346.

PR #929, opened 2017-09-15
Register TypeScript nodes as exports

j-f1:
This ensures that code using `export enum Foo {}` or `export abstract class Bar {}` registers those as exports that can be imported in other files.


j-f1:
Sure, I’ll do that tomorrow. Hopefully.

j-f1:
The parsing error is fixed in #925.

PR #936, opened 2017-10-01
Add `no-default-export` + docs/tests

### LINKED ISSUE #889: Rule to prohibit default exports
TSLint [has a similar rule](https://palantir.github.io/tslint/rules/no-default-export/), where it prohibits default exports. Basically this (second case generating a warning):

```js
// good-1.js
export function foo() {}

// good-2.js
function foo() {}
export {foo}

// bad-1.js
export default function foo() {}

// bad-2.js
function foo() {}
export {foo as default}
```

isiahmeadows:
Fixes #889

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## [2.8.0] - 2017-10-18
 ### Added
+- [`no-default-export`] rule ([#889], thanks [@isiahmeadows])
 - [`exports-last`] rule ([#620] + [#632], thanks [@k15a])
 
 ### Changed
@@ -435,11 +436,13 @@ for info on changes for earlier releases.
 [`exports-last`]: ./docs/rules/exports-last.md
 [`group-exports`]: ./docs/rules/group-exports.md
 [`no-self-import`]: ./docs/rules/no-self-import.md
+[`no-default-export`]: ./docs/rules/no-default-export.md
 
 [`memo-parser`]: ./memo-parser/README.md
 
 [#944]: https://github.com/benmosher/eslint-plugin-import/pull/944
 [#891]: https://github.com/benmosher/eslint-plugin-import/pull/891
+[#889]: https://github.com/benmosher/eslint-plugin-import/pull/889
 [#858]: https://github.com/benmosher/eslint-plugin-import/pull/858
 [#843]: https://github.com/benmosher/eslint-plugin-import/pull/843
 [#871]: https://github.com/benmosher/eslint-plugin-import/pull/871
@@ -670,3 +673,4 @@ for info on changes for earlier releases.
 [@rosswarren]: https://github.com/rosswarren
 [@alexgorbatchev]: https://github.com/alexgorbatchev
 [@robertrossmann]: https://github.com/robertrossmann
+[@isiahmeadows]: https://github.com/isiahmeadows
--- a/docs/rules/no-default-export.md
+++ b/docs/rules/no-default-export.md
@@ -1,0 +1,63 @@
+# no-default-export
+
+Prohibit default exports. Mostly an inverse of [`prefer-default-export`].
+
+[`prefer-default-export`]: ./prefer-default-export.md
+
+## Rule Details
+
+The following patterns are considered warnings:
+
+```javascript
+// bad1.js
+
+// There is a default export.
+export const foo = 'foo';
+const bar = 'bar';
+export default 'bar';
+```
+
+```javascript
+// bad2.js
+
+// There is a default export.
+const foo = 'foo';
+export { foo as default }
+```
+
+The following patterns are not warnings:
+
+```javascript
+// good1.js
+
+// There is only a single module export and it's a named export.
+export const foo = 'foo';
+```
+
+```javascript
+// good2.js
+
+// There is more than one named export in the module.
+export const foo = 'foo';
+export const bar = 'bar';
+```
+
+```javascript
+// good3.js
+
+// There is more than one named export in the module
+const foo = 'foo';
+const bar = 'bar';
+export { foo, bar }
+```
+
+```javascript
+// export-star.js
+
+// Any batch export will disable this rule. The remote module is not inspected.
+export * from './other-module'
+```
+
+## When Not To Use It
+
+If you don't care if default imports are used, or if you prefer default imports over named imports.
--- a/src/rules/no-default-export.js
+++ b/src/rules/no-default-export.js
@@ -1,0 +1,35 @@
+module.exports = {
+  meta: {
+    docs: {},
+  },
+
+  create(context) {
+    // ignore non-modules
+    if (context.parserOptions.sourceType !== 'module') {
+      return {}
+    }
+
+    const preferNamed = 'Prefer named exports.'
+    const noAliasDefault = ({local}) =>
+      `Do not alias \`${local.name}\` as \`default\`. Just export ` +
+      `\`${local.name}\` itself instead.`
+
+    return {
+      ExportDefaultDeclaration(node) {
+        context.report({node, message: preferNamed})
+      },
+
+      ExportNamedDeclaration(node) {
+        node.specifiers.forEach(specifier => {
+          if (specifier.type === 'ExportDefaultSpecifier' &&
+              specifier.exported.name === 'default') {
+            context.report({node, message: preferNamed})
+          } else if (specifier.type === 'ExportSpecifier' &&
+              specifier.exported.name === 'default') {
+            context.report({node, message: noAliasDefault(specifier)})
+          }
+        })
+      },
+    }
+  },
+}
--- a/tests/src/rules/no-default-export.js
+++ b/tests/src/rules/no-default-export.js
@@ -1,0 +1,121 @@
+import { test } from '../utils'
+
+import { RuleTester } from 'eslint'
+
+const ruleTester = new RuleTester()
+    , rule = require('rules/no-default-export')
+
+ruleTester.run('no-default-export', rule, {
+  valid: [
+    test({
+      code: `
+        export const foo = 'foo';
+        export const bar = 'bar';
+      `,
+    }),
+    test({
+      code: `
+        export const foo = 'foo';
+        export function bar() {};
+      `,
+    }),
+    test({
+      code: `export const foo = 'foo';`,
+    }),
+    test({
+      code: `
+        const foo = 'foo';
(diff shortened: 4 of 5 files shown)


isiahmeadows:
Fixes #889

isiahmeadows:
The failing tests appear to be irrelevant (and the same error that appears in `master`).

tests/src/rules/no-default-export.js
```diff
@@ -0,0 +1,123 @@
...
+      code: `
+        import * as foo from './foo';`,
+      }),
+
+    test({
+      code: `export type UserId = number;`,
+      parser: 'babel-eslint',
+      }),
+
+    // ...SYNTAX_CASES,
```
j-f1: What does this mean?

README.md
```diff
@@ -91,6 +91,7 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 [`no-unassigned-import`]: ./docs/rules/no-unassigned-import.md
 [`no-named-default`]: ./docs/rules/no-named-default.md
 [`no-anonymous-default-export`]: ./docs/rules/no-anonymous-default-export.md
+[`no-default-export`]: ./docs/rules/no-default-export.md
```
j-f1: You should mention this under one of the headers so it is visible in the README.

src/rules/no-default-export.js
```diff
@@ -0,0 +1,40 @@
...
+          if (node.type === 'ExportDefaultSpecifier') {
+            context.report({
+              node: ast,
+              message: 'Prefer named exports.',
+            })
+          } else if (node.type === 'ExportSpecifier' &&
+              node.exported.name === 'default') {
+            context.report({
+              node: ast,
+              message: 'Prefer named exports.',
```
j-f1: Can you extract this message to an external variable? Naming it `message` would it be destructured in.

src/rules/no-default-export.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+
+  create: function (context) {
+    // ignore non-modules
+    if (context.parserOptions.sourceType !== 'module') {
+      return {}
+    }
+
+    return {
+      ExportDefaultDeclaration: function (ast) {
```
j-f1: Can you call this `node` to match the other rules?

src/rules/no-default-export.js
```diff
@@ -0,0 +1,40 @@
+module.exports = {
+  meta: {
+    docs: {},
```
j-f1: Can you put a `description` and category in here?

tests/src/rules/no-default-export.js
```diff
@@ -0,0 +1,123 @@
...
+  ],
```
j-f1: Can you add a test for something like `import defaultImport from 'module-from-npm'`? This should be legal, since we can’t control how other libraries are structured.

src/rules/no-default-export.js
isiahmeadows: I didn't include it because it seemed that none of the other rules appeared to use it ([example](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-named-default.js)).

tests/src/rules/no-default-export.js
isiahmeadows: I didn't add a test for it, since it only checks `export` and `export ... from`. I can add a test to clarify this, though.

tests/src/rules/no-default-export.js
isiahmeadows: It was more or less adapted from [`prefer-default-export`'s tests](https://github.com/benmosher/eslint-plugin-import/blob/master/tests/src/rules/prefer-default-export.js#L84).

README.md
isiahmeadows: Good catch! I completely missed that.

isiahmeadows: Fixed!

isiahmeadows:
Ping, `eslint-plugin-import` people? (I feel this might have gotten lost in the weeds a little bit...)

ljharb:
@isiahmeadows if you rebase this (rather than using the "update branch" button), it should rerun the travis tests.

I'll give it a review, but I'm not a fan of even the existence of this rule, so I'm going to leave it to the other collaborators to review/merge.

docs/rules/no-default-export.md
```diff
@@ -0,0 +1,63 @@
+# no-default-export
```
ljharb: i think maybe `forbid-default-export` would be a more typical name for a rule like this, and for the opposite of "prefer"

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## [2.8.0] - 2017-10-18
 ### Added
+- [`no-default-export`] rule ([#889], thanks [@isiahmeadows])
```
j-f1: Don’t forget to add `[@isiahmeadows]: https://github.com/isiahmeadows` to the bottom of this file.

tests/src/rules/no-default-export.js
```diff
@@ -0,0 +1,129 @@
...
+        export const { foo } = { foo: "bar" };`,
+      }),
+    test({
+      code: `
+        export const { foo: { bar } } = { foo: { bar: "baz" } };`,
+      }),
+    test({
+      code: 'export { a, b } from "foo.js"',
+      parser: 'babel-eslint',
+      }),
```
ljharb: all these `}),` need to be de-indented a level (it seems like this is already the way it is in the prefer-default-export tests, but let's fix it here anyways)

src/rules/no-default-export.js
```diff
@@ -0,0 +1,33 @@
...
+  },
+
+  create: function (context) {
+    // ignore non-modules
+    if (context.parserOptions.sourceType !== 'module') {
+      return {}
+    }
+
+    var message = 'Prefer named exports.'
```
ljharb: this should be a `const`.

also, i wonder if this is an opportunity here to make the message a bit more explanatory?

isiahmeadows:
@ljharb I rebased locally and resolved the issue.

FWIW, [this article](https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad?gi=df38c1496fe2) just ran through JavaScript Weekly, and I feel it's highly relevant. I agree with most of what that article states, but it's purely an opinion, and although I happen to agree with most of it, I'm ***very*** against making it the default.

Also, not all languages have default imports/exports (in my experience, most with module systems built-in *don't*).

ljharb:
I've seen it - that article is full of "arguments" supporting named exports that apply identically to default exports; it doesn't make a convincing case at all.

Most languages also don't have a package manager that allows conflicting dependencies, but luckily JS isn't constrained by that :-)

README.md
```diff
@@ -78,7 +78,9 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 * Forbid unassigned imports ([`no-unassigned-import`])
 * Forbid named default exports ([`no-named-default`])
 * Forbid anonymous values as default exports ([`no-anonymous-default-export`])
+* Forbid default exports ([`no-default-export`])
 * Prefer named exports to be grouped together in a single export declaration ([`group-exports`])
+* Forbid default exports ([`no-default-export`])
```
ljharb: this rebase seems to have left this item here twice

src/rules/no-default-export.js
```diff
@@ -0,0 +1,33 @@
...
+
+    return {
+      ExportDefaultDeclaration: function (node) {
+        context.report({node, message})
+      },
+
+      ExportNamedDeclaration: function (node) {
+        for (var i = 0; i < node.specifiers.length; i++) {
+          var specifier = node.specifiers[i]
```
ljharb: this should use `node.specifiers.forEach` instead of a loop (which will allow both `var` instances to be removed)

tests/src/rules/no-default-export.js
ljharb: the indentation here is still off; the `}` should match the indentation level of the line that ends with `}`

isiahmeadows:
@ljharb All fixed (plus some other indentation oddities).

src/rules/no-default-export.js
ljharb: ping

docs/rules/no-default-export.md
ljharb: ping

CHANGELOG.md
isiahmeadows: Oh yeah...oops

docs/rules/no-default-export.md
isiahmeadows: I was following the naming convention the other rules have - there's also `no-dynamic-require`, `no-named-as-default`, and `no-anonymous-default-export`. So IMHO it wouldn't make sense to use `forbid-${name}` when all the others use `no-`. Also, [ESLint recommends this convention](https://eslint.org/docs/developer-guide/working-with-rules#rule-naming-conventions), so I'm kind of hesitant to deviate from it

isiahmeadows: BTW, I kind of missed these comments somehow; sorry about that. 😦

src/rules/no-default-export.js
isiahmeadows: Will do. I'll just note that the code base is a haphazard mix of `var`, `let`, and `const`, and ES6 features are used very inconsistently.

> also, i wonder if this is an opportunity here to make the message a bit more explanatory?

For the case of `export {foo as default}`, sure. For the rest, I'm not sure there is more that needs to be done IMHO. There's not really any extra context needed.

docs/rules/no-default-export.md
ljharb: fair enough; this one isn't a blocker or anything.

isiahmeadows:
@ljharb I added a new commit to address your other feedback (that I somehow missed), and I also fixed a bug in the process (with the proposed `export foo from "bar.js"`, which exports a named binding, not a default one).

isiahmeadows:
Ping, @ljharb/@benmosher?

ljharb:
@isiahmeadows I rebased this branch for you. It's still waiting on reviews from other collabs though.

isiahmeadows:
Ping @benmosher / @jfmengels?

isiahmeadows:
Ping? @benmosher / @jfmengels

sergeysova:
Any progress here?

sergeysova:
Add to readme?

isiahmeadows:
@sergeysova #1026

PR #938, opened 2017-10-05
Recommended a method of handling Webpack function config arguments

robwierzbowski:
I ran into the undefined `env` issue with webpack function configuration, and noticed while searching that plenty of others have had the same problem. Example issues:

https://github.com/benmosher/eslint-plugin-import/issues/398
https://github.com/AtomLinter/linter-eslint/issues/625

The Vue and React boilerplates I've used recently both use function configuration, so I think this is probably a common issue. Adding a recommended way of handling this issue to the documentation could be helpful for webpack users, and save them the time I spent looking for the cause of the problem and a solution. 

Here's a draft of what I believe is the simplest way to handle the issue. Feedback welcome; I'm happy to revise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,6 +336,19 @@ settings:
 [`eslint_d`]: https://www.npmjs.com/package/eslint_d
 [`eslint-loader`]: https://www.npmjs.com/package/eslint-loader
 
+## Webpack configuration arguments
+
+This plugin doesn't pass arguments to [Webpack configuration functions](https://webpack.js.org/configuration/configuration-types/#exporting-a-function). If a Webpack configuration function uses the `env` or `argv` arguments, values must be supplied to avoid `argument undefined` errors. On Node 6.4 or greater we recommend [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters):
+
+```js
+// webpack.config.js
+const defaultEnv = { NODE_ENV: 'development' };
+
+module.exports = (env = defaultEnv) => {
+  // ...configuration
+}
+```
+
 ## SublimeLinter-eslint
 
 SublimeLinter-eslint introduced a change to support `.eslintignore` files


README.md
```diff
@@ -336,6 +336,19 @@ settings:
 [`eslint_d`]: https://www.npmjs.com/package/eslint_d
 [`eslint-loader`]: https://www.npmjs.com/package/eslint-loader
 
+## Webpack configuration arguments
+
+This plugin doesn't pass arguments to [Webpack configuration functions](https://webpack.js.org/configuration/configuration-types/#exporting-a-function). If a Webpack configuration function uses the `env` or `argv` arguments, values must be supplied to avoid `argument undefined` errors. On Node 6.4 or greater we recommend [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters):
```
ljharb: is this something that the plugin could fix?

robwierzbowski: Possibly partially.

#### Case A: Using the resolver during compilation

Since webpack and eslint are running in the same env, it could be possible. The relevant webpack function call is here: https://github.com/benmosher/eslint-plugin-import/blob/a69b77181c423043d39f1c8752b296aaa95facfc/resolvers/webpack/index.js#L84-L86

#### Case B: In the code editor

I don't think this case is solvable. There's no way for the editor to know what env you're running your build process in.

robwierzbowski: The plugin could expose a command line or config option to pass args to the webpack function. But, IMO, people have different enough needs that handling in code is probably best, with a simple template provided for the most common case.

I could see is providing a default `env.NODE_ENV` on line 85, to automatically fix the most common case. That would be pretty prescriptive and could cause unexpected behavior, but maybe a good "works out of the box" solution.

robwierzbowski:
After thinking about this a bit more, maybe we can solve it mostly in the plugin. 


For ESLint as loader, pass in process.env. If possible pass in any args passed to webpack — not sure if they’re available. 

For ESLint as an editor package, what do you think about making the executive decision to pass in a stub/fake env object that sets NODE_ENV as development? The issues I saw on the internet were all for the NODE_ENV variable, so it might be the most common case. For args, don’t pass anything because there’s no way to guess. 

Finally, I can revise the doc addition to say “if you reference any env or agrv values in your webpack config, please stub them for the code editor. The package stubs NODE_ENV=development as a convenience.”

Thoughts?

ljharb:
The way editors should be handles is: if they differ from the command line in any way, they're broken. In other words, pretend editors don't exist.

NODE_ENV can certainly be defaulted to development, yes.

robwierzbowski:
OK, I'll revise this PR, see if I can pass through webpack args, etc.

robwierzbowski:
I'm closing this as I don't really need the functionality anymore and don't think I will complete this PR.

PR #947, opened 2017-10-16
no-common-js: allow conditional requires

jasonxia23:
IMO, `a && require('a')` should be allowed.

### DIFF
--- a/tests/src/rules/no-commonjs.js
+++ b/tests/src/rules/no-commonjs.js
@@ -29,6 +29,7 @@ ruleTester.run('no-commonjs', require('rules/no-commonjs'), {
 
     // allowed requires
     { code: 'function a() { var x = require("y"); }' }, // nested requires allowed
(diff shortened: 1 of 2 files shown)


tests/src/rules/no-commonjs.js
```diff
@@ -20,6 +20,7 @@ ruleTester.run('no-commonjs', require('rules/no-commonjs'), {
 
     // allowed requires
     { code: 'function a() { var x = require("y"); }' }, // nested requires allowed
+    { code: 'var a = c && require("b")' }, // dynamic requires allowed
```
ljharb: this is a conditional require, not a dynamic require.

jasonxia23: sorry for the ambiguity. i mean conditional require.

ljharb:
Took care of the rebase and fixed the comment.

PR #950, opened 2017-10-18
Release 2.8.0

benmosher:


alexgorbatchev:
Did something happen with the rebase or these are all new changes?

benmosher:
Looks like I forgot to update `release` after publishing v2.7.0, and maybe also v2.6.0. so not all the included commits are new.

PR #955, opened 2017-10-19
fix #939 by adding `.mjs` to stock Node resolver

### LINKED ISSUE #939: import statement without `.mjs` file extension causes linter errors
Hi!
I have a simple ESM module file `test.mjs` with this content:
```
export const bla = 'blub';
export default {
  test: true,
};
```
And a file `index.mjs` that loads the module like this:
```
import test, { bla } from './test';
```
ESLint will produce the following errors:
 * Unable to resolve path to module './test' (import/no-unresolved)
 * Missing file extension for "./test" (import/extensions)

Loading the module with file extension fixes the errors:
```
import test, { bla } from './test.mjs';
```
However, omitting the file extension remains best practice, as far as I know.

I am using Node 8.6.0 (with `--experimental-modules` to run ESM), eslint 4.8.0 and eslint-plugin-import 2.7.0.

I tried the following tweaks in `.eslintrc`:
 * rule `"import/extensions": [2, { "js": "never", "json": "always", "mjs": "never" }]` prevents the `import/extensions` error -- which is good
 * setting `"import/extensions": [".js", ".mjs"]` does not help, the error `import/no-unresolved` remains

Thank you.

benmosher:
I'm assuming this is not controversial but thought it'd be good to run it through CI and get feedback just in case.

### DIFF
--- a/resolvers/node/CHANGELOG.md
+++ b/resolvers/node/CHANGELOG.md
@@ -4,7 +4,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## Unreleased
-
+### Added
+- `.mjs` extension detected by default to support `experimental-modules` (#939)
 
 ## v0.3.1 - 2017-06-23
 ### Changed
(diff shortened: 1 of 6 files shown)


benmosher:
canceled the OSX builds until I'm able to rebase onto `ci-cleanup`

PR #959, opened 2017-10-25
newline-after-import: update error message in case multiple empty lines are expected

AdamSelene:
This PR simply updates the `newline-after-import` error message which was misleading when `count` was set to anything other than 1.

### DIFF
--- a/tests/src/rules/newline-after-import.js
+++ b/tests/src/rules/newline-after-import.js
@@ -1,7 +1,10 @@
 import { RuleTester } from 'eslint'
 
-const IMPORT_ERROR_MESSAGE = 'Expected empty line after import statement not followed by another import.'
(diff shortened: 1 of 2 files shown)


tests/src/rules/newline-after-import.js
```diff
@@ -1,7 +1,10 @@
 import { RuleTester } from 'eslint'
 
-const IMPORT_ERROR_MESSAGE = 'Expected empty line after import statement not followed by another import.'
-const REQUIRE_ERROR_MESSAGE = 'Expected empty line after require statement not followed by another require.'
+const IMPORT_ERROR_MESSAGE = 'Expected 1 empty line after import statement not followed by another import.'
+const IMPORT_ERROR_MESSAGE_MULTIPLE = function(count) {
```
ljharb: this might as well be an arrow function

AdamSelene: Updated :)

PR #968, opened 2017-11-16
Fix eslint-import-resolver-webpack with pnpm

bsuh:
Updated resolve to ^1.4.0, where it introduced option to not preserve
symlinks when resolving, matching node's behavior.


ljharb:
This may be a breaking change for the webpack resolver, fwiw.

benmosher:
good point @ljharb, probably should major bump the resolver for this. not a big deal to do that, I think

PR #969, opened 2017-11-16
allow newer version of node-libs-browser

modosc:
this change allows `node-libs-browser@^2.0.0` in  `eslint-import-resolver-webpack`'s deps since that's the version `webpack@^3.0.0` is using (otherwise one ends up with ^1.0.0 and ^2.0.0 installed)

### DIFF
--- a/resolvers/webpack/package.json
+++ b/resolvers/webpack/package.json
@@ -37,7 +37,7 @@
     "interpret": "^1.0.0",
     "is-absolute": "^0.2.3",
     "lodash.get": "^3.7.0",
-    "node-libs-browser": "^1.0.0",
+    "node-libs-browser": "^1.0.0 || ^2.0.0",
     "resolve": "^1.2.0",
     "semver": "^5.3.0"
   },


ljharb:
What is the breaking change in v2? Does it impact the way we use it?

modosc:
> What is the breaking change in v2? Does it impact the way we use it?
i think it's this:

`Update timers-browserify to 2.x
This is a breaking change because timers-browserify 2.x introduces a global setImmediate polyfill`

https://github.com/webpack/node-libs-browser/commit/0dfeab6e72aeb42a37ab1dffa13279982be30a28

modosc:
> whoa, if that's the case, we definitely can't use it. Any package that auto-mutates the global, including transitively, is not a good package and we should not allow it.

seems like it's ok though if you're already using it via webpack3? that's why i tried to make it optional. i don't feel that strongly one way or another, was just auditing our `yarn.lock` file and noticed this.

ljharb:
Hm, that’s a good point. Fair enough!

modosc:
fwiw the appveyor failures don't seem to have anything to do with this change?

modosc:
thanks!

PR #972, opened 2017-11-18
Es6modules only

dominiktilp:
It's really basic implementation. It only checks if AST of imported file contains one of these nodes types
[ 
  'ExportDefaultDeclaration',
  'ExportNamedDeclaration',
  'ExportAllDeclaration',
  'ExportSpecifier',
]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 
 - Add [`group-exports`] rule: style-guide rule to report use of multiple named exports ([#721], thanks [@robertrossmann])
+- Add [`es6modules-only`] rule: module-systems rule to detect import of non-es6 modules ([#971], thanks [@dominiktilp])
 
 ## [2.8.0] - 2017-10-18
 ### Added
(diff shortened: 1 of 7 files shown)


dominiktilp:
Of course, the situation about es6 modules is not ideal. But from my point of view, there is a lot of packages which is not used directly in node or browsers. And at the end, you are always traspiling at least your code to be sure that it will work in your target environment.

ljharb:
It’s never safe to transpile code you didn’t write. If a package ships “not ES5 CJS”in “main”, it’s a bug.

coreyfarrell:
> It’s never safe to transpile code you didn’t write. If a package ships “not ES5 CJS”in “main”, it’s a bug.

@ljharb Sorry for OT question but could you share a link/explanation of how/why it's unsafe?  thank you!

ljharb:
@coreyfarrell quite simply, transpilation is not 1:1; it can change hw the code behaves for certain edge cases (for example, arrow functions aren’t supposed to be constructible, but the default “loose” setting for them in babel allows them to be). You can’t know if the author was relying on all those caveats or not - you can only know that about your own code.

ljharb:
I'm going to close this until other collabs weigh in in favor.

PR #977, opened 2017-11-28
Handle call expressions in the no-anonymous-default-export rule

fatfisz:
Call expressions are not handled by the `no-anonymous-default-export` rule... not anymore!

Previously:
```js
// No error
export default foo(bar)
```

After the PR:
```js
// Error: Assign call result to a variable before exporting as module default
export default foo(bar)
```

### DIFF
--- a/tests/src/rules/no-anonymous-default-export.js
+++ b/tests/src/rules/no-anonymous-default-export.js
@@ -21,6 +21,7 @@ ruleTester.run('no-anonymous-default-export', rule, {
       test({ code: 'export default \'foo\'', options: [{ allowLiteral: true }] }),
       test({ code: 'export default `foo`', options: [{ allowLiteral: true }] }),
       test({ code: 'export default {}', options: [{ allowObject: true }] }),
+      test({ code: 'export default foo(bar)', options: [{ allowCallExpression: true }] }),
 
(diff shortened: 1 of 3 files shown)


tests/src/rules/no-anonymous-default-export.js
```diff
@@ -43,6 +44,7 @@ ruleTester.run('no-anonymous-default-export', rule, {
       test({ code: 'export default \'foo\'', errors: [{ message: 'Assign literal to a variable before exporting as module default' }] }),
       test({ code: 'export default `foo`', errors: [{ message: 'Assign literal to a variable before exporting as module default' }] }),
       test({ code: 'export default {}', errors: [{ message: 'Assign object to a variable before exporting as module default' }] }),
+      test({ code: 'export default foo(bar)', errors: [{ message: 'Assign call result to a variable before exporting as module default' }] }),
```
ljharb: this constitutes a breaking change if it's `false` by default; it'd need to default to `true` so it can be semver-minor

fatfisz: Ahh, yes. Will fix that.

fatfisz:
Apparently the "default" property of the JSON schema is ignored by ESLint, so I added a bit of code that generates defaults.

PR #978, opened 2017-11-29
`no-commonjs`: Fix false positive when `exports` is in scope

### LINKED ISSUE #940: no-commonjs: Allow assigning to a variable called exports
`exports` should only be disallowed assignment if it's the "global" `exports` of CJS, not any old variable named `exports`.

```js
function someFunc() {
  const exports = someComputation();

  expect(exports.someProp).toEqual({ a: 'value' });
}
```

image

adrianhelvik:
I've added a fix for false positives for no-commonjs when `exports` is in scope.

Fixes #940.

### DIFF
--- a/src/rules/no-commonjs.js
+++ b/src/rules/no-commonjs.js
@@ -36,7 +36,12 @@ module.exports = {
 
         // exports.
         if (node.object.name === 'exports') {
-          context.report({ node, message: EXPORT_MESSAGE })
+          const isInScope = context.getScope()
+            .variables
+            .some(variable => variable.name === 'exports')
+          if (! isInScope) {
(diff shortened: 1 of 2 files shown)


src/rules/no-commonjs.js
```diff
@@ -36,7 +36,12 @@ module.exports = {
 
         // exports.
         if (node.object.name === 'exports') {
-          context.report({ node, message: EXPORT_MESSAGE })
+          const isInScope = context.getScope()
+            .variables
+            .some(variable => variable.name === 'exports')
+          if (! isInScope) {
```
ljharb: please remove the space between the `!` and `isInScope`

PR #987, opened 2017-12-21
fixes #954 - strip out webpack loaders for no-extraneous-dependencies

### LINKED ISSUE #954: Strip out webpack loaders from imports
As far as I know I can only _ignore_ imports with webpack loaders, but I'd like them to be supported in certain rules.

E.g. I have something like this:

```
'import-inject-loader?showNotification!..' should be listed in the project's dependencies. Run 'npm i -S import-inject-loader?showNotification!..' to add it (import/no-extraneous-dependencies)
```

With an import like `import {} from 'import-inject-loader?showNotification!../../src/form/FormModel';`. But it would be nice, if it would just be treated as `import {} from '../../src/form/FormModel';` natively.

donaldpipowitch:
Hi everyone!

A small attempt to fix https://github.com/benmosher/eslint-plugin-import/issues/954. I'm not that familiar with eslint-plugin-import so I wasn't sure how "resolvers" should help here or not (https://github.com/benmosher/eslint-plugin-import/issues/954#issuecomment-338195344).

However this could be an intermediate solution.

### DIFF
--- a/tests/src/rules/no-extraneous-dependencies.js
+++ b/tests/src/rules/no-extraneous-dependencies.js
@@ -29,6 +29,7 @@ ruleTester.run('no-extraneous-dependencies', rule, {
     test({ code: 'var foo = require("pkg-up")'}),
     test({ code: 'import "fs"'}),
     test({ code: 'import "./foo"'}),
+    test({ code: 'import "raw-loader!./foo"'}),
     test({ code: 'import "lodash.isarray"'}),
     test({ code: 'import "@org/package"'}),
 
@@ -73,6 +74,13 @@ ruleTester.run('no-extraneous-dependencies', rule, {
   invalid: [
     test({
       code: 'import "not-a-dependency"',
+      errors: [{
+        ruleId: 'no-extraneous-dependencies',
+        message: '\'not-a-dependency\' should be listed in the project\'s dependencies. Run \'npm i -S not-a-dependency\' to add it',
+      }],
+    }),
+    test({
+      code: 'import "raw-loader!not-a-dependency"',
       errors: [{
         ruleId: 'no-extraneous-dependencies',
         message: '\'not-a-dependency\' should be listed in the project\'s dependencies. Run \'npm i -S not-a-dependency\' to add it',
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -55,11 +55,11 @@ function optDepErrorMessage(packageName) {
 }
 
 function stripWebpackLoaders(name) {
-  return name.split('!').pop();
+  return name.split('!').pop()
 }
 
 function reportIfMissing(context, deps, depsOptions, node, name) {
-  name = stripWebpackLoaders(name);
+  name = stripWebpackLoaders(name)
   if (importType(name, context) !== 'external') {
     return
   }
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -54,7 +54,12 @@ function optDepErrorMessage(packageName) {
     `not optionalDependencies.`
 }
 
+function stripWebpackLoaders(name) {
+  return name.split('!').pop();
+}
+
 function reportIfMissing(context, deps, depsOptions, node, name) {
+  name = stripWebpackLoaders(name);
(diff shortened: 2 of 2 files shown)


tests/src/rules/no-extraneous-dependencies.js
```diff
@@ -78,6 +79,13 @@ ruleTester.run('no-extraneous-dependencies', rule, {
         message: '\'not-a-dependency\' should be listed in the project\'s dependencies. Run \'npm i -S not-a-dependency\' to add it',
       }],
     }),
+    test({
+      code: 'import "raw-loader!not-a-dependency"',
```
ljharb: Can’t this syntax take any value? I don’t think it can be generically handled; i think each loader needs it’s own custom resolver.

donaldpipowitch: Hi, could you explain what you mean? I thought this would be a generic solution. It should work with every loader or multiple ones and nothing needs to be resolved.

ljharb: What I mean is; `import "some-loader!not-a-package-name-or-path"` would work too. There's no guarantee that the thing after the `!` is a normal `require.resolve`-able path.

donaldpipowitch: Okay, I think I understand you. But isn't this always the case? Or at least not specific to my PR?

But maybe I just don't understand `eslint-plugin-import` good enough. Is there something I should do differently in my PR?

ljharb: What I'm suggesting is that your entire approach is flawed; it's based on the idea that if you take the part after the `!`, then you can resolve it normally, as if the first part wasn't there.

However, that's just not how webpack's special ! syntax works. Which means that every single custom loader syntax needs its own snowflake package to tell `eslint-plugin-import` how to resolve it - there's no way that the core plugin itself can handle it.

donaldpipowitch: This PR should just fix my reported `no-extraneous-dependencies` problem. AFAIK nothing needs to be resolved here at all...? https://github.com/benmosher/eslint-plugin-import/issues/954#issue-266758465 Everything after the loader will just work as always. Just check if the part behind the loader is an external package and if it is and not listed in the `package.json` report an error.

Maybe we speak about two different topics...? 🤔

ljharb: Right. The part behind the loader might not be a package *at all*, and that should not be an error.

donaldpipowitch: I'm sorry. I don't understand what is wrong 😞

When I write `import "foo"` I _never_ know if `foo` is a package or not. What is the difference between `import "foo"` and `import "bar!foo"` from the point of view of the `no-extraneous-dependencies`? AFAIK it should just ignore the `bar!` part.

Is your objection against stripping the loader part at all or just against the test...? 🤔 Still try to wrap my head around this.

ljharb: When you write `import 'foo'`, you do know, with certainty, that `foo` is an import path - `eslint-plugin-import` has a "node" resolver, for example, that uses node's rules to resolve that path. `import 'bar!foo'` in the "node" resolver would look for a file literally named `bar!foo.js` or `bar!foo/index.js` etc.

Thus, in order to have `!` have different rules, an *entirely distinct resolver package* would need to be created to tell eslint-plugin-import how to resolve them - that knowledge does not belong hardcoded in a core rule.

Separately, you'd want to make a *separate* resolver for every loader. foo-loader and bar-loader might have entirely and wildly different resolving rules. One loader might make `foo!bar` load "bar.js" from "foo.com", let's say. Thus, it's impossible to have any sort of generic way to handle "strings with exclamation points", because there is not a single universal meaning for them, even within webpack.

donaldpipowitch: You're right that `!` as part of a file name/directory name is not possible in webpack, but could be possible in non-webpack projects - and this project is not webpack specific. Maybe I'll create s custom rule which works like this rule, but with webpack in mind.

Feel free to close.

ljharb: I don't think you need a custom *rule* so much as a custom eslint-plugin-import *resolver plugin*.

PR #993, opened 2018-01-05
Fix #666: resolve config from source file

### LINKED ISSUE #666: Webpack resolver looks for custom webpack config in packages with jsnext:main
With a `.eslintrc` like

``` 
"settings": {
  "import/resolver": {
    "webpack": {
      "config": "webpack.karma.config.js"
    }
  }
},
```

when I import

```
import {takeEvery} from 'redux-saga';
```

I get the error

```
 Cannot find module '/path/to/project/node_modules/redux-saga/webpack.karma.config.js
```

When I remove the `jsnext:main` entry from the `package.json` inside `node_modules/redux-saga`, the error goes away. So I guess the plugin is somehow trying to find a Webpack config inside `redux-saga` to load `resolve` configuration for ES6 files? This might be fine, looking for my custom Webpack config does not make sense, though.

Billy-:
when a relative `configPath` is passed, it should be resolved relative to the source file, not the imported file.
Fixes #666

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 
 - Add [`group-exports`] rule: style-guide rule to report use of multiple named exports ([#721], thanks [@robertrossmann])
+- Fix [#666], webpack resolver incorrectly resolving relative paths to the webpack config ([#993])
 
 ## [2.8.0] - 2017-10-18
 ### Added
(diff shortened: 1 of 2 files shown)


Billy-:
Sure, I'll take a look.

Billy-:
Sorry, I think I jumped the gun a bit there. It coincidentally worked (until I ran tests).

Billy-:
Perhaps worth mentioning that the variable naming (`source` and `file`) is very confusing! `source` is the file being imported, and `file` is the source file being linted...

PR #994, opened 2018-01-05
Don't respect config setting if in a node_module

### LINKED ISSUE #666: Webpack resolver looks for custom webpack config in packages with jsnext:main

Billy-:
Fixes #666.

This has made the `resolve.moduleDirectories finds a node module` test to fail - it was the creation of the `package.json` in the fake node_module that prompted it. I have spent quite some time looking at this test alone already, and it appears that resolveSync is not finding it, and I cannot for the life of me work out why this happens...

All I can say is that I have verified that it's not my code that caused it (`checkout master`, create `package.json` in `some-module`, then run tests), just the creation of the package.json (which, of course, is expected of an npm module..)

It is 3:20AM now for me, think I'm going to call it a night. :sleeping:

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -16,12 +16,17 @@ var log = require('debug')('eslint-plugin-import:resolver:webpack')
 exports.interfaceVersion = 2
 
 /**
+ * Settings object
+ * @typedef {Object} settings
+ * @property {string} configPath the path to the webpack config
+ */
+/**
  * Find the full path to 'source', given 'file' as a full reference path.
  *
  * resolveImport('./foo', '/Users/ben/bar.js') => '/Users/ben/foo.js'
  * @param  {string} source - the module to resolve; i.e './some-module'
  * @param  {string} file - the importing file's full path; i.e. '/usr/local/bin/file.js'
- * TODO: take options as a third param, with webpack config file name
+ * @param  {settings} settings - settings object
  * @return {string?} the resolved path to source, undefined if not resolved, or null
  *                   if resolved to a non-FS resource (i.e. script tag at page load)
  */
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -44,12 +44,16 @@ exports.resolve = function (source, file, settings) {
   }
 
   var webpackConfig
-
-  var configPath = get(settings, 'config')
-    , configIndex = get(settings, 'config-index')
+    , configPath
+    , configIndex
     , packageDir
 
-  log('Config path from settings:', configPath)
+  // Only respect webpackConfig settings if not in a node_module
+  if (!~file.indexOf('/node_modules/')) {
+    configPath = get(settings, 'config')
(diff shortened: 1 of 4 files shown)


resolvers/webpack/index.js
```diff
@@ -44,12 +49,16 @@ exports.resolve = function (source, file, settings) {
...
-
-  var configPath = get(settings, 'config')
-    , configIndex = get(settings, 'config-index')
+    , configPath
+    , configIndex
     , packageDir
 
-  log('Config path from settings:', configPath)
+  // Only respect webpackConfig settings if not in a node_module
+  if (!~file.indexOf('/node_modules/')) {
```
ljharb: please never use `!~`, that's far too clever for its own good - explicitly compare to a number instead, since that's what `indexOf` returns.

resolvers/webpack/index.js
```diff
@@ -44,12 +49,16 @@ exports.resolve = function (source, file, settings) {
...
   var webpackConfig
-
-  var configPath = get(settings, 'config')
-    , configIndex = get(settings, 'config-index')
+    , configPath
+    , configIndex
     , packageDir
 
-  log('Config path from settings:', configPath)
+  // Only respect webpackConfig settings if not in a node_module
```
ljharb: i'm not sure why this is something we'd want to do. webpack config applies to *every single JS file*, potentially, including everything inside a node_modules directory.

Billy-: That’s true, for rules, etc., but I wonder if it applies to aliases or other resolve configs, too? I would be surpsied since I can’t think why anyone would want to affect the module resolution within node_modules... and if webpack did do that then surely there would be tonnes of problems as a result..?

I’d also argue that since this is an eslint plugin, the user is never interested in node_modules, thus this change does not have an adverse affect?

ljharb: It applies to everything. Webpack is used *all the time* to modify the resolution of things within node_modules - Airbnb does it at times.

This eslint plugin checks across files - so it's interested in every JS file that's required/imported, full stop.

Billy-: Okay, so is the only solution to resolve the path relative to the user’s project? I can’t see any obvious way to do that..

resolvers/webpack/index.js
benmosher: that is a pretty neat combo operator, but I agree with @ljharb 😁

PR #995, opened 2018-01-08
Ignore node_modules when finding a config by a relative path

Andarist:


resolvers/webpack/index.js
```diff
@@ -57,7 +57,10 @@ exports.resolve = function (source, file, settings) {
       // see if we've got an absolute path
       if (!configPath || !isAbsolute(configPath)) {
         // if not, find ancestral package.json and use its directory as base for the path
-        packageDir = findRoot(path.resolve(file))
+        packageDir = findRoot(path.resolve(file), function(dir) {
```
ljharb: please always add a space after the `function` keyword

Andarist: Sure, fixed here, but that really should be enforced by eslint or prettier being set up. As a newcomer to codebase I have no idea what style rules does it follow, especially if the project is not consistent about them. I've found 55 occurrences of `function(` in most important directories of this package, where 7 of them were **in** the very file I've edited.

ljharb: totally agree that we need stricter linting rules in this repo :-)

resolvers/webpack/test/config.js
```diff
@@ -91,4 +91,13 @@ describe("config", function () {
...
   })
 
+  it("finds the real config when its not an absolute path, ignoring node_modules packages on the way", function () {
+    var reexportingFile = path.join(__dirname, 'files', 'node_modules', 'module-with-reexports', 'lib', 'index.js')
+    var settings = {
+      config: 'webpack.config.js',
+    }
+
+    expect(resolve('./reexported-module', reexportingFile, settings)).to.have.property('path')
+      .and.equal(path.join(__dirname, 'files', 'node_modules', 'module-with-reexports', 'lib', 'reexported-module.js'))
```
ljharb: Can you clarify exactly how this fails when your fix is not included?

Andarist: The webpack resolver tried to load `webpack.config.js` relative to the closest directory with `package.json` - but that might be some node_module and there is obviously no `webpack.config.js` published and even if there was using it wouldnt give proper results. Webpack uses a single config for its resolution algorithm.

swrobel:
Lots of people blocked on using `import/no-unresolved` until this is fixed...

swrobel:
@benmosher @jfmengels ping!

Andarist:
@benmosher Problem is that this plugin finds `webpack.config.js` relative to the package root - this sometimes might be inside of node_modules and in that case it fails to find the project's `webpack.config.js`. 

I've written this PR some time ago so I don't quite remember - but I recall that the call site that I have edited had no sufficient info to detect this in another way and because it's a resolver plugin passing more info to this call site would require changes in the plugin system. This was the simplest fix that I have figured out.

After a quick memory search (😉) I also recall that `file` here is actually not a file that is being linted, but a reexporting file that was requested by a linting file. That's way it's harder to find webpack.config.js relative to it.

swrobel:
@Andarist is it possible to update your PR to fix the conflict so maybe, just maybe, we could get a release with this fix?

swrobel:
@Andarist ping! As a reminder, I'm using a vendored version of this fix in my app with no issues.

Andarist:
@swrobel I've rebased my branch, should be good to go if nothing terrible happened in between. Still need @benmosher approval though

PR #997, opened 2018-01-08
update eslint-import-resolver-webpack homepage to the source of the p…

graingert:
…ackage

### DIFF
--- a/resolvers/webpack/package.json
+++ b/resolvers/webpack/package.json
@@ -27,7 +27,7 @@
   "bugs": {
     "url": "https://github.com/benmosher/eslint-plugin-import/issues"
   },
-  "homepage": "https://github.com/benmosher/eslint-plugin-import#readme",
+  "homepage": "https://github.com/benmosher/eslint-plugin-import/tree/master/resolvers/webpack",
   "dependencies": {
     "array-find": "^1.0.0",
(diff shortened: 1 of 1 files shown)


resolvers/webpack/package.json
```diff
@@ -27,7 +27,7 @@
   "bugs": {
     "url": "https://github.com/benmosher/eslint-plugin-import/issues"
   },
-  "homepage": "https://github.com/benmosher/eslint-plugin-import#readme",
+  "homepage": "https://github.com/benmosher/eslint-plugin-import/tree/master/resolvers/webpack",
```
ljharb: hardcoding "master" here is dangerous; I think maybe https://github.com/benmosher/eslint-plugin-import#resolvers is a better place to link to

graingert: I was just following the pattern in the babel lerna repo. I very much doubt git's going to change the  default branch from master any time soon.

ljharb: github wouldn't, but any maintainer might - that's not the danger i'm concerned with. I don't want to link to a directory that might disappear in a future refactoring.

benmosher: I get what you're saying @ljharb, but this looks reasonable enough, especially if this is a well-known lerna pattern.

PR #998, opened 2018-01-08
Add URL to rule documentation to the metadata

Arcanemagus:
ESLint v4.15.0 added an official location for rules to store a URL to their documentation in the rule metadata in eslint/eslint#9788. This adds the URL to all the existing rules so anything consuming them can
know where their documentation is without having to resort to external packages to guess.

This also fixes the `docs` property of `no-self-import`, which was previously `doc`.

I'm not sure I like how the URL's are being defined here to fit within the 99 character line length limit, if you would like me to move the `ruleDocsUrl` to a single file that they all import, or alternatively do something like https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/commit/eb207d0ef650e9f91f7f7ca3be2ab8cbe762f0d3, just let me know!

Note: `imports-first` has a slightly different format since the documentation for this deprecated rule was removed. If you want that to instead point to `first`'s documentation that can be changed.

### DIFF
--- a/src/docsUrl.js
+++ b/src/docsUrl.js
@@ -1,10 +1,5 @@
 const repoUrl = 'https://github.com/benmosher/eslint-plugin-import'
 
-export default function docsUrl(ruleName, commitHash) {
-  let baseUrl = `${repoUrl}/tree/master/docs/rules`
-  if (commitHash) {
-    baseUrl = `${repoUrl}/blob/${commitHash}/docs/rules`
-  }
-
-  return `${baseUrl}/${ruleName}.md`
+export default function docsUrl(ruleName, commitHash = 'master') {
+  return `${repoUrl}/blob/${commitHash}/docs/rules/${ruleName}.md`
 }
--- a/src/docsUrl.js
+++ b/src/docsUrl.js
@@ -1,5 +1,10 @@
-const ruleDocsUrl = 'https://github.com/benmosher/eslint-plugin-import/tree/master/docs/rules'
+const repoUrl = 'https://github.com/benmosher/eslint-plugin-import'
 
-export default function docsUrl(ruleName) {
-  return `${ruleDocsUrl}/${ruleName}.md`
+export default function docsUrl(ruleName, commitHash) {
+  let baseUrl = `${repoUrl}/tree/master/docs/rules`
+  if (commitHash) {
+    baseUrl = `${repoUrl}/blob/${commitHash}/docs/rules`
+  }
+
+  return `${baseUrl}/${ruleName}.md`
 }
--- a/src/docsUrl.js
+++ b/src/docsUrl.js
@@ -1,0 +1,5 @@
+const ruleDocsUrl = 'https://github.com/benmosher/eslint-plugin-import/tree/master/docs/rules'
+
+export default function docsUrl(ruleName) {
+  return `${ruleDocsUrl}/${ruleName}.md`
+}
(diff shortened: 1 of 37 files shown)


Arcanemagus:
Build failures look to be "normal" (AppVeyor), or related to network timeouts on Travis CI ([#2428.5](https://travis-ci.org/benmosher/eslint-plugin-import/jobs/326501954), [#2428.6](https://travis-ci.org/benmosher/eslint-plugin-import/jobs/326501955), [#2428.13](https://travis-ci.org/benmosher/eslint-plugin-import/jobs/326501962)).

src/rules/default.js
```diff
@@ -1,8 +1,12 @@
 import Exports from '../ExportMap'
 
+const ruleDocsUrl = 'https://github.com/benmosher/eslint-plugin-import/tree/master/docs/rules'
```
ljharb: Can we import this from somewhere rather than repeating it everywhere?

j-f1: Maybe something like `docsUrl('rule-name')`?

Arcanemagus: Yep, can do that for everything other than `first`, since it needs a specific blob.

I didn't really like how this was fitting in the style here which is why I explicitly called it out as something to look at 😉.

Arcanemagus:
Updated, let me know if there are any other changes 😉.

src/rules/imports-first.js
```diff
@@ -1,5 +1,12 @@
 import first from './first'
 
-const newMeta = Object.assign({}, first.meta, { deprecated: true })
+const ruleDocsUrl = 'https://github.com/benmosher/eslint-plugin-import/blob'
+
+const newMeta = Object.assign({}, first.meta, {
+  deprecated: true,
+  docs: {
+    url: `${ruleDocsUrl}/7b25c1cb95ee18acc1531002fd343e1e6031f9ed/docs/rules/imports-first.md`,
```
ljharb: i'm confused, why does this one have a sha but the others don't? could it be an optional second argument passed to `docsUrl`?

Arcanemagus: The rule was deprecated / renamed in https://github.com/benmosher/eslint-plugin-import/commit/8e35cd93322efc478f205056833a17cd039e1902, however the documentation was directly renamed, meaning the straight mapping used elsewhere can't be used here.

An alternative here would be to simply point this rule at `first`'s documentation, as outlined in the PR description.

ljharb: Can we modify `docsUrl` to take an optional second argument that defaults to `master` but can be overridden to this sha?

Arcanemagus: The URL's are slightly different, but sure I can fix it to that.

Arcanemagus: Gave the function an optional second parameter, added tests for it, and used it for `imports-first`.

src/docsUrl.js
```diff
@@ -0,0 +1,10 @@
+const repoUrl = 'https://github.com/benmosher/eslint-plugin-import'
+
+export default function docsUrl(ruleName, commitHash) {
+  let baseUrl = `${repoUrl}/tree/master/docs/rules`
```
j-f1: This can be `blob/master` — `tree` is used for directories, while `blob` is used for files.

Arcanemagus: Wow, it looks like this is a pretty common issue since GitHub automatically redirects the wrong link, but you are absolutely correct.

PR #1000, opened 2018-01-10
Change 1st order's example's label to valid value

NumminorihSF:
This change allow copy-paste value from 1st example into eslintrc. 
It's impossible now cause valid value is `builtin` but not `buildins`.

### DIFF
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -3,7 +3,7 @@
 Enforce a convention in the order of `require()` / `import` statements. The order is as shown in the following example:
 
 ```js
-// 1. node "builtins"
+// 1. node "builtin" modules
 import fs from 'fs';
 import path from 'path';
 // 2. "external" modules


docs/rules/order.md
```diff
@@ -3,7 +3,7 @@
 Enforce a convention in the order of `require()` / `import` statements. The order is as shown in the following example:
 
 ```js
-// 1. node "builtins"
+// 1. node "builtin"
```
ljharb: this is a comment, so changing this doesn't alter it's copy-pasteability, but also it's correct as-is, as a plural. it's describing "node builtins", plural.

NumminorihSF: It's not valid for version `2.8.0`:
```
  2:1  error  Incorrect configuration of the rule: Unknown type `"builtins"`  import/order
```

Is looks like copy-pastable cause it's wrapped with quotes, and all other values are valid. Also it looks very similar to real valid value from documentation, do if I have already read the doc I may open it and copy-paste values from first section. It's just about usability. If `"builtins"` should be a correct value, we have a bug inside plugin and it's undocumented feature.

ljharb: ohhh i'm sorry, i think i get what you mean. yes, "builtin" is the config value, so this is a good change. Maybe `node "builtin" modules`?

NumminorihSF: ok, i'll change it now

PR #1005, opened 2018-01-22
Use link to tagged version for rule docs

sudo-suhas:
This uses the version present in package.json for generating the URL to rule documentation if no [commit-ish](https://stackoverflow.com/questions/23303549/what-are-commit-ish-and-tree-ish-in-git) value is passed. I think this is good because the documentation will match the version the user has installed. This way, even if a rule is changed or removed in the future, the user can find the documentation with ease.

### DIFF
--- a/src/docsUrl.js
+++ b/src/docsUrl.js
@@ -1,5 +1,7 @@
+import pkg from '../package.json'
+
 const repoUrl = 'https://github.com/benmosher/eslint-plugin-import'
 
-export default function docsUrl(ruleName, commitHash = 'master') {
-  return `${repoUrl}/blob/${commitHash}/docs/rules/${ruleName}.md`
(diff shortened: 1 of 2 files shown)


src/docsUrl.js
```diff
@@ -1,5 +1,7 @@
+import pkg from '../package'
```
ljharb: it'd be nice to use `.json` here (unless the linter objects)

sudo-suhas: Done.

tests/src/core/docsUrl.js
```diff
@@ -1,13 +1,14 @@
 import { expect } from 'chai'
 
+import pkg from '../../../package'
```
ljharb: also here :-)

sudo-suhas: Yea.. realised it after the first push 😅. Done.

PR #1007, opened 2018-01-23
Imports that are resolvable should not be reported as builtins

rixth:
`constants` is a builtin (apparently) in node. however, if we can resolve the module name, we should assume that it is _not_ a builtin.

Perhaps we could create a second rule that enforces not clobbering builtin names, but that behaviour shouldn't be part of this rule which is to enforce _ordering_.

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -21,10 +21,11 @@ export function isAbsolute(name) {
   return name.indexOf('/') === 0
 }
 
-export function isBuiltIn(name, settings) {
+export function isBuiltIn(name, settings, path) {
   const base = baseModule(name)
   const extras = (settings && settings['import/core-modules']) || []
-  return builtinModules.indexOf(base) !== -1 || extras.indexOf(base) > -1
+  const isInternal = isInternalModule(name, settings, path)
+  return !isInternal && (builtinModules.indexOf(base) !== -1 || extras.indexOf(base) > -1)
 }
 
 function isExternalPath(path, name, settings) {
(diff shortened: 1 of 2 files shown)


rixth:
Not sure what the difference is between the travis build & the appveyor build (nor why the latter is failing)

ljharb:
> however, if we can resolve the module name, we should assume that it is not a builtin.

Can you elaborate on this? In node, builtin names (like `constants`) take precedence no matter what's on the filesystem.

j-f1:
@ljharb is correct:

> Core modules are always preferentially loaded if their identifier is passed to `require()`. For instance, `require('http')` will always return the built in HTTP module, even if there is a file by that name.
> <sup>[source](https://nodejs.org/dist/latest/docs/api/modules.html#modules_core_modules)</sup>

rixth:
@ljharb that may be so for just `constants`, but deeper paths eg `constants/foo` will be required as expected

ljharb:
Hmm, that's true - also `constants/`.

I wonder if it'd be simpler to change it so that if there's any path at all - ie, anything with a slash - that it's not a builtin?

rixth:
@ljharb Happy to make this change -- would that be acceptable to you, or is there another contributor we should run this past?

ljharb:
I'd be happy to hear from @jfmengels or @benmosher on it, but both of them have been taking open source breaks for awhile now.

If there aren't any node builtins that have a `/` in the name, then this seems like the right way to address the bug.

benmosher:
> I wonder if it'd be simpler to change it so that if there's any path at all - ie, anything with a slash - that it's not a builtin?

this makes sense to me if it matches the spec 👍

ljharb:
@rixth are you no longer interested in updating this PR?

rixth:
@ljharb Correct - it is over a year old, conflicting, and I no longer work at the company/codebase where this was problematic. It was always an edge case at best.

Issue #583, opened 2016-09-24
import/extensions: does not complain about .coffee


ljharb:
I've got the following config:

``` js
['error', 'never', {
  js: 'always',
  jsx: 'always',
}]
```

However, `import '../vendor/blah.coffee';` does not warn. I would expect it to resolve the import, extract everything after the last dot, and then use that extension to determine that it should be omitted.

benmosher:
and there is no `blah.coffee.js` file sibling, I presume?

ljharb:
Correct.

PR #1009, opened 2018-02-02
Add test case for import/resolve extensions order

### LINKED ISSUE #965: This `test/src/rules/extensions.js` test should be failing with 2 errors not 1
This `invalid` test should fail with 2 errors – 1 for `.js` and another for `.jsx`.

https://github.com/benmosher/eslint-plugin-import/blob/a5844d57be0c3757248c7b824e4cc9c2c844db33/tests/src/rules/extensions.js#L129-L144

The error that I believe is missing:

```js
{
    message: 'Unexpected use of file extension "jsx" for "./bar.jsx"',
    line: 2,
    column: 22,
},

silvenon:
When there are multiple files with same names and different extensions, the resolver will go by extensions in the order specified in `import/resolve` settings.

This test case ensures that the `extensions` rule is correctly forcing the extension when omitting it would cause it to resolve to the wrong file.

Closes #965.

### DIFF
--- a/tests/src/rules/extensions.js
+++ b/tests/src/rules/extensions.js
@@ -183,6 +183,22 @@ ruleTester.run('extensions', rule, {
         },
       ],
     }),
+    // extension resolve order (#965)
+    test({
+      code: [
+        'import component from "./bar.jsx"',
+        'import data from "./bar.json"',
+      ].join('\n'),
+      options: [ { json: 'always', js: 'never', jsx: 'never' } ],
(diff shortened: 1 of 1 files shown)


silvenon:
Let me know if more test cases are needed.

silvenon:
@ljharb could you check if this is the same reason behind your #583? My local `.coffee` test is passing when I reorder the extensions.

ljharb:
It might be; but in that example there was no other file except a .coffee.

silvenon:
Ok, if you manage to reproduce it let me know and I'll update my latest PR.

PR #1012, opened 2018-02-02
Support export declarations in extensions rule

### LINKED ISSUE #964: `import/extensions` not flagging errors in `export ... from "./mod"`
As in, it's missing both of these, which should be considered errors:

```js
/* eslint import/extensions: [2, "always"] */
export {foo} from "./foo";

/* eslint import/extensions: [2, "never"] */
export {foo} from "./foo.js";
```

silvenon:
Fixes #964.

### DIFF
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -122,6 +122,10 @@ module.exports = {
 
     function checkFileExtension(node) {
       const { source } = node
+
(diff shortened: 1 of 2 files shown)


silvenon:
Done.

The reason I was minimalistic with testing this is because IMO the `import` tests verify `export` tests because they use the same code, so I only need to test whether different kinds of `export` statements pass most basic tests.

Issue #935, opened 2017-09-28
no-extraneous-dependencies only targets one package.json


Robinfr:
In the Node.js environment, a call using `require()` will walk up the tree to find the dependency, this means that a dependency that is required can be in any `node_modules` folder up the tree. Currently it only looks at one folder.

Could this be adjusted?

Robinfr:
Also related to https://github.com/benmosher/eslint-plugin-import/issues/458 which allows for a single package folder to be set. But in this case, I have multiple..

ljharb:
It walks up the tree until it finds `package.json`, then it stops.

Robinfr:
Yes, it shouldn't. It should keep walking and find all of the `package.json` file and merge them for this rule.

ljharb:
Can you provide a concrete example of what problem you need solved?

Robinfr:
Sure, I have a Lerna monorepo which has several packages that contain tests that make use of Chai and Sinon. Since I don't want to install those in all packages, I installed it at the root level. This works perfectly well, except that the this rule is complaining.

ljharb:
That's not how a lerna monorepo is supposed to be organized tho; lerna bootstrap is supposed to install all those as dev deps. Each package in that multirepo is supposed to be able to function as if it was the root of the repo.

Robinfr:
Hmm thats not true though. Just look at the Babel repository for example. Besides this works because of the way node works. So why shouldn't it work here?

Op 28 sep. 2017 5:52 p.m. schreef Jordan Harband <notifications@github.com>:

That's not how a lerna monorepo is supposed to be organized tho; lerna bootstrap is supposed to install all those as dev deps. Each package in that multirepo is supposed to be able to function as if it was the root of the repo.

ljharb:
The babel repo isn't the best example for the best way to set up a lerna monorepo :-)

Regardless, does that mean you're running the linter per-package?

Robinfr:
No I'm running the linter for all packages at once. Each package has their own .eslintrc file however.

Op 28 sep. 2017 7:06 p.m. schreef Jordan Harband <notifications@github.com>:

The babel repo isn't the best example for the best way to set up a lerna monorepo :-)

Regardless, does that mean you're running the linter per-package?

ljharb:
Gotcha.

[...]

PR #1016, opened 2018-02-07
Improve monorepo support for no-extraneous-dependencies rule

hulkish:
Solves #935

### DIFF
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -51,13 +51,13 @@ function getDependencies(context, packageDir) {
         const manifest = extractDependencies(
           JSON.parse(fs.readFileSync(path.join(dir, 'package.json'), 'utf8'))
         )
-        for (const [depType, deps] of Object.entries(manifest)) {
-          Object.assign(packageContent[depType], deps)
-        }
+        Object.keys(manifest).forEach(depType => {
+          Object.assign(packageContent[depType], manifest[depType])
+        })
       }
     }
 
-    if (!Object.values(packageContent).some(hasKeys)) {
+    if (!Object.keys(packageContent).some(depType => hasKeys(packageContent[depType]))) {
       return null
     }
 
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -7,22 +7,61 @@ import importType from '../core/importType'
 import isStaticRequire from '../core/staticRequire'
 import docsUrl from '../docsUrl'
 
-function getDependencies(context, packageDir) {
-  try {
-    const packageContent = packageDir
-      ? JSON.parse(fs.readFileSync(path.join(packageDir, 'package.json'), 'utf8'))
-      : readPkgUp.sync({cwd: context.getFilename(), normalize: false}).pkg
+function hasKeys(obj) {
+  return Object.keys(obj).length > 0
+}
 
-    if (!packageContent) {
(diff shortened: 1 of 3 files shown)


src/rules/no-extraneous-dependencies.js
```diff
@@ -7,22 +7,66 @@ import importType from '../core/importType'
...
+    const packageContent = {
+      dependencies: {},
+      devDependencies: {},
+      optionalDependencies: {},
+      peerDependencies: {},
+    }
 
-    if (!packageContent) {
-      return null
+    if (Object.prototype.hasOwnProperty.call(context.settings, 'import/paths')) {
```
ljharb: let's use the `has` module for this.

hulkish:
@ljharb It's a documented approach in the readme for lerna: https://github.com/lerna/lerna/blob/master/README.md#common-devdependencies

Furthermore, node itself inherently resolves dependencies from parent `node_modules` dirs. 

When you have... let's say 10-30 packages you're managing in a monorepo - then sure, not a big deal to manage dev deps on a per package basis.

However, at my job for example we manage a monorepo will well over 100 packages.

In this scenario; the task of managing dev deps for this volume of packages quickly outgrows what's practical.

While I agree that this is up for debate on a best practices basis - I'm not sure how valueable it is to have this flexibility prevented at the level of an eslint plugin.

Personally, I've found that hoisting all of these dev deps to the root of monorepo brings a significant benefit and gained sanity.

tizmagik:
Thank you for this PR @hulkish -- this has been a pain point for us so far. Hopefully this gets merged and released 🙏

hulkish:
@ljharb hoping this can land.. is there anything I can improve to allow it?

godban:
Thank for PR. We also reach this problem in monorepo. There even extra plugin in the wild https://github.com/Dreamscapes/eslint-import-resolver-lerna which partially solves it.
But have a solution in `eslint-plugin-import` is definitely better.

dmk255:
I've also run into this issue. I would love to see this merged.

hulkish:
@benmosher hmm, sure - but, how would you configure `packageDir` in your `.eslintrc`?

dmk255:
@hulkish I believe the relevant PR is https://github.com/benmosher/eslint-plugin-import/pull/685

benmosher:
thanks for the PR reference @dmk255! other relevant reading: 

- #1053 
- #458

evocateur:
I just turn off `import/no-extraneous-dependencies` for any test files in the monorepo with an [`overrides`](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns) key. `import/no-unresolved` already handles "is it installed somewhere", so extraneous-ness in test files seems beside the point. `${ROOT}/.eslintrc.yaml` looks like this:

```yaml
---
  extends: airbnb
  root: true
  overrides:
    - files: "**/*{-,.}{test,stories}.js"
      rules:
        # devDependencies are all in the root
        import/no-extraneous-dependencies: off
```

hulkish:
I'll just fork and publish this as a different package.

ljharb:
You’re welcome to do that, but this should stay open.

hulkish:
Just wondering... what's holding this up? What needs to be done to get this merged?

benmosher:
busy week at work, but it's top of my queue for this project when I get time

hulkish:
@benmosher thanks!

benmosher:
@hulkish fyi, merged mine because I had fixed up and added a few more tests, LMK if I missed anything though

PR #1017, opened 2018-02-08
Replaced read-pkg-up with cache-aware package locator

hulkish:


src/core/createPackageLocator.js
```diff
@@ -0,0 +1,69 @@
...
+      try {
+        const location = path.join(dirname, 'package.json')
+        locations.push(location)
+
+        return this.readPackageSync(location)
+      } catch (err) {
+        // if not found, allopw search to continue
+        if (err.code !== 'ENOENT') throw err
+      }
+    } while (dirname !== (dirname = path.dirname(dirname)))
```
ljharb: is there a way this can be written without conflating an assignment with an expression?

src/core/createPackageLocator.js
```diff
@@ -0,0 +1,69 @@
...
+    this.requests = new Map()
+  }
+
+  readPackageSync(location) {
+    const cached = this.store[location]
+    if (cached) return cached
+
+    const response = super.readPackageSync(location)
+    this.store[location] = response.result
+    return response
```
ljharb: if we only ever use the `result`, why return the entire response?

src/core/createPackageLocator.js
hulkish: @ljharb Yes, but this plays nicer with the goal of "fail fast". Meaning, if i were to replace with a typical condition-first loop such as for/while - it would require me to add logic which ensured the initial dirname was searched.

src/core/createPackageLocator.js
hulkish: @ljharb It's needed so that `readPackageUpSync` can report all locations which failed to yield a file - in the event that no fles were located.

hulkish:
@ljharp one thing that occurs to me... is it safe for the cache to exist at module level as i currently have it?

Are there instances where I should be clearing the cache?

hulkish:
@ljharb Ok I think this is ready now. 

One way to easily see the clear benefit here, is by running the following command in both `master` and the branch for this pr:
```bash
BABEL_ENV=test NODE_PATH=./src ./node_modules/.bin/mocha --require babel-register tests/src/rules/no-extraneous-dependencies.js -t 5s --trace-sync-io 2>&1 | grep -c "no-extraneous-dependencies.js"
```

The difference is clear when u see that master produces **524** io calls from this rule. The branch, using same command yields **25**

src/core/CachedPackageLocator.js
```diff
@@ -0,0 +1,61 @@
+import path from 'path'
+import os from 'os'
+import fs from 'fs'
+
+export default class CachedPackageLocator {
```
ljharb: i wonder if this would be better extracted to its own published package with its own tests - is that something you'd be willing to publish and maintain?

src/rules/no-extraneous-dependencies.js
```diff
@@ -131,10 +115,15 @@ module.exports = {
     ],
   },
 
-  create: function (context) {
+  create(context) {
     const options = context.options[0] || {}
     const filename = context.getFilename()
-    const deps = getDependencies(context, options.packageDir)
+    const deps = packageLocator.readUpSync(
```
ljharb: it seems kind of odd to instantiate `packageLocator` only once and then use it only once to get at the `readUpSync` instance method. Perhaps `readUpSync` could just be exported directly as a singleton function, or perhaps `CachedPackageLocator` could be a factory function instead?

tests/src/core/CachedPackageLocator.js
```diff
@@ -0,0 +1,212 @@
...
+      description: 'foo bar',
+    })
+  )
+  const context = {
+    report: sinon.spy(),
+  }
+
+  before(function () {
+    sandbox = sinon.sandbox.create()
+    sandbox.stub(fs, 'readFileSync').reset()
```
ljharb: why would it need to be reset immediately after creating the stub?

hulkish: It won't work without it. Some kind of sinon bug, I gathered.

src/core/CachedPackageLocator.js
hulkish: Hmm sure, I'll try to find some some for this either today or this week.

tests/src/core/CachedPackageLocator.js
ljharb: What happens when it doesn't work?

tests/src/core/CachedPackageLocator.js
```diff
@@ -0,0 +1,212 @@
...
+  })
+
+  after(function () {
+    sandbox.restore()
```
ljharb: this should be done `afterEach`.

hulkish: I understand, that's what I had originally also - it does not initialize the stubbed method without calling reset, though. Try it yourself!

ljharb: I'll play with it; but in the meantime let's change all this stuff so the sandbox is created `before` but everything else is in a beforeEach/afterEach.

ljharb:
@hulkish are you still interested in completing this?

PR #1029, opened 2018-02-22
Release v2.9.0

benmosher:


benmosher:
it's really unclear to me why the windows build is so hosed. seems like not a new issue, nor a great reason to hold off on publishing. worst case, windows users need a patched version later. probably just some issue with the install process during testing.

PR #1044, opened 2018-03-12
add-missing-env-variable-in-webpack-config

kamalbennani:
Following https://github.com/benmosher/eslint-plugin-import/pull/938, I am facing a problem where my project is composed of a several applications and each them might define a "custom" webpack configuration which extends a common configuration which is defined in the root of the project.
```
/
apps/
- app1/
  webpackConfig.js  <-- specific to app1
  .eslintrc
- app2/
  webpackConfig.js  <-- specific to app2
  .eslintrc
- app3/
  webpackConfig.js <-- specific to app3
  .eslintrc
webpack.config.js <-- common configuration
```
A custom webpack configuration looks like so:
__apps/app1/webpackConfig.js__
```JS
export function (baseConfig) => {
 newConfig = baseConfig
 // Some custom manipulation
 ....
 return newConfig
}
```

**Note that the common webpack configuration is responsible for determining which webpack configuration to load based on the application name**.
So for instance, to run the `app1`, we execute the following command :
```SH
webpack-dev-server  --config webpack.config.js --env app1
```

The problem is that the 'env' variable is being ignored, So I can't do something like that:
__apps/app1/.eslintrc__
```JS
const path = require('path');
module.exports = {
  settings: {
    'import/resolver': {
      webpack: {
        config: path.resolve(__dirname, '../../webpack.config.js'),
        env: 'app1'
      }
    }
  },
};
```
The solution is to grab the env from the settings and to pass it to the webpackConfig in the webpack resolver:
__resolvers/webpack/index.js__
```JS
var env = get(settings, 'env')
if (typeof webpackConfig === 'function') {
   webpackConfig = webpackConfig(env)
}
```

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -47,6 +47,7 @@ exports.resolve = function (source, file, settings) {
 
   var configPath = get(settings, 'config')
     , configIndex = get(settings, 'config-index')
+    , env = get(settings, 'env')
     , packageDir
 
   log('Config path from settings:', configPath)
@@ -82,7 +83,7 @@ exports.resolve = function (source, file, settings) {
   }
 
   if (typeof webpackConfig === 'function') {
-    webpackConfig = webpackConfig()
+    webpackConfig = webpackConfig(env)
   }
 
   if (Array.isArray(webpackConfig)) {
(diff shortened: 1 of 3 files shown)


kamalbennani:
Here you are, thanks for accepting the PR :+1:

kamalbennani:
Can you tell me when are you planning to release the resolver-webpack, because as you might now I can't target the commit so I need to wait for the release.
By the way thanks for the merge

ljharb:
@kamalbennani v0.9.0 was published.

Issue #1040, opened 2018-03-09
Auto fix for rule "first"


fengkfengk:
I would like to work on the rule `first` auto fixer. The small and controllable problem should be auto fixed. If can be merged, I may start working now. The strategy is put the no-first import declarations into the bottom of the first import declarations.

ljharb:
That sounds fine, as long as the following criteria apply:
1. imports are never re-ordered relative to each other, even if `absolute-first` is set.
1. if an import creates an identifier, and that identifier is referenced at module level *before* the import itself, that must also not be re-ordered.

fengkfengk:
About the second point.
`const foo = bar;
import bar from './bar.js';`
This won't be re-ordered?

ljharb:
@fengkfengk correct, because that code will currently throw a TDZ error, and that error should be maintained by the autofixer.

PR #1046, opened 2018-03-14
add fixer for `first`

fengkfengk:
Add autofixer for rule `first`. Refer issue #1040.

### DIFF
--- a/tests/src/rules/first.js
+++ b/tests/src/rules/first.js
@@ -20,12 +20,19 @@ ruleTester.run('first', rule, {
                   export { x };\
                   import { y } from './foo';"
          , errors: 1
+         , output: "import { x } from './foo';\
+                  import { y } from './foo';\
+                  export { x };"
          })
   , test({ code: "import { x } from './foo';\
                   export { x };\
                   import { y } from './bar';\
                   import { z } from './baz';"
          , errors: 2
+         , output: "import { x } from './foo';\
+                  import { y } from './bar';\
+                  import { z } from './baz';\
+                  export { x };"
          })
   , test({ code: "import { x } from './foo'; import { y } from 'bar'"
          , options: ['absolute-first']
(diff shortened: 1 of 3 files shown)


tests/src/rules/first.js
```diff
@@ -20,12 +20,19 @@ ruleTester.run('first', rule, {
...
                   export { x };\
                   import { y } from './bar';\
                   import { z } from './baz';"
          , errors: 2
+         , output: "import { x } from './foo';\
+                  import { y } from './bar';\
+                  import { z } from './baz';\
+                  export { x };"
          })
   , test({ code: "import { x } from './foo'; import { y } from 'bar'"
```
ljharb: can we validate that the output matches the input here?

tests/src/rules/first.js
```diff
@@ -35,7 +42,37 @@ ruleTester.run('first', rule, {
...
+                  'use directive';"
          })
+  , test({ code: "var foo = bar;\
+                  import { x } from './foo';\
+                  import { y } from './bar';\
+                  import { z } from './baz';"
+         , errors: 3
+         , output: "import { x } from './foo';\
+                  import { y } from './bar';\
+                  import { z } from './baz';\nvar foo = bar;"
```
ljharb: nbd but could this use a continuation instead of a literal `\n`?

tests/src/rules/first.js
```diff
@@ -35,7 +42,37 @@ ruleTester.run('first', rule, {
...
+  })
+  , test({ code: "var a = x;\
+                  import { x } from './foo';\
+                  import { y } from './bar';\
+                  import { z } from './baz';"
+         , errors: 2
+         , output: "import { y } from './bar';\
+                  import { z } from './baz';\nvar a = x;\
```
ljharb: this is actually incorrect; because `foo` might depend on side effects from `bar` and `baz` - this example can't be reordered at all.

tests/src/rules/first.js
```diff
@@ -35,7 +42,37 @@ ruleTester.run('first', rule, {
...
+                  import { x } from './foo';"
+  })
+  , test({ code: "if (true) { x() };\
+                  import { x } from './foo';\
+                  import { y } from './bar';\
+                  import { z } from './baz';"
+         , errors: 3
+         , output: "import { x } from './foo';\
```
ljharb: same here.

tests/src/rules/first.js
fengkfengk: it must be a seperator here. I thought a line-breaker is suitable, or may be another option?

ljharb: It just looks odd to see the previous lines using \ but this one using \n

ljharb:
Are you no longer interested in completing this PR?

(If so, please reopen this one instead of opening another)

fengkfengk:
now if a specifier is referenced before the import, it won't be reordered.

ljharb:
@fengkfengk now we just need to get coverage metrics back up, and we can merge this!

fengkfengk:
now the coverage improved, some unreachable lines are fixed.

Issue #941, opened 2017-10-09
rule to prohibit circular dependencies


ljharb:
I'd love a rule that errored out whenever any circular dependency exists - supporting both `require` and `import`.

benmosher:
I think this is pretty easily achievable. Would be cool. Probably just need to enhance `ExportMap` to make a note if it finds its way back to itself.

re: supporting both `require`/`import`, `moduleVisitor` exists to abstract away the import type. so in the future we can add dynamic `import()` support there and all rules can pick it up immediately.

Code: https://github.com/benmosher/eslint-plugin-import/blob/master/utils/moduleVisitor.js

see `no-unresolved` rule implementation for an example

aminmarashi:
I need this, so I'm going to give it a shot.

mzedeler:
See also: https://github.com/eslint/eslint/issues/9096

benmosher:
fyi (esp. @aminmarashi) I just roughed this in on the `no-cycles` branch. haven't tested what I've got at all, just jotting down thoughts mostly. will post here if/when I get it completed.

benmosher:
pushed a working version with a handful of tests to `no-cycles` branch. I am concerned it is super inefficient since it may BFS-search the full dependency tree for every import in every file in a project. certainly seems like some broad persistent cache would be helpful.

if anyone can test it out on a decent-sized project and post before/after lint times, that would be great info.

benmosher:
cc @RealBlazeIt: thought you might want to know this is in progress since you suggested it to ESLint core in eslint/eslint#9096, as noted by @mzedeler

benmosher:
First draft increased my test project lint time from 1 to 2 minutes. Pretty substantial increase. That said, also a pretty helpful feature. 

I will keep thinking about cache strategies but I’ll probably go ahead and merge the rule in the mean time, since it is working and useful as is, and not as horrible perf as I had initially worried (for my ~1000 module test project, anyway).

sibelius:
we have 2 rules to handle this already

import/no-cycle
import/no-self-import

can we close this?

jaylattice:
Our codebase uses `import` and `require`, along with module aliases. It looks like `import/no-cycle` does not account for both of these cases. Correct me if I'm wrong.

ljharb:
@jaylattice ideally the rule should handle them all. if it doesn’t, please file an issue/PR for that :-)

wycats:
[...]

PR #1052, opened 2018-03-26
rule: no-cycle

benmosher:
Rule implementation for #941.

I need to do a final smoke test locally before merging, but wanted to open this up for implementation/docs/tests comments. Most notably, this first draft won't detect cycles introduced by deep `require`s. It will only _lint_ `require` calls that import modules that import modules.


docs/rules/no-cycle.md
```diff
@@ -0,0 +1,63 @@
...
+This rule does _not_ detect imports that resolve directly to the linted module;
+for that, see [`no-self-import`].
+
+
+## Rule Details
+
+### Options
+
+By default, this rule only detects cycles for ES6 imports, but see the [`no-unresolved` options](./no-unresolved.md#options) as this rule also supports the same `commonjs` and `amd` flags. However, these flags only impact which import types are _linted_; the
+import/export infrastructure only registers `import` statements in dependencies, so
```
ljharb: It seems super important that this can detect requires as well as imports. What’s involved in making that happen?

tests/files/cycles/depth-three-indirect.js
```diff
@@ -0,0 +1,5 @@
+import './depth-two'
+
+export function bar() {
+    return "side effects???"
+}
```
ljharb: nbd ofc but missing a trailing newline

tests/src/rules/no-cycle.js
```diff
@@ -0,0 +1,89 @@
...
+const ruleTester = new RuleTester()
+    , rule = require('rules/no-cycle')
+
+const error = message => ({ ruleId: 'no-cycle', message })
+
+const test = def => _test(Object.assign(def, {
+  filename: testFilePath('./cycles/depth-zero.js'),
+}))
+
+describe.only("no-cycle", () => {
```
ljharb: this only probably needs to be removed

benmosher: agh! yep 🤕

docs/rules/no-cycle.md
benmosher: Yeah, I knew you'd pick up on that as I realized it and wrote it out. 😅 

Right now, all the normal static analysis inspection of dependencies only picks up imports/exports from the topmost scope of the module. 

Could also potentially nab any single-argument `require`s that are either alone or in `AssignmentExpression`s but everything gets dicier once `require`s are in the mix, since they could be present in any scope in the module, technically.

I want to avoid blocking merging and shipping this on supporting deep registration of CJS `require`s, though.

ljharb: Could you use an ast selector like `CallExpression[callee.name=“require”]` for the visitor?

benmosher: For sure. That's certainly what the `moduleVisitor` implementation does to find and lint `require`s.

It would be a substantial improvement, and I think I am probably overstating the complexity. I'll try to take a second look.

Issue #921, opened 2017-08-24
import/named doesn't detect Flow Opaque types in flow comments


johnhaley81:
Hi all!

When using flow [opaque types](https://flow.org/en/docs/types/opaque-types/) (new in https://github.com/facebook/flow/releases/tag/v0.51.0) via flow [comments](https://flow.org/en/docs/types/comments/) The `import/named` rule doesn't detect the exported types.

Ex:

### Foo.js
```javascript
// @flow

/*::
export opaque type Foo = string;
*/

export const makeFoo = (): Foo => 'new foo';
```

### UseFoo.js
```javascript
// @flow

import { type Foo } from './Foo'; // Foo not found in './Foo' (import/named)
```

A weird workaround is to remove `makeFoo` from `Foo.js` and that will fix the linter rule. Don't know why that is. It's also not a good workaround since that makes the opaque type useless.

SpainTrain:
Is this an issue only with flow comments?

Possibly related to https://github.com/gajus/eslint-plugin-flowtype/issues/260 ?

johnhaley81:
Nope, I'm able to take off the flow comments and this error is still triggering a false-positive.

captainkovalsky:
@johnhaley81 I should notice that eslint couldn't track the flow stubs generated by flow-typed

syymza:
I have run locally the following test on current master (in the valid rules):
```
    test({
      code: 'import type { MyOpaqueType } from "./flowtypes"',
      parser: 'babel-eslint',
    }),
```

It is passing. Can we have a release?

ljharb:
Closed by #1057, in that case.

PR #1057, opened 2018-03-30
Ignore type imports for named rule.

### LINKED ISSUE #931: [named] flow-typed definitions are not used unless node_modules are ignored
eslint-plugin-import: `7.2.0`

I was having the same issue as #708, where modules whose types are defined in a `flow-typed` definition were throwing false positive `import/named` errors about not being found.

I realized that adding the following to my `.eslintrc` solves the problem:

```yaml
settings:
    import/ignore:
      - node_modules
```

So, it seems that if `node_modules` are not ignored, this plugin will look into the node_module for the type definition, and if it doesn't find it, it throws an error.  Ideally, it would also look in `flow-typed` (using whatever method it's currently using when `node_modules` are ignored) before reporting the error.

mattijsbliek:
This fixes #931 for the named rule. This was done to play nice with flow-typed in the following situation:

```js
/*
 * Import an NPM module and grab the type definition from flow-typed
 */
import Something from 'my-npm-package';
// This type isn’t actually in the npm package but in the flow-typed directory, so eslint-import cannot resolve it.
import type { SomethingType } from 'my-npm-package';
```

Closes #1060.

### DIFF
--- a/tests/src/rules/named.js
+++ b/tests/src/rules/named.js
@@ -67,18 +67,10 @@ ruleTester.run('named', rule, {
     test({ code: 'import { deepProp } from "./named-exports"' }),
     test({ code: 'import { deepSparseElement } from "./named-exports"' }),
 
-    // flow types
+    // should ignore imported flow types, even if they don’t exist
     test({
-      code: 'import type { MyType } from "./flowtypes"',
-      'parser': 'babel-eslint',
-    }),
-    test({
-      code: 'import type { MyInterface } from "./flowtypes"',
-      'parser': 'babel-eslint',
-    }),
-    test({
-      code: 'import type { MyClass } from "./flowtypes"',
-      'parser': 'babel-eslint',
+      code: 'import type { MissingType } from "./flowtypes"',
+      parser: 'babel-eslint',
     }),
 
     // TypeScript
@@ -243,16 +235,6 @@ ruleTester.run('named', rule, {
     //     type: 'Literal',
     //   }],
     // }),
-
-    // flow types
-    test({
-      code: 'import type { MissingType } from "./flowtypes"',
-      parser: 'babel-eslint',
-      errors: [{
-        message: "MissingType not found in './flowtypes'",
-        type: 'Identifier',
-      }],
-    }),
 
     // TypeScript
     test({
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+- Ignore type imports for named rule ([#931], thanks [@mattijsbliek])
 - Add documentation for [`no-useless-path-segments`] rule ([#1068], thanks [@manovotny])
 - Fixer for [`first`] ([#1046], thanks [@fengkfengk])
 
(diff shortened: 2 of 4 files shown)


mattijsbliek:
Absolutely, this is not ideal. Flow can resolve the imports because it checks the directories listed under `[libs]` in .flowconfig. Ideally you would just add the flow-typed directory to `modules` array of the import/resolver config options. The reason this doesn’t work is because the location of the typed file does not map 1:1 to the import:

```
// Import from 'bar'
import type { Foo } from 'bar';

// Actual typed file is here:
flow-typed/npm/bar_vx.x.x.js

// And type looks like this
declare module 'bar' {
  declare export type Foo = 'TYPE_HERE';
}
```

I think Flow just goes through all the files in the flow-typed directory and accepts the declared modules. I’m not sure there would be a way for ESLint Plugin Import to work around this without doing the same thing. But this would mean it would have to understand Flow syntax and everything, which I think is not something we would want.

Not sure how to otherwise solve this if not for just skipping typed imports. What do you think?

ljharb:
The current scenario is that the plugin correctly verifies some types, but falsely reports on those in flow-typed - this change removes the usefulness in the former case, correct?

mattijsbliek:
Correct! The plugin can verify all types that are just normally exported from a file.

ljharb:
What about if we did this instead: if the file being imported *exports* types, then do the checking - if it does *not*, then for now, ignore it.

mattijsbliek:
Nice, that could work. I’ll try and implement this today.

mattijsbliek:
Spent the past 4 hours trying to get the export types from the imported file, but couldn’t figure it out. I’m not very familiar with ASTs or ESLint, could you point me in the right direction? If it’s easier for someone else to take over I can create a new issue as well.

tf:
Why does ESLint need to check type imports at all? If we are importing types, doesn't that mean Flow is already in the picture and will check things itself? The only case I currently see that is handled by this plugin but not by Flow is when importing things from a non-Flow file. But then we cannot import types anyway. What am I missing?

mattijsbliek:
I do think that’s a good argument, Flow indeed already does the checking for missing imported types. What are your thoughts on this @ljharb?

ljharb:
Hmm, that's a fair point.

Let's keep this as-is for now - the PR needs a fresh rebase tho.

tests/src/rules/named.js
```diff
@@ -244,14 +244,11 @@ ruleTester.run('named', rule, {
...
+    // should ignore imported flow types, even if they don’t exist
     test({
       code: 'import type { MissingType } from "./flowtypes"',
       parser: 'babel-eslint',
-      errors: [{
-        message: "MissingType not found in './flowtypes'",
-        type: 'Identifier',
-      }],
+      errors: [],
     }),
```
tf: Looking at the overall structure of the test, I suggest moving this into the group of valid examples, instead of having it under invalid but with no errors.

If we'll just ignore type imports, I'd even go a step further and remove the remaining valid `import type` examples since they convey the impression that the plugin still cares about whether a type import is valid or not.

mattijsbliek: Good suggestions, I made it so.

chadlwilson:
Thanks a lot for the work here - this has been causing us some issues. Am I right in thinking this needs a new plugin release (not in 2.11.0)? If so, @benmosher could we plleeeeease get one, even if it's beta? :)

syymza:
This seems to fix #921. 🎺

 Can we have a new release including it?

benmosher:
I thought this was released in the last release but perhaps not?

syymza:
@benmosher this was merged 18 days ago. The last release was 19 days ago.

benmosher:
wah wah. my bad. was hoping for this morning but burned all my time on something else. still likely to be this week.

benmosher:
just published! 😅

PR #1060, opened 2018-04-03
[WIP] Do not report type imports from flow-typed packages

### LINKED ISSUE #960: Error ESLint: <Flow-typed type definition> not found in module (import/named)
I use to flow type checking & flow-typed defs in my project.
I have installed libDef for redux package also
```
import type { Store } from 'redux';
              ^^^ Error:() ESLint: Store not found in 'redux' (import/named)
```
Clicking to Store is pointed me to type definition file from flow-typed directory, but plugin state that such definition has not found in the module.
Though flow correctly identifies Store and displays type info in IDE

tf:
fixes #960

When importing types from packages that are defined in a file inside the `flow-typed` directory, prevent the `named` rule from reporting that the package itself does not provide an export with that name.

Currently there is support for type imports from local files. Restrict these checks to imports with relative paths.

I'm not really sure how much value checking type imports provides since, I suppose, Flow itself is already good at alerting about missing imported types. Maybe type imports should not be checked at
all? At least that appears to be [the route already taken in `no-extraneous-dependencies`](https://github.com/benmosher/eslint-plugin-import/commit/523789f510b0cd1dfd7bbddfd26c1cca82a6d1c0). Does the same hold for Type Scripts?

The alternative of making the `named` rule aware of the top level `flow-typed` directory feels a lot like duplicating functionality that already exists inside Flow.

### DIFF
--- a/src/rules/named.js
+++ b/src/rules/named.js
@@ -18,6 +18,8 @@ module.exports = {
         return // no named imports/exports
       }
 
+      if (isTypeFromPackage(node)) return
+
       const imports = Exports.get(node.source.value, context)
       if (imports == null) return
 
@@ -45,6 +47,14 @@ module.exports = {
           }
         }
       })
+    }
+
+    function isTypeFromPackage(node) {
+      return node.importKind === 'type' && !hasRelativePath(node.source)
+    }
+
+    function hasRelativePath(source) {
+      return source.value.match(/^\.\//)
     }
 
     return {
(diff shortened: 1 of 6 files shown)


mattijsbliek:
I think we’ve opened roughly the same pull request, you might be interested in the discussion here: https://github.com/benmosher/eslint-plugin-import/pull/1057

ljharb:
Indeed; thanks for your contribution, but this is a duplicate of #1057.

PR #1064, opened 2018-04-05
[Fix] `no-cycle`: `create` must *always* return an object, even if no listeners

ljharb:
This is breaking eslint-plugin-airbnb's tests, and will break on any linting that's done on piped-in text.

cc @sharmilajesupaul @not-an-aardvark


ljharb:
The other cases are inside a nested function; there’s only 2 returns inside create.

PR #1065, opened 2018-04-05
minor typo in import/no-cycle rule docs

sharmilajesupaul:
In the  [`import/no-cycle`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-cycle.md#maxdepth) docs, there's an example for using options. 

this is the current text: 

```js
/*eslint import/no-unresolved: [2, { maxDepth: 1 }]*/
```

this PR updates it to: 
```js
/*eslint import/no-cycle: [2, { maxDepth: 1 }]*/
```


benmosher:
@sharmilajesupaul thanks, nice catch!

PR #1067, opened 2018-04-07
Fix link to order autofixer in CHANGELOG

swrobel:


swrobel:
Oops, deleted & re-forked too soon. New PR incoming, and thanks for pointing out the linking issue, I guess that explains why it wasn't linked in my preview.

PR #1069, opened 2018-04-07
[patch] use `resolve` instead of `builtin-modules`

ljharb:
`resolve` has a much more thorough list of builtins, and since I maintain it, we can keep it updated more easily than with `builtin-modules`.


ljharb:
Absent objections, I'm going to merge this once tests pass.

ljharb:
appveyor failure is a fluke.

PR #1070, opened 2018-04-09
Dynamic import chunkname rule

byteme980:
This PR adds a new rule `dynamic-import-chunkname,` which reports on any dynamic-import usages without a properly formatted webpackChunkName leading comment as mentioned in [this tweet](https://twitter.com/lencioni/status/973957708968706050).

This is a useful rule to have, since explicitly defining the webpackChunkName will provide a consistent file name for the code split boundary.

cc @lencioni @ljharb

### DIFF
--- a/docs/rules/dynamic-import-chunkname.md
+++ b/docs/rules/dynamic-import-chunkname.md
@@ -1,0 +1,66 @@
+# dynamic imports require a leading comment with a webpackChunkName (dynamic-import-chunkname)
+
+This rule reports any dynamic imports without a webpackChunkName specified in a leading block comment in the proper format.
+
+This is a useful rule because if the webpackChunkName is not defined in a dynamic import, Webpack will autogenerate the chunk name.
+
+## Rule Details
+This rule runs against `import` by default, but can be configured to also run against an alternative dynamic-import function, e.g. 'dynamicImport.'
+You can also configure the regex format you'd like to accept for the webpackChunkName - for example, we don't want the number 6 to show up in our chunk names.
+ ```javascript
+{
+  "dynamic-import-chunkname": [2, {
+    importFunction: "dynamicImport",
+    webpackChunknameFormat: "[a-zA-Z0-57-9-/_]"
+  }]
+}
+```
+
+### invalid
+The following patterns are invalid:
+
+```javascript
+// no leading comment
+import('someModule');
+
+// incorrectly formatted comment
+import(
+  /*webpackChunkName:"someModule"*/
+  'someModule',
+);
+
+// chunkname contains a 6 (forbidden by rule config)
+import(
+  /* webpackChunkName: "someModule6" */
+  'someModule',
+);
+
+// using single quotes instead of double quotes
+import(
+  /* webpackChunkName: 'someModule' */
+  'someModule',
+);
+
+// single-line comment, not a block-style comment
+import(
+  // webpackChunkName: "someModule"
+  'someModule',
+);
+```
+### valid
+The following patterns are valid:
+
+```javascript
+  import(
+    /* webpackChunkName: "someModule" */
+    'someModule',
+  );
+  import(
+    /* webpackChunkName: "someOtherModule12345789" */
+    'someModule',
+  );
+```
+
+## When Not To Use It
+
+If you don't care that Webpack will autogenerate chunk names and may blow up browser caches and bundle size reports.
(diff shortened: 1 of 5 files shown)


docs/rules/dynamic-import-chunkname.md
```diff
@@ -0,0 +1,66 @@
+# dynamic imports require a leading comment with a webpackChunkName (dynamic-import-chunkname)
+
+This rule reports any dynamic imports without a webpackChunkName specified in a leading block comment in the proper format.
+
+This is a useful rule because if the webpackChunkName is not defined in a dynamic import, Webpack will autogenerate the chunk name.
```
lencioni: This would be better if it explained why it is better to have control over the chunk name vs letting webpack autogenerate one. Basically mention that autogenerated chunk names are not stable across builds, which prevents long-term browser caching.

nit: "webpack" should be lowercase everywhere

docs/rules/dynamic-import-chunkname.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Rule Details
+This rule runs against `import` by default, but can be configured to also run against an alternative dynamic-import function, e.g. 'dynamicImport.'
```
lencioni: `import` here should probably be `import()` to differentiate it as a dynamic import.

docs/rules/dynamic-import-chunkname.md
```diff
@@ -0,0 +1,66 @@
...
+You can also configure the regex format you'd like to accept for the webpackChunkName - for example, we don't want the number 6 to show up in our chunk names.
```
lencioni: "for example, we don't" should maybe be "for example, if we don't". I'd probably also end this sentence with a `:` instead of a `.`

docs/rules/dynamic-import-chunkname.md
```diff
@@ -0,0 +1,66 @@
...
+ ```javascript
+{
+  "dynamic-import-chunkname": [2, {
+    importFunction: "dynamicImport",
```
lencioni: I wonder if this should be `importFunctions` and take an array instead of a string. What do you think?

src/rules/dynamic-import-chunkname.js
```diff
@@ -8,8 +8,11 @@ module.exports = {
     schema: [{
       type: 'object',
       properties: {
-        importFunction: {
-          type: 'string',
+        importFunctions: {
+          type: 'array',
```
ljharb: can we add `uniqueItems: true` to this array?

src/rules/dynamic-import-chunkname.js
```diff
@@ -0,0 +1,70 @@
...
+    const { importFunctions = [] } = config || {}
+    const { webpackChunknameFormat = '[0-9a-zA-Z-_/.]+' } = config || {}
+
+    const commentFormat = ` webpackChunkName: "${webpackChunknameFormat}" `
+    const commentRegex = new RegExp(commentFormat)
+
+    return {
+      [`CallExpression[callee.type="Import"],CallExpression[callee.name]`](node) {
+        const { callee: { name }} = node
+        if (name && !importFunctions.includes(name)) {
```
ljharb: i'm not sure if we can use this in node 4, which eslint supports. you may want to do `new Set(importFunctions)` and then use `.has`?

src/rules/dynamic-import-chunkname.js
```diff
@@ -0,0 +1,71 @@
...
+  create: function (context) {
+    const config = context.options[0]
+    const { importFunctions = [] } = config || {}
+    const { webpackChunknameFormat = '[0-9a-zA-Z-_/.]+' } = config || {}
+
+    const commentFormat = ` webpackChunkName: "${webpackChunknameFormat}" `
+    const commentRegex = new RegExp(commentFormat)
+
+    return {
+      [`CallExpression[callee.type="Import"],CallExpression[callee.name]`](node) {
```
byteme980: @ljharb, let me know if you know of a better way to write these selectors now that we're accepting an array of import functions instead of just a string

ljharb: hmm - i don't think it has to be a blocker or anything, but you *could*, perhaps, dynamically construct the method name like so:
```js
const selector = ['CallExpression[callee.type="Import"]'].concat(importFunctions.map(func => `CallExpression[callee.name="${func}"]`).join(',');

return {
  [selector](node) {
```

but at this point I'm not sure if the readability hit is worth the performance optimization. What do you think?

byteme980: I think this may be what's failing the Travis build for eslint version 2

ljharb: oh, snap - good call. if eslint 2 doesn't have full AST selectors, then we should go back to your original code that just checks `CallExpression` and does the filtering inside the method body :-(

PR #1093, opened 2018-05-02
No relative parent imports rule

chrislloyd:
# no-relative-parent-imports

Use this rule to prevent imports to folds in relative parent paths.

It's useful for large codebases codebases to enforce directed-acyclic-graph like folder structures.

### Examples

Given the following folder structure:

```
my-project
├── lib
│   ├── a.js
│   └── b.js
└── main.js
```

And the .eslintrc file:
```
{
  ...
  "rules": {
    "import/no-relative-parent-imports": "error"
  }
}
```

The following patterns are considered problems:

```js
/**
 *  in my-project/lib/a.js
 */

import bar from '../main'; // Import parent file using a relative path
```

The following patterns are NOT considered problems:

```js
/**
 *  in my-project/main.js
 */

import foo from 'foo'; // Import package using module path
import a from './lib/a'; // Import child file using relative path

/**
 *  in my-project/lib/a.js
 */

import b from './b'; // Import sibling file using relative path
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Added
 - Ignore type imports for [`named`] rule ([#931], thanks [@mattijsbliek])
 - Add documentation for [`no-useless-path-segments`] rule ([#1068], thanks [@manovotny])
+- Add [`no-relative-parent-imports`] rule: disallow relative imports from parent directories.
 
 
 # [2.11.0] - 2018-04-09
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 * Forbid a module from importing itself ([`no-self-import`])
 * Forbid a module from importing a module with a dependency path back to itself ([`no-cycle`])
 * Prevent unnecessary path segments in import and require statements ([`no-useless-path-segments`])
+* Forbid importing modules from parent directories ([`no-relative-parent-imports`])
 
 [`no-unresolved`]: ./docs/rules/no-unresolved.md
 [`named`]: ./docs/rules/named.md
@@ -39,6 +40,7 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 [`no-self-import`]: ./docs/rules/no-self-import.md
 [`no-cycle`]: ./docs/rules/no-cycle.md
 [`no-useless-path-segments`]: ./docs/rules/no-useless-path-segments.md
+[`no-relative-parent-imports`]: ./docs/rules/no-relative-parent-imports.md
 
 ### Helpful warnings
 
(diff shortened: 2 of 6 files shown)


chrislloyd:
As discussed IRL with @ljharb, this is a more extreme version of #966. This helps engineers write code that fits well with DAG based build tools like Bazel or Buck. The inspiration for this change comes from [this essay by the Google Fuschia team](https://fuchsia.googlesource.com/docs/+/master/the-book/dotdot.md) and the [OCAML module system](https://ocaml.org/learn/tutorials/modules.html).

chrislloyd:
My general approach for that would be either a.) create a projects folder relative to the file (`'./projects'`) or b.) create a `projects` package which is somehow available to the file. This rule tends to pair particularly well with yarn workspaces so that common code can be quickly and easily extracted into packages (which indicates explicit share-intent by the author).

Also, holy moly you're the fastest PR reviewer ever.

chrislloyd:
Actually, re-reading your question - I think I misunderstood. An example of what you're asking for could be:

```
project-foo/
    tests/
        a.test.js
    a.js
main.js
```

And you'd like to allow `project-foo/tests/a.test.js` to import `../a.js` but disallow `project-foo/a.js` to import `../main.js`?

ljharb:
@chrislloyd yes, precisely. `project-foo` should be a jail, but everything inside it can import anything inside it.

chrislloyd:
🤔 yep you're right. I think this rule could definitely take options to override some of the behavior but a.) I'd like to understand the use-cases a bit better. The way I intend to use this, exceptions won't be a problem. b.) They can be introduced in further PR's as they'd be non-breaking changes to the rule.

ljharb:
I totally agree that such options can be added later, but I'd like to get an idea of what the schema might look like before merging the rule, to ensure future compatibility :-)

chrislloyd:
Ah great, yeah. I think in those cases I'd like to see a layout like:

```
helpers/
  blah.js
foo.js
foo.test.js
```

That effectively shifts files further up and creates lots of big directories with lots of files in it. The rule is intended to be opinionated. I think what you're asking for could be achieved perhaps with more flexible configuration to `no-restrictive-paths`?

```
"import/no-restricted-paths": [2, { "jails": ["project-*/", "packages/*/"] }]
```

ljharb:
Yes, that’s possible.

Maybe since the rule is intended to be opinionated, we should reply flesh out the docs so the intention and reasoning is clear?

chrislloyd:
Awesome, thanks for the reviews @ljharb & @benmosher. Addressed your feedback in followup commits.

chrislloyd:
The CI failure above seems to be because of a flakey test environment (passing on all other environments, randomly timing out) - could ya'll restart the job? I don't think I have permissions in Travis without being a contributor to the repo.

chrislloyd:
@benmosher review ping!

ljharb:
Released in v2.13.0

Alex-Sokolov:
Thanks for adding rule!

But found a bug:
`import blockIcons from '../../components/test/test.js'` — FAIL
but
`import blockIcons from './../../components/test/test.js'` — OK

ljharb:
@Alex-Sokolov please file a new issue if you’ve found a bug

PR #1095, opened 2018-05-03
exclude tests from coverage + bump TS parser

benmosher:
TIL the test files were included in the coverage stats. 🤦‍♂️


benmosher:
nah, those were old ones. the `test` script runs `nyc` every time. I think I just never cleaned up the pre-nyc istanbul scripts.

also: not sure what I did to screw up ESLint 2 and 3, now...

PR #1104, opened 2018-05-17
fix: allow packageDir and readPkgUp logic to play nice

### LINKED ISSUE #1103: Bug: no-extraneous-packages packageDir should not override readPkgUp logic
Currently, [this logic is flawed](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-extraneous-dependencies.js#L34) so that if you set `packageDir` option to either string or array - it will override the `readPkgUp` logic. 

I think this logic should remain, here is my use case:

1. Mono repo with X packages
2. I am prohibiting use of `devDependencies` for all nested packages.
3. Instead, all `devDependencies` are defined in the workspace/repo root.
4. All circular dependencies on sibling mono repo packages are defined in `peerDependencies`.
5. This means, since it's only a peer dep - using `packageDir` cause this rule to miss the defined peer dep because it will *only* search what is defined in `packageDir`.

hulkish:
Fixes #1103, #1109

### DIFF
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -13,17 +13,12 @@ function hasKeys(obj = {}) {
   return Object.keys(obj).length > 0
 }
 
-function extractDepFields({
-  dependencies,
-  devDependencies,
-  optionalDependencies,
-  peerDependencies,
-}) {
+function extractDepFields(pkg) {
   return {
-    dependencies: dependencies || {},
-    devDependencies: devDependencies || {},
-    optionalDependencies: optionalDependencies || {},
-    peerDependencies: peerDependencies || {},
+    dependencies: pkg.dependencies || {},
+    devDependencies: pkg.devDependencies || {},
+    optionalDependencies: pkg.optionalDependencies || {},
+    peerDependencies: pkg.peerDependencies || {},
   }
 }
 


hulkish:
@benmosher @ljharb This is a fix to what was missed in https://github.com/benmosher/eslint-plugin-import/pull/1085

hulkish:
@benfletcher @ljharb So, I think [this test](https://github.com/hulkish/eslint-plugin-import/blob/master/tests/src/rules/no-extraneous-dependencies.js#L257) is actually incorrect. 

With my fix, this test fails (succeeds where expected to fail).

hulkish:
I've updated this pr to reflect my last comment.

tests/src/rules/no-extraneous-dependencies.js
```diff
@@ -243,24 +248,23 @@ ruleTester.run('no-extraneous-dependencies', rule, {
       options: [{packageDir: packageDirMonoRepoWithNested}],
       errors: [{
         ruleId: 'no-extraneous-dependencies',
-        message: "'left-pad' should be listed in the project's dependencies. Run 'npm i -S left-pad' to add it",
+        message: '\'left-pad\' should be listed in the project\'s dependencies. Run \'npm i -S left-pad\' to add it',
```
ljharb: are all these quotes changes necessary?

tests/src/rules/no-extraneous-dependencies.js
```diff
@@ -243,24 +248,23 @@ ruleTester.run('no-extraneous-dependencies', rule, {
...
       code: 'import react from "react";',
       filename: path.join(packageDirMonoRepoRoot, 'foo.js'),
       errors: [{
         ruleId: 'no-extraneous-dependencies',
-        message: "'react' should be listed in the project's dependencies. Run 'npm i -S react' to add it",
+        message: '\'react\' should be listed in the project\'s dependencies. Run \'npm i -S react\' to add it',
       }],
     }),
     test({
-      code: 'import react from "react";',
```
ljharb: if this test case is supposed to pass, let's move it to "valid"

hulkish: I solved this by:
- keeping this test. since it's also good to be checking
- added exact original test to valid, here https://github.com/benmosher/eslint-plugin-import/pull/1104/files#diff-05140578b7a943405919d8a058934d1dR97

src/rules/no-extraneous-dependencies.js
```diff
@@ -21,66 +20,58 @@ function extractDepFields(pkg) {
...
       })
-    }
-    if (e.name === 'JSONError' || e instanceof SyntaxError) {
+    } else if (e.name === 'JSONError' || e instanceof SyntaxError) {
       context.report({
         message: 'The package.json file could not be parsed: ' + e.message,
         loc: { line: 0, column: 0 },
       })
+    } else {
+      throw e
```
hulkish: @ljharb Thought I should make note of this.. We were swallowing "all else" errors before. I thought that was a mistake.

ljharb: Throwing these errors will crash eslint - I'm not sure if that's better than silently logging them.

tests/src/rules/no-extraneous-dependencies.js
hulkish: yea.. shouldn't it prefer backticks?

ljharb: Only if there's interpolations.

src/rules/no-extraneous-dependencies.js
hulkish: Well, what if its a permission error? Shouldn't that crash eslint?

ljharb: it perhaps should warn on the line, but not crash eslint altogether

hulkish: What way should i do that? I'm thinking:
```js
    } else {
      context.report({
        message: e.message,
        loc: { line: 0, column: 0 },
      })
    }
```

Atlhough, I think this is a package.json file error in this instance, and should probably somehow be attached to that instead?

hulkish: Ok, this is updated.

hulkish:
@dmk255 I think this is all - unless there's more req changes.

src/rules/no-extraneous-dependencies.js
```diff
@@ -1,86 +1,89 @@
...
 
-function extractDepFields(pkg) {
+function extractDepFields({
+  dependencies = {},
+  devDependencies = {},
+  optionalDependencies = {},
+  peerDependencies = {},
+}) {
   return {
-    dependencies: pkg.dependencies || {},
```
ljharb: this isn't quite the same thing; someone might have `"dependencies": ""` or `"dependencies": null`

src/rules/no-extraneous-dependencies.js
```diff
@@ -1,86 +1,89 @@
...
-        message: 'The package.json file could not be found.',
+        message: `Could not find: ${relFiles.join(', ')}`,
         loc: { line: 0, column: 0 },
       })
-    }
-    if (e.name === 'JSONError' || e instanceof SyntaxError) {
+    } else if (e.name === 'JSONError' || e instanceof SyntaxError) {
       context.report({
-        message: 'The package.json file could not be parsed: ' + e.message,
+        message: `Could not parse ${relFiles.pop()}: ${e.message}`,
```
ljharb: it's a bit shady to pop out of this array as part of the message; can it just be `relFiles[relFiles.length - 1]` instead?

tests/src/rules/no-extraneous-dependencies.js
ljharb: This still seems like it should be reverted.

hulkish: https://github.com/hulkish/eslint-plugin-import/blob/2e41085228ddd026ac7735eff36247b7f4d53d61/tests/src/rules/no-extraneous-dependencies.js as u can see, this change is consistent with the rest of the file. It actually fails lint for me with the existing (in master).

hulkish: actually, i can see now  this was corrected in master, so ill revert.

hulkish:
@benfletcher ok I think above concerns are addressed, now.

hulkish:
@benmosher ok i think all concerns mentioned above are resolved, now. No more changes planned for me unless requested.

Tymek:
Might fix https://github.com/benmosher/eslint-plugin-import/issues/1109

dvsekhvalnov:
Hey guys, any update on this one? Will it be merged/released soonish? :)

Thank you.

dmk255:
Just chiming in given that there's been a bit of time since the last comment - any chance this may be merged soonish?

Thank you!

ljharb:
After thinking about this and discussing offline with @hulkish, it kind of seems to me like the current packageDir option is actually not useful.

Perhaps instead of "fixing" it with this new merging behavior, we should add a per-category packageDir option that has the new behavior? Something like this:
```js
'import/no-extraneous-dependencies': [2, {
  dependencies: false, // or true
  optionalDependencies: ['pattern/*/1', 'pattern/**/2'],
  peerDependencies: {
    patterns: ['pattern/*/1', 'pattern/**/2'],
    additionalPackageDirs: ['../', 'somewhere else'],
  },
}]
```

Thoughts?

hulkish:
@ljharb In your example snippet, I'm assuming `patterns` represents package names?

ljharb:
@hulkish it would be the same value that's currently already supported in the schema i indicated under "optionalDependencies".

dmk255:
I'm a little confused as to the use of `optionalDependencies` and `peerDependencies.patterns`.

Also, as a user of a monorepo, my root package.json doesn't contain "peerDependencies" in the traditional sense but rather devDependencies that are shared across my packages.

I would find having the following valuable:
* Each leaf package has the ability to reference the root package.json to get access to the devDependencies. It would be ideal if this could be done without needing to make a custom `.eslintrc.js` in each leaf package but that's not a big deal.
* Each leaf package can resolve my symlinked packages that are included in my root node_modules. Right now I use https://github.com/Dreamscapes/eslint-import-resolver-lerna to help me do this. Otherwise eslint-plugin-import can't resolve my sibling packages and says they are not found.

Presently this PR fixes the first point. I've been able to use it by forking @hulkish's repo and importing that in my package.json. I'm entirely not sure what benefits @ljharb's proposal will grant though I definitely could be missing something.

ljharb:
@dmk255 with my suggestion, you'd configure eslint to allow devDeps to be declared in the root, but regular deps to only be declared in the subproject's package.json.

dmk255:
@ljharb That sounds great - would you be able to give an example of the syntax you're considering to be able to do that? I don't immediately see how it fits into the example above.

ljharb:
@dmk255 
```js
'import/no-extraneous-dependencies': [2, {
  dependencies: true,
  devDependencies: {
    additionalPackageDirs: ['.'],
  },
}]
```
This would require that deps be in the closest package.json, but devDeps could be there OR in the current dir's package.json.

dvsekhvalnov:
Hey, let me ask as well. Here is our use case:
 - we have monorepo, where root package.json defines 'baseline' dependencies (not devDeps) for all subprojects (let's say React15)

- each subproject can override baseline as it moving forward, (e.g. upgrade to React16). 
- normally 90% of subprojects are fine with baseline and simply do not define anything new in dependencies

Will proposed solution be able to handle that case? (e.g. eslint to lookup dependencies in root package.json when linting subproject with empty deps).

Thank you.

ljharb:
@dvsekhvalnov yes, since it's "additional" - it merges - both this PR and my suggestion would address your use case.

hulkish:
@dvsekhvalnov I am going to refactor this pr a bit to meet the functionality described in 
https://github.com/benmosher/eslint-plugin-import/pull/1104#issuecomment-406074658

From there, I think the plan is to merge that functionality (assuming all is well). Since, currently the packageDir option doesn't quite fit the need.

benmosher:
@ljharb @hulkish I am not a user of this rule. it's not clear to me based on discussion if this PR is in a useful releasable state. if it is useful in its current state and solves your problems, I'm fine to merge it.

dvsekhvalnov:
@benmosher and all, i have use case for this change, can try to test if you can explain how i can get workable copy of non-released version.

tizmagik:
@dvsekhvalnov I think you should be able to check out @hulkish branch and then link it in your project. Something like:

```
git clone https://github.com/hulkish/eslint-plugin-import
cd eslint-plugin-import
npm install
npm run prepublish
npm link
cd ~/my-project
npm link eslint-plugin-import
```

Then try running the linter. That should hopefully have it so that your project's eslint-plugin-import is actually linked to @hulkish 's version

tylerlee:
I cloned @hulkish 's fork and this seems to fix my monorepo woes.

benmosher:
I would merge but it's not clear to me why there are conflicts nor how to resolve them for the issues you are all facing. @hulkish are you able to resolve?

DavideDaniel:
is the PR not in a testable state? I cloned @hulkish's fork and attempted seeing if this would resolve our monorepo lint issues particular to eslint-plugin-import and it did not...

ljharb:
ping @hulkish - any interest in rebasing this PR?

PR #1106, opened 2018-05-18
Fixed flow types imports

### LINKED ISSUE #921: import/named doesn't detect Flow Opaque types in flow comments

### LINKED ISSUE #1115: import/named doesn't ignore Flow identifier syntax type import
"Identifier syntax":
```javascript
// @flow
import { type Dispatch } from 'redux';
```
Gives following error:
```
2:15  error    Dispatch not found in 'redux'        import/named
```
Whilst the "declaration syntax":
```javascript
// @flow
import type { Dispatch } from 'redux';
```
Generates no error.

This is especially annoying for users that use the [`flowtype/type-import-style`](https://github.com/gajus/eslint-plugin-flowtype#eslint-plugin-flowtype-rules-type-import-style) rule with the default setting ("identifier") and run ESLint with `--fix`, since that rule will automatically change the syntax, which then triggers `import/named`.

The point of the identifier syntax is to combine type and value imports in the same statement, e.g.

```javascript
// @flow
import { configureStore, type Dispatch } from 'redux';
```

The [docs for this rule](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/named.md) state:
> Note that type imports, as used by Flow, are always ignored.

So I think this is a bug. Can we ignore the identifier syntax for this rule also?

syymza:
#1057 has fixed #921 for the basic case, in which an import is the only thing imported from a specific file.
Sometimes types and functions/classes need to be imported from the same file. This PR fixes also that specific instance:
```
import  { type MyOpaqueType, MyClass } from "./flowtypes"',
```

I have added some tests for all these cases

Fixes #1115.

### DIFF
--- a/src/rules/named.js
+++ b/src/rules/named.js
@@ -31,7 +31,7 @@ module.exports = {
         if (im.type !== type) return
 
         // ignore type imports
-        if(im.importKind === 'type') return
+        if (im.importKind === 'type') return
 
         const deepLookup = imports.hasDeep(im[key].name)
 
--- a/src/rules/named.js
+++ b/src/rules/named.js
@@ -30,6 +30,9 @@ module.exports = {
       node.specifiers.forEach(function (im) {
         if (im.type !== type) return
 
+        // ignore type imports
+        if(im.importKind === 'type') return
+
         const deepLookup = imports.hasDeep(im[key].name)
 
         if (!deepLookup.found) {
--- a/tests/files/flowtypes.js
+++ b/tests/files/flowtypes.js
@@ -10,3 +10,6 @@ export type MyType = {
 export interface MyInterface {}
 
 export class MyClass {}
+
+export opaque type MyOpaqueType: string = string;
+
--- a/tests/src/rules/named.js
+++ b/tests/src/rules/named.js
@@ -270,6 +270,12 @@ ruleTester.run('named', rule, {
       }],
     }),
 
+    test({
+      code: 'import  { type MyOpaqueType, MyMissingClass } from "./flowtypes"',
+      parser: 'babel-eslint',
+      errors: ["MyMissingClass not found in './flowtypes'"],
+    }),
+
     // jsnext
     test({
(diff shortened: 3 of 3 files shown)


src/rules/named.js
```diff
@@ -30,6 +30,9 @@ module.exports = {
       node.specifiers.forEach(function (im) {
         if (im.type !== type) return
 
+        // ignore type imports
+        if(im.importKind === 'type') return
```
ljharb: space after `if`?

syymza: added

syymza:
@benmosher Added the requested test.
`node.importKind === 'type'` checks the imports having the `type` word outside of the `{}` specs. For this reason it cannot be removed, otherwise the other tests I have added in this PR will fail

syymza:
🙄

fizwidget:
Hey @benmosher, could you take another look at this one when you get a chance? Would love to see it merged 🙂

bstuff:
Does this work when there are comments in imported module like:
foo.js:
```
/* flow-include
  export type Foo = string | number;
*/

```

bar.js:
```
import { type Foo } from './foo';
```
?

ljharb:
@bstuff not sure; can you file a new issue if it doesn’t work?

syymza:
@bstuff it should :)

PR #1108, opened 2018-05-28
no-duplicates: add a query-string option to handle false positives wh…

### LINKED ISSUE #1107: no-duplicates rule with query-string params
I'm using the [svg-transform-loader](https://github.com/kisenka/svg-mixer/tree/master/packages/svg-transform-loader) which transforms SVG files on the fly but then I end up importing a file several times but with different options a query string parameter:
```js
import rocketIcon from 'images/rocket.svg'
import blackRocketIcon from 'images/rocket.svg?fill=#000'
```
eslint is not happy with that and raises a `import/no-duplicates` error.

Any idea how I should setup my config so that this plugin actually understands that even if they are based on the same file, it's a valid different import?

pcorpet:
…en using some webpack loaders. Fixes #1107.

### DIFF
--- a/docs/rules/no-duplicates.md
+++ b/docs/rules/no-duplicates.md
@@ -36,6 +36,31 @@ The motivation is that this is likely a result of two developers importing diffe
 names from the same module at different times (and potentially largely different
 locations in the file.) This rule brings both (or n-many) to attention.
 
+### Query Strings
+
+By default, this rule ignores query strings (i.e. paths followed by a question mark), and thus imports from `./mod?a` and `./mod?b` will be considered as duplicates. However you can use the option `considerQueryString` to handle them as different (primarily because browsers will resolve those imports differently).
+
+Config:
+
+```json
+"import/no-duplicates": ["error", {"considerQueryString": true}]
+```
+
+And then the following code becomes valid:
+```js
+import minifiedMod from './mod?minify'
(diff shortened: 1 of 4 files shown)


docs/rules/no-duplicates.md
```diff
@@ -35,6 +35,31 @@ The motivation is that this is likely a result of two developers importing diffe
 names from the same module at different times (and potentially largely different
 locations in the file.) This rule brings both (or n-many) to attention.
 
+### Query Strings
+
+By default, this rule ignores query strings (i.e. paths followed by a question mark), and thus imports from `./mod?a` and `./mod?b` will be considered as duplicates. However you can use the option `query-string` to handle them as different (usually because a webpack loader would actually resolve those imports differently).
```
ljharb: primarily because browsers will resolve them differently; a webpack loader is secondary.

pcorpet:
Schema added to the rule.

I had the impression that several options could be offered, but let's keep it simple for now as I don't have any immediate usecase.

docs/rules/no-duplicates.md
pcorpet: Done.

pcorpet:
@benmosher @jfmengels can I get a review please?

EECOLOR:
@ljharb what do we need to do to get this pull request merged?

PR #1113, opened 2018-06-02
Resolve with dirname

### LINKED ISSUE #828: Why isn't "eslint-import-resolver-node" listed in `eslint-module-utils`'s peerDependencies?
This plugin does not work with `pnpm` because it relies on NPM3's flat `node_modules` structure. Please see: https://github.com/pnpm/pnpm/issues/739, where there is a reproducible repo.

ishitatsuyuki:
Close #828

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -262,7 +262,7 @@ function captureTomDoc(comments) {
 }
 
 ExportMap.get = function (source, context) {
-  const path = resolve(source, context)
+  const path = resolve(source, context, __dirname)
   if (path == null) return null
 
   return ExportMap.for(childContext(path, context))
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -82,5 +82,5 @@ const typeTest = cond([
 ])
 
 export default function resolveImportType(name, context) {
-  return typeTest(name, context.settings, resolve(name, context))
+  return typeTest(name, context.settings, resolve(name, context, __dirname))
 }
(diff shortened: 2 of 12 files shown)


src/ExportMap.js
```diff
@@ -262,7 +262,7 @@ function captureTomDoc(comments) {
 }
 
 ExportMap.get = function (source, context) {
-  const path = resolve(source, context)
+  const path = resolve(source, context, __dirname)
```
ljharb: this would be the dirname of this file - is that intended? I'd expect `process.cwd()`

ishitatsuyuki:
@ljharb All typos fixed. Tests are passing now.

src/ExportMap.js
ishitatsuyuki: This is where we load the "resolver module" from, not the actual module that we're resolving. I've updated the doc comments.

zkochan:
Is there something else that needs to be done in order to get this merged?

ljharb:
@zkochan other maintainers would need to review it.

zkochan:
can we ping them? It would be really great to have this issue solved

@benmosher could you review this PR, please?

donifer:
Friendly bump 🙏

benmosher:
I don't think I like this solution but mostly I think I don't understand why capturing `__dirname` at the calling location and passing it to `resolve` is an ideal solution to #828. I also don't really understand how it even works at all.

ljharb:
@ishitatsuyuki if you're still interested in landing this PR, could you rebase it, address the comments, and mark this as ready for review?

ishitatsuyuki:
Closing as I'm no longer interested in this.

PR #1117, opened 2018-06-07
Fix named flow type imports in specifiers.

### LINKED ISSUE #1115: import/named doesn't ignore Flow identifier syntax type import

rostislav-simonik:
Fixes #1115

### DIFF
--- a/src/rules/named.js
+++ b/src/rules/named.js
@@ -28,7 +28,7 @@ module.exports = {
       }
 
       node.specifiers.forEach(function (im) {
-        if (im.type !== type) return
+        if (im.type !== type || im.importKind === 'type') return
 
         const deepLookup = imports.hasDeep(im[key].name)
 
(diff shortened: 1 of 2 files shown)


syymza:
Duplicates #1106

ljharb:
I’m going to close this in favor of #1106. Thanks for the contribution!

PR #1119, opened 2018-06-08
Use resolved package name

mieszko4:
Should fix https://github.com/benmosher/eslint-plugin-import/issues/496

### DIFF
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -111,7 +111,9 @@ function reportIfMissing(context, deps, depsOptions, node, name) {
   const resolved = resolve(name, context)
   if (!resolved) { return }
 
-  const [, resolvedName = name] = resolved.split(path.join(process.cwd(), 'node_modules/'))
+  // Take last part after node_modules
+  const [resolvedName] = resolved.split('node_modules/').slice(-1)
+
   const splitName = resolvedName.split('/')
   const packageName = splitName[0][0] === '@'
     ? splitName.slice(0, 2).join('/')
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -111,7 +111,8 @@ function reportIfMissing(context, deps, depsOptions, node, name) {
   const resolved = resolve(name, context)
   if (!resolved) { return }
 
-  const splitName = name.split('/')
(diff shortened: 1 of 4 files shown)


src/rules/no-extraneous-dependencies.js
```diff
@@ -111,7 +111,8 @@ function reportIfMissing(context, deps, depsOptions, node, name) {
   const resolved = resolve(name, context)
   if (!resolved) { return }
 
-  const splitName = name.split('/')
+  const [, resolvedName = name] = resolved.split(process.cwd() + "/" + "node_modules/");
```
ljharb: this should use `path.join`, not `+`, for combining paths

mieszko4:
@ljharb I've added the test that would have failed without fix in `no-extraneous-dependencies.js`. Please take a look.

ljharb:
@arcanis i'm intending to merge this soon, does it need any changes to avoid causing berry problems?

arcanis:
From the look of it it should be compatible in most cases, since we store packages inside a standalone `node_modules` path (to make vendor detection easier):

```
/app/.yarn/cache/foo-1.0.0.zip/node_modules/foo/index.js
```

One caveat however: we cannot do that when people use the `workspace:`, `link:`, or `portal:` protocol, because they are resolved to the final path. So given the `package.json`:

```json
{
  "dependencies": {
    "foo": "link:./foo"
  }
}
```

You'd have `/app/foo`, not `/app/node_modules/foo`.

ljharb:
@mieszko4 this needs to be rebased. i _think_ that the part you'll need to update now is either `checkDependencyDeclaration` or `getModuleOriginalName`.

mieszko4:
It's been 6+ years now after I created this PR.
I haven't had a need for this to be merged. And I am not sure if this PR will resolve the mentioned issue.

I think that we should just close this PR. In the future would be great if PRs could be closed or merged quicker :)

ljharb:
@mieszko4 that's always great, but we're all unpaid volunteers here, and things take the time they take :-)

I'll close it then, thanks.

mieszko4:
> @mieszko4 that's always great, but we're all unpaid volunteers here, and things take the time they take :-)

Yeah, I totally get this. I also would check it in more detail if I had time :). Anyway, no hurt feelings, no worries.

PR #1122, opened 2018-06-23
Add ESLint 5 support

ai:
I start to use ESLint 5 on my project with your plugin and didn’t find any problem with compatibility.


jaydenseric:
Need to get rid of the deprecated `experimentalObjectRestSpread`:

screen shot 2018-06-24 at 1 27 18 pm

- https://github.com/benmosher/eslint-plugin-import/blob/v2.12.0/config/recommended.js#L27
- https://github.com/benmosher/eslint-plugin-import/blob/v2.12.0/tests/src/rules/namespace.js#L100

ljharb:
We don't have to get rid of that just yet; deprecation warnings can be ignored. In the test, we can set the ecmaVersion to 2018.

ai:
I updated `ecmaVersion` and removed `experimentalObjectRestSpread` in these two places

ljharb:
Released in v2.13.0

PR #1128, opened 2018-06-29
make rule `no-useless-path-segments` work with commonjs

### LINKED ISSUE #1127: `no-useless-path-segments` does not work with require statements
from rule's [`doc`](https://github.com/benmosher/eslint-plugin-import/blob/37bd5423276abbe6ee2245a3adc5d178fe16121b/docs/rules/no-useless-path-segments.md), it is described that it works with both import and require statements, but it actually work only with import statements

Given the following folder structure:
```
my-project
├── app.js
└── pages
    ├── about.js
    ├── contact.js
    └── index.js
```
tested code
```js
/**
 *  in my-project/app.js
 */
require('./../pages/about.js');
```
**Expectation**: example pattern should be considered problem.
**Actual**: it's not.

You may also resolve this issue by amend the doc, but I hope you don't do it this way. 😛

1pete:
fixes #1127

### DIFF
--- a/src/rules/no-useless-path-segments.js
+++ b/src/rules/no-useless-path-segments.js
@@ -39,6 +39,16 @@ module.exports = {
       url: docsUrl('no-useless-path-segments'),
     },
 
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          commonjs: { type: 'boolean' },
+        },
+        additionalProperties: false,
+      },
+    ],
+
     fixable: 'code',
   },
 
--- a/tests/src/rules/no-useless-path-segments.js
+++ b/tests/src/rules/no-useless-path-segments.js
@@ -7,6 +7,10 @@ const rule = require('rules/no-useless-path-segments')
 function runResolverTests(resolver) {
   ruleTester.run(`no-useless-path-segments (${resolver})`, rule, {
     valid: [
+      // commonjs with default options
+      test({ code: 'require("./../files/malformed.js")' }),
+
+      // esmodule
       test({ code: 'import "./malformed.js"' }),
       test({ code: 'import "./test-module"' }),
       test({ code: 'import "./bar/"' }),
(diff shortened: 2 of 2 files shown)


ljharb:
cc @graingert

graingert:
@ljharb we could make it a config option if we're being really strict with breaking changes

ljharb:
Yeah that’s a good point - worst case we can put it behind an option.

1pete:
Please review again, now it's an option, so it can safely be a patch.
And maybe you can turn `commonjs` on by default on v3.

src/rules/no-useless-path-segments.js
```diff
@@ -39,6 +39,8 @@ module.exports = {
       url: docsUrl('no-useless-path-segments'),
     },
 
+    schema: [ makeOptionsSchema() ],
```
ljharb: this isn’t defining “commonjs” as an option tho

1pete: doesn't line 111 from this function define it?
https://github.com/benmosher/eslint-plugin-import/blob/37554fe9844986a1ab126d2a8f59fe4c7551a2f8/utils/moduleVisitor.js#L107-L122

ljharb: oh, hmm. I don’t think it’s a good idea to add a bunch of options that aren’t being used. Can we make a schema that only has the commonjs option?

1pete: from what i observe, every option is valid, function `makeOptionsSchema` is made specifically for rule that uses util `moduleVisitor`
checkout
https://github.com/benmosher/eslint-plugin-import/blob/37554fe9844986a1ab126d2a8f59fe4c7551a2f8/utils/moduleVisitor.js#L13-L101

but it's your call, i can add only commonjs

ljharb: it’s fine to add support for every option, but then let’s make sure to add tests for all the options.

1pete:
@ljharb now there is one simple option :)

graingert:
We probably want to check amd too

ljharb:
Sure, we just need a test for it.

graingert:
@1pete ^

ljharb:
I'm going to merge this now; if we want to add AMD let's do that in a followup PR.

PR #1135, opened 2018-07-11
[no-relative-parent-imports] Resolve paths

### LINKED ISSUE #1123: no-relative-parent-imports didn't check if path start with './'
When add rule to config
`import blockIcons from '../../components/test/test.js'` — FAIL
but
`import blockIcons from './../../components/test/test.js'` — OK

ref: https://github.com/benmosher/eslint-plugin-import/pull/1093

chrislloyd:
This changes the rule to resolve paths before emitting an error. While this means the error will trigger less often (before we could report an error even if the file didn't exist on disk yet) I think it's a fine tradeoff so that it can be useful in more situations.

Fixes #1123.


chrislloyd:
I only have intermittent time to check this out - just letting any 👀  know that I haven't forgotten about this fix - just busy.

PR #1137, opened 2018-07-12
Fix import/order autofixer when using typescript-eslint-parser

justinanastos:
Addresses https://github.com/benmosher/eslint-plugin-import/issues/1086#issuecomment-383093638

# Why?

When using `typescript-eslint-parser`, AST nodes have a `range` trie but not `start` and `end`. Apparently `eslint` prefers using `range` instead of `.start` and `.end` as well, see https://github.com/eslint/eslint/issues/8956.

# How

Replace use of `.start` and `.end` with `range[0]` and `range[1]` respectively in the `order` rule.

# Test

I added a test to reproduce the error and then a separate commit to prove this fixes it. All other tests pass.


justinanastos:
Closing to refire CI again

PR #1139, opened 2018-07-17
[Fix] `newline-after-import`: recognize decorators

atos1990:
Hi, this is my fix issue https://github.com/benmosher/eslint-plugin-import/issues/1004

### DIFF
--- a/tests/src/rules/newline-after-import.js
+++ b/tests/src/rules/newline-after-import.js
@@ -165,6 +165,16 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
       parserOptions: { ecmaVersion: 2015, sourceType: 'module' },
       parser: require.resolve('babel-eslint'),
     },
+    {
+      code : `// issue 1004\nimport foo from 'foo';\n\n@SomeDecorator(foo)\nexport default class Test {}`,
+      parserOptions: { sourceType: 'module' },
+      parser: require.resolve('babel-eslint'),
+    },
+    {
+      code : `// issue 1004\nconst foo = require('foo');\n\n@SomeDecorator(foo)\nexport default class Test {}`,
+      parserOptions: { sourceType: 'module' },
+      parser: require.resolve('babel-eslint'),
+    },
   ],
 
   invalid: [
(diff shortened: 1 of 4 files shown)


atos1990:
@ljharb, done

tests/src/rules/newline-after-import.js
```diff
@@ -164,6 +164,24 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
       parserOptions: { sourceType: 'module' },
       parser: 'babel-eslint',
     },
+    {
+      code : `// issue 1004
+        import foo from 'foo';\n
+        @SomeDecorator(foo)
```
ljharb: hm, if a newline is required after `import`, this should be invalid - there needs to be a blank line between the `import` line and the decorator.

tests/src/rules/newline-after-import.js
```diff
@@ -164,6 +164,24 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
...
+        export default class Test {}
+      `,
+      parserOptions: { sourceType: 'module' },
+      parser: 'babel-eslint',
+    },
+    {
+      code : `// issue 1004
+        const foo = require('foo');\n
+        @SomeDecorator(foo)
```
ljharb: same here

tests/src/rules/newline-after-import.js
```diff
@@ -349,5 +367,43 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
...
     },
+    {
+      code: `// issue 10042
+        import foo from 'foo';
+        @SomeDecorator(foo)
+        export default class Test {}
+      `,
+      output: `// issue 10042
+        import foo from 'foo';\n
+        @SomeDecorator(foo)
```
ljharb: this should autofix to add a blank line above the decorator

tests/src/rules/newline-after-import.js
```diff
@@ -349,5 +367,43 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
...
+    },
+    {
+      code: `// issue 1004
+        const foo = require('foo');
+        @SomeDecorator(foo)
+        export default class Test {}
+      `,
+      output: `// issue 1004
+        const foo = require('foo');\n
+        @SomeDecorator(foo)
```
ljharb: this should autofix to add a blank line above the decorator

tests/src/rules/newline-after-import.js
atos1990: But at the end of line 169 - '\n', i.e. blank line. Is it not enough?
I think that in this case it is correct, isn't it?

tests/src/rules/newline-after-import.js
```diff
@@ -349,5 +359,27 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
       parserOptions: { sourceType: 'module' },
       parser: 'babel-eslint',
     },
+    {
+      code: `// issue 10042\nimport foo from 'foo';\n@SomeDecorator(foo)\nexport default class Test {}`,
```
ljharb: can these be reverted to use actual newlines instead of `\n`? it's much more readable.

ljharb:
ping @atos1990, this also needs a rebase

TrejGun:
@atos1990, please rebase

shastaxc:
Came here looking for this fix. ETA on merge? With this rule not working with decorators, it pretty much means it won't work with any of my ts files in Angular, which is the only thing I'm using ESLint for.

TrejGun:
thanks @ljharb

elbertcastaneda:
We continue with this error in the latest version, I am thinking that the next version could contain this fix, is it right?

ljharb:
Yes.

ljharb:
@skoblenick as you can see if you click on the [commit that closed the PR](https://github.com/import-js/eslint-plugin-import/commit/b307c7c14bed8b20807cc629911093c5ec47f0ed), it is in both main, and every version v2.21.0 and later.

PR #1141, opened 2018-07-20
[no-relative-parent-imports] support windows

1pete:
make rule `no-relative-parent-imports` support windows

as you see, appveyor ci failed in master branch, i think because windows uses `\` as a directory delimiter
so i change path check to use regular expression `[\\/]`

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -57,7 +57,7 @@ function isInternalModule(name, settings, path) {
 }
 
 function isRelativeToParent(name) {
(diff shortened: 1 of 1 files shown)


benmosher:
@ljharb that is a great point but this PR implies we don't live in that universe--yet. 😄 fwiw I don't recall anything in any resolvers normalizing slashes. but that would be great to avoid stuff like this. 

where "stuff like this" := intermittent AppVeyor gotchas.

thanks @1pete!

Issue #186, opened 2016-02-22
rule: no-unused


benmosher:
On first execution, build an index of imports from some source root(s) pattern (i.e. `/src/.*\.js`) and then flag all exports that are not used.

Subsequent executions using the same base would benefit from a cache.

Could try using a makefile to build the import cache in a file to maintain it across executions + build it fast? Might be a useful strategy for caching exports between executions, too.

FS cache should live in the `node_modules/eslint-plugin-import` folder? or just the system `tmp` dir?

Probably should configure base as a rule parameter? until it benefits some other rule, anyway.

Could also mark full modules unused if nothing imports a given file.

josh:
Definitely ❤️ this idea.

I've implemented something like this for @github but its pretty app specific. I was researching a way it could be extracted into a proper eslint plugin. It'd be rad if it was was part of `eslint-plugin-import`.

My original thought was that *this* plugin wouldn't have to be responsible for building the index per se. If you spec'd out the format of the index file, an application specific Makefile could be responsible for generating it. Currently we have a mix of ES6 imports and older CJS stuff. You'll also want to special case "entry points" somehow since those won't directly be imported, but maybe referenced by some html script tag in the app. Our could see our indexer looking for ES imports, CJS and these custom html templating references.

It'd still be great for a standard index builder to be part of `eslint-plugin-import` itself to all all the easy cases. But I'd interested in having a custom hook into that point.

benmosher:
> My original thought was that this plugin wouldn't have to be responsible for building the index per se

I would be totally up for that if you or someone else can point me to something that would generate an index on the FS I could write the rule against. Less is more. 😎 

> You'll also want to special case "entry points" somehow

I'm imagining this could be a new responsibility of the resolvers, i.e. Node resolver exposes either the `package.json` `main` field(s) or root `index.js`, and the Webpack resolver would read the entry points out of the relevant config file(s). Could be implemented in v3 of resolvers as `getEntryPoints(context|filename) => String[]` or something similar.

Could also allow explicit config of entry points in a setting or rule param, but IMO the less of that, the better.

benmosher:
Idea: the resolvers can provide a list of entry points. Node resolver uses package.json and defaults to index.js. Webpack resolver uses "entry" field. 

Using the memo parser, it shouldn't technically add much time, either.

hjylewis:
Any news or updates on this rule? Having this rule would be super beneficial (yay, deleting code).

Let me know how I can help out!

[...]

PR #1142, opened 2018-07-23
New: `no-unused-modules` rule

rfermann:
As discussed in #186, this rule adds the possibility to find modules without any exports and/or exports within a module, which aren't used in other modules

ToDos:
- [X] implement feature/fix bug
- [X] update docs
- [x] write tests

This is the first rule I created. Writing tests for it may therefor take some additional time.


docs/rules/no-unused-modules.md
```diff
@@ -0,0 +1,80 @@
...
+  - individual exports not being used within other modules
+
+## Rule Details
+
+
+### Options
+
+This rule takes the following option:
+
+- `src`: an array with files/paths to be analyzed. It only for applies for unused exports
```
ljharb: i'm confused about this one - rule schemas don't generally specify their own paths, eslint itself does.

rfermann: For my knowledge, eslint does not share this information with the rules - but I need it for building up a list of existing exports and imports for this rule to work.

If there is a way to get this information from eslint, I would actually be glad to get rid of this option.

ljharb: I'm confused why this wouldn't just use "the whole project" - ie, `process.cwd()` recursively.

rfermann: In my opinion, both solutions aren't optimal...

If we would use `process.cwd()`, the rule would always start its analysis within the root directory, even if eslint is restricted to a single directory within the root directory. In this case, one could of course use the `ignore`-option to restrict the files being linted...but then one could as well just maintain the `src`-option providing the one directory he intends to lint.

rfermann: Good idea. @ljharb: what do you think about this?

docs/rules/no-unused-modules.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+### Options
+
+This rule takes the following option:
+
+- `src`: an array with files/paths to be analyzed. It only for applies for unused exports  
+- `ignore`: an array with files/paths to be ignored. It only for applies for unused exports  
+- `missingExports`: if `true`, files without any exports are reported  
+- `unusedExports`: if `true`, exports without any usage are reported
```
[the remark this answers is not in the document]
rfermann: Thanks for the input. I will update the documentation.

docs/rules/no-unused-modules.md
ljharb: that seems more reasonable.

docs/rules/no-unused-modules.md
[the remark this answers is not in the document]
ljharb: I'm confused, why would you ever want to allow an export that's only used in the same file and never imported?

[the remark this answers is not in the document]
coreyfarrell: I was misunderstanding the purpose of the `ignore` option.  I had expected that `ignore: ['test/**']` would tell this rule to ignore imports performed by tests - thus report anything which is not imported from production code.  A common pattern in my applications is for internal sources to contain `export function defaultOpts() { return {...}; }`, and for that function to be used in exactly two places: the constructor function in the same source and the unit test.

To be honest I don't understand the purpose of the `ignore` option as it's currently implemented.  Why not just use .eslintrc `overrides` to turn this rule off for certain files?  Please correct me if I'm wrong but the `src` option doesn't tell this rule which files to lint, it tells it which files scan for import's, then each file being linted is checked to make sure something from `src` performed the import?  If so I think that `ignore` should counter-act `src`, meaning restrict the list of files scanned for import's - not restrict the list of files to lint.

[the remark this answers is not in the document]
ljharb: The option is for package entry points, which are often untested (because they just re-export something from the build dir).

I expect the spec option to cover files whose *exports* should be checked. Anything that imports it anywhere should count it as used.

ljharb:
Yes, an import * should mark all exports as used, i think.

docs/rules/no-unused-modules.md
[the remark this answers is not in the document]
rfermann: As @ljharb said, the ```ignore``` option is mainly for package entry points. Every file specified in this option will not be linted by this rule. 
However, these files are also analyzed to handle their imports. This means, if an ignored file imports from anywhere else, let's say from ```file-a.js``` the corresponding exports from file  ```file-a.js```  will still be marked as used.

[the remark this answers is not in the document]
coreyfarrell: I misunderstood the purpose of the `ignore` option, IMO it would be clearer if the docs mentioned package entry points.  Renaming it to `ignoreExports` or `entryPoints` could make it's purpose even clearer.

Would you accept a follow-up patch to implement an `ignoreImports` option?  This would allow me to identify when an API is no longer used even though it's still covered by unit tests.  I do test driven development so every export will be imported (directly or indirectly) by unit tests.

Thank you for your responses!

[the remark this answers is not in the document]
ljharb: I agree; both options are needed to be maximally useful.

docs/rules/no-unused-modules.md
rfermann: I added the default src with aa7253baf383f370b3dccefbb11b17cd5db2d8dc. Readme will be updated later on.

tests/src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,234 @@
...
+  missingExports: true,
+}]
+
+const unusedExportsOptions = [{
+  unusedExports: true,
+  src: [testFilePath('./no-unused-modules/**/*.js')],
+  ignore: [testFilePath('./no-unused-modules/*ignored*.js')],
+}]
+
+describe('doPreparation throws correct errors', () => {
```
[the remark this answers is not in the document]
rfermann: I added the additional checks and tests with aa7253baf383f370b3dccefbb11b17cd5db2d8dc.

rfermann:
@ljharb: do you have any ideas how to tackle the failing AppVeyor build?

ljharb:
@rfermann i believe it's on master, and unrelated to this PR.

rfermann:
Okay, good to know.

What else needs to be done from my site to let this PR be merged? Should I adjust the change log? Or is that one of these tasks you will perform?

ljharb:
@rfermann you don't have to worry about the changelog; the appveyor issue should be fixed on master if you rebase.

docs/rules/no-unused-modules.md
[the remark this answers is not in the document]
coreyfarrell: Nothing was ever changed in response to this comment thread.  I'm fine with pushing a follow-up PR to add support for ignoring importers (like tests for example), but I think naming the option `ignore` is confusing.

If you don't want to rename the option maybe some rewording of the docs?  Especially the docs for `ignore`, anyone trying to ignore tests will see the current docs and assume that it allows them to ignore importers (tests), rather than ignoring module entry-points.
```
- `src`: an array with files/paths with exports that must be used.  Defaults to `process.cwd()`, if not provided
- `ignore`: an array with files/paths where unused exports are ignored.
```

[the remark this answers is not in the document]
rfermann: @coreyfarrell: sorry, that wasn't intentionally. I will rename the ```ignore```-option to ```ignoreImports``` and do some rewording in the docs.

rfermann:
@ljharb , @benmosher: what else am I expected to do to get this PR merged? Is it the outstanding rebase which is blocking the merge?

docs/rules/no-unused-modules.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+### Options
+
+This rule takes the following option:
+
+- `src`: an array with files/paths to be analyzed. It only for applies for unused exports. Defaults to `process.cwd()`, if not provided
+- `ignoreExports`: an array with files/paths for which unused exports will not be reported (e.g module entry points) 
+- `missingExports`: if `true`, files without any exports are reported  
+- `unusedExports`: if `true`, exports without any usage within other modules are reported.
```
ljharb: If both missingExports and unusedExports are missing or false, what gets reported?

If nothing, can we ensure it’s a schema error if neither of them are true?

docs/rules/no-unused-modules.md
```diff
@@ -0,0 +1,101 @@
...
+}  // will not be reported
+
+export default 5 // will not be reported
+```
+
+
+
+## When not to use
+
+If you don't mind having unused files or dead code within your codebase, you can disable this rule
```
ljharb: All files should have a trailing newline

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,720 @@
...
+ * modules are consumed within other modules.
+ * @author René Fermann
+ */
+
+import Exports from '../ExportMap'
+import { makeOptionsSchema } from 'eslint-module-utils/moduleVisitor'
+import resolve from 'eslint-module-utils/resolve'
+import docsUrl from '../docsUrl'
+
+// eslint/lib/util/glob-util has been moved to eslint/lib/util/glob-utils with version 5.3
```
ljharb: Can you file an upstream issue? This is a breaking change in a minor version.

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,720 @@
...
+const EXPORT_DEFAULT_DECLARATION = 'ExportDefaultDeclaration'
+const EXPORT_NAMED_DECLARATION = 'ExportNamedDeclaration'
+const EXPORT_ALL_DECLARATION = 'ExportAllDeclaration'
+const IMPORT_DECLARATION = 'ImportDeclaration' 
+const IMPORT_NAMESPACE_SPECIFIER = 'ImportNamespaceSpecifier'
+const IMPORT_DEFAULT_SPECIFIER = 'ImportDefaultSpecifier' 
+const VARIABLE_DECLARATION = 'VariableDeclaration'
+const FUNCTION_DECLARATION = 'FunctionDeclaration'
+const DEFAULT = 'default'
+const UNDEFINED = 'undefined'
```
ljharb: Let’s get rid of this variable - it makes the code harder to read and also defeats engine optimizations which know how to optimize direct comparisons to strings.

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,720 @@
...
+/**
+ * prepare the lists of existing imports and exports - should only be executed once at
+ * the start of a new eslint run
+ */
+const doPreparation = (src, ignoreExports, context) => {
+  const { id } = context
+
+  // do some sanity checks
+  if (typeof src !== UNDEFINED && !Array.isArray(src)) {
+    throw new Error(`Rule ${id}: src option must be an array`)
```
ljharb: ftr this should be ensured by the rule schema

docs/rules/no-unused-modules.md
rfermann: I'm lacking experience with the JSON schema. Do you have an idea or example on how to conditionally require one or the other field?

src/rules/no-unused-modules.js
rfermann: ```listFilesToProcess``` is not part of the official ESLint Node.js API. Hence it is not a breaking change (at least thats how I interpret semantic versioning, please correct me, if I'm wrong in this point).

Additionally, the version of ESLint, in which this change gets introduced, doesn't matter. I still need to make sure ```listFilesToProcess``` gets correctly imported.

src/rules/no-unused-modules.js
rfermann: Added this to schema with 2922910. Also applied id for the second array ("ignoredFiles").

src/rules/no-unused-modules.js
rfermann: Removed with 2922910.

docs/rules/no-unused-modules.md
rfermann: Newline added with 2922910.

docs/rules/no-unused-modules.md
rfermann: Finally figured it out. We now get a schema error, if both are missing/false.

src/rules/no-unused-modules.js
ljharb: It's debatable - to me, if it's reachable, it's part of the public API.

I'd still like to see an upstream issue filed; they should provide the `eslint/lib/util/glob-util` path until v6.

rfermann: What difference does it make for us if they revert the change for now and introduce it again with v6? This rule still needs to handle both cases in order to stay backwards compatible.

Don't get me wrong, I don't refuse to argue for something, which has a benefit for this library. But I struggle to see a benefit in arguing with the eslint team in this particular point.

rfermann: @ljharb: what do you think about my last comment?

ljharb: @rfermann That's a good point - but I think that if they release a bugfix in eslint 5, then we don't actually have to support this rule on an un-patched eslint 5.

docs/rules/no-unused-modules.md
```diff
@@ -0,0 +1,101 @@
+# import/no-unused-modules
+
+Reports:
+  - modules without any or
```
ljharb: modules without any what?

docs/rules/no-unused-modules.md
```diff
@@ -0,0 +1,101 @@
...
+  - individual exports not being used within other modules
```
ljharb:
```suggestion
  - individual exports not being statically `import`ed or `require`ed from other modules in the same project
```

docs/rules/no-unused-modules.md
```diff
@@ -0,0 +1,101 @@
...
+Note: dynamic imports are currently not supported.
+
+## Rule Details
+
+
+### Options
+
+This rule takes the following option:
+
+- `src`: an array with files/paths to be analyzed. It only for applies for unused exports. Defaults to `process.cwd()`, if not provided
```
ljharb:
```suggestion
- `src`: an array with files/paths to be analyzed. It only applies to unused exports. Defaults to `process.cwd()`, if not provided
```

docs/rules/no-unused-modules.md
```diff
@@ -0,0 +1,101 @@
...
+- `ignoreExports`: an array with files/paths for which unused exports will not be reported (e.g module entry points)
```
ljharb:
```suggestion
- `ignoreExports`: an array with files/paths for which unused exports will not be reported (e.g module entry points in a published package)
```

docs/rules/no-unused-modules.md
ljharb:
```suggestion
- `unusedExports`: if `true`, exports without any static usage within other modules are reported.
```

src/ExportMap.js
```diff
@@ -379,6 +379,20 @@ ExportMap.parse = function (path, content, context) {
 
   function captureDependency(declaration) {
     if (declaration.source == null) return null
+    const importedSpecifiers = new Set()
+    if (declaration.specifiers) {
+      declaration.specifiers.forEach(specifier => {
+        if (specifier.type === 'ImportDefaultSpecifier') {
```
ljharb: We could probably shorten this a bit with something like:
```suggestion
        const supportedTypes = new Set(['ImportDefaultSpecifier', 'ImportNamespaceSpecifier', 'ImportSpecifier']);
        if (supportedTypes.has(specifier.type)) {
          importedSpecifiers.add(specifier.type);
        }
```

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,730 @@
...
+ */
+
+import Exports from '../ExportMap'
+import resolve from 'eslint-module-utils/resolve'
+import docsUrl from '../docsUrl'
+
+// eslint/lib/util/glob-util has been moved to eslint/lib/util/glob-utils with version 5.3
+let listFilesToProcess
+try {
+  listFilesToProcess = require('eslint/lib/util/glob-util').listFilesToProcess
```
ljharb: let's put the "utils" require first, since it's the "newer" one and thus will be more common

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,730 @@
...
+const FUNCTION_DECLARATION = 'FunctionDeclaration'
+const DEFAULT = 'default'
+
+let preparationDone = false
+const importList = new Map()
+const exportList = new Map()
+const ignoredFiles = new Set()
+
+const isNodeModule = path => {
+  return path.indexOf('node_modules') > -1
```
ljharb:
```suggestion
  return path.includes('node_modules')
```

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,730 @@
...
+  const srcFiles = new Set()
+  const srcFileList = listFilesToProcess(src)
+
+  // prepare list of ignored files
+  const ignoredFilesList =  listFilesToProcess(ignoreExports)
+  ignoredFilesList.forEach(({ filename }) => ignoredFiles.add(filename))
+
+  // prepare list of source files, don't consider files from node_modules
+  srcFileList.forEach(({ filename }) => {
+    if (isNodeModule(filename)) {
```
ljharb: `srcFileList.filter(({ filename }) => !isNodeModule(filename)).forEach`?

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,730 @@
...
+    if (specifier.type === IMPORT_NAMESPACE_SPECIFIER) {
+      hasNewNamespaceImport = true
+    }
+  })
+  return hasNewNamespaceImport
+}
+
+const newDefaultImportExists = specifiers => {
+  let hasNewDefaultImport = false
+  specifiers.forEach(specifier => {
```
ljharb: if this is an array, `return specifiers.some(({ type }) => type === IMPORT_DEFAULT_SPECIFIER);`?

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,730 @@
...
+  prepareImportsAndExports(srcFiles, context)
+  determineUsage()
+  preparationDone = true
+}
+
+const newNamespaceImportExists = specifiers => {
+  let hasNewNamespaceImport = false
+  specifiers.forEach(specifier => {
+    if (specifier.type === IMPORT_NAMESPACE_SPECIFIER) {
+      hasNewNamespaceImport = true
```
ljharb: same here, can `.some` be used?

rfermann:
Thanks for your latest feedback. I implemented your suggestions. 

> Please ensure all files have trailing newlines.

I don't get that. Did I forgot the trailing newline anywhere? I checked the files I pushed today and all already had a trailing newline :confused:

docs/rules/no-unused-modules.md
```diff
@@ -0,0 +1,101 @@
...
+}  // will not be reported
+
+export default 5 // will not be reported
+```
+
+
+
+## When not to use
+
+If you don't mind having unused files or dead code within your codebase, you can disable this rule
```
ljharb: so passive-aggressive 🤣

tests/files/no-unused-modules/file-b.js
```diff
@@ -0,0 +1 @@
+export const b = 2
```
ljharb: this file, for example, does not have a trailing newline

tests/src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,563 @@
...
+             filename: testFilePath('./no-unused-modules/file-added-4.js.js')}),
+    ],
+    invalid: [],
+  })
+  after(() => {
+    if (fs.existsSync(testFilePath('./no-unused-modules/file-added-4.js.js'))) {
+      fs.unlinkSync(testFilePath('./no-unused-modules/file-added-4.js.js'))
+    }
+  })
+})
```
ljharb: also this one

src/rules/no-unused-modules.js
ljharb: Thanks, filing the issue and linking it here would be great.

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,731 @@
...
+const FUNCTION_DECLARATION = 'FunctionDeclaration'
+const DEFAULT = 'default'
+
+let preparationDone = false
+const importList = new Map()
+const exportList = new Map()
+const ignoredFiles = new Set()
+
+const isNodeModule = path => {
+  return path.includes('node_modules')
```
ljharb: should this ensure it's `(^|\/)node_modules\/|`?

docs/rules/no-unused-modules.md
rfermann: Yeah...well...that's kind of a occupational disease I fear :scream:

What about ```If you have intentionally unused modules within your project, you can disable this rule```?

src/rules/no-unused-modules.js
rfermann: I'm not sure, if I understand that question correctly. What's the background of this question?

docs/rules/no-unused-modules.md
ljharb: lol it's fine as-is, i just thought it was funny :-)

src/rules/no-unused-modules.js
ljharb: A folder named `backup_node_modules` shouldn't return true here, nor should `node_modules_old`. Only a path that is *inside* a directory called, verbatim, `node_modules`, should get a `true`.

rfermann: That would mean, if someone makes a backup of the ```node_modules``` folder and renames the new folder to e.g. ```node_modules_old```, all files within this new folder will be analyzed and perhaps reported, if ```src``` and ```ignoreExports```aren't explicitly defined.

But on the other side it's not up to us covering strange folder naming, is it? I will adjust it later.

package.json
```diff
@@ -52,7 +52,7 @@
     "chai": "^3.5.0",
     "coveralls": "^3.0.0",
     "cross-env": "^4.0.0",
-    "eslint": "2.x - 5.x",
+    "eslint": "^5.11.0",
```
ljharb: This must be reverted, since tests are run on all eslint versions.

src/rules/no-unused-modules.js
ljharb: Yes, that’s exactly right - if they do that, they should see reports until they explicitly ignore the new folder.

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,714 @@
...
+          }
+        })
+      }
+    })
+  })
+}
+
+const getSrc = src => {
+  if (src) {
+    return src
```
ljharb: Should this function be ensuring it’s an array? If so:
```suggestion
     return [].concat(src);
```

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,714 @@
...
+      
+      let value = ''
+      if (exportedValue === IMPORT_DEFAULT_SPECIFIER) {
+        value = DEFAULT
+      } else {
+        value = exportedValue
+      }
+      
+      if (typeof exportStatement !== 'undefined'){
+        if ( exportStatement.whereUsed.size < 1) {
```
ljharb:
```suggestion
        if (exportStatement.whereUsed.size < 1) {
```

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,714 @@
...
+      if (typeof namespaceImports !== 'undefined') {
+        if (namespaceImports.whereUsed.size > 0) {
+          return
+        }
+      }
+
+      const exportStatement = exports.get(exportedValue)
+      
+      let value = ''
+      if (exportedValue === IMPORT_DEFAULT_SPECIFIER) {
```
ljharb: This can just be a const and a ternary

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,714 @@
...
+}
+
+const newNamespaceImportExists = specifiers =>
+  specifiers.some(({ type }) => type === IMPORT_NAMESPACE_SPECIFIER)
+
+const newDefaultImportExists = specifiers =>
+  specifiers.some(({ type }) => type === IMPORT_DEFAULT_SPECIFIER)
+
+module.exports = {
+  isNodeModule,
```
ljharb: why is this exported?

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,714 @@
...
+  doPreparation,
```
ljharb: or this?

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,714 @@
...
+  getSrc,
```
ljharb: or this?

tests/src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,578 @@
...
+  missingExports: true,
+}]
+
+const unusedExportsOptions = [{
+  unusedExports: true,
+  src: [testFilePath('./no-unused-modules/**/*.js')],
+  ignoreExports: [testFilePath('./no-unused-modules/*ignored*.js')],
+}]
+
+describe('getSrc returns correct source', () => {
```
ljharb: I think if these need unit tests, they should be extracted to a utility/helper file and not exported from the rule itself.

src/rules/no-unused-modules.js
rfermann: The json schema already ensures that the ```src``` passed to the rule config is an array.

This function should return an array containing ```process.cwd()``` in case no ```src``` has been defined within the rule config.

ljharb: Makes sense

tests/src/rules/no-unused-modules.js
rfermann: I am not sure, if these really need to be tested explicitly, as they are already implicitly tested by the rule tests and a malfunction should usually be captured here. 

What would be the best choice in your opinion? Would you rather remove the explicit tests or extract the functions into a helper module?

src/rules/no-unused-modules.js
rfermann: Basically for testing, just as you supposed.

src/rules/no-unused-modules.js
```diff
@@ -0,0 +1,706 @@
...
+
+const newNamespaceImportExists = specifiers =>
+  specifiers.some(({ type }) => type === IMPORT_NAMESPACE_SPECIFIER)
+
+const newDefaultImportExists = specifiers =>
+  specifiers.some(({ type }) => type === IMPORT_DEFAULT_SPECIFIER)
+
+module.exports = {
+  meta: {
+    docs: { url: docsUrl('no-unused-modules') },
```
futpib: Isn't the name `no-unused-modules` misleading? This rule reports individual export declarations, but the name suggests it reports whole files as unused.

ljharb: my understanding is that it reports on both - both unused files, and unused exports.

rfermann: Just as @ljharb says, it reports on both:

- modules without any exports are reported [here](https://github.com/rfermann/eslint-plugin-import/blob/d35b7ff37c2778ed43d824cccf9553cd9faafc79/src/rules/no-unused-modules.js#L283)
- and modules with unused exports are reported [here](https://github.com/rfermann/eslint-plugin-import/blob/d35b7ff37c2778ed43d824cccf9553cd9faafc79/src/rules/no-unused-modules.js#L325) and [here](https://github.com/rfermann/eslint-plugin-import/blob/d35b7ff37c2778ed43d824cccf9553cd9faafc79/src/rules/no-unused-modules.js#L331)

However this rule currently does not report, if a module only consists of unused exports. Each of these exports is reported separately. 
I am not sure whether it makes sense to have one single report being triggered if none of the existing exports is used.

ljharb: It does make sense to me in that case to only provide one report; just as I’d expect when a module has no exports.

Issue #1041, opened 2018-03-12
Rule proposal: no-rename-default


steve-taylor:
Consider the following:

components/foo.js:

```jsx
export default class Foo extends React.Component {
    render() {
        // ...
    }
}
```

widgets/foo.js:

```jsx
import FooComponent from '../components/foo';

export default class Foo extends BaseWidget {
    constructor() {
        super(FooComponent);
    }

    // ...
}
```

`Foo` (in components/foo.js) and `FooComponent` are exactly the same thing, but have been renamed via default import/export. When one thing has two or more names, it increases the difficulty of comprehending code, especially when there are many functions, classes, etc. that are renamed on import.

The proposal is to have a rule that generates a warning or error when a default `import` is given a different name than a default `export`.

ljharb:
This is a valuable part of both default and named import syntax (the ability to rename things), and I’m not sure why it’s an issue that needs a rule.

steve-taylor:
It’s an issue for the reasons outlined above. There are already plenty of rules that, when applied, restrict the usage of language capabilities that some people may find useful. It’s a matter of preference.

ljharb:
What would happen when you wanted to use the Foo component and widget in the same file?

Separately, i think that it’s not at all apparent that it’s more difficult to comprehend code when something has a different name in different files; the point of a module is that you think about one module (file) at a time.

steve-taylor:
>What would happen when you wanted to use the Foo component and widget in the same file?

At that point, you might consider naming things more appropriately. For example, you might consider a naming convention whereby widgets are suffixed by `Widget` (e.g. `FooWidget`).

As a last resort, you could use `// eslint-disable-line: import/no-rename-default`.

steve-taylor:
Let's also consider the following case (and I've been guilty of this before):

```js
export default function getDefaultFoo() {
    // ....
}
```

```js
import getMainFoo from '../foos/normal';
```

This can come about as my understanding of the scope, purpose and/or context of a function/class evolves while building it and using it for the first time. It would be nice to have a linting rule to remind me to clean this up by enforcing uniformity between its declared name and its imported name.

For those who don't see this as an issue, they can simply choose not to use this rule, like any other stylistic rule.

steve-taylor:
> the point of a module is that you think about one module (file) at a time.

[...]

PR #1143, opened 2018-07-24
Add rule to enforce default import aliases

mic4ael:
This rule enforces a specific naming for default imports.


docs/rules/rename-default-import.md
```diff
@@ -0,0 +1,56 @@
+# import/rename-default-import
+
+This rule will enforce an alias for a default package import. Only ES6 imports are processed.
```
ljharb: i don't think "an alias" is the proper terminology here; import names are called "bindings", not aliases.

docs/rules/rename-default-import.md
```diff
@@ -0,0 +1,56 @@
...
+Given:
+
+```js
+// ./foo.js
+export default function () { return 'Foo' }
+```
+
+and
+
+```yaml
```
ljharb: the example config should be in json, not yaml.

docs/rules/rename-default-import.md
mic4ael: However, when you take a look at the https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import the docs refer to those parts as `alias{number}`

ThiefMaster:
How is this related to #1041? That issue seems to be about enforcing the same name a library uses to export its default with (e.g. `export default function foo() {}`) while this one is simply about enforcing a name chosen by whoever configures eslint.

mic4ael:
Any update on this?

docs/rules/rename-default-import.md
```diff
@@ -0,0 +1,59 @@
...
+
+and
+
+```json
+// .eslintrc
+{
+  "rules": {
+    "import/rename-default-import": [
+      "warn",
+      {"prop-types": "PropTypes", "Foo": "Foo"}
```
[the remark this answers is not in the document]
mic4ael: sorry, actually there is a mistake here, fixing it

docs/rules/rename-default-import.md
```diff
@@ -0,0 +1,59 @@
+# import/rename-default-import
+
+This rule will enforce a specific binding name for a default package import. Only ES6 imports are processed.
```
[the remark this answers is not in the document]
mic4ael: Currently, we (in our team) are using ES6 only and that is why I envisioned this as only ES6-compatible. I guess at some point if someone needs to use it also for CJS they could improve the rule.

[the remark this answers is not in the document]
ThiefMaster: Does it make much sense to support CJS requires? Default imports aren't that common there, are they?

mic4ael:
Any update on this one?

docs/rules/rename-default-import.md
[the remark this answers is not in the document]
ljharb: CJS only has default exports, so yes, they’re pretty common (100%).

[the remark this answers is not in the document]
ljharb: I think this rule should apply to CJS out of the gate - including when you import a node_module (that’s thus CJS)

docs/rules/rename-default-import.md
```diff
@@ -0,0 +1,61 @@
...
+and
+
+```json
+// .eslintrc
+{
+  "rules": {
+    "import/rename-default-import": [
+      "warn", {
+        "prop-types": "PropTypes", // key: name of the module, value: desired binding for default import
+        "./foo": "Foo"
```
ljharb: “./foo” is a relative path - meaning that it only makes sense inside a specific module. This could map to 10 different foo.js paths, for example. Is the intent to restrict all of them to the same name?

src/rules/rename-default-import.js
```diff
@@ -0,0 +1,93 @@
...
+ * @author Michał Kołodziejski
+ */
+
+import docsUrl from '../docsUrl'
+
+
+function isDefaultImport(specifier) {
+  if (specifier.type === 'ImportDefaultSpecifier') {
+    return true
+  } else if (specifier.type === 'ImportSpecifier' && specifier.imported.name === 'default') {
```
ljharb: There’s no need for the else here; this function can be two if’s and one return false

src/rules/rename-default-import.js
```diff
@@ -0,0 +1,93 @@
...
+  } else {
+    return false
+  }
+}
+
+
+function handleImport(context, specifier, source) {
+  const {value: packageName} = source
+  const {local: {name: importAlias}} = specifier
+  const mappings = context.options[0]
```
ljharb: || {}, or else the Object.keys call will throw

src/rules/rename-default-import.js
```diff
@@ -0,0 +1,93 @@
...
+        let newAlias = mappings[packageName]
+        if (specifier.imported && specifier.imported.name === 'default') {
+          newAlias = `default as ${mappings[packageName]}`
+        }
+
+        return fixer.replaceText(specifier, newAlias)
+      },
+    })
+
+    const declaredVariable = context.getDeclaredVariables(specifier)[0]
```
ljharb: Why is this “the first” of the declared variables?

docs/rules/rename-default-import.md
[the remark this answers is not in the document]
ThiefMaster: good point, indeed makes sense to support cjs then. probably not too much work anyway

docs/rules/rename-default-import.md
ThiefMaster: what about just removing the docs/tests for this case? in the end we just look at it as a string - so it would still work if someone used it, but it'd simply be undocumented (it indeed doesn't make much sense to set this for something relative)

ljharb: If it doesn’t make sense, it shouldn’t be allowed - docs/tests are necessary

src/rules/rename-default-import.js
```diff
@@ -0,0 +1,93 @@
...
+        },
+      })
+    }
+  }
+}
+
+
+module.exports = {
+  meta: {
+      docs: {
```
ThiefMaster: too much indent; the rest of the file is using 4 spaces (also below inside `create`)

src/rules/rename-default-import.js
mic4ael: so, according to `eslint` docs

> If the node is an ImportSpecifier, ImportDefaultSpecifier, or ImportNamespaceSpecifier, the declared variable is returned.

I care about the binding we create here so that later I can get all the references to that binding used in the code.

ljharb: sorry, to clarify - why is this returning an array, but you only care about the first item? is it always an array of length 1?

mic4ael: That depends on the type of the argument passed to the method as far as I understood

mic4ael:
Updated. Let me know what  you think.

tests/src/rules/rename-default-import.js
```diff
@@ -0,0 +1,212 @@
+import { test } from '../utils'
+
+import { RuleTester } from 'eslint'
+
+const ruleTester = new RuleTester(),
+      rule = require('rules/rename-default-import')
```
ljharb: let's do one declaration per variable, variable lists are last-decade's style :-)

tests/src/rules/rename-default-import.js
```diff
@@ -0,0 +1,212 @@
...
+      output: `import PropTypes from 'prop-types';`,
+      errors: [{
+        ruleId: 'rename-default-import',
+        message: `Default import from 'prop-types' should be aliased to PropTypes, not propTypes`,
+      }],
+    }),
+    test({
+      code: `const propTypes = require('prop-types');`,
+      options: [{'prop-types': 'PropTypes'}],
+      output: `const PropTypes = require('prop-types');`,
```
ljharb: this isn't a safe autofix - anything that referenced `propTypes` would have to be updated too.

mic4ael: this is covered below:

```
test({
  code: `
  const propTypes = require('prop-types');

  const obj = {
    foo: propTypes.string
  }
  `,
  options: [{'prop-types': 'PropTypes'}],
  output: `
  const PropTypes = require('prop-types');

  const obj = {
    foo: PropTypes.string
  }
  `,
  errors: [{
    ruleId: 'rename-default-import',
    message: `Default import from 'prop-types' should be aliased to PropTypes, not propTypes`,
  }, {
    ruleId: 'rename-default-import',
    message: `Using incorrect binding name 'propTypes' instead of PropTypes for default import from package prop-types`
  }],
}),
```

mic4ael:
Let me know what you think now

mic4ael:
Any update on this? ;)

tests/src/rules/rename-default-import.js
ljharb: What about if someone has done `export { propTypes }` or referenced it in an object like `{ propTypes }`?

ThiefMaster: I'd ensure the export is unfixable, the object could probably be rewritten to the long form - but I'm not sure if it wouldn't be better to also not provide autofixing there since it's probably more likely that someone wants to manually fix it whatever code is receiving the object.

mic4ael: Updated. Now I don't fix the `export` statements at all.

src/rules/rename-default-import.js
```diff
@@ -0,0 +1,143 @@
...
+        continue
+      }
+
+      context.report({
+        node: variableReference.identifier,
+        message: `Using incorrect binding name '${variableReference.identifier.name}' `
+                + `instead of ${mappings[packageName]} for `
+                + `default import from package ${packageName}`,
+        fix: fixer => {
+          if (variableReference.identifier.parent.type !== 'ExportSpecifier') {
```
ThiefMaster: Am I misreading this or are you now simply skipping exports when fixing things? If a variable gets exported with the wrong name, then no occurrences of it nor its import should be renamed. Renaming it only in some places but not others would result in broken code.

src/rules/rename-default-import.js
```diff
@@ -0,0 +1,143 @@
...
+}
+
+function handleImport(context, node, specifierOrDeclaration, packageName, importAlias) {
+  const mappings = context.options[0] || {}
+
+  if (Object.keys(mappings).indexOf(packageName) === -1) {
+    return
+  }
+
+  if (mappings[packageName] !== importAlias) {
```
ThiefMaster: You can avoid one whole level of indentation further down:

```
if (mappings[packageName] !== importAlias) {
  return
}
```

ThiefMaster: Also, what about combining this and the previous check like this?

```
const expectedName = mappings[packageName]
if (expectedName === undefined) {
  return
}
```

src/rules/rename-default-import.js
```diff
@@ -0,0 +1,143 @@
...
+  if (variableInit.type === 'CallExpression') {
+    return variableInit.callee.name === 'require'
+  }
+  return false
+}
+
+function handleImport(context, node, specifierOrDeclaration, packageName, importAlias) {
+  const mappings = context.options[0] || {}
+
+  if (Object.keys(mappings).indexOf(packageName) === -1) {
```
ThiefMaster: `mappings.hasOwnProperty(packageName)`?

mic4ael:
Updated once again.

docs/rules/rename-default-import.md
```diff
@@ -0,0 +1,56 @@
...
+      }
+    ]
+  }
+}
+```
+
+The following is considered valid:
+
+```js
+import {default as PropTypes} from 'prop-types'
```
ljharb: let's add some `require` examples to the docs as well

src/rules/rename-default-import.js
ljharb: there shouldn't be an "else" after a return. This should just be two if's, and one return false.

src/rules/rename-default-import.js
```diff
@@ -0,0 +1,147 @@
...
+  } else if (specifier.type === 'ImportSpecifier' && specifier.imported.name === 'default') {
+    return true
+  }
+  return false
+}
+
+function isCommonJSImport(declaration) {
+  const variableInit = declaration.init
+  if (variableInit.type === 'CallExpression') {
+    return variableInit.callee.name === 'require'
```
ljharb: we should also ensure that the `require` is a global one, and not `function (require) { require(); }`.

src/rules/rename-default-import.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+  return false
+}
+
+function handleImport(context, node, specifierOrDeclaration, packageName, importAlias) {
+  const mappings = context.options[0] || {}
+  const expectedAlias = mappings[packageName]
+
+  if (expectedAlias === undefined || expectedAlias === importAlias) {
```
ljharb: when would it be `undefined`? is this prevented by a schema error, such that this should really be `has(mappings, packageName)`?

mic4ael: It will be undefined if there is not specific mapping for a given package. For example, I might want to rename `prop-types` but there are other package that i don't care about.

ljharb: right - but i mean, if someone puts `{ "foo": undefined }` in their schema?

mic4ael: I could always try to validate the options so that we don't have accept something like `undefined` which doesn't make sense.

mic4ael:
`import * as Something from 'prop-types'` shouldn't throw any error and that is how it works. Also, I have updated my PR according to the latest comments ;)

src/rules/rename-default-import.js
ljharb: this doesn't seem addressed yet?

src/rules/rename-default-import.js
```diff
@@ -0,0 +1,151 @@
...
+
+module.exports = {
+  meta: {
+    docs: {
+      url: docsUrl('rename-default-import'),
+      recommended: false,
+    },
+    schema: [
+      {
+        type: 'object',
```
ljharb: should this assert `minProperties: 1` as well?

src/rules/rename-default-import.js
mic4ael: I did in the line 136 (didn't want to pass the context so I did it there).

mic4ael:
Updated ;)

mic4ael:
I had to fix a problem with ESLint 4 and now it works fine. Would be nice to know if there is something else to add/fix.

mic4ael:
Hey, what is the state of this PR at the moment?

ljharb:
It's awaiting review from a second collaborator.

mic4ael:
Maybe there is another collaborator that we could ask for a review?

mic4ael:
Hm, just a casual reminder that this is still open ;)

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+- Add [`rename-default-import`] rule: Enforce default import naming
```
```suggestion
- Add [`rename-default-import`] rule: Enforce default import naming ([#1143], thanks [@mic4ael])
```
and the appropriate links for the PR and your username

tests/src/rules/rename-default-import.js
```diff
@@ -0,0 +1,357 @@
...
+      output: `import {default as PropTypes} from 'prop-types';`,
+      errors: [{
+        ruleId: 'rename-default-import',
+        message: `Default import from 'prop-types' should be aliased to PropTypes, not propTypes`,
+      }],
+    }),
+    test({
+      code: `import {default as propTypes, foo} from 'prop-types';`,
+      options: [{'prop-types': 'PropTypes'}],
+      output: `import {default as PropTypes, foo} from 'prop-types';`,
```
ljharb: i'm a little concerned with making this autofixable, if you won't be updating all the other places ijn the file where this variable is referenced (and it might not be safe to update it, if it's used in something like `const obj = { PropTypes };`).

docs/rules/rename-default-import.md
```diff
@@ -0,0 +1,63 @@
...
+
+and
+
+```json
+// .eslintrc
+{
+  "rules": {
+    "import/rename-default-import": [
+      "warn", {
+        "prop-types": "PropTypes", // key: name of the module, value: desired binding for default import
```
ljharb: what validates that the RHS of this object is a valid JS identifier?

Uko:
OMG, guys. You are so slow. This is a super nice rule and it takes you forever to integrate. I guess I'll just copy-paste it to my project…

ljharb:
@Uko the PR is waiting for a response from @mic4ael.

@mic4ael i apologize for the very long delay in getting you a review, but it's been almost as long since I've provided one :-) are you still interested in updating this PR?

mic4ael:
> @Uko the PR is waiting for a response from @mic4ael.
> 
> @mic4ael i apologize for the very long delay in getting you a review, but it's been almost as long since I've provided one :-) are you still interested in updating this PR?

hey @ljharb sorry for the delay, I will update my PR soon ;)

docs/rules/rename-default-import.md
mic4ael: At the moment it is only the user of the plugin who should be aware what they are doing. I could do something like ``eval(`(() => {var ${mapping}})()`)`` or ``new Function(`var ${mapping}`)``

tests/src/rules/rename-default-import.js
mic4ael: As far as I can tell this will update all the references apart from the exported identifiers. I am getting confused. Isn't it something that we agreed on?

docs/rules/rename-default-import.md
mic4ael: Even though it is not the safest way to check for this..

ThiefMaster: is checking that even important? sounds like a case of garbage-in, garbage-out if people put invalid stuff there...

ljharb: I prefer my projects not to ever emit garbage :-)

mic4ael:
> Thanks! I made a few tweaks and added a test case (which is unfortunately failing).
> 
> I'm not convinced on the rule _name_, though - `rename-default-import` is a bit weird; this seems more like `enforce-import-name`. It also kind of seems like the same rule should be preventing named imports from being renamed with `as`?

Ok, I will rename it and will look into the suggestion you made.

tests/src/rules/rename-default-import.js
mic4ael: @ljharb can you please elaborate on this comment since I think it does what you said.

docs/rules/rename-default-import.md
mic4ael: If we decide to validate the user-provided identifier I would like to find out an easy way to do it. Because so far I couldn't find a simple and secure solution.

mic4ael:
Also, as to this:

> It also kind of seems like the same rule should be preventing named imports from being renamed with as

Can you please provide an example of what exactly you mean by that?

mic4ael:
hey @ljharb, can you please take a quick look at my responses? I would like to clarify some of the issues and (maybe) finalize this PR.

benmosher:
FWIW, I agree with @ljharb that the rule's name sounds like it might enforce any import name, not just the default, _but_ I don't know that there is much value in the latter and I think the config could be made smarter if that enhancement is desired later, e.g. `"prop-types": { "default": "PropTypes", "allow-renames": false | ["x", "y"] | ... }`

> OMG, guys. You are so slow. 

Hard to argue with this 😅 since I'm generally MIA and @ljharb is a person dedicated to keeping it clean and understands the cost of adding code that you're not prepared to maintain, which I really appreciate. 

Also, there is always [`patch-package`](https://www.npmjs.com/package/patch-package) in a pinch!

phaux:
I just looked at this rule source code and it looks like it doesn't do at all what was described in the original #1041 feature request? It was supposed to enforce the name in the import statement, but it should go to the imported file and look at the export name and enforce that.

Right now it only enforces the names which you define in the options and you have to manually list all the modules with default exports and also find out what are the names of these default exports. All this information is already available: for external packages it's in their `.d.ts` file and for relative imports it's in the source code of the imported file.

I'm 100% sure it's possible to implement because VSCode can suggest autoimports based on the names of default exports in other files. Also I'm pretty sure that TSLint has a similar rule.

benmosher:
Ah yep, good point @phaux. also @ThiefMaster and @ljharb before you, heh. 😅 I should have re-read the source issue before chiming in at the end. (also: I didn't read enough of the following discussion to know if I'm about to just restate discussion from there 😬)

@mic4ael: is it important for your use that the rule config allow you to lint for a name that is not the original module's default export's name? 

Properly solving #1041 will necessitate the original module code is available but that is typically the expected case for this plugin, and the resolver sub-plugins implement that side of it anyway (e.g. the TS resolver plugin knows where to find `.d.ts` files or TS source, node resolver checks `package.json`, etc.). 

Also consequently, if the default export is some nameless expression, this rule would not apply.

Less config == better config, IMO. I'd be fine with allowing _overrides_ if absolutely necessary, but *not* allowing them and just always reading the source for the default export's true name would be simpler.

ThiefMaster:
> Also consequently, if the default export is some nameless expression, this rule would not apply.

I don't think this is helpful... as a user of a library you have no influence on that. I don't know if e.g. `prop-types` has a named default export... (we want it to be importet consistently as `propTypes` and not `PropTypes`)

benmosher:
Okay, so there _is_ some desire to allow overrides then. That would support unnamed exports, false modules (e.g. CommonJS), and renaming. 

Still seems like you'd want the _default_ behavior to pick up the module's original name if it's available, though?

docs/rules/rename-default-import.md
manuth: Hi @mic4ael
Sorry for reincarnating this thread after such a long time. My take on this would be to parse either `var ${mapping}` or `${mapping}` using `espree` - a module also used internally by `eslint` for parsing code.

In case of `${mapping}`:
https://astexplorer.net/#/gist/94e57078a7e95dbf910a262c645dd3ea/b82ba422cd8e946bee02a1209795a1bbaa9be225

The parsed AST should consist of exactly one `ExpressionStatement` having an `Identifier` set as its `expression`. Everything else (or an exception)  is treated as an incorrect `mapping`.

In case of `var ${mapping}`:
https://astexplorer.net/#/gist/40823431b92aaf01426b255e4ad7163f/24a943d0b759cdf8aff897dfa1c5dcba48cb3847

The parsed AST should consist of exactly one `VariableDeclaration`, which has its `init` set to `null` and its `id` set to an `Identifier`.

That way you can be sure `mapping` is no garbage.

basickarl:
Bump! So, how's it going guys?

tests/src/rules/rename-default-import.js
ljharb: Can we include test cases that have lots of references to the imported thing, so the autofix output can demonstrate that it updates too?

Issue #1146, opened 2018-07-30
Resolving with eslint-import-resolver-webpack silently fails when module is not installed


balazsczap:
After a while of debugging, I found that if a 'webpack' resolver configuration is given in `.eslintrc`, but no installation of `eslint-import-resolver-webpack` is found, the resolve process then tries to use `webpack` itself (since it tries to use the configuration name as a custom package, if a package with the conventional name is not found) as a resolver. 

- If `webpack` is intalled, this obviously fails, since `webpack.resolveImport` is undefined, but instead of an error message, the rule is shown as having failed.
- If `webpack` is not installed, the rule check correctly fails with `Resolve error: unable to load resolver "webpack"`.

ljharb:
How do you suggest fixing this?

balazsczap:
I've started working on a pull request, which should check if the required method (`.resolveImport` for v1 and `.resolve` for v2) is available on the resolver object being passed to withResolver.

I realize this is only a quick fix, but it will at least provide users with info on why their import rules are failing.

helicese:
I just wonder why `eslint-import-resolver-webpack` is not in dependences

balazsczap:
If I understand correctly, the node resolver is default, and works well enough. The webpack resolver is only needed (in some cases) if you're using webpack. 

In my case the webpack resolver was added to devDependencies as it should be (and thus installed locally), but ESLint was being run globally, and tried to load the resolver globally as well, which it couldn't.

ljharb:
Closed by #1148.

PR #1148, opened 2018-07-31
Add error to output when module loaded as resolver has invalid API

balazsczap:
This adds an error to the ESLint context when _something_ was found with the given resolver name, but it has an incorrect API. Previously this failed silently.

Solves #1146.


utils/resolve.js
```diff
@@ -158,10 +158,24 @@ function requireResolver(name, sourceFile) {
...
+  }
+
+  return resolver
+}
+
+function isResolverValid(resolver) {
+  if (!resolver) {
+    return false
+  }
+  switch (resolver.interfaceVersion) {
```
ljharb: can this be an if/else instead of a switch?

ljharb:
Also, is there a way we could add a test for this?

utils/resolve.js
balazsczap: I've used a switch since a switch was already used to decide between versions in the code previously, but I can definitely change if needed.

ljharb: not a blocker, but it'd be nice - switch statements are horrible :-)

PR #1157, opened 2018-08-09
No named export

### LINKED ISSUE #1156: Rule proposal: no-named-export
Although it has uncommon usages, it would be nice to have a `no-named-export` rule to enforce export default only.

This may be useful, for example, when defining that a component file may not expose its helpers, giving this file a single purpose, forcing helpers to be put in a proper file/layer with specific rules and behaviors; and organizing those module exports in an index file.

In a practical example:
```js
// valid
import React from 'react';

const App = () => (<div className="App">App</div>);

export default App;
```

```js
// valid
import React from 'react';

const title = 'App';

const App = () => (<div className="App">{title}</div>);

export default App;
```

```js
// invalid
import React from 'react';

export const title = 'App';

const App = () => (<div className="App">{title}</div>);

export default App;
```

fsmaia:
Fixes #1156.

### DIFF
--- a/src/rules/no-named-export.js
+++ b/src/rules/no-named-export.js
@@ -1,0 +1,33 @@
+import docsUrl from '../docsUrl'
+
+module.exports = {
+  meta: {
+    docs: { url: docsUrl('no-named-export') },
+  },
+
+  create(context) {
+    // ignore non-modules
+    if (context.parserOptions.sourceType !== 'module') {
+      return {}
+    }
+
+    const preferDefault = 'Prefer default export.'
+
+    return {
+      ExportAllDeclaration(node) {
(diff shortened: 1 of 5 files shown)


src/rules/no-named-export.js
```diff
@@ -0,0 +1,33 @@
...
+    docs: { url: docsUrl('no-named-export') },
+  },
+
+  create(context) {
+    // ignore non-modules
+    if (context.parserOptions.sourceType !== 'module') {
+      return {}
+    }
+
+    const preferDefault = 'Prefer default export.'
```
ljharb: This message could be a bit stricter.

fsmaia:
Any news?

Issue #1130, opened 2018-07-04
Why force specific quotes in dynamic-import-chunkname?


st-sloth:
The rule [dynamic-import-chunkname](https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/dynamic-import-chunkname.md) forces to use double quotes in the chunk comment. Is is literally said in docs so:

> ### invalid
> The following patterns are invalid:
> ```javascript
> // using single quotes instead of double quotes
> import(
>   /* webpackChunkName: 'someModule' */
>   'someModule',
> );
> ```

What is the rationale for that? Should I make a PR allowing both single and double quotes by default, or put allowing / forbidding behind an option?

ljharb:
Does webpack support both formats?

st-sloth:
It does. Moreover it supports a couple of other "invalid" cases:

> ```javascript
> // incorrectly formatted comment
> import(
>   /*webpackChunkName:"someModule"*/
>   'someModule',
> );
> 
> // single-line comment, not a block-style comment
> import(
>   // webpackChunkName: "someModule"
>   'someModule',
> );

because [it "parses" these options by inserting comment contents inside an object body](https://github.com/webpack/webpack/blob/eaa5bc80960b70cc01a76fff67007706889fe875/lib/Parser.js#L2089). Which we could also do here in this rule, that is currently too strict and considers cases with multiple "comment options", like `webpackChunkName: "someModule", anotherOption: 123` invalid because of lack of a space after the closing quote. (I guess I should have named the issue after overall specificity of the rule in the first place, not just quotes). So, do I migrate the parsing logic here?

ljharb:
I’m fine adding an option for quotes, but i don’t think there’s any value in allowing “anything”

st-sloth:
1) By option you mean "default is any + restrictive option" or "default stays double + relaxative option?

2) In general, perhaps having comment parsing logic different from that of webpack isn't the best approach? Because, for instance, parsing multiple comment options is currently broken. It reports as invalid proper syntax:

```javascript
import(
    /* webpackChunkName: "a" */
    /* webpackPrefetch: true */
    './a'
)

import(
    /* webpackPrefetch: true */
    /* webpackChunkName: "a" */
    './a'
)

import(/* webpackChunkName: "a", webpackPrefetch: true */ './a')
```

while accepting

```javascript
import(/* webpackPrefetch: true, webpackChunkName: "a" */ './a')

[...]

PR #1163, opened 2018-08-22
fix(dynamic-import-chunkname): Add proper webpack comment parsing

st-sloth:
As briefly discussed in #1130, current webpack comment parsing logic (without taking styling into account) is severely flawed, reporting valid comments as errors while also allowing invalid syntax.

This PR adds
* checking for valid syntax the way webpack itself parses these comments
* support for multiple keys in these comments, both in a single comment and in several comments (see examples in docs or tests)

Styling requirements have not been relaxed by these changes.

The main topic of #1130 - allowing single quotes in strings in webpack comments - will be implemented on top of these changes once they are merged.

### DIFF
--- a/docs/rules/dynamic-import-chunkname.md
+++ b/docs/rules/dynamic-import-chunkname.md
@@ -28,6 +28,10 @@ import(
   /*webpackChunkName:"someModule"*/
   'someModule',
 );
+import(
+  /* webpackChunkName : "someModule" */
+  'someModule',
+);
 
 // chunkname contains a 6 (forbidden by rule config)
 import(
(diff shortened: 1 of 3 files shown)


st-sloth:
> AppVeyor build failed

Locally tests pass.

lencioni:
I investigated why we originally wrote this to enforce double quotes. It turns out that we have an internal babel plugin that only works with double quotes. We originally thought this was webpack's behavior, so we encoded it into this rule.

byteme980:
LGTM, thank you for making these changes 🎉

PR #1176, opened 2018-09-29
Fix packageDir array support

### LINKED ISSUE #1175: Bug: Multiple packageDirs in no-extraneous-dependencies
According to the docs, packageDirs accepts an array of paths, and (presumably) the package.json's found at all the paths have some effect.

In the current implementation, only the last element in the array is used.

This is because [`Object.assign()` is used](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-extraneous-dependencies.js#L45) to merge the dependencies found in each file. As `Object.assign()` performs a shallow merge, and the [extractDepFields function](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-extraneous-dependencies.js#L15-L22) adds default empty objects for fields that are not defined in `package.json`, every field ends up being overridden in each iteration.

I suggest that a deep merge be performed here instead of `Object.assign`.

pzhine:
Fixes #1175

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Fixed
+- [`no-extraneous-dependencies`]: `packageDir` option with array value was clobbering package deps instead of merging them ([#1175] thanks [@aravindet])
+
 
 ## [2.14.0] - 2018-08-13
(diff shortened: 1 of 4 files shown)


benmosher:
looks great! thanks!

ljharb:
@benmosher this was merged while tests were failing, and now master is broken. This needs to be fixed or reverted.

benmosher:
yeah, weird. the original commit built fine though. looks like some deep typescript dependency changed and is why the later (markdown-only) builds failed.

robin-ambachtsheer:
When will this fix be published?

afenton90:
Does this issue need picking up again? As it is broken in the current version of `master`. Happy to re-apply the fix?

benmosher:
published! v2.15.0

PR #1180, opened 2018-10-11
fix typescript build issue

benmosher:


benmosher:
I'm not in a place where I can debug locally. this was my first instinct to see what's up with the build.

benmosher:
fyi @ljharb, early test results show it was indeed typescript resolver v1.1 that introduced the tsconfig.json requirement. (seems reasonable enough, lol)

benmosher:
merged PR build fails seem spurious and unrelated, so I'ma merge

Issue #1035, opened 2018-03-03
Lint Dynamic Imports for no-unresolved


tgreen7:
For code-splitting in our app we have
```
import(/* webpackChunkName: "WelcomePage" */ "./components/WelcomePage")
```

However, we don't get the same awesome eslint error if that path doesn't exist. It would be great to have that linted as well.

ljharb:
Seems reasonable, altho it would require using babel-eslint.

ljharb:
Actually tho - import() uses urls, not paths, so I’m not sure how this would work properly.

vikr01:
> import() uses urls, not paths
Doesn't it use paths and webpack transforms them into urls?

I've also noticed that `import/resolve` doesn't work on `require.*` statements (i.e. `require.resolve` and `require.ensure`). This would be a really neat feature to have.

ljharb:
`require.ensure` is nonstandard webpack nonsense, but `require.resolve` would make some sense.

PR #1216, opened 2018-10-20
Add support for require.resolve

vikr01:
Related to discussion in #1035 ([link to comment](https://github.com/benmosher/eslint-plugin-import/issues/1035#issuecomment-416503352)).

So far I've only added `require.resolve` for `commonjs` with `no-unresolved`, let me know if I should add it anywhere else.

### DIFF
--- a/tests/src/rules/no-unresolved.js
+++ b/tests/src/rules/no-unresolved.js
@@ -150,6 +150,23 @@ function runResolverTests(resolver) {
         }],
       }),
 
+      rest({
+        code: 'var bar = require.resolve("./baz")',
+        options: [{ commonjs: true }],
+        errors: [{
+          message: "Unable to resolve path to module './baz'.",
+          type: 'Literal',
+        }],
+      }),
+      rest({
+        code: 'require.resolve("./baz")',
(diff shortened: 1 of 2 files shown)


vikr01:
@ljharb was thinking the same thing, yeah I'll make a version using an option

vikr01:
Closing in favor of #1217

ljharb:
I'm going to keep the two in sync; orphaned PRs clutter the log.

benmosher:
@ljharb why is this a breaking change if `require.resolve` is part of CommonJS?

feels semver minor to me, in that it is sortof a bugfix, but broad enough that it qualifies as an added feature.

I'd prefer this over require-ing (pun intended) users to discover and enable this behavior.

vikr01:
@benmosher There may be some who are using `require.resolve` on files generated at build time (instead of `path.join`/`path.resolve`).

benmosher:
fair enough, I guess can just try to make the note to bump it to on-by-default in v3.

ljharb:
@benmosher new warnings are almost always a breaking change.

PR #1217, opened 2018-10-21
Add requireResolve option

vikr01:
Alternate to #1216 

Adds an option `requireResolve` to check `require.resolve` statements.

Might want to support `require.resolve` by default in a future major version though.

### DIFF
--- a/utils/moduleVisitor.js
+++ b/utils/moduleVisitor.js
@@ -80,6 +80,19 @@ exports.default = function visitModules(visitor, options) {
     }
   }
 
+  function checkRequireResolve(call) {
+    if (call.callee.type !== 'MemberExpression') return
+    if (call.callee.object.name !== 'require') return
+    if (call.callee.property.name !== 'resolve') return
+    if (call.arguments.length !== 1) return
+
+    const modulePath = call.arguments[0]
+    if (modulePath.type !== 'Literal') return
+    if (typeof modulePath.value !== 'string') return
+
+    checkSourceValue(modulePath, call)
+  }
+
   const visitors = {}
   if (options.esmodule) {
     Object.assign(visitors, {
@@ -97,6 +110,14 @@ exports.default = function visitModules(visitor, options) {
     }
   }
 
+  if (options.requireResolve) {
+    const currentCallExpression = visitors['CallExpression']
+    visitors['CallExpression'] = function (call) {
+      if (currentCallExpression) currentCallExpression(call)
+      checkRequireResolve(call)
+    }
+  }
+
   return visitors
 }
 
@@ -111,6 +132,7 @@ function makeOptionsSchema(additionalProperties) {
       'commonjs': { 'type': 'boolean' },
       'amd': { 'type': 'boolean' },
       'esmodule': { 'type': 'boolean' },
+      'requireResolve': { 'type': 'boolean' },
       'ignore': {
         'type': 'array',
         'minItems': 1,
--- a/tests/src/rules/no-unresolved.js
+++ b/tests/src/rules/no-unresolved.js
@@ -68,6 +68,15 @@ function runResolverTests(resolver) {
            , options: [{ amd: true }]}),
       rest({ code: 'define(["./does-not-exist"], function (bar) {})' }),
 
+      // requireResolve setting
+      rest({ code: 'var foo = require.resolve("./bar")'
+           , options: [{ requireResolve: true }]}),
+      rest({ code: 'require.resolve("./bar")'
+           , options: [{ requireResolve: true }]}),
+      rest({ code: 'require.resolve("./does-not-exist")'
+           , options: [{ requireResolve: false }]}),
+      rest({ code: 'require.resolve("./does-not-exist")' }),
+
       // stress tests
       rest({ code: 'require("./does-not-exist", "another arg")'
            , options: [{ commonjs: true, amd: true }]}),
@@ -175,6 +184,24 @@ function runResolverTests(resolver) {
           type: 'Literal',
         },{
           message: "Unable to resolve path to module './does-not-exist'.",
+          type: 'Literal',
+        }],
+      }),
+
+      // requireResolve setting
+      rest({
+        code: 'var bar = require.resolve("./baz")',
+        options: [{ requireResolve: true }],
+        errors: [{
+          message: "Unable to resolve path to module './baz'.",
+          type: 'Literal',
+        }],
+      }),
+      rest({
+        code: 'require.resolve("./baz")',
+        options: [{ requireResolve: true }],
+        errors: [{
+          message: "Unable to resolve path to module './baz'.",
           type: 'Literal',
         }],
       }),


ljharb:
There was no need for a separate PR, for future reference.

I'll keep both PRs updated.

docs/rules/no-unresolved.md
```diff
@@ -53,6 +53,16 @@ define(['./foo'], function (foo) { /*...*/ }) // reported if './foo' is not foun
 require(['./foo'], function (foo) { /*...*/ }) // reported if './foo' is not found
 ```
 
+If `{requireResolve: true}` is provided, single-argument `require.resolve` calls will be resolved:
+
+```js
+/*eslint import/no-unresolved: [2, { requireResolve: true }]*/
+const { default: x } = require.resolve('./foo') // reported if './foo' is not found
+
+require.resolve(0) // ignored
```
ljharb: why is this ignored? i'd expect this to be coerced to a string and then validated the same.

utils/moduleVisitor.js
```diff
@@ -80,6 +80,19 @@ exports.default = function visitModules(visitor, options) {
     }
   }
 
+  function checkRequireResolve(call) {
+    if (call.callee.type !== 'MemberExpression') return
+    if (call.callee.object.name !== 'require') return
+    if (call.callee.property.name !== 'resolve') return
+    if (call.arguments.length !== 1) return
```
ljharb: i think this should be removed; we should check the first argument in every case that i'm aware of.

docs/rules/no-unresolved.md
vikr01: Attempting to run `require.resolve(0)` gives me this stacktrace:

```bash
TypeError [ERR_INVALID_ARG_TYPE]: The "request" argument must be of type string. Received type number
    at Function.resolve (internal/modules/cjs/helpers.js:28:13)
```

utils/moduleVisitor.js
vikr01: I re-used the code from `require` for `require.resolve` so the same checks happen for both. They should both be used the same way, shouldn't they?

docs/rules/no-unresolved.md
ljharb: it seems useful to me to have the linter warn on it

utils/moduleVisitor.js
ljharb: i'm not sure - i suspect `require` might be used differently in requireJS, but `require.resolve`, i think, only exists in CJS.

vikr01: Maybe, for now, `require.resolve` should work only if `commonjs` and `requireResolve` are both set? That way if a different version of it is added to requireJS, we can handle that too.

vikr01: Maybe, for now, `require.resolve` should work only if `commonjs` and `requireResolve` are both set? That way if a different version of it is added to requireJS, we can handle that too.

vikr01: Maybe, for now, `require.resolve` should work only if `commonjs` and `requireResolve` are both set? That way if a different version of it is added to requireJS, we can handle that too.

vikr01: Maybe, for now, `require.resolve` should work only if `commonjs` and `requireResolve` are both set? That way if a different version of it is added to requireJS, we can handle that combination differently.

vikr01: Maybe, for now, `require.resolve` should work only if `commonjs` and `requireResolve` are both set? That way if a different version of it is added to requireJS, we can handle that combination differently.

vikr01: Maybe, for now, `require.resolve` should work only if `commonjs` and `requireResolve` are both set? That way if a different version of it is added to requireJS, we can handle that combination differently.

docs/rules/no-unresolved.md
vikr01: I agree -- but we should follow the same rules we have for `commonjs` `require` right?

utils/moduleVisitor.js
vikr01: Maybe, for now, `require.resolve` should work only if `commonjs` and `requireResolve` are both set? That way if a different version of it is added to requireJS, we can handle that combination differently.

vikr01: I think this is the best option:

`requireResolve: boolean | { commonjs: ?boolean, amd: ?boolean }`

utils/moduleVisitor.js
```diff
@@ -111,6 +142,17 @@ function makeOptionsSchema(additionalProperties) {
       'commonjs': { 'type': 'boolean' },
       'amd': { 'type': 'boolean' },
       'esmodule': { 'type': 'boolean' },
+      'requireResolve': {
+        'oneOf': [
+          { 'type': 'boolean' },
+          {
+            'type': 'object',
+            'properties': {
+              'commonjs': { 'type': 'boolean' },
```
vikr01: I've made the `requireResolve` option either be a boolean or a shape with a `commonjs` option.

`commonjs: true, requireResolve: true` and `commonjs: false, requireResolve: { commonjs: true }` will both lint `require.resolve`.

As far as I can see `requirejs` doesn't have its own version of `require.resolve`, but perhaps it's better to leave the door open for that?

docs/rules/no-unresolved.md
ljharb: Not necessarily - `require` is used in multiple ways, and so has edge cases - `require.resolve`, however, does not have those edge cases.

utils/moduleVisitor.js
ljharb: That seems reasonable (altho resolution like that is unlikely to ever show up in AMD)

utils/moduleVisitor.js
ljharb: I wouldn't count on AMD getting any updates any time soon; it's effectively dead (even if it's continuing to be maintained).

vikr01: So should I revert this to be just a boolean?

docs/rules/no-unresolved.md
vikr01: I'm not sure what you mean. I've never run into any cases where the arguments of `require` (commonjs) and `require.resolve` work differently.

From [the node documentation](https://nodejs.org/api/modules.html#modules_all_together) it says that `require.resolve` "gets the exact filename that will be loaded when `require()`".

vikr01:
@benmosher Could you review this when you have a chance?

lgandecki:
Hello, any reason why this is not merged? I'd also love to have this functionality. I am also happy to change - can deal with the conflicts and test. Just wouldn't want to do the extra work if there is a separate reason for this not being considered anymore.
Thanks!

ljharb (changes requested):
This needs a rebase, and would need test cases as well (and doc updates).

lgandecki:
@ljharb I'm happy to rebase and do the necessary work.  
It seems to me that this PR does have test cases, are there any specifics that you would like to be added?

ljharb:
@lgandecki that'd be great!

current tests look OK, but the more the better :-) in general, for every new test case, i'd prefer it to be tested in each parser, and both with and without the new option.

PR #1218, opened 2018-10-21
Fix overwriting of dynamic import() CallExpression

### LINKED ISSUE #1035: Lint Dynamic Imports for no-unresolved

### LINKED ISSUE #1166: no-cycle rule not working with babel-eslint parser
Here is a simple repro:

package.json
```
{
  "dependencies": {
    "babel-eslint": "^9.0.0",
    "eslint": "^5.4.0",
    "eslint-plugin-import": "^2.14.0"
  }
}
```
.eslintrc
```
{
	parser: "babel-eslint",
	parserOptions: {
		sourceType: 'module',
	},
	plugins: ['import'],
	rules: { 'import/no-cycle': 'error' },
}
```

a.js
```
import B from './b';
export default "A" + B;
```

b.js
```
import A from './a';
export default A + "B";
```

No error is found by eslint but if default parser is used (by removing line `parser: "babel-eslint",`) then `no-cycle` errors are found in each file.

Is it the expected behaviour ? An issue with `eslint-plugin-import` ? An issue with `babel-eslint` ?

vikr01:
Closes #1035
Closes #1166

So turns out, dynamic imports were already supported in `moduleVisitor` but they were being overwritten if `commonjs` or `amd` options were set.

- Fixed by saving the current `CallExpression` and including it in the function that's overwriting it.
- Added test cases for `import()` in rules using the `moduleVisitor`. Not sure why, but I couldn't get the `no-cycle` rule to throw errors with `import()`.
- Fixes issues with parser `babel-eslint` not working with `no-cycle`, due to the "flow workaround" for `eslint@<5`

This might be breaking for some users, though.

Also since `import()` can work with URLs, maybe we should use something like [`isURL` from `validator`](https://github.com/chriso/validator.js#validators) to ignore URL imports?

### DIFF
--- a/tests/src/rules/no-cycle.js
+++ b/tests/src/rules/no-cycle.js
@@ -37,7 +37,20 @@ ruleTester.run('no-cycle', rule, {
       options: [{ maxDepth: 1 }],
     }),
     test({
+      code: 'import { foo, bar } from "./depth-two"',
+      options: [{ maxDepth: 1 }],
+    }),
+    test({
+      code: 'import("./depth-two").then(function({ foo }){})',
+      options: [{ maxDepth: 1 }],
+      parser: 'babel-eslint',
+    }),
+    test({
       code: 'import type { FooType } from "./depth-one"',
+      parser: 'babel-eslint',
+    }),
+    test({
+      code: 'import type { FooType, BarType } from "./depth-one"',
       parser: 'babel-eslint',
     }),
   ],
@@ -85,8 +98,27 @@ ruleTester.run('no-cycle', rule, {
       errors: [error(`Dependency cycle via ./depth-two:1=>./depth-one:1`)],
     }),
     test({
+      code: 'import one, { two, three } from "./depth-three-star"',
+      errors: [error(`Dependency cycle via ./depth-two:1=>./depth-one:1`)],
+    }),
+    test({
       code: 'import { bar } from "./depth-three-indirect"',
       errors: [error(`Dependency cycle via ./depth-two:1=>./depth-one:1`)],
+    }),
+    test({
+      code: 'import { bar } from "./depth-three-indirect"',
+      errors: [error(`Dependency cycle via ./depth-two:1=>./depth-one:1`)],
+      parser: 'babel-eslint',
+    }),
+    test({
+      code: 'import("./depth-three-star")',
+      errors: [error(`Dependency cycle via ./depth-two:1=>./depth-one:1`)],
+      parser: 'babel-eslint',
+    }),
+    test({
+      code: 'import("./depth-three-indirect")',
+      errors: [error(`Dependency cycle via ./depth-two:1=>./depth-one:1`)],
+      parser: 'babel-eslint',
     }),
   ],
 })
--- a/utils/moduleVisitor.js
+++ b/utils/moduleVisitor.js
@@ -91,7 +91,9 @@ exports.default = function visitModules(visitor, options) {
   }
 
   if (options.commonjs || options.amd) {
+    const currentCallExpression = visitors['CallExpression']
     visitors['CallExpression'] = function (call) {
+      if (currentCallExpression) currentCallExpression(call)
       if (options.commonjs) checkCommon(call)
       if (options.amd) checkAMD(call)
     }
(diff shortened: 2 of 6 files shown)


tests/src/rules/no-cycle.js
```diff
@@ -88,6 +93,16 @@ ruleTester.run('no-cycle', rule, {
       code: 'import { bar } from "./depth-three-indirect"',
       errors: [error(`Dependency cycle via ./depth-two:1=>./depth-one:1`)],
     }),
+    // test({
```
ljharb: i assume these should be uncommented

vikr01: It should. Errors for `import()` with `no-cycle` aren't working because of [the workaround for `flow` import resolution in `eslint@<5.x`](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-cycle.js#L37).

ljharb: That seems like something worth fixing. Why would a flow workaround impact `import()`, which is actual JS?

ljharb: That seems like something worth fixing. Why would a flow workaround impact `import()`, which is actual JS?

ljharb: That seems like something worth fixing. Why would a flow workaround impact `import()`, which is actual JS?

ljharb: That seems like something worth fixing. Why would a flow workaround impact `import()`, which is actual JS?

vikr01: Fixed by using `importer.importKind` instead of `sourceNode.parent && sourceNode.parent.importKind`. Would that break anything? I added a few more tests for `no-cycle` that import multiple variables together just to be careful.

utils/moduleVisitor.js
```diff
@@ -91,12 +91,15 @@ exports.default = function visitModules(visitor, options) {
   }
 
   if (options.commonjs || options.amd) {
+    const currentCallExpression = visitors['CallExpression']
```
ljharb:
```suggestion
    const currentCallExpression = visitors.CallExpression
```

utils/moduleVisitor.js
```diff
@@ -91,12 +91,15 @@ exports.default = function visitModules(visitor, options) {
...
     visitors['CallExpression'] = function (call) {
```
ljharb: i know this predates your PR, but:
```suggestion
    visitors.CallExpression = (call) => {
```

utils/moduleVisitor.js
```diff
@@ -91,12 +91,15 @@ exports.default = function visitModules(visitor, options) {
...
+      if (currentCallExpression) currentCallExpression(call)
       if (options.commonjs) checkCommon(call)
       if (options.amd) checkAMD(call)
     }
   }
 
+
```
ljharb:
```suggestion
```

utils/moduleVisitor.js
vikr01: I agree, but this seems to have done throughout the project.

See:
- [`src/rules/no-commonjs`](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-commonjs.js#L88)
- [`src/rules/default`](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/default.js#L36)

If we do it for this case, shouldn't we do it for all of them?

tests/src/rules/no-cycle.js
vikr01: Worth noting - `no-cycle` was also not working for static `import` with `babel-eslint` because of this, which is why I added a test case for static `import` with the `babel-eslint` parser.

vikr01:
[Here](https://ci.appveyor.com/project/vikr01/eslint-plugin-import/builds/20370307)'s a passing appveyor on the branch of my fork.

edmorley:
@ljharb thank you for reviewing this and the related #1231.
Do we need a second reviewer, if so, who is best to ask? :-)

corradio:
any updates here? Would be great to have this merged

Issue #1219, opened 2018-10-22
ESLint error: Cannot read property 'externals' of undefined


idudinov:
I've come across an issue with ESLint when using `eslint-import-resolver-webpack` and webpack config is an **array of functions**. 
ESLint starts to fail almost on all rules, saying: 
`[eslint] Resolve error: Cannot read property 'externals' of undefined`
I've managed to found the part of sources which responsible for that. The cause is that it first checks whether config is a function (and then executes it) and after that check whether it's an array. 

https://github.com/benmosher/eslint-plugin-import/blob/b4a2f11fcacc6b2f048da4b29cfc896e682f17d1/resolvers/webpack/index.js#L88-L109
When it tries to find proper config with `resolve` field in the array, it finds nothing, because all elements are functions.
And, in my case, it turns that config is undefined which leads to the error I see.

My proposition is to check if config is an array, try to convert every element from function to object.

In case I understood everything correctly and didn't miss anything, I can make a PR for that, just going to add some tests as Contribution Guide claims

idudinov:
Closing the issue because the PR has been merged.

PR #1220, opened 2018-10-22
Webpack Resolver fix for case config is an array of functions

idudinov:
PR for #1219 
+ aded test for the case

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -87,13 +87,13 @@ exports.resolve = function (source, file, settings) {
   }
 
   if (typeof webpackConfig === 'function') {
-    webpackConfig = webpackConfig(env)
+    webpackConfig = webpackConfig(env, {})
   }
 
   if (Array.isArray(webpackConfig)) {
     webpackConfig = webpackConfig.map(cfg => {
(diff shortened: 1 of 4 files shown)


resolvers/webpack/index.js
```diff
@@ -91,6 +91,14 @@ exports.resolve = function (source, file, settings) {
   }
 
   if (Array.isArray(webpackConfig)) {
+    webpackConfig = webpackConfig.map(cfg => {
+      if (typeof cfg === 'function') {
+        return cfg(env)
```
vikr01: I think it's a good idea to use:

```javascript
return cfg(env, {})
```

[Since the second parameter is parsed arguments](https://webpack.js.org/configuration/configuration-types/).

idudinov: 👍 but need to update other calls as well then

idudinov:
Hi @benmosher! Thanks for approving my PR.
Is there any chance to release this to NPM?
I would not bother you with this if there was a way to reference this version package directly in package.json (subdirectories do not work).

benmosher:
just finally released! sorry for the delay!

PR #1224, opened 2018-10-31
Updated lodash in order to fix CVE-2018-3721

adam-golab:
- [x] Fixed CVE-2018-3721 (more information: https://nvd.nist.gov/vuln/detail/CVE-2018-3721)
- [x] Use `prepare` instead of deprecated `prepublish` (more information: https://github.com/npm/npm/issues/10074)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-import",
-  "version": "2.14.1",
+  "version": "2.14.0",
   "description": "Import with sanity.",
   "engines": {
     "node": ">=4"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-import",
-  "version": "2.14.0",
+  "version": "2.14.1",
   "description": "Import with sanity.",
   "engines": {
     "node": ">=4"
@@ -21,7 +21,7 @@
     "test": "cross-env BABEL_ENV=test NODE_PATH=./src nyc -s mocha -R dot --recursive tests/src -t 5s",
     "test-compiled": "npm run prepublish && NODE_PATH=./lib mocha --compilers js:babel-register --recursive tests/src",
     "test-all": "npm test && for resolver in ./resolvers/*; do cd $resolver && npm test && cd ../..; done",
-    "prepublish": "gulp prepublish",
+    "prepare": "gulp prepublish",
     "coveralls": "nyc report --reporter lcovonly && cat ./coverage/lcov.info | coveralls"
   },
   "repository": {
@@ -79,7 +79,7 @@
     "eslint-import-resolver-node": "^0.3.1",
     "eslint-module-utils": "^2.2.0",
     "has": "^1.0.1",
-    "lodash": "^4.17.4",
+    "lodash": "^4.17.11",
     "minimatch": "^3.0.3",
     "read-pkg-up": "^2.0.0",
     "resolve": "^1.6.0"


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-import",
-  "version": "2.14.0",
+  "version": "2.14.1",
```
ljharb: Please revert this; package version numbers should never be updated in PRs.

package.json
```diff
@@ -21,7 +21,7 @@
     "test": "cross-env BABEL_ENV=test NODE_PATH=./src nyc -s mocha -R dot --recursive tests/src -t 5s",
     "test-compiled": "npm run prepublish && NODE_PATH=./lib mocha --compilers js:babel-register --recursive tests/src",
     "test-all": "npm test && for resolver in ./resolvers/*; do cd $resolver && npm test && cd ../..; done",
-    "prepublish": "gulp prepublish",
+    "prepare": "gulp prepublish",
```
ljharb: please revert this; "prepublish" is the correct script name, despite npm's badly worded warning.

package.json
```diff
@@ -79,7 +79,7 @@
     "eslint-import-resolver-node": "^0.3.1",
     "eslint-module-utils": "^2.2.0",
     "has": "^1.0.1",
-    "lodash": "^4.17.4",
+    "lodash": "^4.17.11",
```
ljharb: This is unnecessary, since `^4.17.4` is a semver range that includes `4.17.11` already. There's no CVE to resolve.

adam-golab: I agree, it includes, but version `4.17.4` has vulnerability, so there should be no possibility to install this version, in order to prevent installing buggy code. Leaving it as `^4.17.4` still allows the vulnerable version to be installed (and it's already happening in one of my library that uses `eslint-plugin-import`). For a test, I forked and installed it locally with this change and it resolves the issue. So, at least we should bump this version to `^4.17.5`

package.json
adam-golab: https://www.balena.io/blog/safely-migrating-away-from-prepublish-with-npm-4/

It's deprecated. I don't understand "badly worded warning". NPM's developers clearly described why this script is deprecated and how one can update it to new script name.

package.json
adam-golab: Reverted

package.json
ljharb: There's no value in doing so; "prepublish" works in every version of npm. I doubt npm will ever remove it, and if/when they do, we can handle it then.

package.json
ljharb: If you install a newer version of lodash at the top level, it will dedupe across this package, since it's in the compatible version range. Separately, the CVE doesn't apply to this package's usage of lodash - so there's no actual issue here.

computamike: Just wanted to clarify this with you @ljharb : I realise that it's been a while since you looked at this.

I have a project using eslint-plugin-import, and when I run npm audit  it identifies this package as introducing a version of lodash vulnerable to prototype pollution - but to fix this issue I need to install a new version of lodash - so I'd create a dependancy in my project to lodash (even though it has no dependancy) 

Services such as Snyk would - in all likelihood - identify this as a potential risk - but the way to mitigate that is to replace what was installed with this library to a newer version through my consuming project

ljharb: Since you have a lockfile in order to use npm audit, you can update the version of lodash without adding it to package.json using npm audit’s suggested commands.

package.json
ljharb:
```suggestion
    "prepublish": "gulp prepublish",
```

PR #1226, opened 2018-11-02
named: commonjs exports

### LINKED ISSUE #1145: import/named for `module.exports`?
Hey, is there any way to enable the `import/named` rule for `module.exports`?

Eg:

```
//lib.js
const toExport = true

module.exports = {
  toExport,
}


//user.js
const {toEpxort} = require('./lib') // here I want it to throw an error

console.log(toEpxort)
```

vikr01:
This + #1222 closes #1145 

This adds a new schema to `named`:

`commonjs: boolean | { require: boolean, exports: boolean }`

The option `commonjs` accepts either a boolean or a shape with `require` and `exports` booleans within. Defaults to `false`.

The reason why the shape is allowed is to avoid having commonjs `require` statements linted. Using the boolean only option will apply to both `require` and `exports`.

The following cases of assigning exports are checked:

```javascript
Object.defineProperty(exports, "key", { value: "myValue" });
Object.defineProperty(module.exports, "key", { value: "myValue" });
Object.defineProperty(module["exports"], "key", { value: "myValue" });
Object.defineProperty(exports, "key", { "value": "myValue" });
Object.defineProperty(module.exports, "key", { "value": "myValue" });
Object.defineProperty(module["exports"], "key", { "value": "myValue" });

exports = {...};
module.exports = {...};
module["exports"] = {...};

exports.key = ...;
module.exports.key = ...;
module["exports"].key = ....;
```

All declarations must be at the top level. For example, this will not be checked:

```javascript
if (foo) {
  exports.foo = foo;
}
```

The reason why we check `Object.defineProperty` is that `babel` will build es6 modules that way, i.e.:

```javascript
Object.defineProperty(exports, "__esModule", { value: true });
```

We need to handle those (truthy) `__esModule` properties correctly -- as in the default export will be the entire `module.exports` if that's not set.

`ExportMap.for` and `ExportMap.parse` now take an `options` parameter (defaults to `{}`), with the keys `useCommonjsExports` and `noInterop`. Reason for `noInterop` is that plain `require` statements should not have `.default` being the entire `module.exports` if it doesn't have a truthy `__esModule` value.

TODO: Configure for `babel-plugin-add-module-exports` built files. (Check for `module.exports = exports["default"]` at the end.)


benmosher:
Ah cool! It actually used to do this back in v0.x or v1, but I removed it for efficiency and to narrow the scope. I will have to think about this carefully before reintroducing.

edi9999:
+1 on this "feature", I mostly use exports like this

```
module.exports = {
   foo: true,
   ...
}
```

PR #1228, opened 2018-11-06
feat:add an ignore option for 'no-extraneous-dependencies'

wheasy:
If the alias module starts with "@", the current "no-extraneous-dependencies" rule will report an error in the case of "@custom/XXX".

### DIFF
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -15,10 +15,20 @@ This rule supports the following options:
 
 `peerDependencies`: If set to `false`, then the rule will show an error when `peerDependencies` are imported. Defaults to `false`.
 
+`ignore`:  Some modules can be ignored by the ignore parameter.It supports regular expressions
+
 You can set the options like this:
 
 ```js
-"import/no-extraneous-dependencies": ["error", {"devDependencies": false, "optionalDependencies": false, "peerDependencies": false}]
+"import/no-extraneous-dependencies": [
+  "error", 
+  {
+    "devDependencies": false, 
+    "optionalDependencies": false,
+    "peerDependencies": false,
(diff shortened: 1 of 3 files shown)


ljharb:
Duplicate of #1227 (please try not to open duplicate PRs)

docs/rules/no-extraneous-dependencies.md
```diff
@@ -15,10 +15,20 @@ This rule supports the following options:
 
 `peerDependencies`: If set to `false`, then the rule will show an error when `peerDependencies` are imported. Defaults to `false`.
 
+`ignore`:  Some modules can be ignored by the ignore parameter.It supports regular expressions
```
ljharb: I’m pretty opposed to options that are regexes; globs would be sufficient.

Separately:
```suggestion
`ignore`:  Some modules can be ignored by the `ignore` parameter. It supports regular expressions.
```

src/rules/no-extraneous-dependencies.js
```diff
@@ -185,6 +199,7 @@ module.exports = {
       allowDevDeps: testConfig(options.devDependencies, filename) !== false,
       allowOptDeps: testConfig(options.optionalDependencies, filename) !== false,
       allowPeerDeps: testConfig(options.peerDependencies, filename) !== false,
+      ignore: (options.ignore || []).map(p => new RegExp(p)),
```
ljharb: This isn’t a safe operation.

ljharb:
This seems like a strange approach to fix the problem of being unable to have an alias that’s a valid package name (with an @). Maybe that should be fixed instead?

wheasy:
> This seems like a strange approach to fix the problem of being unable to have an alias that’s a valid package name (with an @). Maybe that should be fixed instead?

This is really not a good solution. I will close this pr.

PR #1238, opened 2018-11-20
[no-restricted-paths] Allow exceptions to zones

### LINKED ISSUE #834: Make no-restricted-paths more generic
`no-restricted-paths` seems like a great rule.
For example, it lets you ensure nothing in `./a` could reach into `./b`.

My use case is a bit different. I want to ensure that nothing in `./a` could import anything outside of `./a` except for `node_modules`. So if I’m in `./a`, I want it to be impossible to import `../index`, `../b`, or `../../whatever`. However I want to preserve the ability to import Node modules, like `'my-package'` which can technically occur at any level above, and could even symlink to `./b`.

Does this request make sense? Basically I’m trying to confine the use of relative paths to a specific folder, and disallow any relative imports outside of it.

rsolomon:
Closes #834

Allow exceptions to be defined for restricted path zones.

### DIFF
--- a/docs/rules/no-restricted-paths.md
+++ b/docs/rules/no-restricted-paths.md
@@ -9,7 +9,7 @@ In order to prevent such scenarios this rule allows you to define restricted zon
 
 This rule has one option. The option is an object containing the definition of all restricted `zones` and the optional `basePath` which is used to resolve relative paths within.
 The default value for `basePath` is the current working directory.
-Each zone consists of the `target` path and a `from` path. The `target` is the path where the restricted imports should be applied. The `from` path defines the folder that is not allowed to be used in an import.
+Each zone consists of the `target` path and a `from` path. The `target` is the path where the restricted imports should be applied. The `from` path defines the folder that is not allowed to be used in an import. An optional `except` may be defined for a zone, allowing exception paths that would otherwise violate the related `from`. Note that `except` is relative to `from` and cannot backtrack to a parent directory.
 
 ### Examples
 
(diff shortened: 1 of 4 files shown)


rsolomon:
Seems like the appveyor failure is unrelated.

docs/rules/no-restricted-paths.md
```diff
@@ -37,3 +37,43 @@ The following patterns are not considered problems when configuration set to `{
...
+
+and the current file being linted is `my-project/server/one/a.js`.
+
+and the current configuration is set to:
+
+```
+{ "zones": [ {
+    "target": "./tests/files/restricted-paths/server/one",
+    "from": "./tests/files/restricted-paths/server",
+    "except": ["./tests/files/restricted-paths/server/one"]
```
ljharb: It seems like these should be relative paths from the “from”, and be disallowed from backtracking out of it - it wouldn’t make sense to have an “except” that wasn’t included in the “from” in the first place.

src/rules/no-restricted-paths.js
```diff
@@ -53,9 +61,32 @@ module.exports = {
...
         matchingZones.forEach((zone) => {
+          const exceptionPaths = zone.except || []
           const absoluteFrom = path.resolve(basePath, zone.from)
 
           if (containsPath(absoluteImportPath, absoluteFrom)) {
+            if (exceptionPaths.some((exceptionPath) => {
+              const absoluteExceptionPath = path.resolve(absoluteFrom, exceptionPath)
+              const relativeExceptionPath = path.relative(absoluteFrom, absoluteExceptionPath)
+
+              if (importType(relativeExceptionPath, context) === 'parent') {
```
rsolomon: @ljharb this will be the same check every time for a config. Do you know of a rule that does this check only once that I could follow as an example, through caching or some other means? Or is this reasonable?

rsolomon: I opted to refactor just to clean up some calls and remove nesting. Otherwise, I figure caching might add unnecessary complexity for what would be a pretty small (albeit repetitive) check.

christophehurpeau:
hello :) I would love to have this feature, is this mergeable ?

markhagemann:
Will this be merged soon?

ljharb:
@rsolomon i've rebased this, but tests are failing locally for me.

PR #1254, opened 2019-01-03
Update deps, fix tests

ljharb:
master is broken. This PR updates dependencies, but it's still broken.

I suspect that fixing the tests will make this PR pass as well; I'm not clear, however, on how to do that.


benmosher:
Bizarre disjoint test failure sets depending on which Node/ESLint pairing you look at in the Travis results. Almost stranger still that for ESLint 4 + Node 8&10, all tests pass.

I want to get back involved and start cleaning up soon. Hopefully I can take a look this week. I remember there being a TypeScript catch-22 but it's been months since I've looked at it. 😞

ljharb:
We shouldn't need to drop eslint 2/3; can we instead skip the TS tests on those eslint versions?

ljharb:
We could explicitly uninstall the typescript parser in travis.yml in v2 and v3, and then auto skip typescript tests whenever the parser isn’t try/catch requireable?

benmosher:
ooo that is an interesting idea.

I just pushed with them wrapped in a `skipESLints` helper and it works as well as expected.

things still failing:

- the `deprecated` rule is broken with ESLint 5 on account of eslint/espree#394
- no-amd + no-commonjs seem to be totally busted for reasons I don't yet understand. maybe something local, will see what happens on CI

benmosher:
another idea: more travis-set versions of packages in the build matrix.

i.e.
- node4 eslint2 babel-eslint@whatever ts-eslint-parser@whatever
- node6 eslint3 babel-eslint@whatever+2 ts-eslint-parser@whatever+4
...

ljharb:
It makes defining the matrix more annoying, but that’s a much better problem to have than “not testing something” or “tests failing” :-) I’m all for it

benmosher:
leaving myself a reminder to use these to fix `deprecated`: https://github.com/eslint/espree/issues/394#issuecomment-454293678

benmosher:
getting different results on Travis than I am seeing locally. in particular, I don't see the same Node <= 6 failures so not sure what's going on there.

benmosher:
alright, winnowed down to just ESLint 4 / Node 4 matrix entry failing, still due to some nonsense with the typescript stuff.

I'm inclined to drop that entry as it doesn't seem super high-value or high-risk and this whole thing is blocking merging PRs / releasing features. do you want to take a look first @ljharb?

tests/dep-time-travel.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# expected: ESLINT_VERSION numeric env var
+
+npm install --no-save eslint@$ESLINT_VERSION --ignore-scripts || true
```
ljharb: why ignore scripts?

ljharb:
I’d say it’s fine to temporarily add it to allowed failures, but continue running it and pursue fixing it separately?

tests/dep-time-travel.sh
benmosher: I don’t remember, I just copied it from the Travis yaml as it was there. Maybe the old versions have prepublish?

benmosher:
Ahhhh great idea, building now!

benmosher:
Thanks for starting this branch, @ljharb! I’m excited to have master clean to start merging and shipping PRs again!

package.json
```diff
@@ -44,45 +44,45 @@
...
+    "nyc": "^11.9.0",
+    "redux": "^3.7.2",
+    "rimraf": "^2.6.3",
+    "sinon": "^2.4.1",
+    "typescript": "^3.2.2",
+    "typescript-eslint-parser": "^21.0.2"
   },
   "peerDependencies": {
-    "eslint": "2.x - 5.x"
+    "eslint": "*"
```
ljharb: This change means that when eslint 6 comes out, we’ll claim compat with it, but won’t necessarily be. I think we should keep this line the same.

benmosher: That is fair, I forgot when I did this that peerDep mismatch is just a warning these days

ljharb: always has been, but `npm ls` will fail (which is good)

ljharb:
I'm going to merge even though coverage went down slightly.

PR #1255, opened 2019-01-06
Update no-named-as-default.md

Linux249:
fix swap of export/import


docs/rules/no-named-as-default.md
```diff
@@ -31,10 +31,10 @@ For post-ES2015 `export` extensions, this also prevents exporting the default fr
 
 ```js
 // valid:
-export foo from './foo.js'
+import foo from './foo.js'
```
ljharb: This is correct as is; export from is a form of importing.

Linux249: Sorry, my bad - pls delete the PR

PR #1261, opened 2019-01-17
Added support for argv parameter of webpack`s config-as-a-function

keann:
Some people may rely on the map of options passed to webpack (as I do), thus it would be nice to be able to provide this second parameter instead of just an empty object.


benmosher:
Great, thanks! Merging because it looks like failures are due to TS shenanigans resolved after you opened this.

PR #1262, opened 2019-01-18
[New] `no-unresolved`: add `caseSensitiveStrict` option

### LINKED ISSUE #1259: Incorrect file resolution for paths higher than CWD
### Env:
Windows 7 x64, NodeJs v10.15.0, eslint 5.11.1, eslint-plugin-import: 2.14.0

### Repro steps: 
Rule **no-unresolved** is turned on with **case sensitive** option
``` javascript
"rules": {
        "import/no-unresolved": ["error", { caseSensitive: true}]
 }
```

Project root is **"B/"**
CWD is **"C/"**

Project Structure (tree)
```
├─ A
   ├─ B
   │  ├─ C
   │  │  ├─ D.js
```

``` javascript
import D from 'B/C/D.js'; // should be no error (casing is preserved correctly)
```

``` javascript
import D from 'B/C/d.js'; // causes error (filename is misspelled)
```

``` javascript
import D from 'B/c/D.js'; // causes error (cwd is misspelled)
```

**Expected result:**
``` javascript
import D from 'b/C/D.js'; // should cause error (project root is misspelled)
```

**Actual result:**
``` javascript
import D from 'b/C/D.js'; // no error (project root is misspelled)
```

### Description:
Eslint task is launched by grunt, where grunt config file is located in _**"C/"**_ folder (CWD). While actual project folder is one ore more levels higher than cwd (_**"B/"**_ ). So imports with such path don't cause error.

It is caused by  _"fileExistsWithCaseSync"_ function
https://github.com/benmosher/eslint-plugin-import/blob/1cd82eb27df85768fbd076e4ff6b7f36d6f652ce/utils/resolve.js#L38
So it basically doesn't go any higher than cwd to cache path.

**Suggestion:**
It either should not be limited at all (because it is relatively cheap operation as it caches previous paths), 
``` javascript
import D from 'a/B/C/D.js'; // should also cause error (it is higher than project root)
```
or at least it should be limited to project dir (something like getBaseDir() is doing - line 153)
``` javascript
import D from 'b/C/D.js'; // should cause error (project root is misspelled)
```

sergei-startsev:
Fix #1259 issue. The code above should be identified as invalid by `import/no-unresolved` rule with enabled `caseSensitiveStrict` option:

```js
/*eslint import/no-unresolved: [2, { caseSensitiveStrict: true }]*/

// Absolute paths
import Foo from `/Users/fOo/bar/file.js` // reported, /Users/foo/bar/file.js
import Foo from `d:/fOo/bar/file.js` // reported, d:/foo/bar/file.js

// Relative paths, cwd is Users/foo/
import Foo from `./../fOo/bar/file.js` // reported
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 
 ### Added
+- [`no-unresolved`]: add `caseSensitiveStrict` option ([#1262], thanks [@sergei-startsev])
 - [`no-unused-modules`]: add eslint v8 support ([#2194], thanks [@coderaiser])
 - [`no-restricted-paths`]: add/restore glob pattern support ([#2219], thanks [@stropho])
 
@@ -1053,6 +1054,7 @@ for info on changes for earlier releases.
 [#1294]: https://github.com/import-js/eslint-plugin-import/pull/1294
 [#1290]: https://github.com/import-js/eslint-plugin-import/pull/1290
 [#1277]: https://github.com/import-js/eslint-plugin-import/pull/1277
+[#1262]: https://github.com/import-js/eslint-plugin-import/pull/1262
 [#1257]: https://github.com/import-js/eslint-plugin-import/pull/1257
 [#1253]: https://github.com/import-js/eslint-plugin-import/pull/1253
 [#1248]: https://github.com/import-js/eslint-plugin-import/pull/1248
--- a/docs/rules/no-unresolved.md
+++ b/docs/rules/no-unresolved.md
@@ -76,10 +76,25 @@ By default, this rule will report paths whose case do not match the underlying f
 const { default: x } = require('./foo') // reported if './foo' is actually './Foo' and caseSensitive: true
 ```
 
+#### `caseSensitiveStrict`
+
+The `caseSensitive` option does not detect case for the current working directory. The `caseSensitiveStrict` option allows checking `cwd` in resolved path. By default, the option is disabled.
+
+
+```js
+/*eslint import/no-unresolved: [2, { caseSensitiveStrict: true }]*/
+
+// Absolute paths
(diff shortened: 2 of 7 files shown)


sergei-startsev:
@benmosher @ljharb Have a chance to take a look? The changes are behind the option, no breaking changes are expected.

benmosher:
Ah interesting, if just plain `caseSensitive` is not properly checking paths with relative bits, I think that is a bug and a new option is not necessary. @ljharb can check my math on that though. 

But my recommendation would be to make `caseSensitive` report on your case as well.

benmosher:
Feels like any explicit path segment should be checked, regardless of the CWD. Not sure how exactly that would work though.

sergei-startsev:
@benmosher right, that's how `caseSensitive` worked initially before ignoring `cwd` part.

ljharb:
@sergei-startsev sorry for the long delay here. if you'd mind rebasing, i'll be happy to take a fresh look.

sergei-startsev:
@ljharb The PR has been rebased, take a look.

ljharb:
@sergei-startsev what happens if someone sets `caseSensitive` to `false`, but `caseSensitiveStrict` to true? Same question if both are true?

sergei-startsev:
@ljharb Good point, rethought `caseSensitiveStrict`, it'll win in both cases now.

ljharb (approved):
Thanks!

sergei-startsev:
@ljharb Do you plan to merge the PR yourself?

ljharb:
@sergei-startsev yep! i'm holding off because i want to find a fix for #2199, so i can release a v2.24 patch release, before merging this in for a v2.25.

ljharb:
@sergei-startsev travis should be working again; i've rebased this branch so you can see the logs.

sergei-startsev:
@ljharb Not sure what went wrong last time, but the checks are green on all platforms now (thanks for fixing travis and adding appveyor). The commit with logs was reverted (not sure why codecov hasn't rerun).

PR #1271, opened 2019-01-27
Fix #1266, make SourceCode only for no-deprecated rule

### LINKED ISSUE #1266: v2.15.0 is leaking memory
After updating to v2.15.0 eslint started running out of memory on our fairly large project.
I was able to reproduce something similar when running the plugin against a dozen copies of [lodash-es](https://github.com/lodash/lodash/tree/es). I can create a repro case, but it's rather trivial.
One rule is enough to see the difference between v2.14.0 and v2.15.0:
```yml
plugins:
  - import

rules:
  import/default: 2
```

This seems to be caused by 05c3935048577bd7b025d6b833d8503807f02189. I've run some memory profiles, but I can't pinpoint the exact source of the leak other than that it seems to originate in [ExportMap.js](https://github.com/benmosher/eslint-plugin-import/blob/master/src/ExportMap.js). It's possible that the leak was always there, but extra parsing and memory allocations made it more evident.

sergei-startsev:
Parsing source code allocates a lot of memory since `SourceCode` instances are built recursively for each node in the dependency graph:
https://github.com/benmosher/eslint-plugin-import/blob/1ec80fa35fa1819e2d35a70e68fb6a149fb57c5e/src/ExportMap.js#L349

`SourceCode` instance is only used to capture docs from leading comments that are actually used **only** for [no-deprecated](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-deprecated.md) rule:
https://github.com/benmosher/eslint-plugin-import/blob/1ec80fa35fa1819e2d35a70e68fb6a149fb57c5e/src/ExportMap.js#L417-L418

https://github.com/benmosher/eslint-plugin-import/blob/1ec80fa35fa1819e2d35a70e68fb6a149fb57c5e/src/ExportMap.js#L453-L454

https://github.com/benmosher/eslint-plugin-import/blob/1ec80fa35fa1819e2d35a70e68fb6a149fb57c5e/src/ExportMap.js#L456-L459

However, `SourceCode` instances are built even if [no-deprecated](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-deprecated.md) rule is turned off (btw the rule is in Stage 0 and is still work in progress). The PR prevents negative impact on other rules.

Here's memory consumption before/after PR for pretty large code base, [no-deprecated](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-deprecated.md) rule is disabled for both cases:

image

The 1st process failed with `heap out of memory` error.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -192,7 +192,8 @@ export default class ExportMap {
 
 /**
  * parse docs from the first node that has leading comments
- * @param  {...[type]} nodes [description]
+ * @param  {SourceCode} source parsed source code
+ * @param  {object} docStyleParsers a set of available docs parses
  * @return {{doc: object}}
  */
 function captureDoc(source, docStyleParsers) {
@@ -202,9 +203,10 @@ function captureDoc(source, docStyleParsers) {
   // 'some' short-circuits on first 'true'
   nodes.some(n => {
     try {
+      let commentsBefore = source && source.getCommentsBefore(n)
       // n.leadingComments is legacy `attachComments` behavior
-      let leadingComments = n.leadingComments || source.getCommentsBefore(n)
-      if (leadingComments.length === 0) return false
+      let leadingComments = n.leadingComments || commentsBefore
+      if (!leadingComments || leadingComments.length === 0) return false
 
       for (let name in docStyleParsers) {
         const doc = docStyleParsers[name](leadingComments)
@@ -274,14 +276,14 @@ function captureTomDoc(comments) {
   }
 }
 
-ExportMap.get = function (source, context) {
+ExportMap.get = function (source, context, options) {
   const path = resolve(source, context)
   if (path == null) return null
 
-  return ExportMap.for(childContext(path, context))
+  return ExportMap.for(childContext(path, context), options)
 }
 
-ExportMap.for = function (context) {
+ExportMap.for = function (context, options) {
   const { path } = context
 
   const cacheKey = hashObject(context).digest('hex')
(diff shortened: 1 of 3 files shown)


src/ExportMap.js
```diff
@@ -346,7 +348,7 @@ ExportMap.parse = function (path, content, context) {
     docStyleParsers[style] = availableDocStyleParsers[style]
   })
 
-  const source = makeSourceCode(content, ast)
+  const source = context.id === 'no-deprecated' ? makeSourceCode(content, ast) : null
```
ljharb: Is there a way to make this metadata the rule can supply rather than hardcoding a single rule name?

sergei-startsev: yeah, I'll take a look

sergei-startsev:
@ljharb any ideas regarding `appveyor` build failure?:
```
Error: Command failed: git cat-file -p f71edacae52498377210e97c5540da50f80f3b2b
fatal: Not a valid object name f71edacae52498377210e97c5540da50f80f3b2b
```
I think I've already seen this in other PRs...

src/ExportMap.js
sergei-startsev: Passing an argument looks better, it seems rule `id` isn't passed in ESlint 2-3.

sergei-startsev: but more changes are required... 😓

ljharb: maybe instead of skipping the comments, we could make them lazy?

ie, instead of using properties, use functions that lazily create the source code object as needed (but memoize it across calls)?

sergei-startsev: I'm afraid it's too late, we don't have required data (content, ast) when we need it, otherwise we have to store them as well

benmosher:
Closing in deference to #1275

benmosher:
Thanks for taking a stab at this, though! 😄

Issue #1269, opened 2019-01-24
[Regression] eslint-module-utils@2.3.0 causes invalid "Resolve error: webpack with invalid interface loaded as resolver"


ronjouch:
See minimal reproduction case with script and pre-run output at [ronjouch/repro-eslint-bug-cra](https://github.com/ronjouch/repro-eslint-bug-cra) .

As demonstrated by the `demo` script, reverting to 2.2.0 inside eslint-plugin-import (`cd node_modules/eslint-plugin-import && npm install eslint-module-utils@2.2.0`) fixes the issue.

Reproduced and fixed on many machines (node 10.15.0 & npm 6.7.0 on Linux, node 8 on macOS, and in a CI Docker). I'm available for more debug info if necessary.

benmosher:
@ronjouch what version of `eslint-import-resolver-webpack` are you using? I don't see it referenced in either the `package-lock.json` or the `package.json`.

ronjouch:
> what version of `eslint-import-resolver-webpack` are you using? I don't see it referenced in either the `package-lock.json` or the `package.json`.

@benmosher thanks for the fast reply 👍. Uh, from what I understand, I'm not using `eslint-import-resolver-webpack`, neither in the minimal scenario, nor in the production web app I extracted it from. 

Here's an `npm ls | grep eslint` for the production web app:

```
├─┬ UNMET PEER DEPENDENCY babel-eslint@8.2.6
│ ├─┬ eslint-scope@3.7.1
│ └── eslint-visitor-keys@1.0.0
├─┬ eslint@4.19.1
│ ├── eslint-scope@3.7.1 deduped
│ ├── eslint-visitor-keys@1.0.0 deduped
├─┬ eslint-config-airbnb-base@13.1.0
│ ├── eslint-restricted-globals@0.1.1
├── eslint-config-react-app@2.1.0
├─┬ eslint-loader@1.9.0
├─┬ eslint-plugin-flowtype@2.39.1
├─┬ eslint-plugin-import@2.15.0
│ ├─┬ eslint-import-resolver-node@0.3.2
│ ├─┬ eslint-module-utils@2.2.0 invalid
├─┬ eslint-plugin-jsx-a11y@5.1.1
├─┬ eslint-plugin-react@7.4.0
npm ERR! peer dep missing: babel-eslint@^7.2.3, required by eslint-config-react-app@2.1.0
[...]

PR #1274, opened 2019-01-28
Patch to create version 2.13.1 to fix breaking dependency after eslint-module-utils ~2.2.0

econne01:
See issue #1269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+## [2.13.1] - 2019-01-28
+### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+## [2.13.1] - 2019-01-28
```
ljharb: version numbers aren't decided in PRs, typically :-)
```suggestion
```

econne01: Ah, ok thanks! I will update to make it more generic

benmosher:
@econne01 AFAICT there is no regression. #1269 was closed as a misunderstanding of config settings.

It is more loudly reporting config issues, though, it seems. I don't think this is a regression so I will close this PR and we can discuss further if you have a different repro.

PR #1283, opened 2019-02-11
Update import/parsers notes in README

### LINKED ISSUE #1273: Update docs about parsers
As typescript-eslint-parser has been superseded by @typescript-eslint/parser, the [docs about import statement parsers](https://github.com/benmosher/eslint-plugin-import/blob/master/README.md#importparsers) should be updated accordingly.

danielmoi:
Fixes #1273

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,31 +342,6 @@ settings:
         directory: "./path/to/folder"
 ```
 
-```json
-// .eslintrc
-{
-  "plugins": ["import"],
-  "rules": {
-    // turn on errors for missing imports
-    "import/no-unresolved": "error"
-  },
-  "settings": {
-    "import/parsers": {
-      "@typescript-eslint/parser": [".ts", ".tsx"]
-    },
-    "import/resolver": {
-      // use <root>/tsconfig.json
-      "typescript": {},
-
-      // use <root>/path/to/folder/tsconfig.json
-      "typescript": {
-        "directory": "./path/to/folder"
-      }
-    }
-  }
-}
-```
-
 In this case, [`eslint-import-resolver-typescript`](https://github.com/alexgorbatchev/eslint-import-resolver-typescript) must be installed and require-able from
 the running `eslint` module's location (i.e., install it as a peer of ESLint).
 
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ rules:
 
 You may use the following shortcut or assemble your own config using the granular settings described below.
 
-Make sure you have installed the [`@typescript-eslint/parser`] which is used in the following configuration. Unfortunately NPM does not allow to list optional peer dependencies.
+Make sure you have installed the [`@typescript-eslint/parser`] which is used in the following configuration. Unfortunately `npm` does not support optional peer dependencies.
 
 ```yaml
 extends:
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ rules:
 
 You may use the following shortcut or assemble your own config using the granular settings described below.
 
-Make sure you have installed the [`@typescript-eslint/parser`] which is used in the following configuration. Unfortunately NPM does not allow to list optional peer dependencies. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -148,7 +148,7 @@ rules:
 
 You may use the following shortcut or assemble your own config using the granular settings described below.
 
-Make sure you have installed the [`@typescript-eslint/parser`] which is used in the following configuration. Unfortunately NPM does not allow to list optional peer dependencies. 
+Make sure you have installed the [`@typescript-eslint/parser`] which is used in the following configuration. Unfortunately NPM does not allow to list optional peer dependencies.
```
ljharb:
```suggestion
Make sure you have installed the [`@typescript-eslint/parser`] which is used in the following configuration. Unfortunately `npm` does not support optional peer dependencies.
```

johman10:
Great that someone is updating the documentation on this! It helped me solve my case, but while I was trying to figure this out, I found out that all you have to do is this:
 - Run: `npm install --save-dev eslint-import-resolver-typescript`
 - Change my `eslintrc.js` to include:
```js
settings: {
  'import/resolver': {
    typescript: {}
  }
}
```

I don't think this is very obvious from the documentation in this PR right now.

ljharb:
@johman10 it'd be great if you reviewed the PR and made a suggestion

johman10:
The config mentioned in the typescript part of the README doesn't work for me. I made a base project with just 2 typescript files and exactly these `extends` values in my `eslintrc` but that doesn't work (see code below).

I think this points to either a problem somewhere else in the project, or a documentation issue. If this project doesn't really support typescript I would suggest pointing to [eslint-import-resolver-typescript](https://github.com/alexgorbatchev/eslint-import-resolver-typescript#readme) to let them explain the typescript support instead and remove all the typescript related code in this project.

If this project is trying to support typescript there should probably be a closer look at the support you offer and whether that all still works. I have been unable to figure out where this issue is coming from for now.

@ljharb what is your view on this? Are you trying to support typescript from this plugin?

**Output**
image

**Command used**
`eslint --ext js,ts .`

**Dependencies**
```
@typescript-eslint/eslint-plugin: 1.4.2
@typescript-eslint/parser: 1.4.2
eslint-config-airbnb-base: 13.1.0
eslint-plugin-import: 2.16.0
eslint: 5.15.1
```

**Code**
`index.ts`
```ts
import module from './module';

module();
```

`module.ts`
```ts
export default function () {
  console.log('Hello from module.ts');
}
```

`eslintrc.js`
```js
module.exports = {
  env: {
    browser: true,
    es6: true,
  },
  extends: [
    'eslint:recommended',
    'plugin:import/errors',
    'plugin:import/warnings',
    'plugin:import/typescript'
  ],
  parserOptions: {
    ecmaVersion: 2018,
    sourceType: 'module',
  },
};
```

README.md
```diff
@@ -333,13 +333,21 @@ directly using webpack, for example:
 # .eslintrc.yml
 settings:
   import/parsers:
-    typescript-eslint-parser: [ .ts, .tsx ]
+    @typescript-eslint/parser: [ .ts, .tsx ]
```
johman10: I don't think this is valid YAML. It gets confused because of the `@`, so you would have to make sure this key is a string by doing:
```suggestion
    '@typescript-eslint/parser': [ .ts, .tsx ]
```

ljharb:
Yes, this plugin does and will support typescript. That's not the same thing as us having the expertise to fix the docs, or the implementation, or to figure out where the problem lies.

gretzky:
I'm getting the same errors as @johman10, tried adding resolvers as well and no dice.

Primajin:
When using `'@typescript-eslint/parser'` I get tons of `import/named` errors while `'typescript-eslint-parser'` worked totally fine. So I can't upgrade to the new name / version until this is solved here.

benmosher:
@johman10 / @gretzky I think the TS config is updated to reference the modern TS parser in the latest version.

Closing because I think we got these updates in a later PR, thanks though everyone!

PR #1112, opened 2018-05-30
doc(no-relative-parent-imports): rename title

feychenie:
Renamed the title of this page to be the full name of the rule, like on the rest of the docs.

PR #1287, opened 2019-02-13
docs: fix no-default-export rule name

tiii:
Change rule name form `no-default-export` to `import/no-default-export` in documentation.


golopot:
Somehow this is already fixed in another commit. [link to doc](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-default-export.md).

ljharb:
Indeed; this seems covered by #1112.

PR #1290, opened 2019-02-18
No useless index in import paths

### LINKED ISSUE #394: Proposal: disallow "index" in import paths
Importing the directory will resolve to index, so including index in your path is unnecessary. I think this would make a good rule for this project, what do you think?

Bad:

```js
import foo from 'foo/index';
```

Good:

```js
import foo from 'foo';
```

timkraut:
As discussed in #394, this PR adds an option `noUselessIndex` to detect (and fix) paths containing unnecessary `index` parts (with or without file extension):

Given the following file structure:
```
my-project
├── foo.js
└── foo
    └── index.js
└── bar
    └── index.js
```

Before:
```
import 'foo/index' // No warning
import 'bar/index' // No warning
```

After:
```
import './foo/index.js' // Warning (fixable to `./foo/`)
import './foo/index' // Warning (fixable to `./foo/`)
import './bar/index' // Warning (fixable to `./bar`)
```

#### Implementation detail:
I've replaced the `sumBy` function with `reduce` to get rid of Lodash (not necessary in this case IMO). A benchmark which compares these 2 alternatives shows no difference between them or a slight advantage for `reduce`: https://www.measurethat.net/Benchmarks/ShowResult/46787

Fixes #394.

### DIFF
--- a/docs/rules/no-useless-path-segments.md
+++ b/docs/rules/no-useless-path-segments.md
@@ -11,6 +11,9 @@ my-project
 ├── app.js
 ├── footer.js
 ├── header.js
+└── helpers.js
+└── helpers
+    └── index.js
 └── pages
     ├── about.js
     ├── contact.js
@@ -30,6 +33,8 @@ import "../pages/about.js"; // should be "./pages/about.js"
 import "../pages/about"; // should be "./pages/about"
 import "./pages//about"; // should be "./pages/about"
 import "./pages/"; // should be "./pages"
+import "./pages/index"; // should be "./pages" (except if there is a ./pages.js file)
+import "./pages/index.js"; // should be "./pages" (except if there is a ./pages.js file)
 ```
 
 The following patterns are NOT considered problems:
@@ -46,3 +51,25 @@ import ".";
 import "..";
 import fs from "fs";
 ```
+
+## Options
+
+### noUselessIndex
+
+If you want to detect unnecessary `/index` or `/index.js` (depending on the specified file extensions, see below) imports in your paths, you can enable the option `noUselessIndex`. By default it is set to `false`:
+```js
+"import/no-useless-path-segments": ["error", {
+  noUselessIndex: true,
+}]
+```
+
+Additionally to the patterns described above, the following imports are considered problems if `noUselessIndex` is enabled:
+
+```js
+// in my-project/app.js
+import "./helpers/index"; // should be "./helpers/" (not auto-fixable to `./helpers` because this would lead to an ambiguous import of `./helpers.js` and `./helpers/index.js`)
+import "./pages/index"; // should be "./pages" (auto-fixable)
+import "./pages/index.js"; // should be "./pages" (auto-fixable)
+```
+
+Note: `noUselessIndex` only avoids ambiguous imports for `.js` files if you haven't specified other resolved file extensions. See [Settings: import/extensions](https://github.com/benmosher/eslint-plugin-import#importextensions) for details.
--- a/src/rules/no-useless-path-segments.js
+++ b/src/rules/no-useless-path-segments.js
@@ -3,6 +3,7 @@
  * @author Thomas Grainger
  */
 
+import { getFileExtensions } from 'eslint-module-utils/ignore'
 import moduleVisitor from 'eslint-module-utils/moduleVisitor'
 import resolve from 'eslint-module-utils/resolve'
 import path from 'path'
@@ -31,9 +32,9 @@ function normalize(fn) {
   return toRelativePath(path.posix.normalize(fn))
 }
 
-const countRelativeParents = (pathSegments) => pathSegments.reduce(
-  (sum, pathSegment) => pathSegment === '..' ? sum + 1 : sum, 0
-)
+function countRelativeParents(pathSegments) {
+  return pathSegments.reduce((sum, pathSegment) => pathSegment === '..' ? sum + 1 : sum, 0)
+}
 
 module.exports = {
   meta: {
@@ -47,33 +48,28 @@ module.exports = {
         type: 'object',
         properties: {
           commonjs: { type: 'boolean' },
+          noUselessIndex: { type: 'boolean' },
         },
         additionalProperties: false,
       },
     ],
 
     fixable: 'code',
-    messages: {
-      uselessPath: 'Useless path segments for "{{ path }}", should be "{{ proposedPath }}"',
-    },
   },
 
   create(context) {
     const currentDir = path.dirname(context.getFilename())
-    const config = context.options[0]
+    const options = context.options[0]
 
     function checkSourceValue(source) {
       const { value: importPath } = source
 
-      function report(proposedPath) {
+      function reportWithProposedPath(proposedPath) {
         context.report({
           node: source,
-          messageId: 'uselessPath',
-          data: {
-            path: importPath,
-            proposedPath,
-          },
-          fix: fixer => fixer.replaceText(source, JSON.stringify(proposedPath)),
+          // Note: Using messageIds is not possible due to the support for ESLint 2 and 3
+          message: `Useless path segments for "${importPath}", should be "${proposedPath}"`,
+          fix: fixer => proposedPath && fixer.replaceText(source, JSON.stringify(proposedPath)),
         })
       }
 
@@ -87,7 +83,28 @@ module.exports = {
       const normedPath = normalize(importPath)
       const resolvedNormedPath = resolve(normedPath, context)
       if (normedPath !== importPath && resolvedPath === resolvedNormedPath) {
-        return report(normedPath)
+        return reportWithProposedPath(normedPath)
+      }
+
+      const fileExtensions = getFileExtensions(context.settings)
+      const regexUnnecessaryIndex = new RegExp(
+        `.*\\/index(\\${Array.from(fileExtensions).join('|\\')})?$`
+      )
+
+      // Check if path contains unnecessary index (including a configured extension)
+      if (options && options.noUselessIndex && regexUnnecessaryIndex.test(importPath)) {
+        const parentDirectory = path.dirname(importPath)
+
+        // Try to find ambiguous imports
+        if (parentDirectory !== '.' && parentDirectory !== '..') {
+          for (let fileExtension of fileExtensions) {
+            if (resolve(`${parentDirectory}${fileExtension}`, context)) {
+              return reportWithProposedPath(`${parentDirectory}/`)
+            }
+          }
+        }
+
+        return reportWithProposedPath(parentDirectory)
       }
 
       // Path is shortest possible + starts from the current directory --> Return directly
@@ -113,7 +130,7 @@ module.exports = {
       }
 
       // Report and propose minimal number of required relative parents
-      return report(
+      return reportWithProposedPath(
         toRelativePath(
           importPathSplit
             .slice(0, countExpectedRelativeParents)
@@ -123,6 +140,6 @@ module.exports = {
       )
     }
 
-    return moduleVisitor(checkSourceValue, config)
+    return moduleVisitor(checkSourceValue, options)
   },
 }
--- a/src/rules/no-useless-path-segments.js
+++ b/src/rules/no-useless-path-segments.js
@@ -3,10 +3,9 @@
  * @author Thomas Grainger
  */
 
-import path from 'path'
-import sumBy from 'lodash/sumBy'
-import resolve from 'eslint-module-utils/resolve'
 import moduleVisitor from 'eslint-module-utils/moduleVisitor'
+import resolve from 'eslint-module-utils/resolve'
+import path from 'path'
 import docsUrl from '../docsUrl'
 
 /**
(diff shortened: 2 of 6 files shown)


timkraut:
@ljharb I've updated the PR to use an option `detectUselessIndex` to detect these issue. The rule tries to find ambiguous imports for all extensions provided in:
```js
    'import/resolver': {
      'node': {
        'extensions': [
          // some extensions
        ],
      },
    },
```

If this setting is not set, `['.js']` will be assumed.

Let me know if that looks like a reasonable approach to you. I'm totally open for different ideas.

src/rules/no-useless-path-segments.js
```diff
@@ -45,65 +62,99 @@ module.exports = {
...
+      // Report rule violation if path is not the shortest possible
+      const resolvedPath = resolve(importPath, context)
+      const normedPath = normalize(importPath)
+      const resolvedNormedPath = resolve(normedPath, context)
+      if (normedPath !== importPath && resolvedPath === resolvedNormedPath) {
+        return report(normedPath)
+      }
+
+      // Path contains unnecessary /index
+      if (options && options.detectUselessIndex && importPath.endsWith('index')) {
```
ljharb: should this be checking importPath, or something else? what happens if they’ve specified `index.js`?

src/rules/no-useless-path-segments.js
```diff
@@ -45,65 +62,99 @@ module.exports = {
...
         context.report({
           node: source,
-          message: `Useless path segments for "${value}", should be "${proposed}"`,
-          fix: fixer => fixer.replaceText(source, JSON.stringify(proposed)),
+          messageId: 'uselessPath',
+          data: {
+            path: importPath,
+            proposedPath,
+          },
+          fix: fixer => proposedPath && fixer.replaceText(source, JSON.stringify(proposedPath)),
```
ljharb: i think that when proposedPath is not present, `fix` shouldn’t be either

src/rules/no-useless-path-segments.js
```diff
@@ -19,19 +18,37 @@ import docsUrl from '../docsUrl'
...
-  return toRel(path.posix.normalize(fn))
+  return toRelativePath(path.posix.normalize(fn))
 }
 
-const countRelParent = x => sumBy(x, v => v === '..')
+function countRelativeParents(pathSegments) {
+  return pathSegments.reduce((sum, pathSegment) => pathSegment === '..' ? sum + 1 : sum, 0)
+}
+
+function getFileExtensions(settings) {
```
ljharb: this seems like a useful helper to share in multiple rules (probably one already defined somewhere)

src/rules/no-useless-path-segments.js
timkraut: What do you mean by "should this be checking importPath, or something else?"?

I've added a test for `../index.js`. Importing something like `../index.js` doesn't cause any warnings. The [extensions](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md) rule can find those issues. Apparently, the `extension` rule is not able to auto-fix violations. Fixing `../index.js` to be `../` would require 2 steps if both rules are enabled.

src/rules/no-useless-path-segments.js
timkraut: I've updated the implementation for this case and added a better output handling:
- `fix` is only set if there is a possible fix
- If the auto-fix is not possible because of an ambiguous path, a dedicated message which contains the colliding paths is shown

src/rules/no-useless-path-segments.js
timkraut: Good point! I've found a function to do that - I only had to export it. I've also updated the `README.md` section to clarify how to specify the list of file extensions.

docs/rules/no-useless-path-segments.md
```diff
@@ -46,3 +50,24 @@ import ".";
...
+
+Additionally to the patterns described above, the following imports are considered problems if `detectUselessIndex` is enabled:
+
+```js
+// in my-project/app.js
+import "./helpers/index"; // should be "./helpers" (not auto-fixable because of helpers.js)
+import "./pages/index"; // should be "./pages" (auto-fixable)
+```
+
+Note: `detectUselessIndex` detects only ambiguous imports for `.js` files if you haven't specified all resolved file extensions. See [Settings: import/extensions](https://github.com/benmosher/eslint-plugin-import#importextensions) for details.
```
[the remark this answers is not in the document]
timkraut: Hm... that's difficult to implement. Currently, I loop through all specified extensions for every path which ends in `/index` to make sure there is no ambiguous import (except if the parent directory is `.` or `..` as it's not possible to have ambiguous imports in these cases). If you have an idea how we could do this efficiently for arbitrary file extensions, let me know.

.gitignore
```diff
@@ -39,3 +39,6 @@ lib/
 yarn.lock
 package-lock.json
 npm-shrinkwrap.json
+
+# IDEs
+.idea/
```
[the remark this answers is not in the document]
timkraut: That's a great idea. Never thought about this one. I've updated the PR to exclude this change.

src/rules/no-useless-path-segments.js
ljharb: That means that if a user has `.js` required, they can't get `index.js` auto-removed?

timkraut: Yes, exactly. They would get a warning but the `extension` rule currently is not able to auto-fix those things.

ljharb: It seems like the extensions rule should be unrelated here; any path that could be shortened should be caught by this rule.

timkraut: Ok, I'll add this feature then. Would you consider it part of the `detectUselessIndex` option or add a new option for this?

timkraut: I've updated the PR to find and fix `index.js` as well (depending on the configured file extensions, other file extensions would be found, too). I haven't added a new option for this. Let me know if you prefer an additional option.

ljharb: I think it should be part of `detectUselessIndex`, yes.

docs/rules/no-useless-path-segments.md
```diff
@@ -46,3 +50,24 @@ import ".";
...
+"import/no-useless-path-segments": ["error", {
+  detectUselessIndex: true,
+}]
+```
+
+Additionally to the patterns described above, the following imports are considered problems if `detectUselessIndex` is enabled:
+
+```js
+// in my-project/app.js
+import "./helpers/index"; // should be "./helpers" (not auto-fixable because of helpers.js)
```
ljharb: this should be autofixable to `./helpers/`

docs/rules/no-useless-path-segments.md
```diff
@@ -46,3 +50,24 @@ import ".";
...
+import "./pages/index"; // should be "./pages" (auto-fixable)
+```
```
ljharb: can we include an example that has `index.js` as well?

docs/rules/no-useless-path-segments.md
timkraut: Ah. I didn't even know that there is a difference between `./helpers/` --> `./helpers/index.js` and `./helpers` --> `./helpers.js`. 

Just to make sure we don't miss an option, these are all possible import paths:
```
./helpers.js --> ./helpers.js
./helpers --> ./helpers.js
./helpers/ --> ./helpers/index.js
./helpers/index --> ./helpers/index.js
./helpers/index/ --> ./helpers/index/index.js
./helpers/index.js --> ./helpers/index.js
```

If I'm not mistaken, this means that there is no option for ambiguous imports. Either it's an `index.js` file in an directory (in this case, we have a trailing `/`) or it's a file (in this case, we have no trailing `/`). Am I missing something?

ljharb: Yes.

In the presence of both `helpers.js` and `helpers/index.js`:
```sh
./helpers.js → ./helpers.js
./helpers → ./helpers.js (ambiguous)
./helpers/ → ./helpers/index.js
./helpers/index.js → ./helpers/index.js
./helpers/index → ./helpers/index.js
/.helpers/index/ → ./helpers/index/index.js (error, nonexistent here)
```

When there is only `helpers.js`:
```sh
./helpers.js → ./helpers.js
./helpers → ./helpers.js
./helpers/ → ./helpers/index.js (error, nonexistent here)
./helpers/index.js → ./helpers/index.js (error, nonexistent here)
./helpers/index → ./helpers/index.js (error, nonexistent here)
/.helpers/index/ → ./helpers/index/index.js (error, nonexistent here)
```

When there is only `helpers/index.js`:
```sh
./helpers.js → ./helpers.js  (error, nonexistent here)
./helpers → ./helpers/index.js
./helpers/ → ./helpers/index.js
./helpers/index.js → ./helpers/index.js
./helpers/index → ./helpers/index.js
/.helpers/index/ → ./helpers/index/index.js (error, nonexistent here)
```

timkraut: Thank you very much! That helped a lot!

I'll extend the list by the expected result:

### `helpers.js` and `helpers/index.js` exist
Import path (`import '...'`) | Resolves to | Shortest possible path? (Explanation)
-- | -- | --
`./helpers.js` | `./helpers.js` | ✅ (changing the path to `./helpers` would make it ambiguous)
`./helpers` | `./helpers.js` or `./helpers/index.js` (**ambiguous**) | ✅ (path is ambiguous but contains no useless path segments)
`./helpers/` | `./helpers/index.js` | ✅ (changing the path to `./helpers` would make it ambiguous)
`./helpers/index.js` | `./helpers/index.js` | ❌ (change to `./helpers/` possible)
`./helpers/index` | `./helpers/index.js` | ❌ (change to `./helpers/` possible)
`./helpers/index/` | `./helpers/index/index.js` | ~~(Not existent)~~

### Only `helpers.js` exists:
Import path (`import '...'`) | Resolves to | Shortest possible path? (Explanation)
-- | -- | --
`./helpers.js` | `./helpers.js` | ❌ (change to `./helpers` possible)
`./helpers` | `./helpers.js` | ✅ 
`./helpers/` | `./helpers.js` | ❌ (change to `./helpers` possible)
`./helpers/index.js` | `./helpers/index.js` | ~~(Not existent)~~
`./helpers/index` | `./helpers/index.js` | ~~(Not existent)~~
`./helpers/index/` | `./helpers/index/index.js` | ~~(Not existent)~~

### Only `helpers/index.js` exists:
Import path (`import '...'`) | Resolves to | Shortest possible path? (Explanation)
-- | -- | --
`./helpers.js` | `./helpers.js` | ❌ (change to `./helpers` possible)
`./helpers` | `./helpers.js` | ✅ 
`./helpers/` | `./helpers.js` | ❌ (change to `./helpers` possible)
`./helpers/index.js` | `./helpers/index.js` | ❌ (change to `./helpers` possible)
`./helpers/index` | `./helpers/index.js` | ❌ (change to `./helpers` possible)
`./helpers/index/` | `./helpers/index/index.js` | ~~(Not existent)~~

I'll add tests for all these cases and adapt the implementation to that.

timkraut:
I've changed the name of the new option from `detectUselessIndex` to `noUselessIndex` and rebased the branch to the latest `master` commit.

timkraut:
I've updated the PR to include tests for `getFileExtensions`. As this function returns a `Set`, I've updated `chai.js` to the latest version to test for that. This resulted in failing tests which I've fixed as well. See https://github.com/chaijs/chai/issues/781 for the breaking changes (in the case of this plugin, I had to rename `.deep` to `.nested`)

timkraut:
I'm struggling a little bit with the support for ESLint 2 and 3. I've fixed the issue that `message` is missing which resulted in failing tests. ESLint 4 und 5 seem to understand the `messageId` property. It looks like the failing tests are mostly due to the TypeScript support which isn't working in ESLint 2 and 3 apparently.

I get messages like the following:
```js
errors:
   [ TypeError: traverser.traverse is not a function
         at Object.parseForESLint (/Users/timkraut/Development/eslint-plugin-import/node_modules/typescript-eslint-parser/parser.js:41:15)
         at Object.exports.parse (/Users/timkraut/Development/eslint-plugin-import/node_modules/typescript-eslint-parser/parser.js:79:17)
         at parse (/Users/timkraut/Development/eslint-plugin-import/utils/parse.js:38:17)
         at Function.ExportMap.parse (/Users/timkraut/Development/eslint-plugin-import/src/ExportMap.js:339:15)
         at Function.ExportMap.for (/Users/timkraut/Development/eslint-plugin-import/src/ExportMap.js:323:25)
         at Function.ExportMap.get (/Users/timkraut/Development/eslint-plugin-import/src/ExportMap.js:286:10)
         at Context.<anonymous> (/Users/timkraut/Development/eslint-plugin-import/tests/src/core/getExports.js:332:31)
         at callFn (/Users/timkraut/Development/eslint-plugin-import/node_modules/mocha/lib/runnable.js:348:21)
         at Hook.Runnable.run (/Users/timkraut/Development/eslint-plugin-import/node_modules/mocha/lib/runnable.js:340:7)
         at next (/Users/timkraut/Development/eslint-plugin-import/node_modules/mocha/lib/runner.js:309:10)
         at Immediate.<anonymous> (/Users/timkraut/Development/eslint-plugin-import/node_modules/mocha/lib/runner.js:339:5)
         at processImmediate (timers.js:632:19) ],
```

Do you have an idea how I might fix those?

docs/rules/no-useless-path-segments.md
ljharb: I've updated your table for two cases, otherwise looks good.

ljharb:
@timkraut i've pulled out your chai change into a separate commit on master (it'll be pushed shortly)

If older eslints don't support messageId, let's not use it yet.

timkraut:
Is there anything else I can do to get this merged/released?

timkraut:
@ljharb Could you release this, too? Currently the documentation states something which isn't possible with the release from january.

ljharb:
Nope, only one person can release this package.

PR #1294, opened 2019-02-21
closes #1293. allows aliases that start with @ to be caught as internal

### LINKED ISSUE #1293: [import/order] Aliases that start with @ get incorrectly grouped
Given that Typescript is going to switch over to ESLint as its main linting tool, I wanted to try to get my project at work switched over. In Typescript, you can define paths. Paths allow you to define import paths, i.e. `@my-alias/fn` could actually point to `src/my-folder/fn.ts`. 

In my project, currently using TSLint, we have external, aliases, parent, sibling as our import sort order. After trying to get a test case working with ESLint, it seems like this plugin incorrectly associates aliases that start with `@` to external modules, since it thinks they are "scoped".

I don't know the solution to this, however, I would be open to putting in a PR if someone can provide some insight into solving this? 

I have tried to create a minimal repro, but it seems like code-sandbox won't give access to the terminal... But this example does fail, saying that you can't have newlines between groups on the lodash import. https://codesandbox.io/s/7m5zpmxo90

jeffshaver:
@ljharb as we talked about in the issue, here is the fix. Please let me know if there are any issues. Or if you think we need more tests for this case. I added one. There were already a ton that didn't pass and didn't think I should fix them in this PR.

Fixes #1293. Unblocked by #1295.

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -57,7 +57,8 @@ export function isScopedMain(name) {
 }
 
 function isInternalModule(name, settings, path) {
-  return externalModuleRegExp.test(name) && !isExternalPath(path, name, settings)
+  const matchesScopedOrExternalRegExp = scopedRegExp.test(name) || externalModuleRegExp.test(name)
+  return (matchesScopedOrExternalRegExp && !isExternalPath(path, name, settings))
 }
 
 function isRelativeToParent(name) {
@@ -76,9 +77,9 @@ function isRelativeToSibling(name) {
 const typeTest = cond([
   [isAbsolute, constant('absolute')],
   [isBuiltIn, constant('builtin')],
+  [isInternalModule, constant('internal')],
   [isExternalModule, constant('external')],
   [isScoped, constant('external')],
-  [isInternalModule, constant('internal')],
   [isRelativeToParent, constant('parent')],
   [isIndex, constant('index')],
   [isRelativeToSibling, constant('sibling')],
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -50,6 +50,11 @@ describe('importType(name)', function () {
     const pathContext = testContext({ "import/resolver": { node: { paths: [ path.join(__dirname, '..', '..', 'files') ] } } })
     expect(importType('@importType/index', pathContext)).to.equal('internal')
   })
+    
+  it("should return 'internal' for internal modules that are referenced by aliases", function () {
+    const pathContext = testContext({ 'import/resolver': { node: { paths: [path.join(__dirname, '..', '..', 'files')] } } })
+    expect(importType('@my-alias/fn', pathContext)).to.equal('internal')
+  })
 
   it("should return 'parent' for internal modules that go through the parent", function() {
     expect(importType('../foo', context)).to.equal('parent')


src/core/importType.js
```diff
@@ -8,8 +8,8 @@ function constant(value) {
   return () => value
 }
 
-function baseModule(name) {
-  if (isScoped(name)) {
+function baseModule(name, path) {
+  if (isScoped(name, path)) {
```
ljharb: i don't think it makes sense to make `isScoped` check `isExternalPath`; i think all the places currently calling `isScoped` should be explicitly handling `isExternalPath` first.

jeffshaver: I can do that. It does seem that "scoped" packages have a connotation meaning "external". So I thought it would be fine putting it in `isScoped` because scoped modules have to be external. Will update shortly

jeffshaver: Actually... I don't even know how to solve this? Move the `isScoped` check down underneath `isInternalModule`in the lodash.cond call? The problem I am seeing is that isScoped means external, but isExternalModule doesn't check it. It is checked by the lodash.cond call

jeffshaver: Seems like moving the internal check above scoped and external solved that problem and let me revert most of the other changes

jeffshaver:
@ljharb updated per your comments. Appreciate you looking at this!

src/core/importType.js
ljharb: Scoped packages are the same as anything in node_modules; they're things installed from npm. I don't think "scoped" means "external".

jeffshaver:
@ljharb anything I can do to help unbreak master?

ljharb:
@jeffshaver open another PR that fixes the tests? :-D

jeffshaver:
@ljharb. I'll take a look later today to see if I can fix some

jeffshaver:
@ljharb since my other pr got merged (thanks! sorry we couldn't fix that one test), i rebased my branch with upstream master. just wanted to let you know

ljharb:
@jeffshaver oh we fixed it! thanks for the PR, and the ping.

src/core/importType.js
```diff
@@ -76,9 +77,9 @@ function isRelativeToSibling(name) {
 const typeTest = cond([
   [isAbsolute, constant('absolute')],
   [isBuiltIn, constant('builtin')],
+  [isInternalModule, constant('internal')],
   [isExternalModule, constant('external')],
   [isScoped, constant('external')],
-  [isInternalModule, constant('internal')],
```
ljharb: This looks good, but there is the possibility this will be a breaking change - however, since it doesn't break any tests, and since it's unlikely anyone is depending on the *inability* to mark an external module as "internal", it's probably fine.

jeffshaver:
@ljharb appreciate it! When do you think a new version will be released with these changes?

ljharb:
Probably not super quickly; ping me again in a week if it’s not done by then.

laozhu:
Have the PR been released? My '@' alias imports are still 'unknown' type with the latest eslint-plugin-import.

ljharb:
@laozhu yes, if you click on the commit https://github.com/benmosher/eslint-plugin-import/commit/651829d6a2862b8fa2e33426421a5dba526e4cdb you can see it's in 2.17.0 and higher.

Issue #1034, opened 2018-02-28
Local imports using webpack aliases are being incorrectly identified as separate groups


cheapsteak:
Using a webpack config using aliases (and eslint-import-resolver-webpack), each import for an internal modules is being treated as a separate "group". With `'newlines-between': 'always-and-inside-groups'`, they each have to be separated by a new line

```js
resolve: {
  alias: {
    '~': path.resolve(__dirname, 'src'),
  },
},
```

```js
import React from 'react';

import ComponentA from '~/components/ComponentA'; // [eslint] There should be at least one empty line between import groups (import/order)
import ComponentB from '~/components/ComponentB'; // [eslint] There should be at least one empty line between import groups (import/order)
import ComponentC from '~/components/ComponentC'; // no error

import ComponentD from '~/components/ComponentD'; // no error
```

Reproduced here: https://github.com/cheapsteak/eslint-import-order-bug-reproduction
Pull locally and check the lint messages for this file - https://github.com/cheapsteak/eslint-import-order-bug-reproduction/blob/master/src/App.js

fsmaia:
I'm having a different issue, but I think they are related.

In my case, I have some aliases set in webpack, and resolver is properly configured:
```js
resolve: {
  alias: {
    actions: path.join(__dirname, './src/actions'),
    components: path.join(__dirname, './src/components'),
    constants: path.join(__dirname, './src/constants'),
    //...
  },
},
```

And `import/order` is configured with:
```json
{
  "groups": ["builtin", "external", "internal", "parent", "sibling", "index"],
  "newlines-between": "always"
}
```

I'm trying to import a external followed by internal module:
```js
import { createSelector } from 'reselect';

import { MY_CONSTANT_EXAMPLE } from 'constants/my-constant';
```

However, I'm getting this error:
> [eslint] `constants/my-constant` import should occur before import of `reselect` (import/order)

The example above works fine if I replace `constants` with `src/constants`.

laugri:
Hi @cheapsteak , did you find a workaround since you ran into this issue ?

cheapsteak:
Unfortunately not

doits:
+1 named imports should be all in the internal group

echenley:
@fsmaia I am having the exact same issue with a `constants` folder in particular. All of my other internal aliases seem to work.

echenley:
@ljharb It seems like `isBuiltIn` needs to be updated to check for a resolved path:

[...]

PR #1297, opened 2019-03-06
fix: aliased internal modules that look like core modules

echenley:
## What

Update `isBuiltIn` to return `false` if a path is returned from a resolver. 
Previously an alias like 'constants/foo' would be classified as a builtin rather than internal since 'constants' is a core module of node. Fixes one of the issues mentioned in #1034.

Update webpack resolver to attempt to resolve modules internally before classifying them as builtin.

Add several tests to cover these cases.

## Example

```
project
└── src
    └── constants
        └── index.js
```

```json
{
  "settings": {
    "import/resolver": {
      "node": {
        "paths": ["/project/src"]
      }
    }
  }
}
```

Result:

```js
// BEFORE
import { FOO } from 'constants/index'
import { isObject } from 'lodash'

// AFTER
import { isObject } from 'lodash'
import { FOO } from 'constants/index'
```

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -21,7 +21,9 @@ export function isAbsolute(name) {
   return name.indexOf('/') === 0
 }
 
-export function isBuiltIn(name, settings) {
+// path is defined only when a resolver resolves to a non-standard path
+export function isBuiltIn(name, settings, path) {
+  if (path) return false
   const base = baseModule(name)
   const extras = (settings && settings['import/core-modules']) || []
   return coreModules[base] || extras.indexOf(base) > -1
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -70,8 +70,7 @@ describe('importType(name)', function () {
     const pathContext = testContext({ 'import/resolver': { webpack: { config: webpackConfig } } })
     expect(importType('constants/index', pathContext)).to.equal('internal')
     expect(importType('constants/', pathContext)).to.equal('internal')
-    // the following assertion fails because the webpack resolver runs an resolve.isCore('constants') without first checking paths config
-    // expect(importType('constants', pathContext)).to.equal('internal')
+    expect(importType('constants', pathContext)).to.equal('internal')
   })
 
   it("should return 'parent' for internal modules that go through the parent", function() {
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -7,6 +7,7 @@ import { testContext } from '../utils'
 
 describe('importType(name)', function () {
   const context = testContext()
+  const pathToTestFiles = path.join(__dirname, '..', '..', 'files')
 
   it("should return 'absolute' for paths starting with a /", function() {
     expect(importType('/', context)).to.equal('absolute')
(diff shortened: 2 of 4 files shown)


tests/src/core/importType.js
```diff
@@ -42,20 +43,35 @@ describe('importType(name)', function () {
...
-    const pathContext = testContext({ 'import/resolver': { node: { paths: [path.join(__dirname, '..', '..', 'files')] } } })
+    const pathContext = testContext({ 'import/resolver': { node: { paths: [pathToTestFiles] } } })
     expect(importType('@my-alias/fn', pathContext)).to.equal('internal')
   })
 
+  it("should return 'internal' for aliased internal modules that look like core modules (node resolver)", function () {
+    const pathContext = testContext({ 'import/resolver': { node: { paths: [pathToTestFiles] } } })
+    expect(importType('constants/index', pathContext)).to.equal('internal')
+    // the following assertion fails because the node resolver runs an resolve.isCore('constants') without first checking paths config
+    // expect(importType('constants', pathContext)).to.equal('internal')
```
echenley: With my changes, this will still return `'builtin'` unless changes are made to the node resolver.

tests/src/core/importType.js
```diff
@@ -42,20 +43,35 @@ describe('importType(name)', function () {
...
+  })
+
+  it("should return 'internal' for aliased internal modules that look like core modules (webpack resolver)", function () {
+    const webpackConfig = { resolve: { modules: [pathToTestFiles, 'node_modules'] } }
+    const pathContext = testContext({ 'import/resolver': { webpack: { config: webpackConfig } } })
+    expect(importType('constants/index', pathContext)).to.equal('internal')
+    // the following assertion fails because the webpack resolver runs an resolve.isCore('constants') without first checking paths config
+    // expect(importType('constants', pathContext)).to.equal('internal')
```
echenley: With my changes, this will still return `'external'` unless changes are made to the webpack resolver.

tests/src/core/importType.js
echenley: In particular: https://github.com/benmosher/eslint-plugin-import/blob/master/resolvers/node/index.js#L12-L15

tests/src/core/importType.js
echenley: In particular: https://github.com/benmosher/eslint-plugin-import/blob/master/resolvers/webpack/index.js#L41-L43

src/core/importType.js
```diff
@@ -21,10 +21,11 @@ export function isAbsolute(name) {
   return name.indexOf('/') === 0
 }
 
-export function isBuiltIn(name, settings) {
+// path is defined only when a resolver resolves to a non-standard path
+export function isBuiltIn(name, settings, path) {
   const base = baseModule(name)
   const extras = (settings && settings['import/core-modules']) || []
-  return coreModules[base] || extras.indexOf(base) > -1
+  return !path && coreModules[base] || extras.indexOf(base) > -1
```
ljharb: let's change this to an early return (at the top of the function, ideally):
```suggestion
  if (!path) return false
  return coreModules[base] || extras.indexOf(base) > -1
```

tests/src/core/importType.js
ljharb: Can that be changed in this PR? We can always merge, release the node resolver first, and release the main plugin after.

tests/src/core/importType.js
ljharb: ditto

tests/src/core/importType.js
echenley: I did a little local testing and I think the current behavior here is actually correct. If you `require('constants')` in node, it appears to always resolve to the core module rather than anything matching NODE_PATH. `require('constants/index')` however, does resolve the local file. For that reason, I think it's better to leave the node resolver as-is.

ljharb: Similarly, `require(‘constants/‘)` should always match the local file - if that’s the case then this is fine.

echenley: That does appear to be the case. Added a test.

tests/src/core/importType.js
ljharb: let's uncomment this, at least, just like above.

echenley: It's a little crude, but moving the coreLibs check until after module resolution in the webpack resolver does appear to be a solution. Added that fix to the PR.

That said, I know webpack has had kind of a weird relationship with core node.js modules, so it's hard to tell if the fix is always going to be correct.

https://webpack.js.org/configuration/node/
https://blog.logrocket.com/new-features-in-webpack-5-2559755adf5e (removing default core polyfills in v5)

ljharb: If that's the case, we'd want to pivot the behavior based on webpack version, i think?

echenley: Since webpack 5 isn't out yet, I think just moving the coreLibs check works for now.

karuana:
When will this PR be merged?
I am waiting for this PR to be merged. :sob::sob::sob:

PR #1304, opened 2019-03-19
support export type named exports from typescript

### LINKED ISSUE #1282: TypeScript export and import/named rule
TypeScript's exported types/interfaces does not work with `import/named` rule.

File 'xxx.ts':

```ts
export type ScpRouteProps = ReactRouter.RouteProps & {
    access?: boolean;
    title: string;
};
```

Then import it in different file:
```ts
import {ScpRouteProps} from './xxx';
```

Will result in error:

```
7:5  error  ScpRouteProps not found in './xxx'    import/named
```

Same error when importing:

```ts
import {WrappedFieldProps} from 'redux-form';
// 1:9  error  WrappedFieldProps not found in 'redux-form'                import/named
```

Or

```ts
import {AutoSizer, Column, ColumnProps, Table, TableProps} from 'react-virtualized';
// /xxx/web/src/components/VirtualizedTable.tsx
//  3:28  error  ColumnProps not found in 'react-virtualized'  import/named
//  3:48  error  TableProps not found in 'react-virtualized'   import/named
```

bradennapier:
Adds support so that `export type` named imports work with this plugin.

Resolves #1282

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   },
   "homepage": "https://github.com/benmosher/eslint-plugin-import",
   "devDependencies": {
+    "@typescript-eslint/parser": "^1.5.0",
     "babel-eslint": "^8.2.6",
     "babel-plugin-istanbul": "^4.1.6",
     "babel-preset-es2015-argon": "latest",
@@ -69,7 +70,7 @@
     "rimraf": "^2.6.3",
     "sinon": "^2.4.1",
     "typescript": "^3.2.2",
-    "typescript-eslint-parser": "^21.0.2"
+    "typescript-eslint-parser": "^22.0.0"
   },
   "peerDependencies": {
     "eslint": "2.x - 5.x"
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "nyc": "^11.9.0",
     "redux": "^3.7.2",
     "rimraf": "^2.6.3",
+    "semver": "^6.0.0",
     "sinon": "^2.4.1",
     "typescript": "^3.2.2",
     "typescript-eslint-parser": "^22.0.0"
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ rules:
 
 You may use the following shortcut or assemble your own config using the granular settings described below.
 
-Make sure you have installed the [`@typescript-eslint/parser`] which is used in the following configuration. Unfortunately NPM does not allow to list optional peer dependencies. 
+Make sure you have installed [`@typescript-eslint/parser`] which is used in the following configuration. Unfortunately NPM does not allow to list optional peer dependencies.
 
 ```yaml
 extends:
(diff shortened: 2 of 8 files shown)


kjleitz:
@ljharb I was checking out the bug this fixes, and I'm just curious—what would a test case look like for this? I'm not sure I see any existing examples of specs for the other named declaration exports (just grepped for a few of them)

bradennapier:
Yeah I’ve been meaning to ask that.  In fact it would seem this somehow increased coverage although not sure how or why. 

It’s a single case statement added and only covers a typescript type alias.

ljharb:
The OP in the [linked issue](https://github.com/benmosher/eslint-plugin-import/issues/1282#issue-407406248) seems to have a test case?

schmod:
Here's [a commit](https://github.com/benmosher/eslint-plugin-import/commit/7b5e6c56f5895550199a5bd88226f9fb4f6eab97) that adds test-coverage. (GitHub won't let me PR it directly into @bradennapier's branch for some reason)
 
This turned out to be a weird one -- there are already tests for this functionality, but they were passing.  The root issue actually appears to be a regression that only appears when switching from `typescript-eslint-parser` to `@typescript-eslint/parser`.  I've updated the tests to run against both parsers.  

`typescript-eslint-parser` was recently deprecated, but is widely-used to the extent where we should avoid deliberately breaking support for it in the short-term.

ljharb:
@schmod thanks; i've added that commit to this branch.

package.json
```diff
@@ -68,8 +69,7 @@
     "redux": "^3.7.2",
     "rimraf": "^2.6.3",
     "sinon": "^2.4.1",
-    "typescript": "^3.2.2",
-    "typescript-eslint-parser": "^21.0.2"
```
ljharb: does this need to be retained here, so that the test case will work?

ljharb:
@schmod looks like tests are failing in eslint 2 and 3.

schmod:
Huh. I'm guessing that we've uncovered an existing incompatibility between `@typescript-eslint/parser` and eslint 3.  

Given that `@typescript-eslint/parser` has a peerDependency on eslint 5.x, I guess we'll just want to skip any tests using that parser on ESLint <5.

schmod:
@ljharb https://github.com/schmod/eslint-plugin-import/commit/8355aaf0937497add49ad702753e532bf85a6047

ljharb:
@schmod eslint 2 and 3 tests are still failing with that commit :-/

schmod:
Huh, I'll have to dig in.  They passed on my local machine.

schmod:
Fun times.  The tests pass on MacOS (using the `dep-time-travel.sh` script and all), but fail on Linux/Docker, and I have no clue why.

I've narrowed this down to `@eslint-typescript/parser` merely being installed (which shouldn't matter?!).  I'll add a workaround to dep-time-travel to remove the new parser when it isn't being used by the tests (ie. on ESLint <5), but this is a pretty spooky bug, TBH.  

Commit: https://github.com/benmosher/eslint-plugin-import/pull/1316/commits/1914fd62bbe34e1a9c7fd094621acd1370eebf5e

tpakhoa1996:
Does this one also include export import SomeEnum = SomeModule.SomeEnum;

PR #1312, opened 2019-03-30
no-duplicates: Add autofix

lydell:
This adds autofix to `no-duplicates`. The autofix merges duplicate imports into the first one. It bails if there are multiple default import names, or comments associated with the (non-first) duplicate imports (since it's unclear how the user wants to move them).

### DIFF
--- a/src/rules/no-duplicates.js
+++ b/src/rules/no-duplicates.js
@@ -26,7 +26,12 @@ function checkImports(imported, context) {
 }
 
 function getFix(first, rest, sourceCode) {
-  // Sorry ESLint <= 3 users, no autofix for you.
+  // Sorry ESLint <= 3 users, no autofix for you. Autofixing duplicate imports
+  // requires multiple `fixer.whatever()` calls in the `fix`: We both need to
+  // update the first one, and remove the rest. Support for multiple
+  // `fixer.whatever()` in a single `fix` was added in ESLint 4.1.
+  // `sourceCode.getCommentsBefore` was added in 4.0, so that's an easy thing to
+  // check for.
   if (typeof sourceCode.getCommentsBefore !== 'function') {
     return undefined
   }
--- a/src/rules/no-duplicates.js
+++ b/src/rules/no-duplicates.js
@@ -36,6 +36,11 @@ function getFix(first, rest, sourceCode) {
     return undefined
   }
 
+  // If the first import is `import * as ns from './foo'` there's nothing we can do.
+  if (hasNamespace(first)) {
+    return undefined
+  }
+
   const defaultImportNames = new Set(
     [first, ...rest].map(getDefaultImportName).filter(Boolean)
   )
@@ -47,11 +52,14 @@ function getFix(first, rest, sourceCode) {
   }
 
   // It's not obvious what the user wants to do with comments associated with
-  // duplicate imports, so skip imports with comments when autofixing.
+  // duplicate imports, so skip imports with comments when autofixing. Also skip
+  // `import * as ns from './foo'` imports, since they cannot be merged into
+  // another import.
   const restWithoutComments = rest.filter(node => !(
       hasCommentBefore(node, sourceCode) ||
       hasCommentAfter(node, sourceCode) ||
-      hasCommentInsideNonSpecifiers(node, sourceCode)
+      hasCommentInsideNonSpecifiers(node, sourceCode) ||
+      hasNamespace(node)
   ))
 
   const specifiers = restWithoutComments
@@ -75,6 +83,7 @@ function getFix(first, rest, sourceCode) {
 
   const unnecessaryImports = restWithoutComments.filter(node =>
     !hasSpecifiers(node) &&
+    !hasNamespace(node) &&
     !specifiers.some(specifier => specifier.importNode === node)
   )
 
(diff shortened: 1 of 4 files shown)


lydell:
The ESLint v3 and v2 tests are failing. Guidance on how to solve that would be appreciated.

lydell:
I disabled autofix for ESLint <= 3.

src/rules/no-duplicates.js
```diff
@@ -2,21 +2,198 @@ import resolve from 'eslint-module-utils/resolve'
 import docsUrl from '../docsUrl'
 
 function checkImports(imported, context) {
-  for (let [module, nodes] of imported.entries()) {
-    if (nodes.size > 1) {
-      for (let node of nodes) {
-        context.report(node, `'${module}' imported multiple times.`)
+  for (const [module, nodes] of imported.entries()) {
+    if (nodes.length > 1) {
```
ljharb: shouldn't this be `nodes.size`?
```suggestion
    if (nodes.size > 1) {
```

src/rules/no-duplicates.js
```diff
@@ -2,21 +2,198 @@ import resolve from 'eslint-module-utils/resolve'
...
+          node: node.source,
+          message,
+        })
+      }
+    }
+  }
+}
+
+function getFix(first, rest, sourceCode) {
+  // Sorry ESLint <= 3 users, no autofix for you.
```
ljharb: how is this handled elsewhere in the codebase that might use `getCommentsBefore`?

src/rules/no-duplicates.js
```diff
@@ -2,21 +2,198 @@ import resolve from 'eslint-module-utils/resolve'
...
+  if (typeof sourceCode.getCommentsBefore !== 'function') {
+    return undefined
+  }
+
+  const defaultImportNames = new Set(
+    [first, ...rest].map(getDefaultImportName).filter(Boolean)
+  )
+
+  // Bail if there are multiple different default import names – it's up to the
```
ljharb: is there a way we could reduce the ones that are the same? ie, if the list was `a, a, b, c`, it could be autofixed to `a, b, c`?

src/rules/no-duplicates.js
```diff
@@ -2,21 +2,198 @@ import resolve from 'eslint-module-utils/resolve'
...
+
+  const shouldAddDefault = getDefaultImportName(first) == null && defaultImportNames.size === 1
+  const shouldAddSpecifiers = specifiers.length > 0
+  const shouldRemoveUnnecessary = unnecessaryImports.length > 0
+
+  if (!(shouldAddDefault || shouldAddSpecifiers || shouldRemoveUnnecessary)) {
+    return undefined
+  }
+
+  return function* (fixer) {
```
ljharb: does this really need to be a generator?

src/rules/no-duplicates.js
lydell: I changed to an array, since it does not make sense to use a `Set` for storing AST objects – they will never deduplicate anyway. And even if they did, that's not what we want here – we need all the duplicate imports.

src/rules/no-duplicates.js
lydell: `src/ExportMap.js` does this:

```js
      // n.leadingComments is legacy `attachComments` behavior
      if ('leadingComments' in n) {
        leadingComments = n.leadingComments
      } else if (n.range) {
        leadingComments = source.getCommentsBefore(n)
      }
```

However, I tried something like this and that at least makes the autofix not crash, but instead the autofix did nothing. As far as I know, ESLint <= 3 does not support multiple `fixer.whatever()` in the `fix`, which is required for `no-duplicates` – we both update the first import and remove one or more others. I can add a comment explaining all of this.

src/rules/no-duplicates.js
lydell: I don't think it's possible to end up with `a, a, b, c`, because duplicate identifiers are a syntax error: https://astexplorer.net/#/gist/bcba02f3c293bd99d3e00b3061befcfb/89b6673f21acde39324bb7063394b56834ede08e

src/rules/no-duplicates.js
lydell: I think it's possible to make an array, `const fixes = []`, and `.push()` to it instead of `yield` and return that at the end. I can do that if you think that's better (and it works).

lydell: Confirmed – the array version works, too. Personally I like the generator here, but let me know which way you want it.

src/rules/no-duplicates.js
lydell: (Also, note that the particular line you commented on isn't about the specifiers inside `{...}`, but about _default_ imports: `import a from "a"; import b from "a"` cannot be autofixed.)

src/rules/no-duplicates.js
lydell: Comment added.

lydell:
Ready for another round of review.

lydell:
@ljharb Friendly ping. If you have the time to review again, that'd be great!

src/rules/no-duplicates.js
ljharb: I'd prefer the array version, I find it easier to understand.

lydell: Pushed the array version!

lydell:
Thanks! :tada:

PR #1319, opened 2019-04-12
fix: false positives for typescript type + value export

### LINKED ISSUE #1318: import/export: false positives for typescript type + value export
Reporting from https://github.com/typescript-eslint/typescript-eslint/issues/10
The following code is valid typescript (even though it's a bit weird).

```TS
// unexected error: Multiple exports of name 'Foo'. (import/export)
export const Foo = 1;
// unexected error: Multiple exports of name 'Foo'. (import/export)
export type Foo = string;
```

When importing this, it causes the name to work as both a type and a value.
```TS
import { Foo } from './foo';

const x: Foo = 'string val';
const y = Foo; // === 1
```

bradzacher:
Fixes #1318

### DIFF
--- a/src/rules/export.js
+++ b/src/rules/export.js
@@ -1,5 +1,6 @@
 import ExportMap, { recursivePatternCapture } from '../ExportMap'
 import docsUrl from '../docsUrl'
+import includes from 'array-includes'
 
 module.exports = {
   meta: {
@@ -12,12 +13,13 @@ module.exports = {
   create: function (context) {
     const named = new Map()
 
-    function addNamed(name, node) {
-      let nodes = named.get(name)
+    function addNamed(name, node, type) {
+      const key = type ? `${type}:${name}` : name
+      let nodes = named.get(key)
 
       if (nodes == null) {
         nodes = new Set()
-        named.set(name, nodes)
+        named.set(key, nodes)
       }
 
       nodes.add(node)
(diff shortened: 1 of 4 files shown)


ljharb:
This seems like it might overlap with #923?

bradzacher:
@ljharb to prevent double commenting - https://github.com/benmosher/eslint-plugin-import/issues/1318#issuecomment-482433011

src/rules/export.js
```diff
@@ -34,7 +35,14 @@ module.exports = {
         if (node.declaration == null) return
 
         if (node.declaration.id != null) {
-          addNamed(node.declaration.id.name, node.declaration.id)
+          if ([
+            'TSTypeAliasDeclaration',
+            'TSInterfaceDeclaration',
+          ].includes(node.declaration.type)) {
```
ljharb: `includes` isn't available on node 4; we'll need to add the `array-includes` package here.

tests/src/rules/export.js
```diff
@@ -106,3 +107,43 @@ ruleTester.run('export', rule, {
...
+    }
+
+    ruleTester.run('export', rule, {
+      valid: [
+        test({
+          code: `
+            export const Foo = 1;
+            export type Foo = number;
+          `,
+          ...parserConfig,
```
ljharb: also, spread isn't on node < 8.

PR #1320, opened 2019-04-12
fix(export): Support typescript namespaces

### LINKED ISSUE #1300: BUG? Typescript export namespace nested `Multiple exports of name`
`import/export`
```ts
export namespace API {
    //Multiple exports of name 'User' 👇
    export namespace User {
        export const A = 0;
    }

    export namespace M {
        //Multiple exports of name 'User' 👇
        export const User = {};
    }
}
```
But they are not the same `User`

bradzacher:
Fixes #1300

- Added nested export scope support
- Fixed broken test on ESLint v3 caused by old typescript parser

### DIFF
--- a/src/rules/export.js
+++ b/src/rules/export.js
@@ -2,6 +2,28 @@ import ExportMap, { recursivePatternCapture } from '../ExportMap'
 import docsUrl from '../docsUrl'
 import includes from 'array-includes'
 
+/*
+Notes on Typescript namespaces aka TSModuleDeclaration:
+
+There are two forms:
+- active namespaces: namespace Foo {} / module Foo {}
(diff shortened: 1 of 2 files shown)


ljharb:
Hmm, I just pushed up a commit to deal with "Fixed broken test on ESLint v3 caused by old typescript parser" by skipping the test on eslint < 4. Does that mean this PR fixes the test so it passes? If so, please rebase and revert my change in the process :-)

bradzacher:
@ljharb - rebased on top!

Yeah the error was happening because in the old version of typescript which got installed with eslint v3 produced a different AST, hence that case wasn't handled. Not sure why it worked fine on the PR then failed on master.

PR #1330, opened 2019-04-15
no-unused-modules: make sure that rule with no options will not fail 

Kiwka:
Currently rule is failing in console, if no options are provided, but none of them is required, as `src` has preset fallback for it, other params are optional (even though without setting `missingExports` or `unusedExports`, to true this rule is not really used).


ljharb:
Thanks!

PR #1344, opened 2019-04-24
Support CommonJS in Extensions

sedenardi:
This applies the same logic of the `import/extensions` rule to CommonJS-style require paths. Similar to the `import/no-unresolved` rule, it's enabled with a `{ commonjs: true }` option so existing setups aren't disrupted.

I've added notes to the rule's docs, as well as 4 tests (in the style as similar tests for this rule). Please let me know if you have any questions/comments/concerns.

### DIFF
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -2,6 +2,7 @@ import path from 'path'
 
 import resolve from 'eslint-module-utils/resolve'
 import { isBuiltIn, isExternalModuleMain, isScopedMain } from '../core/importType'
+import isStaticRequire from '../core/staticRequire'
 import docsUrl from '../docsUrl'
 
 const enumValues = { enum: [ 'always', 'ignorePackages', 'never' ] }
@@ -23,6 +24,7 @@ function buildProperties(context) {
       defaultConfig: 'never',
       pattern: {},
       ignorePackages: false,
+      commonjs: false,
     }
 
     context.options.forEach(obj => {
@@ -47,6 +49,11 @@ function buildProperties(context) {
       // If ignorePackages is provided, transfer it to result
       if (obj.ignorePackages !== undefined) {
         result.ignorePackages = obj.ignorePackages
+      }
+
+      // If commonjs is provided, transfer it to result
+      if (obj.commonjs !== undefined) {
+        result.commonjs = obj.commonjs
       }
     })
 
@@ -121,14 +128,7 @@ module.exports = {
       return resolvedFileWithoutExtension === resolve(file, context)
     }
 
-    function checkFileExtension(node) {
-      const { source } = node
-
-      // bail if the declaration doesn't have a source, e.g. "export { foo };"
-      if (!source) return
-
-      const importPath = source.value
-
+    function checkFileExtension(importPath, node) {
       // don't enforce anything on builtins
       if (isBuiltIn(importPath, context.settings)) return
 
@@ -147,7 +147,7 @@ module.exports = {
         const extensionForbidden = isUseOfExtensionForbidden(extension)
         if (extensionRequired && !extensionForbidden) {
           context.report({
-            node: source,
+            node,
             message:
               `Missing file extension ${extension ? `"${extension}" ` : ''}for "${importPath}"`,
           })
@@ -155,16 +155,38 @@ module.exports = {
       } else if (extension) {
         if (isUseOfExtensionForbidden(extension) && isResolvableWithoutExtension(importPath)) {
           context.report({
-            node: source,
+            node,
             message: `Unexpected use of file extension "${extension}" for "${importPath}"`,
           })
         }
       }
     }
 
+    function checkImportFileExtension(node) {
+      const { source } = node
+
+      // bail if the declaration doesn't have a source, e.g. "export { foo };"
+      if (!source) return
+
+      const importPath = source.value
+
+      checkFileExtension(importPath, source)
+    }
+
+    function checkCommonJSFileExtension(node) {
+      if (!props.commonjs) return
+
+      if (!isStaticRequire(node)) return
+
+      const importPath = node.arguments[0].value
+
+      checkFileExtension(importPath, node)
+    }
+
     return {
-      ImportDeclaration: checkFileExtension,
-      ExportNamedDeclaration: checkFileExtension,
+      ImportDeclaration: checkImportFileExtension,
+      ExportNamedDeclaration: checkImportFileExtension,
+      CallExpression: checkCommonJSFileExtension,
     }
   },
 }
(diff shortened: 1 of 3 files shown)


src/rules/extensions.js
```diff
@@ -147,24 +147,46 @@ module.exports = {
...
+
+      checkFileExtension(importPath, node)
+    }
+
     return {
-      ImportDeclaration: checkFileExtension,
-      ExportNamedDeclaration: checkFileExtension,
+      ImportDeclaration: checkImportFileExtension,
+      ExportNamedDeclaration: checkImportFileExtension,
+      CallExpression: checkCommonJSFileExtension,
```
ljharb: maybe we could set this to null if props.commonjs is falsy, and avoid the visitor entirely?

tests/src/rules/extensions.js
```diff
@@ -359,5 +379,48 @@ ruleTester.run('extensions', rule, {
...
+        },
+      ],
+    }),
+
+    test({
+      code: `
+        const foo = require('./foo.js')
+        const bar = require('./bar.json')
+        const Component = require('./Component.jsx')
+        const express = require('express')
```
ljharb: Let’s also include a test for a package that has an extension-like name

src/rules/extensions.js
sedenardi: Will make the change.

tests/src/rules/extensions.js
sedenardi: Do you mean like `const dot = require('./file.with.dot')` ?

ljharb: and like `require('asn1.js')`, a real package.

sedenardi: Done. Also added tests for `asn1.js` for `import` statements.

src/rules/extensions.js
sedenardi: Rather than setting it to `null` (this generated some errors on CI) or use `undefined` in a ternary operator, the assignment is wrapped in an `if` statement.

benmosher:
haven't looked closely at code but I like the idea 👍

src/rules/extensions.js
ljharb: aw, i'd actually prefer the inline `undefined` to mutating the object.

src/rules/extensions.js
```diff
@@ -34,7 +36,9 @@ function buildProperties(context) {
       }
 
       // If this is not the new structure, transfer all props to result.pattern
-      if (obj.pattern === undefined && obj.ignorePackages === undefined) {
+      if (obj.pattern === undefined &&
```
ljharb: is this the style the linter wants? i'd expect:
```js
if (
  obj.pattern === undefined &&
  obj.ignorePackages === undefined &&
  obj.commonjs === undefined
) {
```

src/rules/extensions.js
```diff
@@ -147,24 +149,49 @@ module.exports = {
...
+      checkFileExtension(importPath, node)
+    }
+
+    const ext = {
+      ImportDeclaration: checkImportFileExtension,
+      ExportNamedDeclaration: checkImportFileExtension,
     }
+
+    if (props.commonjs) {
+      ext.CallExpression = checkCommonJSFileExtension
```
ljharb: def would prefer avoiding mutation here and using an inline undefined, if it would work

src/rules/extensions.js
sedenardi: It didn't complain about that style, just with the length of the line when I had them all inline. Changed.

src/rules/extensions.js
sedenardi: Changed to inline.

sedenardi: After changing the line to 

```
CallExpression: props.commonjs ? checkCommonJSFileExtension : undefined,
```

I see CI broke with the following error (https://travis-ci.org/benmosher/eslint-plugin-import/jobs/524246160):

```
1) extensions valid import a from "a":
     TypeError [ERR_INVALID_ARG_TYPE]: Error while loading rule 'extensions': The "listener" argument must be of type Function. Received type undefined
      at Object.keys.forEach.selector (node_modules/eslint/lib/eslint.js:873:29)
      at Object.keys.filter.forEach.key (node_modules/eslint/lib/eslint.js:872:39)
      at EventEmitter.module.exports.api.verify (node_modules/eslint/lib/eslint.js:841:93)
      at runRuleForItem (node_modules/eslint/lib/testers/rule-tester.js:387:38)
      at testValidTemplate (node_modules/eslint/lib/testers/rule-tester.js:420:28)
      at Context.RuleTester.it (node_modules/eslint/lib/testers/rule-tester.js:548:25)
```

ljharb: gotcha, i guess this approach is fine then.

sedenardi: Reverted and pushed.

sedenardi:
Hey guys, no rush to get this reviewed and merged, just wanted to make sure you weren't waiting on anything on my end. Thanks again for all your help and feedback.

alexnewmannn:
Thanks for picking this up @sedenardi - this will be a big help within our team if this gets released 👍

benmosher:
ideally everything that is acting on arbitrary imports/requires should go through moduleVisitor so we can centralize the detection. I'd also like to (if it doesn't already) allow setting es/common/amd settings globally for all rules which would make most sense to support from `moduleVisitor` vs. replicating everywhere.

sorry to hold this up! 😅

sedenardi:
@benmosher I've cutover rule to use `moduleVisitor` as you suggested. I've also remove all the Line/Column declarations from the error tests, as they've changed (due to the reported node when erroring). I feel this is safe because the test for the rule you gave as an example (`no-unresolved`) does not use Line/Column declarations for their error testing.

Thank you again for your time.

sedenardi:
Hi all, just bumping this PR to see if we can get someone to take a look at it again since I think I've satisfied all the requirements.

src/rules/extensions.js
```diff
@@ -147,24 +153,21 @@ module.exports = {
         const extensionForbidden = isUseOfExtensionForbidden(extension)
         if (extensionRequired && !extensionForbidden) {
           context.report({
-            node: source,
```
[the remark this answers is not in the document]
benmosher: Actually IIRC the highlighting might be better in editors/IDEs using the entire node instead of just the source, even though one could argue that the error is only in the `source` bit. is that why?

I'd like to merge this, just want to understand this change first.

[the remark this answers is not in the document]
sedenardi: You're right, not sure why I wasn't using the `source` node still. I've added it back in and added back the line/column number references to the existing `import` and new `require` tests.

ljharb:
@sedenardi it'd be great if you could rebase this on top of latest master

sedenardi:
> @sedenardi it'd be great if you could rebase this on top of latest master

Apologies for the delayed reply. Feel free to close/reject this PR. To be honest, it's been too long since I've been familiar with the code to rebase 300+ commits over almost 2 years, and I no longer rely on this PR's feature in my projects.

PR #1354, opened 2019-05-05
Fix apparent corner case with `import/no-commonjs` rule

charlessuh:
`var x = require("x")` will throw an error, but `var x; x = require("x")` does not.

### DIFF
--- a/tests/src/rules/no-commonjs.js
+++ b/tests/src/rules/no-commonjs.js
@@ -60,6 +60,7 @@ ruleTester.run('no-commonjs', require('rules/no-commonjs'), {
 
     // imports
     { code: 'var x = require("x")', errors: [ { message: IMPORT_MESSAGE }] },
+    { code: 'x = require("x")', errors: [ { message: IMPORT_MESSAGE }] },
     { code: 'require("x")', errors: [ { message: IMPORT_MESSAGE }] },
 
     // exports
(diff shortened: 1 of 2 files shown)


tests/src/rules/no-commonjs.js
```diff
@@ -28,27 +28,27 @@ ruleTester.run('no-commonjs', require('rules/no-commonjs'), {
...
-    { code: 'var a = c && require("b")' }, // conditional requires allowed
-    { code: 'require.resolve("help")' }, // methods of require are allowed
-    { code: 'require.ensure([])' }, // webpack specific require.ensure is allowed
-    { code: 'require([], function(a, b, c) {})' }, // AMD require is allowed
-    { code: "var bar = require('./bar', true);" },
-    { code: "var bar = proxyquire('./bar');" },
-    { code: "var bar = require('./ba' + 'r');" },
-    { code: 'var zero = require(0);' },
-    { code: 'require("x")', options: [{ allowRequire: true }] },
+    { code: 'function a() { var x = require("y"); }', parserOptions: { sourceType: 'module' } }, // nested requires allowed
```
ljharb: why are all these parserOptions required?

charlessuh: This [line](https://github.com/benmosher/eslint-plugin-import/blob/5101b73effbf3706495f62121a8719f0ea0e2c68/src/rules/no-commonjs.js#L90) causes the check for `require` to only occur in a module scope.

ljharb:
I think this can be called a bugfix, even though it will cause more warnings to be issued.

Issue #1154, opened 2018-08-03
New rule for monorepos: no relative import between packages


tamlyn:
I'm not sure if this is possible with any existing rule.

I have a monorepo and I would like to ensure that packages don't import each others files using relative imports. For example:

```
├── packages
│   ├── first
│   │   ├── index.js
│   │   └── package.json
│   └── second
│       ├── b.js
│       └── package.json
├── base.js
└── package.json
```

- From `packages/first/index.js` it disallows `require('../../base')` 
- From `packages/first/index.js` it disallows `require('../second/b')` but allows `require('second/b')` (assuming `second` is the package name).
- From `packages/second/b.js` it disallows `require('../first')` but allows `require('first')` (assuming `first` is the package name)
- From `base.js` it disallows `require('./packages/second/b')` but allows `require('second/b')` (this is less important than the above cases).

I think the formulation of the rule would be that any folder that contains a `package.json` forms a boundary that no relative imports can cross.

ljharb:
I don’t think the rule should handle bare imports of other sub projects, but a rule that specifies boundaries - dirs from which relative imports can not escape - seems quite useful.

SebT:
Totally agree with @ljharb, need this too.

kwelch:
It looks like this may cover this case pretty well. 

https://github.com/azz/eslint-plugin-monorepo

wendellhu95:
> It looks like this may cover this case pretty well.
> 
> https://github.com/azz/eslint-plugin-monorepo

This doesn't work for me but https://github.com/sterlingwes/eslint-plugin-monorepo-cop does.

PR #1355, opened 2019-05-09
Add new rule: max-depth

kwelch:
We are trying to actively avoid imports that dig into package folders, due to a habit we created with some internal tooling. To help us in that effort I decided to create this rule. 

I am not tied to any of this to please let me know if you have a better name or configuration for this.

Thanks 👋!

- [x] write tests
- [x] update docs
- [x] make a note in change log

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 
 - [`named`]: ignore Flow `typeof` imports and `type` exports ([#1345], thanks [@loganfsmyth])
+- Added new rule [`max-depth`]. ([#1355], thanks [@kwelch])
 
 ## [2.17.2] - 2019-04-16
 
--- a/docs/rules/max-depth.md
+++ b/docs/rules/max-depth.md
@@ -1,0 +1,50 @@
+# import/max-dependencies
+
+Forbid importing into nested folders of a module.
+
+This is a useful rule as import from packages you should stick to the public API (which is typically exported from the root of the package). Depending on specific folders in a package that are not public API exposes you to a greater potential of breaking changes.
+
+### Options
+
+This rule takes the following options:
+
+#### `max`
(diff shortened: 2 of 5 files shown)


ljharb:
I'm very conflicted about this rule - deep imports from third-party packages is a *good* practice, and prevents treeshaking from being necessary. Put another way, treeshaking only provides a benefit when imports are too-shallow.

Can you explain a bit more about your use case, ie, why this rule provides value? What "habit" and "internal tooling" do you have in mind?

kwelch:
I agree deep imports can be beneficially in certain cases where it is the public API. Especially when done for treeshaking, my thought is a package like lodash, where it is more beneficially to use `lodash/map` vs `{map} from 'lodash'` (at least that is my understanding). 

However our use case it more to avoid importing files that are not part of the public API and could cause breaking changes since we are depending on internal tooling. For instance:

`import colorHelper from 'react-color/lib/helpers/color';`

This was taking from one of our components. If the library decides to rename `helpers` to `utils`. In the eyes of semver it could not even a release just a chore, but when it is released it because a breaking change on this file. 

If you don't think this is a fit that is fine, I have it working with an internal plugin and thought it could be useful for the larger community. Totally up to you.

ljharb:
Everything reachable is the public API :-)

Separately, you can already use `no-restricted-imports`, a core rule, to achieve this - we do this internally at airbnb for anything under `internal-package-name/private/**`.

kwelch:
Oh interesting I had not thought of configuring it that way. I had used that for more path based imports. 

With that rule in mind, is it possible to config it in a way to not allow importing for any folder of a certain level other than your own? (Thinking with regards to monorepo, where you don't want relative imports into other packages)

ljharb:
That seems like a different use case, but i think #1154 would cover that?

kwelch:
Cool! I agree that `no-restricted-imports` makes for a simpler solution for my case. I am good with closing this, unless feel there is something of use here.

ljharb:
Let's close this for now, then. Thanks for the contribution!

PR #1356, opened 2019-05-11
Improve support for Typescript declare structures

christophercurrie:
This change improves support various forms used in Typescript declaration files. It's often necessary to change import resolution to prefer '.d.ts' files over '.js' files, but doing so breaks existing import resolution for certain types of UMD modules (React, for example).

This patch has two improvements:
* It adds support for the `export declare function foo()` syntax, which wasn't previously recognized
* It adds support for the `export = Namespace` syntax for exporting everything in a namespace.

The `import/named` tests were updated to add both variations as additional test files.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -517,8 +517,6 @@ ExportMap.parse = function (path, content, context) {
       const moduleDecl = ast.body.find((bodyNode) =>
         bodyNode.type === 'TSModuleDeclaration' && bodyNode.id.name === n.expression.name
       )
-      log(moduleDecl)
-      log(moduleDecl.body)
       if (moduleDecl && moduleDecl.body && moduleDecl.body.body) {
         moduleDecl.body.body.forEach((moduleBlockNode) => {
           // Export-assignment exports all members in the namespace, explicitly exported or not.
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -532,7 +532,8 @@ ExportMap.parse = function (path, content, context) {
               )
             )
           } else {
-            m.namespace.set(exportedDecl.id.name,
+            m.namespace.set(
+              exportedDecl.id.name,
               captureDoc(source, docStyleParsers, moduleBlockNode))
           }
         })
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -527,7 +527,8 @@ ExportMap.parse = function (path, content, context) {
           if (exportedDecl.type === 'VariableDeclaration') {
             exportedDecl.declarations.forEach((decl) =>
               recursivePatternCapture(decl.id,(id) => m.namespace.set(
-                id.name, captureDoc(source, docStyleParsers, decl, exportedDecl, moduleBlockNode))
+                id.name,
+                captureDoc(source, docStyleParsers, decl, exportedDecl, moduleBlockNode))
               )
             )
           } else {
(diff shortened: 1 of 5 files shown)


utils/unambiguous.js
```diff
@@ -18,7 +18,7 @@ exports.test = function isMaybeUnambiguousModule(content) {
 }
 
 // future-/Babel-proof at the expense of being a little loose
-const unambiguousNodeType = /^(Exp|Imp)ort.*Declaration$/
+const unambiguousNodeType = /^(((Exp|Imp)ort.*Declaration)|TSExportAssignment)$/
```
ljharb: does this need capturing groups? (also i think it might not need this many parens)
```suggestion
const unambiguousNodeType = /^(?:(?:Exp|Imp)ort.*Declaration|TSExportAssignment)$/
```

christophercurrie:
Thanks. Incorporated your change and fixed the issues with older typescript parsers.

src/ExportMap.js
```diff
@@ -513,14 +514,20 @@ ExportMap.parse = function (path, content, context) {
 
     // This doesn't declare anything, but changes what's being exported.
     if (n.type === 'TSExportAssignment') {
-      const d = ast.body.find(
-        b => b.type === 'TSModuleDeclaration' && b.id.name === n.expression.name
+      const md = ast.body.find(
```
ljharb: it might be nice if we used whole words for these var names, it’s hard to know what’s going on right now :-)

christophercurrie:
I don’t disagree, but I was trying to follow the style of the file. Happy to make a change, just let me know.

ljharb:
Thanks, it'd be great to improve the readability in this PR :-) it's not a blocker, tho.

src/ExportMap.js
```diff
@@ -509,6 +511,32 @@ ExportMap.parse = function (path, content, context) {
...
+        moduleDecl.body.body.forEach((moduleBlockNode) => {
+          // Export-assignment exports all members in the namespace, explicitly exported or not.
+          const exportedDecl = moduleBlockNode.type === 'ExportNamedDeclaration' ?
+            moduleBlockNode.declaration :
+            moduleBlockNode
+
+          if (exportedDecl.type === 'VariableDeclaration') {
+            exportedDecl.declarations.forEach((decl) =>
+              recursivePatternCapture(decl.id,(id) => m.namespace.set(
+                id.name, captureDoc(source, docStyleParsers, decl, exportedDecl, moduleBlockNode))
```
ljharb: if each item can’t fit on one line, please put each on its own line:
```suggestion
                id.name,
                captureDoc(source, docStyleParsers, decl, exportedDecl, moduleBlockNode))
```

src/ExportMap.js
```diff
@@ -509,6 +511,32 @@ ExportMap.parse = function (path, content, context) {
...
+              )
+            )
+          } else {
+            m.namespace.set(exportedDecl.id.name,
```
ljharb:
```suggestion
            m.namespace.set(
              exportedDecl.id.name,
```

christophercurrie:
Applied.

joaovieira:
@christophercurrie this does not cover situations like:

```ts
export = supertest;

declare function supertest(app: any): supertest.SuperTest<supertest.Test>;
declare namespace supertest { ... }
```

and

```ts
import { snakeCase } from "./index";
export = snakeCase;
```

I've raised https://github.com/benmosher/eslint-plugin-import/issues/1527 for that.

PR #1105, opened 2018-05-17
Alphabetize imports within groups

duncanbeevers:
In order to provide absolute ordering, rather than just group-wise ordering, this PR offers an `alphabetize` option that enforces locale-aware alphabetical intra-group ordering.

It makes no effort to provide a more robust `fix` experience, so multiple passes may be necessary to bring a set of imports into a stable configuration, though since an absolute ordering is available, a single-pass `fix` probably isn't unreasonable.

Fixes #1406.

PR #629, opened 2016-10-17
Fix #389: Allow secondary alphabetical sorting

randallreedjr:
- After sorting imports by group, allow sorting alphabetically within a group
- import/order rule now accepts sort-order option that may be 'ignore' (default) or 'alphabetical'
- Fixes #389

Issue #389, opened 2016-06-21
[import/order] sort order within groups


nevir:
It'd be handy if [`import/order`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md) either played nice with eslint's [`sort-imports`](https://github.com/eslint/eslint/blob/master/docs/rules/sort-imports.md) rule (if that's feasible?) - or implemented something along the same lines, but on a per-group basis

benmosher:
Can you provide an example of what you want that isn't working?

nevir:
Yeah, say for example you have the following files:

```
a.js
c.js
subdir/
  b.js
  d.js
  index.js
```

`index.js` has the following:
```
import a from '../a';
import c from '../c';

import b from './b';
import d from './d';
```

And eslint is configured with:
```
import/order: [2, {groups: ['parent', 'sibling'], newlines-between: 'always'}]
sort-imports: 2
```

When running eslint on `subdir/index.js`, you receive a `Imports should be sorted alphabetically` error on the `import b from '.b/';` line

(e.g `sort-imports` is checking the imports regardless of groups, but really we need inner-group sorting)

jfmengels:
I like the idea, but I'm worried that we'd get a lot of requests to support plenty of types to order the imports.

I know of a few that are requested already
- Alphabetically, by import name
- By length of the line (to have a stair-case like code)

I could very well see an alphabetical order too, but by the name of the variable it will be affected to (not sure how that'd work with destructuring), which makes more sense IMO as that makes looking for a variable declaration easier.

Don't know if people will want other types of sort, but I think we could be surprised.

We'd probably need support for exceptions too (personally, I always like to have `lodash` as the first import in my files. Don't ask me why). Like when using a test runner like `tape`/`ava`, "I want that to always be the first import", would make sense I think.

Maybe I'm overthinking it, but this seems like potentially quite a lot of work to maintain. That said, if we're willing to commit to this work and/or accept only much requested sort options, I'm all for it.

benmosher:
+1 to @jfmengels on this being a lot of work.

The simplest thing I can imagine doing would be to allow a group-sort comparator function to be configured as a rule option. Such a function would take the two `ImportDeclaration` nodes to compare, and would be on its own from there.

Then the community could support modules or shared configs with comparators that folks like. 

In the event that one rose to the top, we could integrate it (but the beauty would be: we don't need to).

[...]

PR #1360, opened 2019-05-14
Alphabetize imports within groups

### LINKED ISSUE #1406: An alphabetical order rule?
Hi!
I found that the alphabetical order import rule is very useful.
Do you mind I add a new rule for that?
Or an option to existed order rule?

stropho:
I'm not the real author of these commits. I just rebased the existing https://github.com/benmosher/eslint-plugin-import/pull/1105 . I'm looking forward to use this new feature.

Closes #1105. Fixes #1406.

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -180,12 +180,12 @@ function fixOutOfOrder(context, firstNode, secondNode, order) {
   const sourceCode = context.getSourceCode()
 
   const firstRoot = findRootNode(firstNode.node)
-  let firstRootStart = findStartOfLineWithComments(sourceCode, firstRoot)
+  const firstRootStart = findStartOfLineWithComments(sourceCode, firstRoot)
   const firstRootEnd = findEndOfLineWithComments(sourceCode, firstRoot)
 
   const secondRoot = findRootNode(secondNode.node)
-  let secondRootStart = findStartOfLineWithComments(sourceCode, secondRoot)
-  let secondRootEnd = findEndOfLineWithComments(sourceCode, secondRoot)
+  const secondRootStart = findStartOfLineWithComments(sourceCode, secondRoot)
+  const secondRootEnd = findEndOfLineWithComments(sourceCode, secondRoot)
   const canFix = canReorderItems(firstRoot, secondRoot)
 
   let newCode = sourceCode.text.substring(secondRootStart, secondRootEnd)
@@ -241,6 +241,63 @@ function makeOutOfOrderReport(context, imported) {
     return
   }
   reportOutOfOrder(context, imported, outOfOrder, 'before')
+}
+
+function importsSorterAsc(importA, importB) {
+  if (importA < importB) {
+    return -1
+  }
+
+  if (importA > importB) {
+    return 1
+  }
+
+  return 0
+}
+
+function importsSorterDesc(importA, importB) {
+  if (importA < importB) {
+    return 1
+  }
+
+  if (importA > importB) {
+    return -1
+  }
+
+  return 0
+}
+
+function mutateRanksToAlphabetize(imported, order) {
+  const groupedByRanks = imported.reduce(function(acc, importedItem) {
+    if (!Array.isArray(acc[importedItem.rank])) { 
+      acc[importedItem.rank] = []
+    }
+    acc[importedItem.rank].push(importedItem.name)
+    return acc
+  }, {})
+
+  const groupRanks = Object.keys(groupedByRanks)
+
+  const sorterFn = order === 'asc' ? importsSorterAsc : importsSorterDesc
+  // sort imports locally within their group
+  groupRanks.forEach(function(groupRank) {
+    groupedByRanks[groupRank].sort(sorterFn)
+  })
+
+  // assign globally unique rank to each import
+  let newRank = 0
+  const alphabetizedRanks = groupRanks.sort().reduce(function(acc, groupRank) {
+    groupedByRanks[groupRank].forEach(function(importedItemName) {
+      acc[importedItemName] = newRank
+      newRank += 1
+    })
+    return acc
+  }, {})
+
+  // mutate the original group-rank with alphabetized-rank
+  imported.forEach(function(importedItem) {
+    importedItem.rank = alphabetizedRanks[importedItem.name]
+  })
 }
 
 // DETECTING
@@ -427,6 +484,13 @@ function makeNewlinesBetweenReport (context, imported, newlinesBetweenImports) {
   })
 }
 
+function getAlphabetizeConfig(options) {
+  const alphabetize = options.alphabetize || {}
+  const order = alphabetize.order || 'ignore'
+
+  return {order}
+}
+
 module.exports = {
   meta: {
     type: 'suggestion',
@@ -473,6 +537,16 @@ module.exports = {
               'never',
             ],
           },
+          alphabetize: {
+            type: 'object',
+            properties: {
+              order: {
+                enum: ['ignore', 'asc', 'desc'],
+                default: 'ignore',
+              },
+            },
+            additionalProperties: false,
+          },
         },
         additionalProperties: false,
       },
@@ -482,6 +556,7 @@ module.exports = {
   create: function importOrderRule (context) {
     const options = context.options[0] || {}
     const newlinesBetweenImports = options['newlines-between'] || 'ignore'
+    const alphabetize = getAlphabetizeConfig(options)
     let ranks
 
     try {
@@ -524,6 +599,10 @@ module.exports = {
         registerNode(context, node, name, 'require', ranks, imported)
       },
       'Program:exit': function reportAndReset() {
+        if (alphabetize.order !== 'ignore') {
+          mutateRanksToAlphabetize(imported, alphabetize.order)
+        }
+
         makeOutOfOrderReport(context, imported)
 
         if (newlinesBetweenImports !== 'ignore') {
--- a/tests/src/rules/order.js
+++ b/tests/src/rules/order.js
@@ -519,6 +519,47 @@ ruleTester.run('order', rule, {
         },
       ],
     }),
+    // Option alphabetize: {order: 'ignore'}
+    test({
+      code: `
+        import a from 'foo';
+        import b from 'bar';
+
+        import index from './';
+      `,
+      options: [{
+        groups: ['external', 'index'],
+        alphabetize: {order: 'ignore'},
+      }],
+    }),
+    // Option alphabetize: {order: 'asc'}
+    test({
+      code: `
+        import c from 'Bar';
+        import b from 'bar';
+        import a from 'foo';
+
+        import index from './';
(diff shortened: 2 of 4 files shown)


stropho:
Can anybody help? Is there any way to test incorrect configuration passed to the eslint rule? It seems like such errors are handled already by the `RuleTester` and not the rule itself, so I don't see i simple way to test it at the moment.

Or should we just delete the checks to make coveralls happy ?

ljharb:
Please don’t open duplicate PRs; now have to maintain both in sync.

Instead, spot a link to your commits on the original PR, and a maintainer will update it.

stropho:
ok, got it

VincentLanglet:
Is this ready to merge ? This feature will be great.

stropho:
I believe it is ready. I created a fix based on the suggestion in https://github.com/benmosher/eslint-plugin-import/pull/1105 . And now, probably some of the maintainers have to say "it's ok" and merge it.

syashenkov:
I think example still needs correction:
```js
import assert from 'assert';
import buffer from 'buffer';
import foo from 'foo';
import bar from 'bar';
import Baz from 'Baz';
```
It's unclear if imports should be sorted by module name or imported names. Also, it's uncertain what behaviour is expected using star imports (import * as) and symbol imports (import { A, B })

It would be best to make example with some real life imports like
```js
import * as classnames from 'classnames';
import PropTypes from 'prop-types';
import React, { PureComponent } from 'react';
import { compose } from 'recompose';
```

syashenkov:
Or a more clear example like

Pass:
```js
import c from 'a';
import b from 'b';
import a from 'c';
```
Fail:
```js
import a from 'c';
import b from 'b';
import c from 'a';
```

syashenkov:
@ljharb Could you please look at this PR?

stropho:
Fixed the example. 
Rebased on master branch. 
Hope everybody is happy now.

Travis CI failed but it has the same errors as master branch did https://travis-ci.org/benmosher/eslint-plugin-import/builds/553254212 . So I don't think there is anything I can do about it right now.

deb0ch:
The tests don't seem to be failing on `master` anymore, is this mergeable soon before there are conflicts ?

stropho:
@deb0ch should be mergable; thanks for the reminder, just rebased :champagne: :tada: :champagne: :tada:

luczsoma:
@ljharb Could you look into merging this PR?

It is a long-wanted feature, and also would come with a few benefits, since it:
- Fixes #389 
- Closes #629 
- Closes #1105 

Thank you.

src/rules/order.js
```diff
@@ -240,17 +241,61 @@ function makeOutOfOrderReport(context, imported) {
...
+  }, {})
+
+  const groupRanks = Object.keys(groupedByRanks)
+
+  // sort imports locally within their group
+  groupRanks.forEach(function(groupRank) {
+    groupedByRanks[groupRank].sort(function(importA, importB) {
+      return ignoreCase
+        ? importA.localeCompare(importB)
+        : (importA < importB ? -1 : (importA === importB ? 0 : 1))
```
syabro: Scary. Could it be replaced with `if... return`?

luczsoma: Btw it is also incorrect I think. The default `sensitvity` option is `variant` in `localeCompare` for the default `sort`ing usage (applies in the above snippet), which [does not ignore casing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#Parameters).

I don't want to meddle in this PR, but I suggest the following:
```
groupedByRanks[groupRank].sort(function(importA, importB) {
  return importA.localeCompare(importB, undefined, { sensitivity: ignoreCase ? 'accent' : 'case' })
})
```

luczsoma:
@stropho, since I have no write access to this, in order to get this PR finally merged before any further conflicts emerge, I would like to ask you to

- either commit the following two minor changes, please,
- or invite me as a contributor into `stropho/eslint-plugin-import-1` so I am able to commit them.

1. In `src/rules/order.js`, change this:

```
groupedByRanks[groupRank].sort(function(importA, importB) {
  return ignoreCase
    ? importA.localeCompare(importB)
    : (importA < importB ? -1 : (importA === importB ? 0 : 1))
})
```

to this:

```
groupedByRanks[groupRank].sort(function(importA, importB) {
  return importA.localeCompare(importB, undefined, { 
    sensitivity: ignoreCase ? 'accent' : 'case' 
  })
})
```

2. Add the following to `CHANGELOG.md`, into the Unreleased section:

```
- [`order`]: Add support for alphabetical sorting within import groups ([#1360], thanks [@duncanbeevers] & [@stropho])
```

Thank you!

stropho:
@luczsoma  I haven't really investigated the code much, just wanted to help to merge it.... anyway, I just tried to fix it myself since you've told me exactly what to do :-D About 4 tests fail now. That is ok but it seems to be nearly impossible for me to find which tests are failing. 
I added you as collaborator... Feel free to add changes. I could find some spare time at the end of the week.

luczsoma:
Thank you @stropho!

As I tried to extend the functionality with configurable numeric sorting (1 < 2 < 10 vs 1 < 10 < 2) and configurable casing options (a-z < A-Z vs a-z > A-Z), I found out that `localeCompare` doesn't really work as common sense expects, see https://stackoverflow.com/questions/57671159. For the sake of simplicity, I removed `localeCompare`.

I also fixed @syabro's concerns.

luczsoma:
I think this should be mergeable now.

syabro:
> As I tried to extend the functionality with configurable numeric sorting (1 < 2 < 10 vs 1 < 10 < 2) and configurable casing options (a-z < A-Z vs a-z > A-Z)

I bet it could be done later :) We badly need a basic sort

luczsoma:
> > As I tried to extend the functionality with configurable numeric sorting (1 < 2 < 10 vs 1 < 10 < 2) and configurable casing options (a-z < A-Z vs a-z > A-Z)
> 
> I bet it could be done later :) We badly need a basic sort

Planning to do it some time later. :)

stropho:
thx @luczsoma  - I'm not a fan of rewriting code to make it bigger but as long as it is more readable and everybody's happy about it.. why not...

What got my attention though is that the `ignoreCase` probably does not work as it was intended. 
My understanding when `ignoreCase=true`: it doesn't matter if `a` is before or after `A`
My understanding of how it works now: sorting treats `a` that it equals `A` but the tested code needs to be equal to the result of sorting :-/
So I changed the tests a bit to actually test the `ignoreCase` option and had to comment 2 of them because it seems they are not fixable with the current implementation. So what now ?  

1. easier choice - remove `ignoreCase` option. Why would anyone create files with capital or non-ascii letters ? I understand that this is strongly opinionated. But if anyone will ever want such feature, it could be added later. As mentioned before:
> We badly need a basic sort
2. fix the `ignoreCase` case. This does not quite fit my schedule. But maybe someone else can try ?

Let's vote!!! 
:+1: for option 1.
 :rocket:  for option 2.

stropho:
alright, `ignoreCase` option is removed. Less options --> smaller chance something is not working :wink:

stropho:
@ljharb any news? any suggestions to get it merged ?

aunpongsit:
any updates on this? really want it to be merged.

src/rules/order.js
```diff
@@ -240,6 +240,50 @@ function makeOutOfOrderReport(context, imported) {
...
+    } else if (importA > importB) {
+      result = resultModifier
+    }
+
+    return result
+  }
+}
+function mutateRanksToAlphabetize(imported, order) {
+  const groupedByRanks = imported.reduce(function(acc, importedItem) {
+    acc[importedItem.rank] = acc[importedItem.rank] || []
```
ljharb:
```suggestion
    if (!Array.isArray(acc[importedItem.rank]) { acc[importedItem.rank] = [] }
```

src/rules/order.js
```diff
@@ -384,6 +435,15 @@ module.exports = {
               'never',
             ],
           },
+          alphabetize: {
+            type: 'object',
```
ljharb:
```suggestion
            type: 'object',
            additionalProperties: false,
```

tests/src/rules/order.js
```diff
@@ -335,10 +335,10 @@ ruleTester.run('order', rule, {
     // Option newlines-between: 'always' with multiline imports #3
     test({
       code: `
-        import foo
+        import a
```
ljharb: why does this need to be changed? can this test be reverted?

ljharb:
(additionally, please rebase this on latest master)

src/rules/order.js
luczsoma: What is the change request here? To move the `additionalProperties: false` directly under `type: 'object'`, instead of its current position at the bottom of the schema (under `properties`)?

tests/src/rules/order.js
luczsoma: I think the intention of this commit (which is not my own) was to show how the alphabetical sorting in a clearer way, with minimal examples. I would not revert that.

luczsoma: If your comment is concerned about only this particular test case (which tests multiline, not ordering), I will revert that part to its original state.

luczsoma:
Requested changes are done, rebased onto master.

luczsoma:
@ljharb Is there anything else we can do to get this merged? I cannot dismiss the change request (although everything requested is done), only you (or the author of this PR, @stropho) can.

nmorel:
I tried the PR and it breaks the `newlines-between` rules.  
With `always`, I have an empty line between all my imports and not just groups.  
From what I see, `newlines-between` uses rank to determine empty lines to add and alphabetize modifies the ranks to order.

nmorel:
Putting the newlines test first seems to fix it :   
```javascript
if (newlinesBetweenImports !== 'ignore') {
  makeNewlinesBetweenReport(context, imported, newlinesBetweenImports)
}

if (alphabetize.order !== 'ignore') {
  mutateRanksToAlphabetize(imported, alphabetize.order)
}

makeOutOfOrderReport(context, imported)
```

src/rules/order.js
dietergeerts: > 

I was just looking into this PR because I can use it, though I wonder for what reason you find the suggested change better? For my personally, what's there is more readable.

cameronmurphy:
Any news on this? Very keen for this functionality!

kondei:
I want this

iippis:
You guys are so close to the finish line, don't quit now! I'm eagerly waiting for this feature.

tests/src/rules/order.js
ljharb: Yes, my comment is because for it not to be a breaking change, no existing test cases should be altered.

src/rules/order.js
ljharb: avoiding unconditional assignment is more correct, which is more important than what's readable.

PR #1366, opened 2019-05-28
feat: support import ts definition

davidNHK:
Typescript support extension with [.d.ts](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html)

But in current config of `import` , it should complain error when try import custom type.

### DIFF
--- a/config/typescript.js
+++ b/config/typescript.js
@@ -2,7 +2,7 @@
  * Adds `.jsx`, `.ts` and `.tsx` as an extension, and enables JSX/TSX parsing.
  */
 var jsExtensions = ['.js', '.jsx'];
-var tsExtensions = ['.ts', '.tsx'];
+var tsExtensions = ['.d.ts','.ts', '.tsx'];
 var allExtensions = jsExtensions.concat(tsExtensions);
 
 module.exports = {


config/typescript.js
```diff
@@ -2,7 +2,7 @@
  * Adds `.jsx`, `.ts` and `.tsx` as an extension, and enables JSX/TSX parsing.
  */
 var jsExtensions = ['.js', '.jsx'];
-var tsExtensions = ['.ts', '.tsx'];
+var tsExtensions = ['.d.ts','.ts', '.tsx'];
```
ljharb: i'm concerned this might be a breaking change.

leoyli: This should be:

```js
var tsExtensions = ['.ts', '.d.ts', '.tsx'];
```

benmosher: I think it should be `['.ts', '.tsx', '.d.ts']`, no? A full React module should override a definition file, I would think?

In practice I would expect it to matter <<0.1% of the time, though. Seems like a weird case to have definitions and full TS modules at the same filename location. Whereas a `.d.ts` _should_ override a `.js` file, right?

So maybe the correct thing would be 
```js
var allExtensions = ['.ts', '.tsx', '.d.ts', '.js', '.jsx'];
```

PR #1372, opened 2019-05-29
[New] `order`: add fixer for destructuring commonjs import

### LINKED ISSUE #1337: Fixer of `import/order` does not work when destructuring a commomjs module
```js
/* eslint import/order: 2 */

const foo = require('./foo')
const {promisify} = require('util')
```

**Expected**:
Fixer fixes.
**Actual**:
Fixer does not fix.

golopot:
fixes #1337

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -140,7 +140,8 @@ function isPlainRequireModule(node) {
     return false
   }
   const decl = node.declarations[0]
-  const result = (decl.id != null &&  decl.id.type === 'Identifier') &&
+  const result = decl.id &&
+    (decl.id.type === 'Identifier' || decl.id.type === 'ObjectPattern') &&
     decl.init != null &&
     decl.init.type === 'CallExpression' &&
     decl.init.callee != null &&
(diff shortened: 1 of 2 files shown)


src/rules/order.js
```diff
@@ -140,7 +140,7 @@ function isPlainRequireModule(node) {
     return false
   }
   const decl = node.declarations[0]
-  const result = (decl.id != null &&  decl.id.type === 'Identifier') &&
```
golopot: `variableDeclarator.id` is non-null by estree spec.

ljharb: estree isn't the only parser. please include the null check, in case other parsers fail to ensure it's non-null.

src/rules/order.js
```diff
@@ -140,7 +140,7 @@ function isPlainRequireModule(node) {
...
+  const result = (decl.id.type === 'Identifier' || decl.id.type === 'ObjectPattern') &&
```
golopot:
```suggestion
  const result = decl.id &&
    (decl.id.type === 'Identifier' || decl.id.type === 'ObjectPattern') &&
```

src/rules/order.js
golopot: updated

PR #1375, opened 2019-06-06
Added support for correctly ordering unknown types e.g. custom aliases

swernerx:
We figured out that the auto-fix feature of the "order" rule destroyed our custom imports which are using a custom `babel-module` resolver. Settings for the resolver:

```js
'import/resolver': {
  'babel-module': {
    extensions: ['.js', '.jsx', '.ts', '.tsx', '.json'],
    alias: {
      '-': './src/',
    },
  },
},
```

Until now, unknown import types were sorted problematically by the "order" feature. These were not assigned to a common group but loosely added or kept between the other imports. This was especially problematic when using the 'newlines-between'-feature. In this case there was an additional line feed after each unknown import. See examples below.

Example:

```js
import { get } from 'lodash'
import { FormError } from '-/components/Product/Configuration/FormError'
import React, { PureComponent } from 'react'
import { ModalContainer } from '../../../Modal/ModalContainer'
import { addEvent, removeEvent } from '-/common/tracking/basket'
import { ChangeItem } from '-/components/Basket/ChangeItem'
import getBasketProduct from './BasketProduct.gql'
import UpdateItemInBasket from './UpdateItemInBasket.gql'
```

Previous version output:

```js
import { get } from 'lodash'

import { FormError } from '-/components/Product/Configuration/FormError'

import React, { PureComponent } from 'react'

import { ModalContainer } from '../../../Modal/ModalContainer'

import { addEvent, removeEvent } from '-/common/tracking/basket'

import { ChangeItem } from '-/components/Basket/ChangeItem'

import getBasketProduct from './BasketProduct.gql'
import UpdateItemInBasket from './UpdateItemInBasket.gql'
```

New version output:

```js
import { get } from 'lodash'
import React, { PureComponent } from 'react'

import { FormError } from '-/components/Product/Configuration/FormError'
import { addEvent, removeEvent } from '-/common/tracking/basket'
import { ChangeItem } from '-/components/Basket/ChangeItem'

import { ModalContainer } from '../../../Modal/ModalContainer'

import getBasketProduct from './BasketProduct.gql'
import UpdateItemInBasket from './UpdateItemInBasket.gql'
```

I updated the docs accordingly, added new tests and made sure that nothing else broke. The change itself is pretty small but it dramatically changes the result for these situations.

### DIFF
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -86,7 +86,7 @@ How groups are defined, and the order to respect. `groups` must be an array of `
   // Then the rest: internal and external type
 ]
 ```
-The default value is `["builtin", "external", "unknown", "parent", "sibling", "index"]`.
+The default value is `["builtin", "external", "parent", "sibling", "index"]`.
 
 You can set the options like this:
 
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -77,7 +77,7 @@ This rule supports the following options:
 
 ### `groups: [array]`:
 
-How groups are defined, and the order to respect. `groups` must be an array of `string` or [`string`]. The only allowed `string`s are: `"builtin"`, `"external"`, `"internal"`, `"parent"`, `"sibling"`, `"index"`. The enforced order is the same as the order of each element in a group. Omitted types are implicitly grouped together as the last element. Example:
+How groups are defined, and the order to respect. `groups` must be an array of `string` or [`string`]. The only allowed `string`s are: `"builtin"`, `"external"`, `"internal"`, `"unknown"`, `"parent"`, `"sibling"`, `"index"`. The enforced order is the same as the order of each element in a group. Omitted types are implicitly grouped together as the last element. Example:
 ```js
 [
   'builtin', // Built-in types are first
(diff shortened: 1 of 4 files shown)


docs/rules/order.md
```diff
@@ -86,7 +86,7 @@ How groups are defined, and the order to respect. `groups` must be an array of `
   // Then the rest: internal and external type
 ]
 ```
-The default value is `["builtin", "external", "parent", "sibling", "index"]`.
+The default value is `["builtin", "external", "unknown", "parent", "sibling", "index"]`.
```
ljharb: i'm concerned about changing the defaults; can you add `unknown` without changing the defaults, for now?

swernerx: I've changed this in the last commit. Could you take a look again?

swernerx: I've changed this in the last commit. Could you take a look again?

ljharb:
I’m a bit confused - what code is generating “unknown”? Is this a type that we’ve always generated but simply not allowed to be in the config list?

swernerx:
Yes, it seems like this is exactly the case.

swernerx:
It produced here: https://github.com/benmosher/eslint-plugin-import/blob/master/src/core/importType.js#L88

swernerx:
Thanks! Can you tell when the new version will be released?

ljharb:
i can't, no, but this PR will be in the next one, whenever that is :-)

Issue #1332, opened 2019-04-16
`prefer-default-export` fails on Typescript type declaration


sharmilajesupaul:
`import/prefer-default-export` erroneously warns on files that contain a single exported type.

eg: 
```
// foo.ts 

export type Foo = { // `prefer-default-export` fails on this line
 bar: string;
};
```

Related issue: https://github.com/benmosher/eslint-plugin-import/issues/1164

thisissami:
Not sure if this helps or not, but this was addressed once a while back: https://github.com/benmosher/eslint-plugin-import/pull/639 seems like this bug was reintroduced.

ljharb:
That was for flow, this is for typescript.

thisissami:
@ljharb my assumption based off of https://github.com/benmosher/eslint-plugin-import/issues/1164 is that it is broken for both `typesript` and `flow`

thisissami:
and both `typescript` and `flow` use the identical `export type insert_type_name_here` syntax, so unless the parser was explicitly looking for `// @flow` at the top of a file - it should work identically for each.

mkaraula:
👋 are there any news on this?

PR #1377, opened 2019-06-10
[bugfix] Re: #1332 - skip warning on a type declaration or alias. 

sharmilajesupaul:
Attempt at fixing #1332 

Skips warning on a single type declaration or type alias. ~Remove the check for `node.exportKind` as it is `undefined` in all test cases.~ I believe this is used in eslint v2 & 3

This should work with `@typescript-eslint/parser` too, but I'm not testing that here because this repo is still using `typescript-eslint-parser`.

cc. @ljharb

### DIFF
--- a/src/rules/prefer-default-export.js
+++ b/src/rules/prefer-default-export.js
@@ -47,8 +47,15 @@ module.exports = {
         // if there are specifiers, node.declaration should be null
         if (!node.declaration) return
 
-        // don't count flow types exports
+        // don't warn on single type aliases or declarations
         if (node.exportKind === 'type') return
+
+        if (
+          node.declaration.type === 'TSTypeAliasDeclaration' ||
+          node.declaration.type === 'TypeAlias'
+        ) {
+          return
+        }
 
         if (node.declaration.declarations) {
           node.declaration.declarations.forEach(function(declaration) {
--- a/tests/src/rules/prefer-default-export.js
+++ b/tests/src/rules/prefer-default-export.js
@@ -1,8 +1,6 @@
-import { test } from '../utils'
+import { test, getNonDefaultParsers } from '../utils'
 
 import { RuleTester } from 'eslint'
-import eslintPkg from 'eslint/package.json'
-import semver from 'semver'
 
 const ruleTester = new RuleTester()
     , rule = require('rules/prefer-default-export')
@@ -133,18 +131,7 @@ ruleTester.run('prefer-default-export', rule, {
 });
 
 context('Typescript', function() {
-  // Typescript
-  const parsers = [require.resolve('babel-eslint')];
-
-  if (semver.satisfies(eslintPkg.version, '>=4.0.0 <6.0.0')) {
-    parsers.push(require.resolve('typescript-eslint-parser'));
-  }
-
-  if (semver.satisfies(eslintPkg.version, '>5.0.0')) {
-    parsers.push(require.resolve('@typescript-eslint/parser'));
-  }
-
-  parsers.forEach((parser) => {
+  getNonDefaultParsers().forEach((parser) => {
     const parserConfig = {
       parser: parser,
       settings: {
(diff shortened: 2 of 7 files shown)


sharmilajesupaul:
I came across another the failing case with a single named export 

i.e. this errors: 
```js
type foo = string;
export { foo }
```

could we merge those fixes in a separate PR?

cc. @ljharb

ljharb:
@sharmilajesupaul if this PR, once rebased, passes tests, then it's up to you if you want to defer those fixes to another PR - but it's also great if you can include them in this PR too :-)

ljharb:
@sharmilajesupaul i've fixed the test issues around eslint and typescript parser versions; but now there's some seemingly legit failures. Can you look into them?

tests/src/rules/prefer-default-export.js
```diff
@@ -129,4 +128,62 @@ ruleTester.run('prefer-default-export', rule, {
...
     }),
   ],
-})
+});
+
+context('Typescript', function() {
+  // Typescript
+  const parsers = ['typescript-eslint-parser', 'babel-eslint'];
+
+  if (semver.satisfies(eslintPkg.version, '>5.0.0')) {
```
ljharb: missing an `import semver from 'semver'` and `import eslintPkg from 'eslint/package.json'` at the top, i think?

sharmilajesupaul: yup just updated, forgot these in my `git add -p`

tests/src/rules/prefer-default-export.js
```diff
@@ -129,4 +130,62 @@ ruleTester.run('prefer-default-export', rule, {
...
+    parsers.push('@typescript-eslint/parser');
```
ljharb: eslint 6 now requires absolute paths here
```suggestion
    parsers.push(require.resolve('@typescript-eslint/parser'));
```

tests/src/rules/prefer-default-export.js
```diff
@@ -129,4 +130,62 @@ ruleTester.run('prefer-default-export', rule, {
       }],
     }),
   ],
-})
+});
+
+context('Typescript', function() {
+  // Typescript
+  const parsers = ['typescript-eslint-parser', 'babel-eslint'];
```
ljharb: eslint 6 now requires absolute paths here
```suggestion
  const parsers = [require.resolve('typescript-eslint-parser'), require.resolve('babel-eslint')];
```

tests/src/rules/prefer-default-export.js
```diff
@@ -129,4 +130,62 @@ ruleTester.run('prefer-default-export', rule, {
       }],
     }),
   ],
-})
+});
+
+context('Typescript', function() {
+  // Typescript
+  const parsers = [require.resolve('typescript-eslint-parser'), require.resolve('babel-eslint')];
```
ljharb: oops, i think typescript-eslint-parser can only work in eslint < 6

sharmilajesupaul: that makes sense, i updated this with another conditional that checks for <v6

src/rules/prefer-default-export.js
```diff
@@ -47,9 +47,16 @@ module.exports = {
         // if there are specifiers, node.declaration should be null
         if (!node.declaration) return
 
-        // don't count flow types exports
+        // don't warn on single type aliases or declarations
         if (node.exportKind === 'type') return
 
+        if (
+          node.declaration.type === 'TSTypeAliasDeclaration' ||
+          node.declaration.type === 'TypeAlias'
```
lencioni: @sharmilajesupaul @brieb should this also ignore interfaces? I believe the Babel/TypeScript AST type is `InterfaceDeclaration` and the typescript eslint parser type is `TSInterfaceDeclaration`.

lencioni: https://github.com/benmosher/eslint-plugin-import/pull/1425

PR #1386, opened 2019-06-20
`order` rule: add pathGroups option to add support to order by paths

Mairu:
Well there are multiple issues and pull requests open regarding the ordering of groups by paths.
Because I saw that the proposed approach in https://github.com/benmosher/eslint-plugin-import/issues/795 was "accepted" I implemented it this way.

But if @ljharb has changed his mind in the mean time and would prefer something like custom groups as proposed in https://github.com/benmosher/eslint-plugin-import/issues/1015#issuecomment-388573720 I can also rework this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - support `parseForESLint` from custom parser ([#1435], thanks [@JounQin])
 - [`no-extraneous-dependencies`]: Implement support for [bundledDependencies](https://npm.github.io/using-pkgs-docs/package-json/types/bundleddependencies.html) ([#1436], thanks [@schmidsi]))
 - [`no-unused-modules`]: add flow type support ([#1542], thanks [@rfermann])
+- [`order`]: Adds support for pathGroups to allow ordering by defined patterns ([#795], [#1386], thanks [@Mairu])
 
 ### Fixed
 - [`default`]: make error message less confusing ([#1470], thanks [@golopot])
@@ -644,6 +645,7 @@ for info on changes for earlier releases.
 [#1401]: https://github.com/benmosher/eslint-plugin-import/pull/1401
 [#1393]: https://github.com/benmosher/eslint-plugin-import/pull/1393
 [#1389]: https://github.com/benmosher/eslint-plugin-import/pull/1389
+[#1386]: https://github.com/benmosher/eslint-plugin-import/pull/1386
 [#1377]: https://github.com/benmosher/eslint-plugin-import/pull/1377
 [#1375]: https://github.com/benmosher/eslint-plugin-import/pull/1375
 [#1372]: https://github.com/benmosher/eslint-plugin-import/pull/1372
@@ -788,6 +790,7 @@ for info on changes for earlier releases.
 [#863]: https://github.com/benmosher/eslint-plugin-import/issues/863
 [#842]: https://github.com/benmosher/eslint-plugin-import/issues/842
 [#839]: https://github.com/benmosher/eslint-plugin-import/issues/839
+[#795]: https://github.com/benmosher/eslint-plugin-import/issues/795
 [#793]: https://github.com/benmosher/eslint-plugin-import/issues/793
 [#720]: https://github.com/benmosher/eslint-plugin-import/issues/720
 [#717]: https://github.com/benmosher/eslint-plugin-import/issues/717
@@ -1025,3 +1028,4 @@ for info on changes for earlier releases.
 [@Taranys]: https://github.com/Taranys
 [@maxmalov]: https://github.com/maxmalov
 [@marcusdarmstrong]: https://github.com/marcusdarmstrong
+[@Mairu]: https://github.com/Mairu
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -94,6 +94,32 @@ You can set the options like this:
 "import/order": ["error", {"groups": ["index", "sibling", "parent", "internal", "external", "builtin"]}]
 ```
 
+### `pathGroups: [array of objects]`:
+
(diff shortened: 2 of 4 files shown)


docs/rules/order.md
```diff
@@ -94,6 +94,31 @@ You can set the options like this:
...
+
+To be able so group by paths mostly needed with aliases pathGroups can be defined.
+
+Properties of the objects
+
+| property | required | type    | description   |
+|----------|:--------:|---------|---------------|  
+| pattern  |     x    | string  | minimatch pattern for the paths to be in this group (will not be used for builtins or externals) |
+| group    |     x    | string  | one of the allowed groups, the pathGroup will be positioned relative to this group |
+| position |          | integer | defines where around the group the pathGroup will be positioned <br> negative -> before, 0 = like the group, positive -> after <br> default: 1 <br> maximal magnitude: 1000 |
```
ljharb: i think "like the group" should be the default; and i think that the alternative should be "before" or "after", and ordering in the config array itself should determine relative priority.

src/rules/order.js
```diff
@@ -242,9 +243,28 @@ function makeOutOfOrderReport(context, imported) {
 
 // DETECTING
 
+function computePathRank(ranks, pathGroups, path) {
+  for (const { pattern, patternOptions, group, position = 1 } of pathGroups) {
```
ljharb: can we avoid using for..of here?

src/rules/order.js
```diff
@@ -242,9 +243,28 @@ function makeOutOfOrderReport(context, imported) {
...
+    if (minimatch(path, pattern, patternOptions ||{ nocomment: true })) {
```
ljharb:
```suggestion
    if (minimatch(path, pattern, patternOptions || { nocomment: true })) {
```

docs/rules/order.md
```diff
@@ -94,6 +94,32 @@ You can set the options like this:
...
+### `pathGroups: [array of objects]`:
+
+To be able so group by paths mostly needed with aliases pathGroups can be defined.
+
+Properties of the objects
+
+| property       | required | type   | description   |
+|----------------|:--------:|--------|---------------|  
+| pattern        |     x    | string | minimatch pattern for the paths to be in this group (will not be used for builtins or externals) |
+| patternOptions |          | object | options for minimatch, default: { nocomment: true } |
```
ljharb: there seem to be no tests using this option - can you add some? without them i have no idea why this is needed.

Mairu: Added a test using another option. I could also see to add this as default options and remove the options. But perhaps somebody has a usecase for negated pattern. So I though it would not hurt to give the possibility to adjust the options.

Mairu:
Anything else needed for this to be merged? @ljharb

docs/rules/order.md
```diff
@@ -94,6 +94,32 @@ You can set the options like this:
...
+| group          |     x    | string | one of the allowed groups, the pathGroup will be positioned relative to this group |
+| position       |          | string | defines where around the group the pathGroup will be positioned, can be 'after' or 'before', if not provided pathGroup will be positioned like the group |
+
+```json
+{
+  "import/order": ["error", {
+    "pathGroups": [
+      {
+        "pattern": "~/**",
+        "external": "external"
```
cartogram:
```suggestion
        "group": "external"
```

fundlg:
+1, need this feature

osdiab:
Any reason this has stalled? Would love to use this

zaqqaz:
> Any reason this has stalled? Would love to use this @ljharb


@benmosher, @ljharb ⬆️

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`no-namespace`]: Make rule fixable ([#1401], thanks [@TrevorBurnham])
 - support `parseForESLint` from custom parser ([#1435], thanks [@JounQin])
 - [`no-extraneous-dependencies`]: Implement support for [bundledDependencies](https://npm.github.io/using-pkgs-docs/package-json/types/bundleddependencies.html) ([#1436], thanks [@schmidsi]))
+- [`order`]: Adds support for pathGroups to allow ordering by defined patterns ([#795], thanks [@Mairu]
```
```suggestion
- [`order`]: Adds support for pathGroups to allow ordering by defined patterns ([#795], thanks [@Mairu])
```

You'll need to add links for 795, 1386, and for @Mairu, in this file.

Mairu: Added them, though that was some kind of magic :)

osdiab:
ping on if this is gonna get merged...

osdiab:
@benmosher, @ljharb ⬆️

oliverdolgener:
I'd really like to see this getting merged :D

osdiab:
I was wondering if I was misusing this new rule.

I configured my `import/order` like so:

```js
{
  rules: {
    "import/order": [
      "warn",
      {
        // i disallow all relative imports, so I'm using `parent` to mean other packages in my monorepo,
        // and sibling to be auxiliary code in my repo
        "groups": ["builtin", "external", "parent", "internal", "index", "sibling"],
        "pathGroups": [
          {
            "pattern": "@app/**",
            "group": "internal"
          },
          {
            "pattern": "@my.company/**",
            "group": "parent"
          },
          {
            "pattern": "__tests__/**",
            "group": "sibling"
          },
        ]
      }
    ]
  }
}
```

But when I have this code, it doesn't mark it as an error

```tsx
import { RecurringStatus } from "@my.company/common/es6/entity/types"; // this should be in the middle
import { ChartistGraph } from "@app/components/common/ChartistGraph"; // this should be last
import { ILineChartOptions } from "chartist"; // this should be first
```

osdiab:
I edited the above for my second try

ljharb:
Please file a new issue so we can track it properly :-)

jgabriele:
Thank you ❤️!

PR #1393, opened 2019-06-23
Support ESLint v6

### LINKED ISSUE #1362: support for eslint 6
Tests currently run, but are allowed to fail, on the eslint v6 alpha.

We should figure out what changes are needed to make them pass.

(see #1361)

sheepsteak:
The main issues are:
- [x] Use `require.resolve` when passing the `parser` option to `RuleTester` (https://eslint.org/docs/user-guide/migrating-to-6.0.0#rule-tester-parser)
- [x] Add `ecmaVersion` when `sourceType` is `module` (https://eslint.org/docs/user-guide/migrating-to-6.0.0#-the-default-parser-now-validates-options-more-strictly)
- [x] Transform output of the `FileEnumerator` into what the `import/no-unused-modules` rule expects
- [x] `@typescript-eslint/parser` is using an internal module (`eslint/lib/util/traverser`) which has been removed in ESLint@6 and therefore causing an error. It's been fixed in https://github.com/typescript-eslint/typescript-eslint/pull/628 but not been released yet. **(Updated to canary, thanks @golopot)**
- [x] Same as above for `typescript-eslint-parser` but this is an archived package. As I write this I realise I should probably use the `testVersion` function so that tests for this are skipped on `eslint>=6`

To test:
```
npm i eslint@6
npm test
```

The only failing tests should be related to the TypeScript issue above.

Related to https://github.com/airbnb/javascript/issues/2036.
Fixes #1362.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
   "homepage": "https://github.com/benmosher/eslint-plugin-import",
   "devDependencies": {
     "@eslint/import-test-order-redirect-scoped": "file:./tests/files/order-redirect-scoped",
-    "@typescript-eslint/parser": "^1.5.0",
+    "@typescript-eslint/parser": "1.10.3-alpha.13",
     "babel-cli": "^6.26.0",
     "babel-core": "^6.26.3",
     "babel-eslint": "^8.2.6",
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   },
   "homepage": "https://github.com/benmosher/eslint-plugin-import",
   "devDependencies": {
+    "@eslint/import-test-order-redirect-scoped": "file:./tests/files/order-redirect-scoped",
     "@typescript-eslint/parser": "^1.5.0",
     "babel-cli": "^6.26.0",
     "babel-core": "^6.26.3",
@@ -58,13 +59,12 @@
     "chai": "^4.2.0",
     "coveralls": "^3.0.2",
     "cross-env": "^4.0.0",
-    "eslint": "2.x - 5.x",
+    "eslint": "2.x - 6.x",
     "eslint-import-resolver-node": "file:./resolvers/node",
     "eslint-import-resolver-typescript": "^1.0.2",
     "eslint-import-resolver-webpack": "file:./resolvers/webpack",
-    "eslint-module-utils": "file:./utils",
     "eslint-import-test-order-redirect": "file:./tests/files/order-redirect",
-    "@eslint/import-test-order-redirect-scoped": "file:./tests/files/order-redirect-scoped",
+    "eslint-module-utils": "file:./utils",
     "eslint-plugin-import": "2.x",
     "linklocal": "^2.8.2",
     "mocha": "^3.5.3",
@@ -77,7 +77,7 @@
     "typescript-eslint-parser": "^22.0.0"
   },
   "peerDependencies": {
-    "eslint": "2.x - 5.x"
+    "eslint": "2.x - 6.x"
   },
   "dependencies": {
     "array-includes": "^3.0.3",
--- a/src/rules/no-unused-modules.js
+++ b/src/rules/no-unused-modules.js
@@ -15,7 +15,10 @@ try {
   var FileEnumerator = require('eslint/lib/cli-engine/file-enumerator').FileEnumerator
   listFilesToProcess = function (src) {
     var e = new FileEnumerator()
-    return Array.from(e.iterateFiles(src))
+    return Array.from(e.iterateFiles(src), ({ filePath, ignored }) => ({
+      ignored,
+      filename: filePath,
+    }))
   }
 } catch (e1) {
   try {
--- a/tests/src/rules/default.js
+++ b/tests/src/rules/default.js
@@ -28,19 +28,19 @@ ruleTester.run('default', rule, {
 
     // es7 export syntax
     test({ code: 'export bar from "./bar"'
-         , parser: 'babel-eslint' }),
+         , parser: require.resolve('babel-eslint') }),
     test({ code: 'export { default as bar } from "./bar"' }),
     test({ code: 'export bar, { foo } from "./bar"'
-         , parser: 'babel-eslint' }),
+         , parser: require.resolve('babel-eslint') }),
     test({ code: 'export { default as bar, foo } from "./bar"' }),
     test({ code: 'export bar, * as names from "./bar"'
-         , parser: 'babel-eslint' }),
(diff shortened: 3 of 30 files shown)


sheepsteak:
It's also having some sort of whitespace issue with a couple of tests as you can see here - https://ci.appveyor.com/project/benmosher/eslint-plugin-import/builds/25480510/job/i7a5k9x6j6s1hg8a#L3072. I think I've fixed it and then it comes back again haha

src/rules/no-unused-modules.js
```diff
@@ -15,7 +15,10 @@ try {
   var FileEnumerator = require('eslint/lib/cli-engine/file-enumerator').FileEnumerator
   listFilesToProcess = function (src) {
     var e = new FileEnumerator()
-    return Array.from(e.iterateFiles(src))
+    return Array.from(e.iterateFiles(src)).map(({ filePath, ignored })=>({
```
ljharb: Array.from takes a second mapper argument:
```suggestion
    return Array.from(e.iterateFiles(src), ({ filePath, ignored })=>({
```

src/rules/no-unused-modules.js
```diff
@@ -347,9 +350,9 @@ module.exports = {
           context.report(
             node,
             `exported declaration '${value}' not used within other modules`
-          )
-        }
-      } else {
+            )
```
ljharb: this indentation is wrong
```suggestion
          )
```

src/rules/no-unused-modules.js
```diff
@@ -347,9 +350,9 @@ module.exports = {
...
+          }
```
ljharb:
```suggestion
        }
```

src/rules/no-unused-modules.js
```diff
@@ -347,9 +350,9 @@ module.exports = {
...
+        } else {
```
ljharb:
```suggestion
      } else {
```

tests/src/rules/default.js
```diff
@@ -28,19 +28,19 @@ ruleTester.run('default', rule, {
 
     // es7 export syntax
     test({ code: 'export bar from "./bar"'
-         , parser: 'babel-eslint' }),
+         , parser: require.resolve('babel-eslint') }),
```
ljharb: why is this needed? eslint should be resolving whatever’s passed to parser

tests/src/rules/default.js
```diff
@@ -67,37 +67,37 @@ ruleTester.run('default', rule, {
...
     test({
-      code: "import bar from './default-export-from-ignored.js';",
+      code: 'import bar from \'./default-export-from-ignored.js\';',
       settings: { 'import/ignore': ['common'] },
-      parser: 'babel-eslint',
+      parser: require.resolve('babel-eslint'),
     }),
     test({
-      code: "export bar from './default-export-from-ignored.js';",
+      code: 'export bar from \'./default-export-from-ignored.js\';',
```
ljharb: Please revert these changes; the double quotes are intentional to avoid having to deal with escaping backslashes

ljharb:
Are you perhaps running prettier on autosave in your editor or something? If so, disable that - prettier should only be run via eslint autofix, and only on repos that use it.

package.json
```diff
@@ -47,6 +47,7 @@
   },
   "homepage": "https://github.com/benmosher/eslint-plugin-import",
   "devDependencies": {
+    "@eslint/import-test-order-redirect-scoped": "file:./tests/files/order-redirect-scoped",
     "@typescript-eslint/parser": "^1.5.0",
```
golopot:
```suggestion
     "@typescript-eslint/parser": "1.10.3-alpha.13",
```
Regarding the `@typescript-eslint/parser` bug, `@typescript-eslint/parser` releases every commit to master as a canary version. You can use it for now.

tests/src/rules/default.js
sheepsteak: https://eslint.org/docs/user-guide/migrating-to-6.0.0#-ruletester-now-requires-an-absolute-path-on-parser-option

If I remove it I get an error in line with the above:
```
1) default valid export bar from "./bar":

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Parsers provided as strings to RuleTester must be absolute paths
      + expected - actual

      -false
      +true
```

src/rules/no-unused-modules.js
```diff
@@ -15,7 +15,10 @@ try {
   var FileEnumerator = require('eslint/lib/cli-engine/file-enumerator').FileEnumerator
   listFilesToProcess = function (src) {
     var e = new FileEnumerator()
-    return Array.from(e.iterateFiles(src))
+    return Array.from(e.iterateFiles(src), ({ filePath, ignored })=>({
```
ljharb:
```suggestion
    return Array.from(e.iterateFiles(src), ({ filePath, ignored }) => ({
```

sheepsteak:
Sorry, about the double quote changes. I've reverted all of them. I have Prettier set to only work when there is a `.prettierrc` present so it not sure what's causing that 🤷‍♂️

golopot:
I suppose it is eslint in your editor doing that changes. The ci does not lint over `./tests`, so there are many lines in `./tests` not passing eslint.

sheepsteak:
Added one last commit to only run tests for `typescript-eslint-parser` when ESLint is `<6` where appropriate given that it's archived and will probably never get the required fix.

Would you like me to update Travis CI config to use `6.0.0` instead of `6.0.0-alpha`?

ljharb:
Yes, please.

sheepsteak:
@ljharb OK I had to one more thing and make `testVersion` take a function instead of an object so that `require.resolve` didn't get run and throw an error. This happens when trying to resolve the TypeScript parsers on test runs where they've been removed in `dep-time-travel.sh`. All tests pass on Travis now 🎉

frederikhors:
Please @ljharb can you merge and release this. This is urgent now with eslint 6 in production everywhere (unfortunately).

`"eslint": "2.x - 6.x"`

ljharb:
@frederikhors If `npm ls` exits nonzero, your dep graph is invalid, so eslint 6 *shouldn't* be in production when you have peer dep conflicts with it. If it's urgent for you, it's a problem of your own making.

frederikhors:
@ljharb the problem is not mine but it is here: https://github.com/prettier/prettier-vscode/issues/672

ljharb:
Their problem is the same; upgrading to eslint v6 prior to peer deps upgrading is invalid.

frederikhors:
@ljharb still same problem today.

PR #1401, opened 2019-06-29
 Make no-namespace rule fixable

TrevorBurnham:
This PR adds support for `--fix` to the `no-namespace` rule, in the case where all references to the namespace can be statically resolved:

### Input

```js
import * as foo from 'foo';
florp(foo.bar);
florp(foo.baz);
```

### Output

```js
import { bar, baz } from 'foo';
florp(bar);
florp(baz);
```

The motivation for this is to make it easier to opt in to `no-namespace` in codebases that are already using this pattern. It sure beats having to do it all by hand! 😄 

Note that this will only work under ESLint 5+, as older versions don't allow for traversing upward from variable references. That is, in this code `identifier.parent` is `undefined`. I haven't been able to figure out a good way to support this under ESLint 4 (let me know if you have any ideas!), so I've added a note for this in the rule's documentation.

### DIFF
--- a/src/rules/no-namespace.js
+++ b/src/rules/no-namespace.js
@@ -16,13 +16,143 @@ module.exports = {
     docs: {
       url: docsUrl('no-namespace'),
     },
+    fixable: 'code',
   },
 
   create: function (context) {
     return {
       'ImportNamespaceSpecifier': function (node) {
-        context.report(node, `Unexpected namespace import.`)
+        const scopeVariables = context.getScope().variables
+        const namespaceVariable = scopeVariables.find((variable) =>
+          variable.defs[0].node === node
+        )
+        const namespaceReferences = namespaceVariable.references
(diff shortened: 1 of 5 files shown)


src/rules/no-namespace.js
```diff
@@ -16,13 +16,150 @@ module.exports = {
...
     }
   },
 }
+
+/**
+ * @param {Identifier[]} namespaceIdentifiers
+ * @returns {boolean} `true` if the namespace variable is more than just a glorified constant
+ */
+function usesNamespaceAsObject(namespaceIdentifiers) {
+  for (let i = 0; i < namespaceIdentifiers.length; i++) {
```
ljharb: can this be written as a `.some` instead of a for loop?

TrevorBurnham: Sure thing. ✅

Issue #1327, opened 2019-04-14
Exclude package main files by default from `no-unused-modules`


ehmicky:
`unusedExports: true` with the `no-unused-modules` rule should never report files that are exported as the main file (`main`, `browser` or `bin` fields in `package.json`). What do you think?

ljharb:
That sounds like a good enhancement, no need for it to be behind an option, even.

rfermann:
What package.json should be used for this purpose? 

- the package.json from the working directory,
- the package.json from the folder specified as `src` in the rule options (if existing) or
- the package.json closest to the file currently being linted (if existing and not the same as in the working directory)?

ljharb:
For a given file, the closest package.json to it, that has a main, does not have private true, and points to it, marks it as used.

ehmicky:
Thanks for implementing this @ljharb! :tada:

ljharb:
Thank @rfermann :-)

PR #1404, opened 2019-07-02
Exclude pkg main files from `no-unused-modules`

rfermann:
Implements a solution for #1327

### DIFF
--- a/docs/rules/no-unused-modules.md
+++ b/docs/rules/no-unused-modules.md
@@ -105,7 +105,8 @@ export function doAnything() {
 export default 5 // will not be reported
 ```
 
-
+#### Important Note
+Exports from files listed as a main file (`main`, `browser`, or `bin` fields in `package.json`) will be ignored by default. This only applies if the `package.json` is not set to `private: true`
 
 ## When not to use
 
--- a/src/rules/no-unused-modules.js
+++ b/src/rules/no-unused-modules.js
@@ -7,6 +7,10 @@
 import Exports from '../ExportMap'
 import resolve from 'eslint-module-utils/resolve'
 import docsUrl from '../docsUrl'
+import { dirname, join } from 'path'
+import readPkgUp from 'read-pkg-up'
+import values from 'object.values'
+import includes from 'array-includes'
 
 // eslint/lib/util/glob-util has been moved to eslint/lib/util/glob-utils with version 5.3
 // and has been moved to eslint/lib/cli-engine/file-enumerator in version 6
@@ -80,7 +84,7 @@ const prepareImportsAndExports = (srcFiles, context) => {
     if (currentExports) {
       const { dependencies, reexports, imports: localImportList, namespace  } = currentExports
 
-      // dependencies === export * from 
+      // dependencies === export * from
       const currentExportAll = new Set()
       dependencies.forEach(value => {
         currentExportAll.add(value().path)
@@ -146,7 +150,7 @@ const prepareImportsAndExports = (srcFiles, context) => {
 }
 
 /**
- * traverse through all imports and add the respective path to the whereUsed-list 
+ * traverse through all imports and add the respective path to the whereUsed-list
  * of the corresponding export
  */
 const determineUsage = () => {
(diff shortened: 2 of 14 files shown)


docs/rules/no-unused-modules.md
```diff
@@ -105,7 +105,8 @@ export function doAnything() {
 export default 5 // will not be reported
 ```
 
-
+#### Important Note
+Exports from files listed as a main file (`main`, `browser` or `bin` fields in package.json) will be ignored by default. This does only apply, if the package.json is not set to `private: true`
```
ljharb:
```suggestion
Exports from files listed as a main file (`main`, `browser`, or `bin` fields in package.json) will be ignored by default. This does only apply, if the package.json is not set to `private: true`
```

src/rules/no-unused-modules.js
```diff
@@ -201,6 +203,42 @@ const newNamespaceImportExists = specifiers =>
...
+  }
+
+  if (pkg.bin && typeof pkg.bin === 'object') {
+    const binFiles = Object.values(pkg.bin).map(value => join(basePath, value))
+    if (binFiles.includes(file)) {
+      return true
+    }
+  }
+
+  if (pkg.browser) {
```
ljharb: The browser field can also be an object.

rfermann: Thanks for this info. This wasn't mentioned in the [npm documentation](https://docs.npmjs.com/files/package.json). I will quickly add this part.

ljharb: https://github.com/defunctzombie/package-browser-field-spec

src/rules/no-unused-modules.js
```diff
@@ -201,6 +203,49 @@ const newNamespaceImportExists = specifiers =>
...
 
+const fileIsInPkg = file => {
+  const { path, pkg } = readPkgUp.sync({cwd: file, normalize: false})
+  const basePath = dirname(path)
+
+  if (pkg.private === true) {
+    return false
+  }
+
+  if (pkg.bin && typeof pkg.bin === 'string') {
```
ljharb: seems like we can consolidate the logic for “check string field X” and “check object field Y”, so t doesn’t have to be repeated for main/bin/browser and bin/browser?

rfermann: Good hint. I mad the logic a bit more generic with 3ed222d.

src/rules/no-unused-modules.js
```diff
@@ -207,38 +207,39 @@ const fileIsInPkg = file => {
...
-    if (browserFiles.includes(file)) {
+  if (pkg.browser) {
+    if (checkPkgField(pkg.browser)) {
       return true
     }
   }
 
   if (pkg.main) {
-    if (join(basePath, pkg.main) === file) {
+    if (checkPkgField(pkg.main)) {
```
ljharb: the only remaining problem here is that main can’t be an object :-/

rfermann: Maybe I don't see the issue here. `main` will either return `true` within the first `if` branch or just return at the end of the function.
Would it be better to return `false` in the first `if` branch to make `main` not even pass the `if` branch for object handling?

ljharb: Yes, that's what I'm suggesting. an object "main" field is just invalid and shouldn't be considered.

docs/rules/no-unused-modules.md
```diff
@@ -105,7 +105,8 @@ export function doAnything() {
 export default 5 // will not be reported
 ```
 
-
+#### Important Note
+Exports from files listed as a main file (`main`, `browser`, or `bin` fields in package.json) will be ignored by default. This does only apply, if the package.json is not set to `private: true`
```
ljharb:
```suggestion
Exports from files listed as a main file (`main`, `browser`, or `bin` fields in `package.json`) will be ignored by default. This only applies if the `package.json` is not set to `private: true`
```

tests/files/no-unused-modules/main/index.js
```diff
@@ -0,0 +1 @@
+export const main = 'main'
```
ljharb: please ensure every file has a trailing newline

rfermann: I always miss that. In my other projects it gets added automatically by ESLint/Prettier...
Can we probably add `eol-last` to the eslint-config of this project?

ljharb: Sure, sounds great.

ljharb: (but, your editor should add a trailing newline by default in every file, entirely unrelated to eslint)

src/rules/no-unused-modules.js
```diff
@@ -201,6 +203,58 @@ const newNamespaceImportExists = specifiers =>
...
+  const basePath = dirname(path)
+
+  const checkPkgFieldString = pkgField => {      
+    if (join(basePath, pkgField) === file) {
+        return true
+      }
+  }
+
+  const checkPkgFieldObject = pkgField => {    
+      const pkgFieldFiles = Object.values(pkgField).map(value => join(basePath, value))
```
ljharb: `Object.values` is not available on all supported nodes; please use https://npmjs.com/object.values instead

rfermann: I would rather use `_.values(object)`, if it is okay for you. Lodash already is a project dependency, with object.values we would add a new dependency.

ljharb: I'd prefer to use the standard method; adding a dependency isn't a problem.

src/rules/no-unused-modules.js
```diff
@@ -201,6 +204,58 @@ const newNamespaceImportExists = specifiers =>
...
+
+  const checkPkgFieldString = pkgField => {
+    if (join(basePath, pkgField) === file) {
+        return true
+      }
+  }
+
+  const checkPkgFieldObject = pkgField => {
+      const pkgFieldFiles = values(pkgField).map(value => join(basePath, value))
+      if (pkgFieldFiles.includes(file)) {
```
ljharb: includes also isn't in node 4; i'll update the PR with a dep for that too.

PR #1409, opened 2019-07-08
fix: Improve parse perf when using `@typescript-eslint/parser`

### LINKED ISSUE #1408: Slowness when using typescript parser with `project` option
A few of our users are reporting slowness when using our parser (`@typescript-eslint/parser`) with this plugin.

Having a quick look at the codebase, I notice that you invoke the parser yourself so you can gather information about other files.

https://github.com/benmosher/eslint-plugin-import/blob/05c3935/utils/parse.js#L9-L39

As part of this, you pass in the `parserOptions` from the root config. If the `project`/`projects` option is set in the `parserOptions`, it will cause our parser to do a complete typechecking parse cycle, which is often rather slow.

I'm still gathering information from users, we have some caching and such in the parser to attempt to circumvent this, but I think there's a double parse happening because you have a new instance of everything.

My hypothesis is that if we add the following lines at line 25 in the above file, it will make your parser invocation be a single-file, non type-checked parse, which should obviously be fast!

```diff
+  delete parserOptions.project
+  delete parserOptions.projects
```

Once I get some data from users I'll happily raise a PR for this - I just wanted to bring it to your attention first.

bradzacher:
Fixes #1408
See the above issue for justification.

### DIFF
--- a/utils/parse.js
+++ b/utils/parse.js
@@ -31,7 +31,14 @@ exports.default = function parse(path, content, context) {
   // provide the `filePath` like eslint itself does, in `parserOptions`
   // https://github.com/eslint/eslint/blob/3ec436ee/lib/linter.js#L637
   parserOptions.filePath = path
-
+  
+  // @typescript-eslint/parser will parse the entire project with typechecking if you provide
+  // "project" or "projects" in parserOptions. Removing these options means the parser will
+  // only parse one file in isolate mode, which is much, much faster.
+  // https://github.com/benmosher/eslint-plugin-import/issues/1408#issuecomment-509298962
+  delete parserOptions.project;
+  delete parserOptions.projects;
+  
   // require the parser relative to the main module (i.e., ESLint)
   const parser = moduleRequire(parserPath)
 


utils/parse.js
```diff
@@ -31,7 +31,14 @@ exports.default = function parse(path, content, context) {
   // provide the `filePath` like eslint itself does, in `parserOptions`
   // https://github.com/eslint/eslint/blob/3ec436ee/lib/linter.js#L637
   parserOptions.filePath = path
-
+  
+  // @typescript-eslint/parser will parse the entire project with typechecking if you provide
+  // "project" or "projects" in parserOptions. Removing these options means the parser will
+  // only parse one file in isolate mode, which is much, much faster.
+  // https://github.com/benmosher/eslint-plugin-import/issues/1408#issuecomment-509298962
+  delete parserOptions.project;
```
deser: Seems like a dirty hack :)

bradzacher: `@typescript-eslint/parser` does exactly what it is told to do.
If you give it `project` or `projects`, it will do an entire workspace parse so it can provide complete type information.

This is a bit of a hack, sure, but then again ESLint doesn't provide a mechanism for parsing files up-front / accessing parsed files, so `eslint-plugin-import` having to do a manual parse itself is a hack as well.

deser: Should we rise the feature proposal?)

bradzacher: It's been talked about a bit by the eslint folks, but it's not high priority. We talked about it with them when we were adding type information to the parser.

Ultimately it's a pretty big edge case - the _vast_ majority of plugins (and parsers) do not need to break module boundaries.

ljharb:
LGTM; i'm not sure if appveyor is failing on master or only on this PR.

bradzacher:
@ljharb - appveyor was passing with my first commit - it appears using `Object.assign` breaks some tests.
It's a bit hard to investigate, because locally no matter what I do, there are many failing tests locally for some reason...
I'll just revert the commit.

ljharb:
That's a shame. We can fix that in a followup PR.

benmosher:
No need to re-`Object.assign`, the object is already shallowly cloned on [line 19](https://github.com/benmosher/eslint-plugin-import/pull/1409/files#diff-0b7444adb90a4cb9726b357167de6e97L19). LGTM! thanks!

deser:
Guys, do you have vision when it would be delivered\published?

ljharb:
@benmosher it’s still better to create a new object all at once instead of mutating it at any point.

benmosher:
fair enough, but even if we used destructuring to remove the fields (iff this file was being transpiled) the Babel'd version would start with an empty object and add fields. 

given that we can't predict all the fields we need to pass through, this seems acceptable to me. I'm good to merge if you are, @ljharb.

ljharb:
Definitely, we can look into creating the object differently in a follow up. Once tests are passing, it’s good to go.

bradzacher:
> Once tests are passing, it’s good to go.

@ljharb - having a look, the travis CI tests that are failing are the same ones that are failing in master right now (tests against eslint v2/v3)

ljharb:
We should fix those in master first, then, and rebates this PR

deser:
Or just merge and make all happy!)

deser:
If tests are failing in master can we just merge this and release?

ljharb:
No, we shouldn’t merge and must not release without passing tests.

deser:
ok. sorry

benmosher:
sorry @deser, test failures were on me but they are fixed now 👍🏼

bradzacher:
AppVeyor failed during install time on node8.
It looks like a transient error, but I don't have permission to retry the build.
Could someone with permissions please retry that build?

```
[00:00:33] npm ERR! code Z_BUF_ERROR
[00:00:33] npm ERR! errno -5
[00:00:33] npm ERR! zlib: unexpected end of file
```

hayes:
thanks for getting this merged!

@ljharb can you cut a release of [eslint-module-utils](https://github.com/benmosher/eslint-plugin-import/blob/master/utils/package.json) as well? it looks like only [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import/blob/master/package.json) got released after this change

ljharb:
Oops, yes will do today.

ljharb:
v2.4.1 of eslint-module-utils is published.

PR #1411, opened 2019-07-08
[WIP] Fix dynamic-import-chunkname validation regex

### LINKED ISSUE #1407: dynamic-import-chunkname does not handle webpackInclude/webpackExclude comments correctly
`dynamic-import-chunkname` rule does not handle `webpackInclude`/`webpackExclude` comments correctly.

`.eslintrc.js`:
```js
module.exports = {
	parser: "babel-eslint",
	parserOptions: {
		ecmaVersion: 2018,
		sourceType: "module",
	},
	plugins: ["import"],
	rules: {
		"import/dynamic-import-chunkname": "error",
	},
};
```
`index.js`
```js
import(
	/* webpackChunkName: "foo" */
	/* webpackInclude: /foo/ */
	"./foo"
);
```
Output:
```
  1:1  error  dynamic imports require a leading comment in the form /* webpackChunkName: "[0-9a-zA-Z-_/.]+",? */  import/dynamic-import-chunkname

```

Package name|Package version
-|-
`eslint` | `6.0.1`
`eslint-plugin-import` | `2.18.0`

Minimal reproducible example: https://github.com/vkrol/eslint-plugin-import-webpack-comment-with-regex-bug

vkrol:
Fixes #1407.

### DIFF
--- a/src/rules/dynamic-import-chunkname.js
+++ b/src/rules/dynamic-import-chunkname.js
@@ -30,7 +30,7 @@ module.exports = {
     const { webpackChunknameFormat = '[0-9a-zA-Z-_/.]+' } = config || {}
 
     const paddedCommentRegex = /^ (\S[\s\S]+\S) $/
-    const commentStyleRegex = /^( \w+: ("[^"]*"|\d+|false|true),?)+ $/
+    const commentStyleRegex = /^( \w+: ("[^"]*"|\/.*\/|\d+|false|true),?)+ $/
     const chunkSubstrFormat = ` webpackChunkName: "${webpackChunknameFormat}",? `
     const chunkSubstrRegex = new RegExp(chunkSubstrFormat)
 
--- a/tests/src/rules/dynamic-import-chunkname.js
+++ b/tests/src/rules/dynamic-import-chunkname.js
@@ -34,6 +34,30 @@ ruleTester.run('dynamic-import-chunkname', rule, {
     },
     {
       code: `dynamicImport(
+        /* webpackInclude: /some-regex/ */
+        /* webpackChunkName: "someModule" */
+        'test'
+      )`,
+      options,
+    },
+    {
+      code: `dynamicImport(
+        /* webpackChunkName: "someModule" */
+        /* webpackInclude: /some-regex/ */
+        'test'
+      )`,
+      options,
+    },
+    {
+      code: `dynamicImport(
+        /* webpackExclude: /some-regex/ */
+        /* webpackChunkName: "someModule" */
+        'test'
+      )`,
+      options,
+    },
+    {
+      code: `dynamicImport(
         /* webpackChunkName: "Some_Other_Module" */
         "test"
       )`,


tests/src/rules/dynamic-import-chunkname.js
```diff
@@ -32,6 +32,14 @@ ruleTester.run('dynamic-import-chunkname', rule, {
       )`,
       options,
     },
+    {
+      code: `dynamicImport(
+        /* webpackInclude: /some-regex/ */
+        /* webpackChunkName: "someModule" */
```
ljharb: Must the comments be in this order, or can they be swapped?

Either way, please also add a test for the reversed order, and for "webpackExclude"

src/rules/dynamic-import-chunkname.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
     const { webpackChunknameFormat = '[0-9a-zA-Z-_/.]+' } = config || {}
 
     const paddedCommentRegex = /^ (\S[\s\S]+\S) $/
-    const commentStyleRegex = /^( \w+: ("[^"]*"|\d+|false|true),?)+ $/
+    const commentStyleRegex = /^( \w+: ("[^"]*"|\/.*\/|\d+|false|true),?)+ $/
```
ljharb: let's not just allow "anything", let's be explicit about webpackInclude/webpackExclude.

tests/src/rules/dynamic-import-chunkname.js
vkrol: > Must the comments be in this order, or can they be swapped?
They can be swapped.
> Either way, please also add a test for the reversed order, and for "webpackExclude"
OK.

src/rules/dynamic-import-chunkname.js
vkrol: We do not just allow anything. We allow anything between slashes.

vkrol: If we want to be explicit about webpackInclude/webpackExclude then we need to be explicit about webpackPreload + true/false too for example. In this case we need to write more sophisticated parser of magic comments.

ljharb: fair point, and reasonable to say it's out of scope for this PR - but i'd expect that even if the stuff after the `:` is "anything", that the part before it would have to be more explicit.

vkrol: Would this be the appropriate solution?
```
-    const commentStyleRegex = /^( \w+: ("[^"]*"|\/.*\/|\d+|false|true),?)+ $/
+    const commentStyleRegex = /^( (webpackChunkName|webpackInclude|webpackExclude|...): ("[^"]*"|\/.*\/|\d+|false|true),?)+ $/
```

ljharb: That looks great to me - does that mirror what webpack supports? (like, is the space after the colon required, is the space between `/*` and `w` required or optional, etc)

vkrol: > like, is the space after the colon required, is the space between /* and w required or optional, etc)

The space after the colon and the space between /* and w are optional. Do we need to change the regex in this PR or it is out of scope?

ljharb: It seems better to change it in this PR, so that after this PR lands, it actually matches what webpack supports and no more.

ljharb:
@vkrol are you still interested in completing this PR?

vkrol:
@ljharb Yes.

vkrol:
Obsoleted by https://github.com/import-js/eslint-plugin-import/pull/2330.

ljharb:
That hasn’t landed yet tho.

vkrol:
This was fixed in https://github.com/import-js/eslint-plugin-import/pull/2330/.

PR #1412, opened 2019-07-12
[fix] `export`: false positive for typescript overloads

### LINKED ISSUE #1357: import/export: false positives for typescript overloads
Minimal example to reproduce (not a sensible use case for overloads, but anyway):

```
export function consume(val: number): void;
export function consume(val: string): void;
export function consume(val: number | string): void {
    // Do something.
}
```

All three `consume` will be annotated with ` error  Multiple exports of name 'consume'   import/export`.

golopot:
Fixes #1357.

Fixes false positives like this: 
```ts
export function foo(a: number) {}
export function foo(a: string) {}
 ```

### DIFF
--- a/src/rules/export.js
+++ b/src/rules/export.js
@@ -24,6 +24,21 @@ ambient namespaces:
 const rootProgram = 'root'
 const tsTypePrefix = 'type:'
 
+/**
+ * Detect function overloads like:
+ * ```ts
+ * export function foo(a: number);
+ * export function foo(a: string);
+ * export function foo(a: number|string) { return a; }
+ * ```
+ * @param {Set<Object>} nodes
+ * @returns {boolean}
+ */
+function isTypescriptFunctionOverloads(nodes) {
+  const types = new Set(Array.from(nodes, node => node.parent.type))
+  return types.size === 2 && types.has('TSDeclareFunction') && types.has('FunctionDeclaration')
+}
+
 module.exports = {
   meta: {
     type: 'problem',
@@ -122,6 +137,8 @@ module.exports = {
         for (let [, named] of namespace) {
           for (let [name, nodes] of named) {
             if (nodes.size <= 1) continue
+
+            if (isTypescriptFunctionOverloads(nodes)) continue
 
             for (let node of nodes) {
               if (name === 'default') {
--- a/tests/src/rules/export.js
+++ b/tests/src/rules/export.js
@@ -118,7 +118,7 @@ context('Typescript', function () {
     parsers.push(require.resolve('@typescript-eslint/parser'))
   }
 
-  if (semver.satisfies(eslintPkg.version, '<6.0.0')) {
+  if (semver.satisfies(eslintPkg.version, '>=4.0.0 <6.0.0')) {
     parsers.push(require.resolve('typescript-eslint-parser'))
   }
 
@@ -144,6 +144,14 @@ context('Typescript', function () {
           code: `
             export const Foo = 1;
             export interface Foo {}
+          `,
+        }, parserConfig)),
+
+        test(Object.assign({
+          code: `
+            export function fff(a: string);
+            export function fff(a: number);
+            export function fff(a: string|number) {};
           `,
         }, parserConfig)),
 

### CODE REVIEW

File: src/rules/export.js [hunk 1]
```diff
@@ -34,6 +47,7 @@ module.exports = {
 
   create: function (context) {
     const namespace = new Map([[rootProgram, new Map()]])
+    const isTypescriptFile = /\.ts|\.tsx$/.test(context.getFilename())
```
ljharb: this seems like something that would need to be determined based on parser settings, not hardcoded extensions - also, what about flow?

src/rules/export.js
```diff
@@ -24,6 +24,19 @@ ambient namespaces:
 const rootProgram = 'root'
 const tsTypePrefix = 'type:'
 
+/**
+ * Detect function overloads like:
+ * ```js
+ * export function foo(a: number) {}
+ * export function foo(a: string) {}
```
ljharb: don’t overloads look like:
```suggestion
 * export function foo(a: string);
```
ie, the body is supposed to be omitted?

src/rules/export.js
golopot: 1. Will it suffice to do something like `\/@typescript-eslint\/parser\/.test(context.parserPath)`?(`context.parserPath` is like `/home/some-user/some-project/node_modules/@typescript-eslint/parser/dist/parser.js`) 2. It seems that flow does not accept function overload like the case here.

ljharb: that would force people to install the typescript parser even if they weren’t using typescript, i think - ideally we’d find a more reliable mechanism than “its typescript” to handle this case.

golopot: Just find out find that these function nodes without body has node type `TSDeclareFunction`. Therefore I change the bail out logic to `some nodes is TSDeclareFunction && every nodes is TSDeclareFunction or FunctionDeclaration`.

src/rules/export.js
golopot: Sure.

src/rules/export.js
```diff
@@ -137,7 +141,7 @@ module.exports = {
           for (let [name, nodes] of named) {
             if (nodes.size <= 1) continue
 
-            if (isTypescriptFile && isTypescriptFunctionOverloads(nodes)) continue
+            if (isTypescriptFunctionOverloads([...nodes])) continue
```
benmosher: why the spread into an array? is `nodes` not an Array, just an iterable? `Set` maybe? if so I might `Array.from(nodes)` to make it a little clearer. but definitely nitpicking, this is a great improvement

ljharb: why not pass the Set in directly, and convert it inside the function?

golopot: Ready for review

src/rules/export.js
```diff
@@ -24,6 +24,24 @@ ambient namespaces:
...
+ * export function foo(a: number);
+ * export function foo(a: string);
+ * export function foo(a: number|string) { return a; }
+ * ```
+ * @param {Set<Object>} nodes
+ * @returns {boolean}
+ */
+function isTypescriptFunctionOverloads(nodes) {
+  return Array.from(nodes).some(node => node.parent.type === 'TSDeclareFunction') &&
+    Array.from(nodes).every(node => (
```
ljharb: let's definitely not do Array.from twice. maybe something like this for the body of the function?
```js
const types = new Set(Array.from(nodes, node => node.parent.type));
if (!types.has('TSDeclareFunction') || types.size > 2) {
  return false;
}
return types.size === 1 || types.has('FunctionDeclaration');
```

golopot: I end up with this:
```js
const types = new Set(Array.from(nodes, node => node.parent.type))
return types.size === 2 && types.has('TSDeclareFunction') && types.has('FunctionDeclaration')
```
The case when `types` is the same as `new Set("TSDeclareFunction")` is when the function lacks implementation, which is a tsc error.

ljharb: That isn't the same logic though - your current logic allows it to be only "TSDeclareFunction". Functions don't need an implementation in tsc if they're just defining a type overload.

golopot: That case is a tsc error https://www.typescriptlang.org/play/#code/KYDwDg9gTgLgBAMwK4DsDGMCWEWIBQCUQA . And for the purpose of this rule, that case does not matter.

PR #1416, opened 2019-07-15
New: Add order-member rule

fa93hws:
Add a new rule to order the import member.

e.g.
`import { c, b, d, a, e } from 'foo'`
becomes
`import { a, b, c, d, e } from 'foo'`

It's case sensitive by default.
e.g.
`import { c, b, d, A, a, Z, e } from 'foo'`
becomes
`import { A, Z, a, b, c, d, e } from 'foo'`

There is a option to make it case insensitive
e.g.
`import { c, b, d, A, a, B, e } from 'foo'`
becomes (It's stable sort)
`import { A, a, b, B, c, d, e } from 'foo'`

**The doc will be updated once approved**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -30,6 +30,7 @@ export const rules = {
   'no-nodejs-modules': require('./rules/no-nodejs-modules'),
   'no-webpack-loader-syntax': require('./rules/no-webpack-loader-syntax'),
   'order': require('./rules/order'),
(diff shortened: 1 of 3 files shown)


ljharb:
This already exists in eslint core: https://eslint.org/docs/rules/sort-imports#import-sorting-sort-imports

PR #1419, opened 2019-07-17
[Refactor] use native JS instead of lodash

ljharb:
This removes a large dep that often has CVEs, in exchange for using native JS which is clearer and more maintainable.

Closes #1224.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,6 @@
     "eslint-import-resolver-node": "^0.3.2",
     "eslint-module-utils": "^2.4.0",
     "has": "^1.0.3",
-    "lodash": "^4.17.11",
     "minimatch": "^3.0.4",
     "read-pkg-up": "^2.0.0",
     "resolve": "^1.11.0"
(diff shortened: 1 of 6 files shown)


ljharb:
Going to go ahead and merge once tests all pass; the changes are minimal.

The last commit also fixes the failing tests on master.

ljharb:
Some eslint 3 tests still failing; I'll fix those and get tests passing on this PR prior to merging.

benmosher:
The test failures are pretty weird. No idea there.

FWIW the webpack resolver still has a lodash reference, but it's declared in its `package.json` so I would expect that to be unrelated?

golopot:
https://github.com/benmosher/eslint-plugin-import/blob/6a07c8142c79044a83a1949a17d41a2390833b59/tests/src/rules/no-deprecated.js#L212

`typescript-eslint-parser@22.0.0` requires eslint >= 4, maybe that's it?

https://unpkg.com/typescript-eslint-parser@22.0.0/package.json

golopot:
This pr closes #1224 ?

ljharb:
I can repro these failures locally - but when I check out latest master, they still fail :-/

Issue #1413, opened 2019-07-12
run eslint on the entire project instead of only on `src`


golopot:
Currently ci only run eslint on `./src`, therefore lots of codes outside of `./src` (especially .`/tests`) are unlinted. This causes lots of friction in development. In my case, when I edit a test file, I have to disable eslint in my editor in order to avoid unrelated changes.

golopot:
There is a technical problem: Since we don't want to lint `tests/files`, an `.eslintignore` file needs to be added. But doing that alters the behavior of unit tests for `no-unused-modules`, the rule will consider `tests/files/**/*` to be ignored, hence failing the tests.

ljharb:
we could provide a custom eslintignore file for tests (or for not tests), perhaps?

ljharb:
Fixed in #1475.

PR #1435, opened 2019-08-03
feat: support `parseForESLint` from custom parser

### LINKED ISSUE #1433: Parse error in mdx files
Hi, I'm reporting here too, because I realized [an issue](https://github.com/tleunen/eslint-import-resolver-babel-module/issues/106) I originally reported at `babel-plugin-module-resolver` might not be specific to that resolver. Please read it for more information. 

As I commented there, I also get parse errors when using the default node resolver, but on a different import.

import:
```jsx 
import styled from "styled-components";
```

Error:
```
21:20  error    Parse errors in imported module 'styled-components': parser.parse is not a function (undefined:undefined)  import/namespace
21:20  error    Parse errors in imported module 'styled-components': parser.parse is not a function (undefined:undefined)  import/default
21:20  warning  Parse errors in imported module 'styled-components': parser.parse is not a function (undefined:undefined)  import/no-named-as-default
21:20  warning  Parse errors in imported module 'styled-components': parser.parse is not a function (undefined:undefined)  import/no-named-as-default-member
```

JounQin:
close #1433

### DIFF
--- a/tests/src/core/parse.js
+++ b/tests/src/core/parse.js
@@ -9,6 +9,8 @@ describe('parse(content, { settings, ecmaFeatures })', function () {
   const path = getFilename('jsx.js')
   const parseStubParser = require('./parseStubParser')
   const parseStubParserPath = require.resolve('./parseStubParser')
+  const eslintParser = require('./eslintParser')
+  const eslintParserPath = require.resolve('./eslintParser')
   let content
 
   before((done) =>
@@ -41,6 +43,15 @@ describe('parse(content, { settings, ecmaFeatures })', function () {
     expect(parseSpy.args[0][1], 'custom parser to get parserOptions.tokens equal to true').to.have.property('tokens', true)
     expect(parseSpy.args[0][1], 'custom parser to get parserOptions.range equal to true').to.have.property('range', true)
     expect(parseSpy.args[0][1], 'custom parser to get parserOptions.filePath equal to the full path of the source file').to.have.property('filePath', path)
+  })
+
+  it('passes with custom `parseForESLint` parser', function () {
+    const parseForESLintSpy = sinon.spy(eslintParser, 'parseForESLint')
+    const parseSpy = sinon.spy()
+    eslintParser.parse = parseSpy
+    parse(path, content, { settings: {}, parserPath: eslintParserPath })
+    expect(parseForESLintSpy.callCount, 'custom `parseForESLint` parser to be called once').to.equal(1)
+    expect(parseSpy.callCount, '`parseForESLint` takes higher priority than `parse`').to.equal(0)
   })
 
   it('throws on context == null', function () {
--- a/utils/parse.js
+++ b/utils/parse.js
@@ -42,6 +42,24 @@ exports.default = function parse(path, content, context) {
   // require the parser relative to the main module (i.e., ESLint)
   const parser = moduleRequire(parserPath)
 
+  if (typeof parser.parseForESLint === 'function') {
+    let ast
+    try {
+      ast = parser.parseForESLint(content, parserOptions).ast
+    } catch (e) {
+      //
+    }
+    if (!ast || typeof ast !== 'object') {
+      console.warn(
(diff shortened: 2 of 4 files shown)


cdfa:
So I ran `npm i -D JounQin/eslint-plugin-import#feat/parse_for_eslint`, but it's not working. I looked for the `utlis/parse.js` file you modified and it seems to be located at `node_modules/eslint-module-utils/parse.js`, but it doesn't have your patch. How do I test it?

JounQin:
@cdfa There is no need to wait for this PR, I've exported `parse` from `@rxts/eslint-plugin-mdx`, you can simply upgrade to the latest version.

cdfa:
Ah great!

tests/src/core/parse.js
```diff
@@ -43,6 +45,16 @@ describe('parse(content, { settings, ecmaFeatures })', function () {
     expect(parseSpy.args[0][1], 'custom parser to get parserOptions.filePath equal to the full path of the source file').to.have.property('filePath', path)
   })
 
+  it('passes with custom `parseForESLint` parser', function () {
+    const parseForESLintSpy = sinon.spy(eslintParser.parseForESLint)
```
ljharb: please avoid passing just an object into `sinon.spy`; that implicitly spies on all properties. Instead, let's pass explicit property names.

utils/parse.js
```diff
@@ -1,4 +1,4 @@
-"use strict"
+'use strict'
```
ljharb: please revert all unrelated formatting changes

tests/src/core/parse.js
JounQin: It's a function

utils/parse.js
JounQin: OK, it must be related to my editor `formatOnSave`.

Besides, why the usage of quotes are inconsistent through the project.

tests/src/core/parse.js
ljharb: i'm a bit confused; sinon is for spying on properties. what does passing the function in here do, as opposed to `sinon.spy(eslintParser, 'parseForEslint')`?

utils/parse.js
ljharb: #1413; however, either way, what eslint doesn't make consistent, shouldn't be autoformatted :-)

tests/src/core/parse.js
JounQin: It can spy function (wrapping the original one) directly. The tests are passing.😆

ljharb: Iiiiiii

ljharb: right, but if you followed my suggestion, you wouldn’t have to reassign it on the next lines.

JounQin: > right, but if you followed my suggestion, you wouldn’t have to reassign it on the next lines.

Seems more clear, I will change it.

tests/src/core/parse.js
```diff
@@ -43,6 +45,15 @@ describe('parse(content, { settings, ecmaFeatures })', function () {
     expect(parseSpy.args[0][1], 'custom parser to get parserOptions.filePath equal to the full path of the source file').to.have.property('filePath', path)
   })
 
+  it('passes with custom `parseForESLint` parser', function () {
+    const parseForESLintSpy = sinon.spy(eslintParser, 'parseForESLint')
+    const parseSpy = sinon.spy()
+    eslintParser.parse = parseSpy
```
ljharb: this can also be `const parseSpy = sinon.spy(eslintParser, 'parse');`

JounQin: > this can also be `const parseSpy = sinon.spy(eslintParser, 'parse');`

Nope, `eslintParser.parse` is `undefined`, `sinon` will throw an error.

See https://travis-ci.org/benmosher/eslint-plugin-import/builds/567436253?utm_source=github_status&utm_medium=notification

ljharb: ah, fair enough :-)

PR #1438, opened 2019-08-03
Add support for bundledDependencies

### LINKED ISSUE #1436: Support for bundledDependencies
I have some auto-generated code (through https://github.com/prisma/photonjs) that is located in `node_modules/@generated/photon`. 

How can I tell `eslint-plugin-import` to not throw a `import/no-extraneous-dependencies` error? The best solution I can think of so far is to put it into the [bundledDependencies](https://npm.github.io/using-pkgs-docs/package-json/types/bundleddependencies.html). But it seems like `eslint-plugin-import` does not look at these.

Related: https://github.com/prisma/prisma2/issues/314

schmidsi:
Fixes #1436

### DIFF
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -1,6 +1,6 @@
 # import/no-extraneous-dependencies: Forbid the use of extraneous packages
 
-Forbid the import of external modules that are not declared in the `package.json`'s `dependencies`, `devDependencies`, `optionalDependencies` or `peerDependencies`.
+Forbid the import of external modules that are not declared in the `package.json`'s `dependencies`, `devDependencies`, `optionalDependencies`, `peerDependencies`, or `bundledDependencies`.
 The closest parent `package.json` will be used. If no `package.json` is found, the rule will not lint anything. This behaviour can be changed with the rule option `packageDir`.
 
 Modules have to be installed for this rule to work.
@@ -14,6 +14,8 @@ This rule supports the following options:
 `optionalDependencies`: If set to `false`, then the rule will show an error when `optionalDependencies` are imported. Defaults to `true`.
 
 `peerDependencies`: If set to `false`, then the rule will show an error when `peerDependencies` are imported. Defaults to `false`.
+
+`bundledDependencies`: If set to `false`, then the rule will show an error when `bundledDependencies` are imported. Defaults to `true`.
 
 You can set the options like this:
 
@@ -70,7 +72,10 @@ Given the following `package.json`:
   },
   "peerDependencies": {
     "react": ">=15.0.0 <16.0.0"
-  }
+  },
+  "bundledDependencies": [
+    "@generated/foo",
+  ]
 }
 ```
 
@@ -90,6 +95,10 @@ var test = require('ava');
 /* eslint import/no-extraneous-dependencies: ["error", {"optionalDependencies": false}] */
 import isArray from 'lodash.isarray';
 var isArray = require('lodash.isarray');
+
+/* eslint import/no-extraneous-dependencies: ["error", {"bundledDependencies": false}] */
+import foo from '"@generated/foo"';
+var foo = require('"@generated/foo"');
 ```
 
 
@@ -103,6 +112,7 @@ var foo = require('./foo');
 import test from 'ava';
 import find from 'lodash.find';
 import isArray from 'lodash.isarray';
+import foo from '"@generated/foo"';
 
 /* eslint import/no-extraneous-dependencies: ["error", {"peerDependencies": true}] */
 import react from 'react';
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -11,12 +11,19 @@ function hasKeys(obj = {}) {
   return Object.keys(obj).length > 0
 }
 
+function arrayOrKeys(arrayOrObject) {
+  return Array.isArray(arrayOrObject) ? arrayOrObject : Object.keys(arrayOrObject)
+}
+
 function extractDepFields(pkg) {
   return {
     dependencies: pkg.dependencies || {},
     devDependencies: pkg.devDependencies || {},
     optionalDependencies: pkg.optionalDependencies || {},
     peerDependencies: pkg.peerDependencies || {},
+    // BundledDeps should be in the form of an array, but object notation is also supported by
+    // `npm`, so we convert it to an array if it is an object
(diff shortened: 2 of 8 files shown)


tests/src/rules/no-unresolved.js
```diff
@@ -26,10 +26,10 @@ function runResolverTests(resolver) {
       test({ code: 'import "./malformed.js"' }),
 
       rest({ code: 'import foo from "./bar";' }),
-      rest({ code: "import bar from './bar.js';" }),
-      rest({ code: "import {someThing} from './test-module';" }),
-      rest({ code: "import fs from 'fs';" }),
-      rest({ code: "import('fs');"
+      rest({ code: 'import bar from \'./bar.js\';' }),
```
ljharb: please revert unrelated formatting changes (if you're using prettier in your editor, you should disable it, and only run prettier via eslint autofix)

schmidsi:
Fixes my original issue (https://github.com/prisma/prisma2/issues/314) through `npm link`. Should work now. Thanks for the help.

docs/rules/no-extraneous-dependencies.md
```diff
@@ -1,6 +1,6 @@
 # import/no-extraneous-dependencies: Forbid the use of extraneous packages
 
-Forbid the import of external modules that are not declared in the `package.json`'s `dependencies`, `devDependencies`, `optionalDependencies` or `peerDependencies`.
+Forbid the import of external modules that are not declared in the `package.json`'s `dependencies`, `devDependencies`, `optionalDependencies`, `peerDependencies` or `bundledDependencies`.
```
ljharb: i'm sad this wasn't using the Oxford comma before, but never too late to write it correctly :-)
```suggestion
Forbid the import of external modules that are not declared in the `package.json`'s `dependencies`, `devDependencies`, `optionalDependencies`, `peerDependencies`, or `bundledDependencies`.
```

src/rules/no-extraneous-dependencies.js
```diff
@@ -17,6 +17,7 @@ function extractDepFields(pkg) {
     devDependencies: pkg.devDependencies || {},
     optionalDependencies: pkg.optionalDependencies || {},
     peerDependencies: pkg.peerDependencies || {},
+    bundledDependencies: pkg.bundledDependencies ||pkg.bundleDependencies || [],
```
ljharb:
```suggestion
    bundledDependencies: pkg.bundledDependencies || pkg.bundleDependencies || [],
```
it's super crazy to me that both of these keys [are supported by npm](http://npm.github.io/using-pkgs-docs/package-json/types/bundleddependencies.html). Can we confirm which one takes precedence from npm itself, when both are present?

ljharb: also, if this is mistakenly defined as an object, then the code will crash later when it assumes it's an array. I think we need an `Array.isArray` check here.

schmidsi: > it's super crazy to me that both of these keys [are supported by npm](http://npm.github.io/using-pkgs-docs/package-json/types/bundleddependencies.html). Can we confirm which one takes precedence from npm itself, when both are present?

Yes, I was also a bit surprised. [Looks like bundleDependencies takes precedence](https://github.com/npm/cli/blob/656bce7dd0f9753a273912e803261ed246593924/lib/install/save.js#L74) I have to change my code then.

- [ ] `bundleDependencies` takes precedence over `bundledDependencies`

> also, if this is mistakenly defined as an object, then the code will crash later when it assumes it's an array. I think we need an `Array.isArray` check here.

Good catch. On top of that, [it seems like it is also possible to define the bundleDependencies as an object](https://github.com/npm/cli/blob/c1522be2406a0ea4f14c85753edd42ddd8d7e180/test/tap/bundled-dependencies-nonarray.js#L37). Although, my VS code shows a type error (see screenshot below). Anyways, I think it would be safe to handle both cases

- [ ] Normalize array & object notation

Screenshot 2019-08-05 at 12 38 10

I'll add those fixes later today or tomorrow.

schmidsi:
It's ready to merge in my opinion. Any more comments?

src/rules/no-extraneous-dependencies.js
```diff
@@ -17,6 +17,7 @@ function extractDepFields(pkg) {
     devDependencies: pkg.devDependencies || {},
     optionalDependencies: pkg.optionalDependencies || {},
     peerDependencies: pkg.peerDependencies || {},
+    bundledDependencies: pkg.bundleDependencies || pkg.bundledDependencies || [],
```
ljharb: this still isn't ensuring it's an array?

src/rules/no-extraneous-dependencies.js
```diff
@@ -121,11 +124,16 @@ function reportIfMissing(context, deps, depsOptions, node, name) {
   const isInDevDeps = deps.devDependencies[packageName] !== undefined
   const isInOptDeps = deps.optionalDependencies[packageName] !== undefined
   const isInPeerDeps = deps.peerDependencies[packageName] !== undefined
+  // BundledDeps should be in the form of an array, but object notation is also supported by `npm`
+  const isInBundledDeps =
+    deps.bundledDependencies.indexOf(packageName) !== -1 ||
```
ljharb: objects don't have an `indexOf` method, so this will crash.

src/rules/no-extraneous-dependencies.js
schmidsi: I transform the object to an array now.

src/rules/no-extraneous-dependencies.js
schmidsi: Hmm, I actually had tests for this scenario. Funnily, the object `deps.bundledDependencies` was something in between an object and an array, which can be reconstructed like this:

```javascript
const arrayObject = Object.assign([], {a: 3});
console.log(Array.isArray(arrayObject)) // true
console.log(typeof arrayObject) // "object"
console.log(arrayObject.c) // 3
console.log(arrayObject) // [a: 3]
console.log(arrayObject[0]) // undefined 
```

I'm not sure if this is a bug in JavaScript by itself or just a weird edge-case, I've never saw something like this.

Anyways, I clean the `bundledDependencies` now so it is always an array and go with `indexOf`.

schmidsi:
I think now it should be good to merge? 😊

src/rules/no-extraneous-dependencies.js
ljharb: That's not a bug; all arrays are objects.

PR #1439, opened 2019-08-04
[no-commonjs] add allowConditionalRequire option

### LINKED ISSUE #1437: [no-commonjs] does not catch conditional require
The example here: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-commonjs.md#allow-require

cannot be reported even if `allowRequire: false`, because of this line:

https://github.com/benmosher/eslint-plugin-import/blob/1edbbd03ec636469328ad59da922ed7edc8cd36d/src/rules/no-commonjs.js#L90

The context scope is `block` in this case.

Pessimistress:
Fixes #1437

Currently:

1. `if (typeof window !== "undefined") require("x")` is reported by default
1. `if (typeof window !== "undefined") { require("x") }` is never reported (scope)
1. `typeof window !== "undefined" && require("x")` is never reported (conditional require)

After this PR, all cases will be treated the same (as conditional requires). A follow up PR will add an option for users to configure whether conditional requires should be reported.

### DIFF
--- a/src/rules/no-commonjs.js
+++ b/src/rules/no-commonjs.js
@@ -25,6 +25,26 @@ function allowRequire(node, options) {
   return options.allowRequire
 }
 
+function allowConditionalRequire(node, options) {
+  return options.allowConditionalRequire !== false
+}
+
+function validateScope(scope) {
+  return scope.variableScope.type === 'module'
+}
+
+// https://github.com/estree/estree/blob/master/es5.md
+function isConditional(node) {
+  if (
+    node.type === 'IfStatement'
+    || node.type === 'TryStatement'
+    || node.type === 'LogicalExpression'
+    || node.type === 'ConditionalExpression'
+  ) return true
+  if (node.parent) return isConditional(node.parent)
+  return false
+}
+
 //------------------------------------------------------------------------------
 // Rule Definition
 //------------------------------------------------------------------------------
@@ -35,6 +55,7 @@ const schemaObject = {
   properties: {
     allowPrimitiveModules: { 'type': 'boolean' },
     allowRequire: { 'type': 'boolean' },
+    allowConditionalRequire: { 'type': 'boolean' },
   },
   additionalProperties: false,
 }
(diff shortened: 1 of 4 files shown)


src/rules/no-commonjs.js
```diff
@@ -25,6 +25,11 @@ function allowRequire(node, options) {
   return options.allowRequire
 }
 
+function validateScope(scope) {
+  if (scope.variableScope.type === 'module') return true
```
Pessimistress: I think this might break the current behavior with try... catch. I'm AFK right now, will add a test case later.

Pessimistress:
I have no idea how to add the option without making these changes... Do you mind if I just add it in the same PR?

Pessimistress:
> those who are relying on these warnings have an existing solution to enable them

I highly doubt if anyone is relying on these warnings. The only breaking change is that `if (typeof window !== "undefined") require("x")` is no longer reported. Considering `if (typeof window !== "undefined") { require("x") }` was never reported, I consider the existing behavior already broken.

ljharb:
@Pessimistress adding the option in this PR sounds fine to me.

src/rules/no-commonjs.js
```diff
@@ -25,6 +25,27 @@ function allowRequire(node, options) {
   return options.allowRequire
 }
 
+function allowConditionalRequire(node, options) {
+  return options.allowConditionalRequire === undefined ? true : options.allowConditionalRequire
```
ljharb:
```suggestion
  return options.allowConditionalRequire !== false
```

src/rules/no-commonjs.js
```diff
@@ -25,6 +25,27 @@ function allowRequire(node, options) {
...
+}
+
+function validateScope(scope) {
+  if (scope.variableScope.type === 'module') return true
+  return false
```
ljharb:
```suggestion
  return scope.variableScope.type === 'module'
```

src/rules/no-commonjs.js
```diff
@@ -35,6 +56,7 @@ const schemaObject = {
   properties: {
     allowPrimitiveModules: { 'type': 'boolean' },
     allowRequire: { 'type': 'boolean' },
+    allowConditionalRequire: { 'type': 'boolean' },
```
ljharb:
```suggestion
    allowConditionalRequire: { 'type': 'boolean', default: true },
```
(altho for back compat we can't rely on the default getting applied)

src/rules/no-commonjs.js
```diff
@@ -25,6 +25,27 @@ function allowRequire(node, options) {
   return options.allowRequire
 }
 
+function allowConditionalRequire(node, options) {
+  return options.allowConditionalRequire !== false
+}
+
+function validateScope(scope) {
+  if (scope.variableScope.type === 'module') return true
```
ljharb:
```suggestion
```
now that the other change is done :-)

PR #1440, opened 2019-08-05
Handle nested main packages as main packages.

ivo-stefchev:
There are cases like "redux-saga/effects" or "lodash/get" where you cannot or don't want to put extensions. Theoretically all folders of a package can have 'package.json' with `main` clause.

In the case of *redux-saga/effects* it also conflicts with rule `import/no-unresolved`.

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -34,7 +34,7 @@ function isExternalPath(path, name, settings) {
 }
 
 const externalModuleRegExp = /^\w/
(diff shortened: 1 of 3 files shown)


ljharb:
@ivo-stefchev sorry for the lack of review here, are you no longer interested in completing this PR?

ljharb:
Deleting the fork means this can never land :-(

Issue #1248, opened 2018-12-19
import/extensions ignores extension overrides if ignorePackages is set in object options


le0nik:
node: v10.14.1
eslint: v5.10.0
eslint-plugin-import: v2.14.0

.eslintrc.js

```js
module.exports = {
  plugins: ['import'],
  rules: {
    'import/extensions': ['error', 'always', {
      ignorePackages: true,
      js: 'never',
    }]
  }
}
```

./src/index.js

```js
import app from './app';
```

Running eslint results in an error:

```
/Volumes/Projects/project/src/index.js
1:22  error  Missing file extension "js" for "./app"  import/extensions
```

The problem seems to be here https://github.com/benmosher/eslint-plugin-import/blob/1cd82eb27df85768fbd076e4ff6b7f36d6f652ce/src/rules/extensions.js#L36-L45

Plugin expects extensions overrides to be specified inside `pattern` property inside the options object if `ignorePackages` property is also specified. It's not documented [here](https://github.com/benmosher/eslint-plugin-import/blob/1cd82eb27df85768fbd076e4ff6b7f36d6f652ce/docs/rules/extensions.md) though.

So this configuration works correctly:

```js
{
  rules: {
    'import/extensions': ['error', 'always', {
      ignorePackages: true,
      pattern: { // wrap extensions in the `pattern` object
        js: 'never',
      }
    }]
  }
}
```

Either documentation is lacking or `pattern` property should not be expected in options.

ljharb:
Ths doesn’t seem resolved yet?

ivo-stefchev:
Oh, how I love explicit configs.
So the reason why this `module.exports = {
  plugins: ['import'],
  rules: {
    'import/extensions': ['error', 'always', {
      ignorePackages: true,
      js: 'never',
    }]
  }
}`
is because in order to apply the **js** part - you need to have **ingnorePackages** to be **undefined**. So basically if **ignorePackages** is present in the object - it disables all extension specific options. For me at least this makes very little sense.

ivo-stefchev:
Also this defaultConfig is enum of three possible values.
```
"import/extensions": [<severity>, "never" | "always" | "ignorePackages"]
```
**ignorePackages** doesn't do anything when you set it, hence I propose to be removed from the enumValues and left only as boolean property part of the object.

ljharb:
Seems like a good PR.

PR #1442, opened 2019-08-05
[extensions] Better document. Add unit tests. (#1248)

ivo-stefchev:
I was curious of how this rule worked, so I went the easiest way and wrote a bunch of unit tests for the buildProperties (a method that returns the options from the config for the extensions rule) method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`no-extraneous-dependencies`]: ensure `node.source` is truthy ([#1589], thanks [@ljharb])
 - [`extensions`]: Ignore query strings when checking for extensions ([#1572], thanks [@pcorpet])
 
+### Docs
+- [`extensions`]: improve `ignorePackages` docs ([#1248], thanks [@ivo-stefchev])
+
 ### Added
 - [`order`]: add option pathGroupsExcludedImportTypes to allow ordering of external import types ([#1565], thanks [@Mairu])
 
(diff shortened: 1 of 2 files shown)


tests/src/rules/extensions.js
```diff
@@ -1,9 +1,79 @@
 import { RuleTester } from 'eslint'
 import rule from 'rules/extensions'
 import { test, testFilePath } from '../utils'
+import { expect } from 'chai'
 
 const ruleTester = new RuleTester()
 
+describe('extensions buildProperties', function() {
```
ljharb: i'm confused what the purpose of these tests are. can the tests be exercised with ruleTester test cases, rather than hardcoding to things like `__test__`?

ivo-stefchev: Sorry, I deleted my previous comment by accident. I did it like that, because buildProperties is private function and needs to be tested.
So I guess it's a no go for this PR?

ljharb: in general, private things shouldn't be tested - in this plugin's case, everything should be tested via the RuleTester.

soryy708: I'm not sure why you're saying "private things shouldn't be tested ".

soryy708: But when you refactor internals and break something, you wish you had tests to catch that early.

ljharb: The tests should be on the public pieces, and should transparently exercise all the private pieces.

Any behavior of your private code that is not observable via your public code *should not be tested because it doesn't matter if it works or not*

Issue #1446, opened 2019-08-07
[import/named] fails to track deep exports when re-exporting from a module in node_modules


kentcdodds:
Here's a reproduction: https://github.com/kentcdodds/eslint-export-named-issue

```javascript
// exporter.js
// NOTE: I observed that it's only a problem if you export * from a file in node_modules
// and it doesn't matter module it is. In my case, I'm trying to re-export from `@testing-library/react`.
// here I'll just use eslint-plugin-import so we don't need to install anything else.

import {config} from 'eslint-plugin-import' // <-- no error here (so it definitely exports `config`)

export * from 'eslint-plugin-import'

// importer.js

import {configs} from './exporter' // <-- import/named fails
```

I'm not sure why this is an issue, but like I said, it appears to only happen when you re-export from a `node_modules` module. Seems to work fine if you re-export from one of your own files.

redbugz:
We have the same problem, and I started poking around. Not entirely familiar with the code, but it seems to be the module is ignored due the unambiguous check. When importing the node_module module, it gets the compiled es5 source code, which fails the unambiguous test. So hasDeep has a null innerMap, and returns no exports, so the import is not found.

It seems possibly to be a similar condition to #1460.

We have gotten some modules do work with `export * from 'modulename'`, and I'm guessing they are resolving to the es6 module source instead of the es5.

Using Kent's repo, here is the log output:
```
% DEBUG=eslint-plugin-import* npm run lint    

> @ lint /Users/me/dev/sandbox/eslintexport2
> eslint .

  eslint-plugin-import:resolver:node Resolving: eslint-plugin-import from: /Users/me/dev/sandbox/eslintexport2/exporter.js +0ms
  eslint-plugin-import:resolver:node Resolved to: /Users/me/dev/sandbox/eslintexport2/node_modules/eslint-plugin-import/lib/index.js +3ms
  eslint-plugin-import:ExportMap ignored path due to unambiguous regex: /Users/me/dev/sandbox/eslintexport2/node_modules/eslint-plugin-import/lib/index.js +0ms
  eslint-plugin-import:resolver:node Resolving: ./exporter from: /Users/me/dev/sandbox/eslintexport2/importer.js +6ms
  eslint-plugin-import:resolver:node Resolved to: /Users/me/dev/sandbox/eslintexport2/exporter.js +1ms
[...]

PR #1447, opened 2019-08-07
fix [import/named] for importing from a module which re-exports named exports from a node_modules module

kentcdodds:
This adds a failing test for #1446

I'm not sure how to proceed, but hopefully this helps someone diagnose the issue.

### DIFF
--- a/tests/src/rules/named.js
+++ b/tests/src/rules/named.js
@@ -26,6 +26,7 @@ ruleTester.run('named', rule, {
     test({code: 'import { destructingRenamedAssign } from "./named-exports"'}),
     test({code: 'import { ActionTypes } from "./qc"'}),
     test({code: 'import {a, b, c, d} from "./re-export"'}),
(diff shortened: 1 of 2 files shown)


tests/src/rules/named.js
```diff
@@ -26,6 +26,7 @@ ruleTester.run('named', rule, {
     test({code: 'import { destructingRenamedAssign } from "./named-exports"'}),
     test({code: 'import { ActionTypes } from "./qc"'}),
     test({code: 'import {a, b, c, d} from "./re-export"'}),
+    test({code: 'import {RuleTester} from "./re-export-node_modules"'}),
```
kentcdodds: I chose `eslint` as the `node_module` and `RuleTester` as the import because I figured that'd be pretty stable. It's unlikely that export would change, and in the event that it does, we'll need to update this test file quite a bit anyway.

PR #1472, opened 2019-09-07
Make aggregating modules valid in group-exports

### LINKED ISSUE #1471: group-exports aggregating modules should be valid
I would like to make [aggregating modules](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) valid when using the `group-exports` rule.

**Use-case**
When aggregating modules in a single file the `group-exports` rule fails on because of multiple export declarations. This makes it invalid to re-export submodules from the root of a module.

**Steps to reproduce**
- create a file with the following exports
```js
export { default as module1 } from 'module-1'
export { default as module2 } from 'module-2'
```
- validate the file using the `group-exports` rule

**Expected results**
No linting error.

**Actual result**
Linting error.

atikenny:
Fixes #1471

### DIFF
--- a/docs/rules/group-exports.md
+++ b/docs/rules/group-exports.md
@@ -27,6 +27,12 @@ export {
 ```
 
 ```js
+// Aggregating exports -> ok
+export { default as module1 } from 'module-1'
+export { default as module2 } from 'module-2'
+```
+
+```js
 // A single exports assignment -> ok
 module.exports = {
   first: true,
@@ -61,6 +67,12 @@ module.exports = test
 // Multiple named export statements -> not ok!
 export const first = true
 export const second = true
+```
+
+```js
+// Aggregating exports from the same module -> not ok!
+export { module1 } from 'module-1'
+export { module2 } from 'module-1'
 ```
 
 ```js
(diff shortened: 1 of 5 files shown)


tests/src/rules/group-exports.js
```diff
@@ -45,6 +45,10 @@ ruleTester.run('group-exports', rule, {
       // test
       export default {}
     ` }),
+    test({ code: `
+      export { default as module1 } from './module-1'
+      export { default as module2 } from './module-2'
```
ljharb: let's add another test to ensure that this:
```jsx
       export { a as module1 } from './module-1'
       export { b as module2 } from './module-1'
```
still warns?

docs/rules/group-exports.md
```diff
@@ -69,6 +69,12 @@ export const first = true
 export const second = true
 ```
 
+```js
+// Aggregating exports from the same module -> ok
```
ljharb: this should be "not ok", since you can do `export { module1, module2 } from 'module-1'`

src/rules/group-exports.js
```diff
@@ -86,6 +95,18 @@ function create(context) {
         })
       }
 
+      // Report multiple `aggregated exports` from the same module (ES2015 modules)
+      Object.keys(nodes.sources)
```
ljharb: You can use the object.values package here, instead of having to use Object.keys.

PR #1475, opened 2019-09-14
[eslint] make CI lint tests, utils, resolvers, ...etc

golopot:
First steps for fixing #1413.


.eslintignore
```diff
@@ -0,0 +1,7 @@
+lib
+coverage
+.nyc_output
+node_modules
+# we want to ignore "tests/files" here, but unfortunately doing so would
+# interfere with unit test and fail it for some reason.
+# tests/files
```
golopot: See also https://github.com/benmosher/eslint-plugin-import/issues/1413#issuecomment-531471382 :
> There is a technical problem: Since we don't want to lint `tests/files`, an `.eslintignore` file needs to be added. But doing that alters the behavior of unit tests for `no-unused-modules`, the rule will consider `tests/files/**/*` to be ignored, hence failing the tests.

.eslintrc.yml
```diff
@@ -11,7 +11,7 @@ env:
 
 parserOptions:
   sourceType: module
-  ecmaVersion: 6
+  ecmaVersion: 8
```
golopot: This is needed because somewhere in tests we uses trailing comma in function call.

.eslintrc.yml
```diff
@@ -20,13 +20,19 @@ rules:
...
   no-shadow: 1
+  no-unused-vars: 1
+  no-console: 1
+  no-extra-semi: 2
+  no-useless-escape: 1
   quotes:
     - 2
     - single
     - allowTemplateLiterals: true
+      avoidEscape: true
```
golopot: This avoids large git diff in tests.

.eslintrc.yml
```diff
@@ -20,13 +20,19 @@ rules:
   comma-dangle: [2, always-multiline]
   eqeqeq: [2, "allow-null"]
   no-shadow: 1
+  no-unused-vars: 1
```
golopot: The severities will be increased in follow ups.

.eslintrc.yml
ljharb: that means it should be changed in an overrides, just for tests - but not for the whole repo.

.eslintrc.yml
ljharb: similarly, all test-specific changes should go in an "overrides", not apply to the whole repo

.eslintrc.yml
golopot: overrides is unavailable in eslint <=3, and the Travis script run lint on all versions.

.eslintrc.yml
golopot: ditto

golopot:
The current approach has the disadvantage that contributors 1. cannot run `eslint .`, but have to  run `eslint . --ignore-path .eslintignore-real`. 2. will need to disable editors' eslint when working on `tests/files/**`.

I found another approach: add the file `tests/files/.eslintrc`, which disables all rules by `rules: { "some-rule": 0, ... }`. This approach do not suffer from the aforementioned problems.

I am going to force push with the later approach.

.eslintrc.yml
ljharb: in that case, a nested config file inside the `test` directory would work in all eslint versions.

golopot: I chose not to do to that because then I need do that in these 5 directories `tests`, `utils`, `resolvers`, `config`, `memo-parser`. But maybe that's not that bad, what do you think?

ljharb: that’s a fair point - maybe we could make a separate file at the root, `.eslintrc-test` and then in those 5 dirs, use `extends: [‘../.eslintrc-test’]`?

golopot: That is a good idea.

ljharb:
I think the latter approach you suggest sounds better.

.eslintrc.yml
ljharb: this should be moved to the test-only config file?

memo-parser/.eslintrc.yml
```diff
@@ -0,0 +1,2 @@
+---
+extends: '../.eslintrc-test.yml'
```
ljharb: memo-parser is a published package; it should use the non-test settings.

resolvers/.eslintrc
```diff
@@ -1,3 +1,4 @@
 ---
+extends: '../.eslintrc-test.yml'
```
ljharb: these are all published packages; they should use non-test settings.

utils/.eslintrc.yml
```diff
@@ -0,0 +1,2 @@
+---
+extends: '../.eslintrc-test.yml'
```
ljharb: this also is a published package, and should use non-test config.

.eslintrc.yml
golopot: done

ljharb:
We shouldn't be autofixing warnings; looks like the second commit is only doing that.

I'll rebase and clean it up a bit, thanks! <3

PR #1476, opened 2019-09-15
[New]: add rule `default-import-match-filename`

### LINKED ISSUE #325: Proposal: `default-import-match-filename`
I often see things such as:

```js
import parseDate from './types/date';
import float from './types/parseFloat';
```

The disparity between the default import name (`parseDate`) and file name (`date`) often indicates that one or the other is named inappropriately. In the first case, thats the file name. In the second case, thats the local variable name.

Valid code:

```js
import parseFloat from './types/parseFloat';
import parseDate from './types/parseDate';
```

Invalid code:

```js
import parseDate from './types/date';
import float from './types/parseFloat';
```

There is no way to enforce a right name, e.g. user could just as well end up with:

```js
import date from './types/date';
import float from './types/float';
```

However, I would argue that these issues are easy to notice in the code. Having this rule would prevent lazy developers from using a better variable name in a local file and ignore the need to rename the file and other references.

Furthermore, this enables easier code inspection (because it is easy to recognise what a variable represents when it is being used consistently between many files).

golopot:
Closes #325 .

* Import name is considered to matche filename if they are equal after `toLowercase`, and striping extensions, and striping characters "._-". 
* CommonJs is also checked.
* Cases like "./", "../", are not checked.


docs/rules/default-import-match-filename.md
```diff
@@ -0,0 +1,27 @@
...
+Enforces default import name to match filename. Name matching is case-insensitive, and characters `._-` are stripped.
+
+## Rule Details
+
+### Fail
+
+```js
+import bar from './foo';
+import utilsFoo from '../utils/foo';
+import foo from '../foo/index.js';
```
ljharb: this should not be a failure; the proper name for any `foo/index` is `foo`.

src/rules/default-import-match-filename.js
```diff
@@ -0,0 +1,120 @@
+import docsUrl from '../docsUrl'
+import isStaticRequire from '../core/staticRequire'
+
+/**
+ * @param {string} filename
+ * @returns {string}
+ */
+function removeExtension(filename) {
+  return filename.replace(/\.[^.]+$/, '')
```
ljharb:
```suggestion
  return path.basename(filename, path.extname(filename))
```
with an `import path from 'path'` at the top

src/rules/default-import-match-filename.js
```diff
@@ -0,0 +1,120 @@
...
+  return /^(\.\/|\.\.\/)/.test(path)
+}
+
+/**
+ * Get filename from a path.
+ * @param {string} path
+ * @returns {string | undefined}
+ */
+function getFilename(path) {
+  if (!isLocalModule(path)) return undefined
```
ljharb: why would we not want this enforced on things from node_modules? `import foo from 'package/bar'` should be an error.

src/rules/default-import-match-filename.js
```diff
@@ -0,0 +1,120 @@
...
+  const [, filename] = /\/([^/]*)$/.exec(path)
+  if (filename === '' || filename === '.' || filename === '..') {
```
ljharb: with either `.` or `../`, i would expect the import name to be "the current directory name" (which would prefer `package.json`'s name, if there was a `package.json` in the current directory).

golopot: Done. File IO scares me.

golopot:
What does the "exception" option looks like?
```js
// When "name" and "path" both globly match an item in whitelist, the error is ignored.
{
  whitelist: [
    {
      name: '*', // any name is allowed
      path: '*/models/*',
    },
    {
      name: 'reactDomSever',
      path: 'react-dom/sever',
    },
  ];
}
```

src/rules/default-import-match-filename.js
golopot: I think this use case brings lots of complexity (if then else, file IO, and reading `package.json`) with little benefit. Using `require('..')` or `require('../..')` happens rare enough and is an anti pattern IMO. I would like to ignore this case.

And a hairy case: `import src from '..';`, when `..` resolves to `src`.

src/rules/default-import-match-filename.js
golopot: The gist is that this that rule can be annoying if you cannot control the filename, like when it a file in other peoples' package, which does not adhere to this rule.

src/rules/default-import-match-filename.js
ljharb: In no way is it an antipattern; and it happens in every `test` dir in all my projects.

The recursive case is just as hairy with `from '.'`, and both need to be handled.

ljharb:
This seems like a duplicate of #1143. @golopot, could you pull in whatever changes are relevant from that PR, ideally in a separate commit with that PR's author, and update this PR?  I'll then keep both in sync.

docs/rules/default-import-match-filename.md
```diff
@@ -0,0 +1,44 @@
...
+
+Enforces default import name to match filename. Name matching is case-insensitive, and characters `._-` are stripped.
+
+## Rule Details
+
+### Options
+
+#### `ignorePaths`
+
+Set this option to `['some-dir/', 'bb']` to ignore import statements whose path contains either `some-dir/` or `bb` as a substring.
```
ljharb: should this support globs, and that should be in the example?

src/rules/default-import-match-filename.js
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Match paths consisting of only '.' and '..', like '.', './', '..', '../..'.
+ * @param {string} path
+ * @returns {boolean}
+ */
+function isAncestorRelativePath(path) {
+  return (
+    path.length > 0 &&
+    path[0] !== '/' &&
```
ljharb:
```suggestion
    !path.startsWith('/') &&
```

src/rules/default-import-match-filename.js
```diff
@@ -0,0 +1,206 @@
...
+  )
+}
+
+/**
+ * @param {string} packageJsonPath
+ * @returns {string | undefined}
+ */
+function getPackageJsonName(packageJsonPath) {
+  try {
+    const packageJsonContent = fs.readFileSync(packageJsonPath).toString()
```
ljharb:
```suggestion
    const packageJsonContent = String(fs.readFileSync(packageJsonPath))
```
always String(x), never x.toString()

src/rules/default-import-match-filename.js
```diff
@@ -0,0 +1,206 @@
...
+      .every(segment => segment === '..' || segment === '.' || segment === '')
+  )
+}
+
+/**
+ * @param {string} packageJsonPath
+ * @returns {string | undefined}
+ */
+function getPackageJsonName(packageJsonPath) {
+  try {
```
ljharb: why can't this do a `require(packageJsonPath).name`?

src/rules/default-import-match-filename.js
armano2:
```suggestion
    (path[0] !== '/' || path[0] !== '\\') &&
```

src/rules/default-import-match-filename.js
```diff
@@ -0,0 +1,206 @@
...
+ * Match paths consisting of only '.' and '..', like '.', './', '..', '../..'.
+ * @param {string} path
+ * @returns {boolean}
+ */
+function isAncestorRelativePath(path) {
+  return (
+    path.length > 0 &&
+    path[0] !== '/' &&
+    path
+      .split('/')
```
armano2:
```suggestion
      .split(/[/\\]/)
```

src/rules/default-import-match-filename.js
ljharb: On a windows machine, an absolute path would start with `c://`, no?

ljharb:
ping @golopot, it'd be great to get this in :-)

src/rules/default-import-match-filename.js
golopot: This should be resolved.

src/rules/default-import-match-filename.js
golopot: I think both ways are okay.

src/rules/default-import-match-filename.js
golopot: Is this needed? I have never seen people using backslash in an import path.

docs/rules/default-import-match-filename.md
golopot: It seems that globs is not a good tool against relative paths.

ljharb: can you elaborate? it seems like a good tool to me.

src/rules/default-import-match-filename.js
ljharb: I'd assume Windows users might?

src/rules/default-import-match-filename.js
ljharb: it seems a lot simpler to use the require approach than fs + parse?

src/rules/default-import-match-filename.js
```diff
@@ -0,0 +1,203 @@
...
+              type: 'string',
+            },
+          },
+        },
+      },
+    ],
+  },
+
+  create(context) {
+    const ignorePaths = context.options[0]
```
ljharb: could this be made into a Set, so `has` operations on it are `O(1)`?

src/rules/default-import-match-filename.js
armano2: its allowed thus itshould be supported, normally no one should do this as its making code work only on windows machines

ljharb: Arguably that's true for absolute linux paths, just in reverse.

docs/rules/default-import-match-filename.md
golopot: To perform a glob matching a cwd is required, for example:
```js
// Using a hypothetical glob library
glob.match({
  pattern: 'tests/**/*.js'
  target: '/home/john/project-a/tests/a.js'
  cwd: '/home/john/project-a'
})
// true
```
What should the cwd be in this rule? The first thought is to use the location of `eslintrc` as cwd, but it is problematic when there is another `eslintrc` in a subdirectory, or when `overrides` is used.

golopot:
Currently the rule is case-insesative, and underscores are ignored. Variable name `foo_BarBAeZ` is allowed for file `foo_bar_baz`.

https://github.com/benmosher/eslint-plugin-import/blob/dfdd7d43a06c1f449788a9921b2973c7f13e0eab/tests/src/rules/default-import-match-filename.js#L50-L53

docs/rules/default-import-match-filename.md
ljharb: the cwd for any eslint rule is always `process.cwd()`, since you can't be sure where the config file is located.

src/rules/default-import-match-filename.js
golopot: There is no usage of `.has` for now, and there is usage of `.some`, so it is more convenient to leave it as an array.

docs/rules/default-import-match-filename.md
golopot: Using `process.cwd()` taken as the cwd means that the rule will give different result depending on the cwd of the shell, is that fine? 

```
// option for this rule
{
  ignoredPaths: ['tests/**/*.js']
}

project-root$ npx eslint . # will ignore files in tests/ 
project-root/tests$ npx eslint . # will **not** ignore files in tests/
```

src/rules/default-import-match-filename.js
ljharb: I’m not sure i understand why it’s more convenient, nor why existing usage of .has is relevant.

src/rules/default-import-match-filename.js
armano2: yes, you can use both `C:\` or `C://` or even `C:/`

docs/rules/default-import-match-filename.md
```diff
@@ -0,0 +1,44 @@
...
+
+Enforces default import name to match filename. Name matching is case-insensitive, and characters `._-` are stripped.
+
+## Rule Details
+
+### Options
+
+#### `ignorePaths`
+
+This option accepts an array of glob patterns. The glob patterns are to be matched against the resolved **abosolute** path of import statements. As an example, with the option `{ignorePaths: ['**/foo.js']}`, the statement `import whatever from './foo.js'` is ignored, since `./foo.js` resolves to, say, `/home/me/thing/foo.js`, which matches the glob pattern `**/foo.js`.
```
golopot: This needs re-review.

src/rules/default-import-match-filename.js
golopot: Since the test cases for absolute paths passes, I will leave the implementation as is.
https://github.com/benmosher/eslint-plugin-import/blob/64c6999a4627d2ea70fbb174194d66fc8b6af699/tests/src/rules/default-import-match-filename.js#L55-L57

golopot:
I am having trouble in designing the option `ignorePatterns`. This option is meant to ignore import statements when the import source matches a configured pattern. There are three choices:

1. (Current implementation) Use the resolved absolute path to match against the patterns. For example, in `import a from './foo'`, the source resolves to absolute path `/home/me/my-project/src/foo.js`, and that path is used to match against a pattern, say `**/foo.js`. In this case the pattern `**/foo.js` matches. This approach has the drawback that only patterns starting with `**/` can ever match, and "relative" patterns like `src/foo.js` or `foo.js` will never match in general.

2. Use the directory containing the closest `package.json` as the "cwd" for glob matching, With this approach, patterns `src/foo.js` or `foo.js` are considered to be relative to `/home/me/my-project`, which is the location of the closest `package.json`.

3. Use `process.cwd` as the "cwd" for glob matching. This approach has the drawback that the lint result depends on the cwd used for running the linter.

ljharb:
Typically, process.cwd() is the current dir - because a rule can't know if its config came from a file (or which one), or from command line config, or an inline comment.

With option 1, however - could that resolved path have `process.cwd()` stripped off it, and all patterns applied relative to that?

golopot:
> With option 1, however - could that resolved path have process.cwd() stripped off it, and all patterns applied relative to that?

It should be doable using the function path.relative(). That sounds like option 3 in https://github.com/benmosher/eslint-plugin-import/pull/1476#issuecomment-592915191?

ljharb:
Yes, option 3 sounds right to me.

stropho:
Looking forward to this rule. Is there any autofix provided? Or not for now ?

Pearce-Ropion:
It would be great to see a little bit more control over how this rule operates in the parserOptions. For example, the ability to make this case sensitive or the ability to whitelist which characters are ignored.
```
// These would be the defaults
{
    caseSensitive: false,
    allowedChars: [".", "_", "-"],
}
```

For example:
My repo uses absolute imports and I would like to ensure that the component is imported exactly as the filename is styled. Lets say I have a component called `GridColumn`:
```
parserOptions: {
    caseSensitive: true,
    ignoredChars: [],
}
```
**pass - `import GridColumn from "components/GridColumn";`**
fail - `import gridColumn from "components/GridColumn";`
fail - `import Gridcolumn from "components/GridColumn";`
fail - `import gridcolumn from "components/GridColumn";`
fail - `import Grid-Column from "components/GridColumn";`
fail - `import Grid_Column from "components/GridColumn";`
fail - `import GRIDCOLUMN from "components/GridColumn";`
etc...

ljharb:
@Pearce-Ropion while the casing should ideally match, remember that not all filesystems are case-sensitive.

Pearce-Ropion:
> @Pearce-Ropion while the casing should ideally match, remember that not all filesystems are case-sensitive.

Of course, thats why I suggested it as an optional setting. My reasoning here that some people have the luxury of knowing what systems their code is being developed on (for example, every engineer at my company is required to use a linux machine). Also, after having a glance at the code, it looks like the `isCompatible()` fn is doing a string equality check which would make it relatively easy to add these options.

Suggested changes snippet
```
/**
 * @param {string} filename
 * @param {object} options
 * @returns {string}
 */
function normalizeFilename(filename, options) {
  const allowedCharsRegex = new RegExp(`[${options.allowedChars.map(c => `\${c}`)}]`, 'g');

  if (options.caseSensitive) {
    return filename.replace(allowedCharsRegex, '');
  }

  return filename.replace(allowedCharsRegex, '').toLowerCase()
}

/**
 * Test if local name matches filename.
 * @param {string} localName
 * @param {string} filename
 * @param {object} options
 * @returns {boolean}
 */
function isCompatible(localName, filename, options) {
  const allowedCharsRegex = new RegExp(`[${options.allowedChars.map(c => `\${c}`)}]`, 'g');

  let normalizedLocalName = localName.replace(allowedCharsRegex, '');
  if (!options.caseSensitive) {
    normalizedLocalName = normalizedLocalName.toLowerCase();
  }

  return (
    normalizedLocalName === normalizeFilename(filename, options) ||
    normalizedLocalName === normalizeFilename(removeExtension(filename), options)
  )
}
```

docs/rules/default-import-match-filename.md
ljharb: yes, that seems fine to me.

granmoe:
It will be awesome to have this rule 👍  Let's merge this thing 🚀

epiqueras:
https://github.com/benmosher/eslint-plugin-import/issues/325#issuecomment-280840161

I think that should be part of this. When would you ever want to enforce this for an import name and not the export name, granted it's an internal module.

fregante:
Does this need further work or is it done?

ljharb:
It needs further work, as far as I'm aware. I'll mark it as a draft.

dword-design:
I'm also working on an NPM package. But is quite some work because you need lots of options for customization. Will write when it's published.

giladgd:
Is there any update on this? It would be great to have this as part of `eslint-plugin-import` IMO

PR #1484, opened 2019-09-18
Feature/support infinity in json for no cycle

### LINKED ISSUE #1482: no-cycle Infinity not usable in JSON config
The following JSON config can't be applied, because JSON does not support `Infinity` nor `NaN`.

```json
{
  "rules": {
    "import/no-cycle": ["error", { "maxDepth": Infinity }]
  }
}
```

https://stackoverflow.com/questions/1423081/json-left-out-infinity-and-nan-json-status-in-ecmascript

> Configuration for rule "import/no-cycle" is invalid:
	Value "Infinity" should be integer.

I suggest to support a string `"Infinity"`.

AndyOGo:
fixes #1482

### DIFF
--- a/src/rules/no-cycle.js
+++ b/src/rules/no-cycle.js
@@ -36,7 +36,7 @@ module.exports = {
     if (myPath === '<text>') return {} // can't cycle-check a non-file
 
     const options = context.options[0] || {}
-    const maxDepth = options.maxDepth || Infinity
+    const maxDepth = +options.maxDepth || Infinity
 
     function checkSourceValue(sourceNode, importer) {
       const imported = Exports.get(sourceNode.value, context)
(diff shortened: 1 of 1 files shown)


src/rules/no-cycle.js
```diff
@@ -26,7 +35,7 @@ module.exports = {
     if (myPath === '<text>') return {} // can't cycle-check a non-file
 
     const options = context.options[0] || {}
-    const maxDepth = options.maxDepth || Infinity
```
ljharb: this means that if you simply don’t set the maxDepth option, it will default to infinity. I don’t think we need an extra schema value to cover this.

AndyOGo: @ljharb Thank you, yeah I understand the meaning of this conditional assignment.

As you may know, sometimes folks wish to explicitly set options, like here:
https://github.com/airbnb/javascript/blob/7932a52e4b8f5a96089b245e2086769bf30cf0e4/packages/eslint-config-airbnb-base/rules/imports.js#L233-L235

PR #1491, opened 2019-09-29
[core]: regex marking packages as internal

### LINKED ISSUE #1125: How to treat imports inside a monorepo as internal
I am working on a monorepo where I have some files with a list of imports that looks like this:

```
import React from 'react';

import MyComponent from '@my/src/MyComponent';

import styles from './styles.scss';
```

In this example, the second import refers to a package inside the monorepo and I would like it to be treated as an internal import by `import/order`. Is this possible to achieve?

Librazy:
fixes #1125

We are developing packages in a monorepo and packages depend on each other so we need to separate our package from external dependency. We used to [tslint/ordered-imports](https://palantir.github.io/tslint/rules/ordered-imports/) which allow custom group by regex pattern.

This PR added a new setting which allowed user to provide a regex pattern and packages that match that pattern will be marked internal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 
 ### Added
+- [`internal-regex`]: regex pattern for marking packages "internal"  ([#1491], thanks [@Librazy])
+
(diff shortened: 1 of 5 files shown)


vincaslt:
Yes please, useful for monorepos, what's the plan for this?

PR #1492, opened 2019-10-01
fix: utils requires for resolver-node to be dependent #828

anion155:
fix https://github.com/benmosher/eslint-plugin-import/issues/828.
Tested with pnpm by linking resolver-node to module-utils's node_modules folder as if it was declared in peerDependencies, it worked. So I'm assuming just putting it in peerDependencies would work too.

### DIFF
--- a/utils/package.json
+++ b/utils/package.json
@@ -27,5 +27,8 @@
   "dependencies": {
     "debug": "^2.6.8",
     "pkg-dir": "^2.0.0"
+  },
+  "peerDependencies": {
+    "eslint-import-resolver-node": "^0.3.2"
   }
 }


utils/package.json
```diff
@@ -27,5 +27,8 @@
   "dependencies": {
     "debug": "^2.6.8",
     "pkg-dir": "^2.0.0"
+  },
+  "peerDependencies": {
+    "eslint-import-resolver-node": "^0.3.2"
```
ljharb: this definitely isn't the right solution; eslint-module-utils does not require the node resolver.

anion155:
@ljharb Sorry for long reaction. So then add `resolve` package and wrap it to match `utils/resolve`'s exports would be a good approach?

ljharb:
I'm still not sure what the underlying issue is; but anything that works with `npm` but doesn't work with an alternative client, is broken in the alternative client by definition.

https://github.com/benmosher/eslint-plugin-import/issues/828#issuecomment-310159079 is the expected solution here. Want to update the PR to do only that?

ljharb:
This seems to have been fixed by #1591.

PR #1493, opened 2019-10-01
fix(importType): Accept '@example' as internal

### LINKED ISSUE #1379: [import/order]: Typescript path incorrect groups
I ran into a weird case of import/order with new lines between groups.
I'm using typescript & the typescript resolver (apologies, if this issue is on the resolver side, just tell me and I'll open it there).

I created a repo for this as that's simpler than trying to explain: https://github.com/tomaskallup/eslint-import-typescript-bug

Just clone, `yarn/npm install` & `yarn/npm run lint`. It will report lines in `atomics/box/box-article/box-article.ts` as wrong, because it assings different groups to imports of same type (they should be internal export according to documentation of the import/order rule) and the rule requires space between them.

Since the imports start with '@', this might be related to #1293, but I didn't really find anything helpful to my case.

AamuLumi:
Hi !

Fix #1379 

The bug was related to the `internal` regex which didn't accept a module with only one word.
This is the case in Typescript when you use the tconfig `paths` to accept the import of `@myModule` in replacement of `/src/someFolder/myModule`.

Thanks for developing this library!

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -43,7 +43,7 @@ export function isExternalModuleMain(name, settings, path) {
   return externalModuleMainRegExp.test(name) && isExternalPath(path, name, settings)
 }
 
-const scopedRegExp = /^@[^/]+\/[^/]+/
+const scopedRegExp = /^@[^/]+\/?[^/]+/
 function isScoped(name) {
   return scopedRegExp.test(name)
 }
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -30,6 +30,7 @@ describe('importType(name)', function () {
   })
 
   it("should return 'external' for scopes packages", function() {
+    expect(importType('@cycle/', context)).to.equal('external')
     expect(importType('@cycle/core', context)).to.equal('external')
     expect(importType('@cycle/dom', context)).to.equal('external')
     expect(importType('@some-thing/something', context)).to.equal('external')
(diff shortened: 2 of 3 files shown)


src/core/importType.js
```diff
@@ -43,7 +43,7 @@ export function isExternalModuleMain(name, settings, path) {
   return externalModuleMainRegExp.test(name) && isExternalPath(path, name, settings)
 }
 
-const scopedRegExp = /^@[^/]+\/[^/]+/
+const scopedRegExp = /^@[^/]+\/?[^/]+/
```
ljharb: hmm, this also seems to allow `@foo/` - can you add an explicit test for that?

AamuLumi: When I tested, the '@foo/` case is recognized as `external`.
Is it correct, or must I change that?

tests/src/core/importType.js
```diff
@@ -96,7 +98,6 @@ describe('importType(name)', function () {
   })
 
   it("should return 'unknown' for any unhandled cases", function() {
-    expect(importType('@malformed', context)).to.equal('unknown')
```
ljharb: why does this need to be removed?

AamuLumi: `@malformed` is not anymore an unhandled case. This is the purpose of this PR to accept `@foo` pattern.
The test has been replaced by [this one](https://github.com/benmosher/eslint-plugin-import/pull/1493/files#diff-410595ac51ca4f4f193da8d5704b1b0bR59).

src/core/importType.js
AamuLumi: @ljharb I re-up this question

ljharb: I think that because it's got the `/`, it's a folder in node_modules, so external seems right?

AamuLumi: I have the same opinion on this point.
So it seems ok since there's a test for `@cycle/`.

Menci:
But it doesn't accept "@/aaa/bbb" paths.

ljharb:
@Menci `@/` isn't a valid scoped package.

Please file a new issue though, we'll see what we can do.

Menci:
Sorry, I just noticed this commit is so old that this RegExp has been
modified to support `@/`. It's my other configurations that caused my error.

PR #1494, opened 2019-10-05
fix: import/no-cycle triggering errors for Flow imports

### LINKED ISSUE #1343: import/no-cycle should not trigger for Flow imports
This is a regression of #1098 (fix in PR #1126) that was re-introduced in v2.17.0.

v2.16.0 works as expected. v2.17.2 still has this regression.

maxmalov:
The changes fix a complex circular dependency where Flow type imports are transient dependencies of the checked module. Basically such imports can be ignored too since they will be stripped anyway

Fixes #1343

### DIFF
--- a/tests/files/cycles/flow-types.js
+++ b/tests/files/cycles/flow-types.js
@@ -1,0 +1,6 @@
+// @flow
+
+import type { FooType } from './flow-types-depth-two';
+import { type BarType } from './flow-types-depth-two';
+
+export const bar = 1;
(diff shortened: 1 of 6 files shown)


tests/files/cycles/flow-types.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+import type { FooType } from './flow-types-depth-two';
```
102: Should this also work for `import { type FooType } ...` syntax?

maxmalov: This complicates things a little, since import kind becomes the value. I didn’t discover how to identify such type imports yet

dominicfraser:
@102 @maxmalov - could we have an update on the status of this PR? :)

maxmalov:
@dominicfraser yeah, the fix supposed to handle both type import variants `import type { Foo }` and `import { type Bar }`

PR #1501, opened 2019-10-11
Fix typo

sonicdoe:


tests/src/rules/order.js
```diff
@@ -164,7 +164,7 @@ ruleTester.run('order', rule, {
         var index = require('./');
       `,
     }),
-    // Addijg unknown import types (e.g. using an resolver alias via babel) to the groups.
+    // Adding unknown import types (e.g. using an resolver alias via babel) to the groups.
```
ljharb:
```suggestion
    // Adding unknown import types (e.g. using a resolver alias via babel) to the groups.
```

sonicdoe:
Interesting, I was not aware of this (and GitHub even still shows the “Allow edits from maintainers” option). Anyway, I’ve force-pushed your “a resolver” suggestion and also fixed it in two other lines.

ljharb:
I think it's a bug in GitHub - ie, they didn't account for this use case because virtually nobody makes PRs from any fork but their personal fork.

PR #1514, opened 2019-10-24
Delete unnecesary, malformed .babelrc

soryy708:
Opening the deleted `.babelrc` file in my editor yields the following error: "Expected a JSON object, array or literal    .jsonc(0)". (sort of linting the .babelrc)

This file is unnecessary because babel already uses the one in the root.

If I checkout the code with Windows line ends (CRLF), `npm install` where appropriate, and `npm test` - the tests fail. Deleting this `.babelrc` file makes the test pass.


soryy708:
Ok, CI is nice. Perhaps a different solution...

soryy708:
Trying to fix it breaks CI too. Hmm...
According to the [.babelrc docs](https://babeljs.io/docs/en/6.26.3/babelrc), I still think there's no reason for this `.babelrc` file to exist. According to "Lookup behavior", it should be able to find the `.babelrc` at the root.

soryy708:
`es2015-argon` ???

ljharb:
If you cd into the folder and run Babel, i don’t believe it finds the config in the root.

PR #1516, opened 2019-10-24
Fix broken .babelrc

soryy708:
Reason for change:
If I checkout the code with Windows line ends (CRLF), `npm install` where appropriate, and `npm test` - the tests fail.

Deleting this `.babelrc` file makes the test pass, but [breaks CI](https://github.com/benmosher/eslint-plugin-import/pull/1514).

Opening the old `.babelrc` file in my editor yields the following error: "Expected a JSON object, array or literal .jsonc(0)". (sort of linting the .babelrc)

### DIFF
--- a/resolvers/webpack/.babelrc
+++ b/resolvers/webpack/.babelrc
@@ -1,1 +1,3 @@
-../../.babelrc
+{
+  "extends": "../../.babelrc"
+}
--- a/resolvers/webpack/.babelrc
+++ b/resolvers/webpack/.babelrc
@@ -1,3 +1,3 @@
 {
-  "extends": "../../.babelrc "
+  "extends": "../../.babelrc"
 }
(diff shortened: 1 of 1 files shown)


soryy708:
> Resolve error: ENOENT: no such file or directory, open 'C:\\projects\\eslint-plugin-import\\resolvers\\webpack\\.babelrc

... But that's false. The file is not deleted, it's here. Just contents changed.
I wonder if it's a CI bug & a re-run will fix?

soryy708:
According to the [.babelrc docs](https://babeljs.io/docs/en/6.26.3/babelrc), I still think there's no reason for this `.babelrc` file to exist. According to "Lookup behavior", it should be able to find the `.babelrc` at the root.

ljharb:
This is Babel 6.

PR #1524, opened 2019-10-29
add external modules setting

### LINKED ISSUE #1510: Add setting to specify external modules by package name pattern
I'm proposing a new setting to specify external modules by a package name pattern to satisfy the following use case:

We have a monorepo consisting of multiple packages that depend on each other. We'd like to enforce that every package that is imported by another package in the same repository is listed as a dependency in package.json. We want to use the import/no-extraneous-dependencies rule to enforce this, however, we can't specify that each package in the monorepo is an external dependency because:
1. We resolve package dependencies using a custom resolver that resolves packages to their source location
2. Eslint detects these dependencies as internal modules because their resolved path does not fall under node_modules

In the past, our packages were scoped which resulted in them being treated as external modules because of #1293. However, since that has been fixed we now run into this problem.

The existing `import/external-module-folders` setting isn't sufficient for us because:

1. Our packages don't reside as a flat list under a single directory such as node_modules
2. The package dir name does not match the package name in all cases, especially not with scoped prefixes

Instead, a new `import/external-module-patterns` setting that allowed one to specify an array of string/regex patterns to match against package names, could be used to match all of our packages as external since they all share a common scope.

---

A second option could be to add some further metadata in the returned value from custom resolvers to specify whether a module is external or not.

---

Alternatively, it would be worth revisiting the decision made in #943 to only report errors for dependencies that can be resolved and provide a config option to opt-in to reporting errors for unresolved dependencies. As then, we would no longer need to mark our modules as external and would just remove our custom resolver altogether so that they are always unresolved and therefore marked as external.

ValeryVS:
fixes #1510


README.md
```diff
@@ -341,6 +341,24 @@ Contribution of more such shared configs for other platforms are welcome!
 
 An array of folders. Resolved modules only from those folders will be considered as "external". By default - `["node_modules"]`. Makes sense if you have configured your path or webpack to handle your internal paths differently and want to considered modules from some folders, for example `bower_components` or `jspm_modules`, as "external".
 
+#### `import/external-modules`
```
ljharb: https://github.com/benmosher/eslint-plugin-import#importexternal-module-folders already exists - how is this different?

src/core/importType.js
```diff
@@ -25,12 +25,15 @@ export function isBuiltIn(name, settings, path) {
...
 function isExternalPath(path, name, settings) {
+  const modules = (settings && settings['import/external-modules']) || []
   const folders = (settings && settings['import/external-module-folders']) || ['node_modules']
 
   // extract the part before the first / (redux-saga/effects => redux-saga)
   const packageName = name.match(/([^/]+)/)[0]
 
-  return !path || folders.some(folder => -1 < path.indexOf(join(folder, packageName)))
+  return !path
+    || folders.some(folder => -1 < path.indexOf(join(folder, packageName)))
```
ljharb: no need for a yoda condition, and "includes" exists:
```suggestion
    || folders.some(folder => path.includes(join(folder, packageName)))
```

PR #1526, opened 2019-10-30
[Typescript] Lookup for external modules in @types folder.

### LINKED ISSUE #1525: TypeScript import/order regression
I've just upgraded to `eslint-plugin-import@2.18.2` and `eslint-import-resolver-typescript@2.0.0` and started to get `import/order` errors:

Screen Shot 2019-10-29 at 17 28 05

Any module defined in `@types/*` (e.g. `@types/express` in this case) is seen as "internal", while modules with their own types (e.g. `http-status-codes`) are seen as "external". I've tried with other modules as well.

Changing the default order to:

```
    'import/order': [
      'error',
      {
        groups: ['internal', 'external'],
      },
    ],
```

Gets rid of the error.

Can't understand the root cause though, but here's some logging from the resolver if it helps:
Screen Shot 2019-10-29 at 22 09 36

The [`import/order` rule history](https://github.com/tihonove/eslint-plugin-import/commits/master/src/rules/order.js) doesn't seem to have anything suspicious. Could the new version of the resolver be causing this?

Re: https://github.com/alexgorbatchev/eslint-import-resolver-typescript/issues/33

### LINKED ISSUE #1541: cannot resolve node_modules/@types modules
thanks for the plugin first of all!

typescript support work nice, except for `import/no-unresolved` when you only install types (and not the actual package)

e.g. with `npm i @types/aws-lambda`, the following failes `import {Context} from 'aws-lambda'` in a typescript file

joaovieira:
### Why
When using TypeScript an external module can be resolved to its type definition file located in `node_modules/@types` from `eslint-import-resolver-typescript@2.0.0`. At the moment, that module would be marked as "internal" because the module is not found directly inside `node_modules`, but instead nested inside `@types`.

### What
This PR just adds the `node_modules/@types` folder to lookup for external modules when using the TypeScript config.

Fixes #1525. Fixes #1541.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,8 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Fixed
 - [`no-unused-modules`]: fix usage of `import/extensions` settings ([#1560], thanks [@stekycz])
-
-### Fixed
 - [`import/extensions`]: ignore non-main modules ([#1563], thanks [@saschanaz])
+- TypeScript config: lookup for external modules in @types folder ([#1526], thanks [@joaovieira])
 
 ## [2.19.1] - 2019-12-08
 ### Fixed
@@ -634,6 +633,7 @@ for info on changes for earlier releases.
 [#1560]: https://github.com/benmosher/eslint-plugin-import/pull/1560
 [#1551]: https://github.com/benmosher/eslint-plugin-import/pull/1551
 [#1542]: https://github.com/benmosher/eslint-plugin-import/pull/1542
+[#1526]: https://github.com/benmosher/eslint-plugin-import/pull/1526
 [#1521]: https://github.com/benmosher/eslint-plugin-import/pull/1521
 [#1519]: https://github.com/benmosher/eslint-plugin-import/pull/1519
 [#1507]: https://github.com/benmosher/eslint-plugin-import/pull/1507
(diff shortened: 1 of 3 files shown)


joaovieira:
@benmosher @ljharb any time to review this one liner?

(and maybe while you're at it, also #1528. These two are breaking for TS users with the new `eslint-import-resolver-typescript@2.0.0`).

joaovieira:
@ljharb I assumed we didn't test configs, as there are no tests for any of them. 

The `import/external-module-folders` setting has already been tested in https://github.com/benmosher/eslint-plugin-import/blob/master/tests/src/core/importType.js#L144. 

Happy to test that the config sets that setting?

ljharb:
@joaovieira while that is quite true that we don't already (my bad) it's still better to add one, as if something's not tested, that tells me it's ok to break :-)

ljharb:
I'm not sure why the travis tests are failing; this test passes for me locally.

joaovieira:
@ljharb it was already failing before this PR, and [has been failing for all PRs ever since](https://ci.appveyor.com/project/benmosher/eslint-plugin-import/history). I looked at it a while back but couldn't understand the issue as well :/

joaovieira:
@ljharb fixed. Mocha doesn't seem to play nicely with globs. Decided to put the config tests next to all the others to simplify.

Issue #1527, opened 2019-10-30
Handle TSExportAssignment exports other than namespaces


joaovieira:
At the moment `eslint-plugin-import` only handles the `export = ` syntax in TypeScript if it references a namespace. 

Using [`@types/supertest`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/supertest/index.d.ts#L13) as an example. The module definition is:

```ts
import * as superagent from "superagent"

export = supertest;

declare function supertest(app: any): supertest.SuperTest<supertest.Test>;
declare namespace supertest {...}
```

And the generated `ExportMap` is:

```js
ExportMap {
  path: '/Users/joaovieira/.../node_modules/@types/supertest/index.d.ts',
  namespace: Map {
    'Response' => {},
    'Request' => {},
    'CallbackHandler' => {},
    'Test' => {},
    'AgentOptions' => {},
    'agent' => {},
    'SuperTest' => {}
  },
  reexports: Map {},
  dependencies: Set {},
  imports: Map {
    '/Users/joaovieira/.../node_modules/@types/superagent/index.d.ts' => { getter: [Function], source: [Object], importedSpecifiers: [Set] }
  },
  errors: []
}
```

Only the namespace is being exported. The declared function with the same name is not added to the exported namespace as 'default'.

Looking at the code, it is explicitly [only handling namespaces](https://github.com/benmosher/eslint-plugin-import/blob/c14c9bd665525865f023b7214bbaf3d0124563dd/src/ExportMap.js#L517).

This means, user code such as:

```ts
import request, { Test } from 'supertest';
```

Raises an eslint error `No default export found in module.eslint(import/default)`.

___

The same issue happens when the module re-exports an identifier as in [`@types/lodash`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash/snakeCase.d.ts). The module definition is:

```ts
import { snakeCase } from "./index";
export = snakeCase;
```

The ExportMap in this case is an empty object:

```
ExportMap {
  path: '/Users/joaovieira/.../node_modules/@types/lodash/snakeCase.d.ts',
  namespace: Map {},
  reexports: Map {},
  dependencies: Set {},
  imports: Map {
    '/Users/joaovieira/.../node_modules/@types/lodash/index.d.ts' => { getter: [Function], source: [Object], importedSpecifiers: [Set] }
  },
  errors: []
}
```

So, this fails:

```js
[...]

PR #1528, opened 2019-10-30
[TypeScript] Support default exports in TSExportAssignment.

joaovieira:
Allows TypeScript export assignments of default identifiers and/or mixed merged declarations. E.g.:

```ts
declare const foobar: number;
export = foobar;
```

```ts
declare function foobar(): void;
declare namespace foobar {
  type MyType = string
}
export = foobar;
```

```ts
import { foobar } from './module';
export = foobar;
```

```ts
// No longer reports missing default export
import foobar, { MyType } from './foobar';
```

Fixes: #1527


src/ExportMap.js
```diff
@@ -514,30 +514,50 @@ ExportMap.parse = function (path, content, context) {
...
+      const exportedDecls = ast.body.filter(({ type, id, declarations }) => 
+        declTypes.includes(type) && 
+        (id && id.name === exportedName || declarations.find(d => d.id.name === exportedName))
       )
-      moduleDecls.forEach((moduleDecl) => {
-        if (moduleDecl && moduleDecl.body && moduleDecl.body.body) {
-          moduleDecl.body.body.forEach((moduleBlockNode) => {
+      if (exportedDecls.length === 0) {
+        // Export is not referencing any local declaration, must be re-exporting
+        m.namespace.set('default', captureDoc(source, docStyleParsers, n))
```
joaovieira: I'm not sure about `n` here. Would it be worth looking through the `m.imports` to see if any matches the exported name? We don't do that on [regular exports](https://github.com/benmosher/eslint-plugin-import/pull/1528/files#diff-e991b36c6628220049668b5e2d2c43d8R501). E.g.:

```
import { test } from './test';
export { foo }; // all good...
```

JounQin:
It should read about `esModuleInterop` option or add an equal option for this plugin instead of hard coding the logic.

ljharb:
@JounQin can you elaborate?

TS's module system is broken without synthetic imports and esModuleInterop enabled; I'm completely comfortable telling people they need to enable those if they want stuff to work properly.

Maxim-Mazurok:
What a great PR! I was searching for a solution all day, finally gave up and started to dig into the sources. Strangely enough, this PR didn't come up while I was searching for similar issues.
So, I started coding basically the same thing: #1689

@JounQin I agree with you, in up PR I'm reading from `tsconfig.json`. But, there might be multiple tsconfigs and they might include/exclude specific files. So, the best approach would be to first, figure out which tsconfig is covering file where the import is made and then check if that config has `esModuleInterop` enabled.
Here's an [example](https://github.com/alexgorbatchev/eslint-import-resolver-typescript/blob/585509e95f93adf8b7ef5839029c19c55edbe76e/src/index.ts#L148-L171) of code from `eslint-import-resolver-typescript` that does kind of similar thing: 
```js
mappers = configPaths!
  // turn glob patterns into paths
  .reduce<string[]>(
    (paths, path) => paths.concat(isGlob(path) ? globSync(path) : path),
    [],
  )
  .map(loadConfig)
  .filter(isConfigLoaderSuccessResult)
  .map(configLoaderResult => {
    const matchPath = createMatchPath(
      configLoaderResult.absoluteBaseUrl,
      configLoaderResult.paths,
    )

    return (source: string, file: string) => {
      // exclude files that are not part of the config base url
      if (!file.includes(configLoaderResult.absoluteBaseUrl)) {
        return undefined
      }

      // look for files based on setup tsconfig "paths"
      return matchPath(source, undefined, undefined, extensions)
    }
  })
```

That being said, we should either do the right thing and figure out how to handle complex cases. Or, we can do the simple thing and only cover simple case (only root tsconfig) or introduce new leg in `parserOptions` that will also act globally but will be more explicit than assuming that everybody has only one tsconfig.

Also, I can't disagree with @ljharb about assuming that everyone uses `esModuleInterop` nowadays. It seems to be a pretty popular flag to use. And in this case, I would be very happy to see this PR merged, thanks!

src/ExportMap.js
```diff
@@ -530,30 +530,50 @@ ExportMap.parse = function (path, content, context) {
...
+        'TSDeclareFunction',
+        'TSEnumDeclaration',
+        'TSTypeAliasDeclaration',
+        'TSInterfaceDeclaration',
+        'TSAbstractClassDeclaration',
+        'TSModuleDeclaration',
+      ]
+      const exportedDecls = ast.body.filter(({ type, id, declarations }) => 
+        declTypes.includes(type) && 
+        (id && id.name === exportedName || declarations.find(d => d.id.name === exportedName))
```
julien1619, line 546: Hey! Here `declarations` is undefined unless `type === 'VariableDeclarations'` so it crashes the linter in some cases. I'll create another PR to fix it.

ljharb: Thanks!

richard-lopes-ifood:
Nice fixing :rocket:! When is it going to be released?

PR #1529, opened 2019-10-30
Fix linting errors.

joaovieira:
Run `eslint . --fix`. There are still 11 warnings. Not entirely sure if they were already there before. Happy to fix those too.


ljharb:
Things that are warnings shouldn't be autofixed; can you revert and make sure you're running `eslint . --fix --quiet`?

joaovieira:
There was only one fixable/fixed warning https://github.com/benmosher/eslint-plugin-import/pull/1529/files#diff-48ef751391942af34a1ce845e0e31b27R26, still comma-dangle related.

Also, just noticed this is also a duplicate of https://github.com/benmosher/eslint-plugin-import/pull/1520. I can revert that one warning fix here if you want.

PR #1534, opened 2019-11-02
Exclude flow `typeof` imports from extraneous dependencies

devongovett:
Type imports were already ignored, but `typeof` imports should also be ignored.

For example:

```javascript
// @flow
import typeof TypeScriptModule from 'typescript';
```


ljharb (approved):
Rebased, and went ahead and used the code in your OP as a test case.

PR #1551, opened 2019-12-03
[Refactor] Avoid superfluous calls and code

brettz9:
LGTM.com often identifies some issues not found by ESLint; this PR fixes the issues found at https://lgtm.com/projects/g/benmosher/eslint-plugin-import/?mode=list . In this case, they are just some harmless superfluous arguments and redundant checks, but helpful to fix both for clarity and to reduce noise among the lgtm alerts.

(The site has a [Github integration](https://lgtm.com/help/lgtm/github-apps-integration) which I find worth adding to check PRs.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,6 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Added
 - [`internal-regex`]: regex pattern for marking packages "internal"  ([#1491], thanks [@Librazy])
-
-### Added
 - [`group-exports`]: make aggregate module exports valid ([#1472], thanks [@atikenny])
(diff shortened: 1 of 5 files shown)


src/core/importType.js
```diff
@@ -72,14 +72,14 @@ function isRelativeToSibling(name) {
 }
 
 function typeTest(name, settings, path) {
-  if (isAbsolute(name, settings, path)) { return 'absolute' }
```
ljharb: these are all intentional, not bugs, to minimize future refactoring.

brettz9: Yeah, I figured they may be. I've reverted for this file. Unfortunately, LGTM's disabling only works at the individual line or file level, so disabling these warnings from LGTM would probably be considered too ugly, triggering as they would `max-len` warnings.

PR #1560, opened 2019-12-09
[Fix] `no-unused-modules`: fix usage of `import/extensions` settings

### LINKED ISSUE #1555: no-unused-modules does not work for modules  in files other than .js
I have a few Typescript only projects I would like to turn the rule on. However, I am not able to do that because the rule loads files with `.js` extension only.

It is caused by `FileEnumarator` default settings that are used in [the rule implementation](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-unused-modules.js#L21).

I have tried to use `import/extensions` settings, however, that does not work (obviously when you see the code).

I can see 3 possible solutions:

1. use the default `FileEnumarator` from the eslint itself
2. use the `import/extensions` settings for the `FileEnumarator`
3. use another settings option for the `FileEnumarator`

I believe there is a reason why the default `FileEnumarator` from the eslint itself is not used so the first option is invalid. Adding another settings option might be unnecessary because it would be probably duplication of `import/extensions`.

stekycz:
Fixes #1555

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Fixed
+- [`no-unused-modules`]: fix usage of `import/extensions` settings ([#1560], thanks [@stekycz])
 
 ### Fixed
 - [`import/extensions`]: ignore non-main modules ([#1563], thanks [@saschanaz])
@@ -624,6 +626,7 @@ for info on changes for earlier releases.
 [`memo-parser`]: ./memo-parser/README.md
 
 [#1563]: https://github.com/benmosher/eslint-plugin-import/pull/1563
+[#1560]: https://github.com/benmosher/eslint-plugin-import/pull/1560
 [#1551]: https://github.com/benmosher/eslint-plugin-import/pull/1551
 [#1542]: https://github.com/benmosher/eslint-plugin-import/pull/1542
 [#1521]: https://github.com/benmosher/eslint-plugin-import/pull/1521
@@ -1042,3 +1045,4 @@ for info on changes for earlier releases.
 [@christophercurrie]: https://github.com/christophercurrie
 [@randallreedjr]: https://github.com/randallreedjr
 [@Pessimistress]: https://github.com/Pessimistress
+[@stekycz]: https://github.com/stekycz
--- a/src/rules/no-unused-modules.js
+++ b/src/rules/no-unused-modules.js
@@ -5,6 +5,7 @@
  */
 
 import Exports from '../ExportMap'
+import { getFileExtensions } from 'eslint-module-utils/ignore'
 import resolve from 'eslint-module-utils/resolve'
 import docsUrl from '../docsUrl'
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+- [`no-unused-modules`]: fix usage of `import/extensions` settings ([#1560], thanks [@stekycz])
```
```suggestion
### Fixed
- [`no-unused-modules`]: fix usage of `import/extensions` settings ([#1560], thanks [@stekycz])
```

src/rules/no-unused-modules.js
```diff
@@ -17,18 +18,27 @@ import includes from 'array-includes'
 let listFilesToProcess
 try {
   var FileEnumerator = require('eslint/lib/cli-engine/file-enumerator').FileEnumerator
-  listFilesToProcess = function (src) {
-    var e = new FileEnumerator()
+  listFilesToProcess = function (src, extensions) {
+    var e = new FileEnumerator({
```
ljharb: weird that these are still `var`s, could we change them to const and let?

tests/src/rules/no-unused-modules.typescript.js
```diff
@@ -0,0 +1,42 @@
+import { test, testFilePath } from '../utils'
+import typescriptConfig from '../../../config/typescript'
+
+import { RuleTester } from 'eslint'
+
+// We need to clear the cache to reinitialize the rule for Typescript only
```
ljharb: seems like maybe instead of requiring this (since the rule will run on both kinds of files in the same run), we should move the listFilesToProcess code in the rule from module level down to the rule logic itself?

src/rules/no-unused-modules.js
stekycz: I was surprised as well. I update these 2 usages of `var`, however, there are more of them in the repository. I would leave the rest as is in the scope of this PR.

ljharb: agreed, thanks

tests/src/rules/no-unused-modules.typescript.js
stekycz: That would require moving more functions in the module to the logic itself. The `FileEnumerator` class [uses its own cache in the module level](https://github.com/eslint/eslint/blob/master/lib/cli-engine/file-enumerator.js#L195-L212) and that could cause an unexpected increase in memory usage. I might dig deeper to find a better solution. However, that might end in greater refactoring of the rule itself. Do you want it? Or is it a good idea for another PR instead?

ljharb: fair, let's work on that in a followup PR instead of this one.

stekycz: I have added JSX only test cases. However, I have had to do a few adjustments as well to make it work as you suggested. So consider the work half-done. I believe there is still space for improvement.

ljharb:
Tests seem to be failing in eslint 2-5.

stekycz:
Should be fixed. I have tested it for eslint 2-6 on Node 10 on OSX.

PR #1569, opened 2019-12-12
fix [import/named] for importing from a module which re-exports `*` from node_modules or commonjs module

### LINKED ISSUE #1446: [import/named] fails to track deep exports when re-exporting from a module in node_modules

redbugz:
Fixes #1446 
Based on #1447 

When re-exporting `*` from a commonjs module, or a module in node_modules which comes through as a commonjs module, do not error since the commonjs exports cannot be evaluated. Treat with similar logic to ignored modules or directly importing from a commonjs module.

When re-exporting `*`, the module is listed in dependencies instead of `reexports`. There is similar logic for when re-exporting a named export, to check for a null from the exportCache, which I understand to mean an ignored or unparseable es5/commonjs module:
https://github.com/benmosher/eslint-plugin-import/blob/master/src/ExportMap.js#L91

There is a todo in this location, and I'm not sure if this code satisfies that todo and it can be removed or not.
Also, when logging the value of `innerMap`, I never it saw it be undefined. So it's possible the `continue` line is also no longer needed, but I am not sure.

This fixes `hasDeep`, but there is similar logic in `has` and `get`:
https://github.com/benmosher/eslint-plugin-import/blob/master/src/ExportMap.js#L68
https://github.com/benmosher/eslint-plugin-import/blob/master/src/ExportMap.js#L146
I'm not sure if those should get similar treatment, or if those are not relevant for this issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`extentions`]: Fix scope regex ([#1611], thanks [@yordis])
 - [`no-duplicates`]: allow duplicate imports if one is a namespace and the other not ([#1612], thanks [@sveyret])
 - Add some missing rule meta schemas and types ([#1620], thanks [@bmish])
+- [`named`]: for importing from a module which re-exports named exports from a `node_modules` module ([#1569], [#1447], thanks [@redbugz], [@kentcdodds])
 
 ### Changed
 - [`import/external-module-folders` setting] behavior is more strict now: it will only match complete path segments ([#1605], thanks [@skozin])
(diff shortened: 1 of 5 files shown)


yuri-sakharov:
Looks like build checks failed. I also need this fix. When will it be in release  version. Thanks a lot.

PR #1572, opened 2019-12-15
[extensions] Ignore query strings when checking for extensions.

### LINKED ISSUE #1567: [extensions] Alerting on missing extension when using query string
Since last release 2.19.1 (although I did not check 2.19.0), I get some new errors due to my imports using query strings:
```
error  Missing file extension "svg" for "images/logo.svg?fill=#fff"  import/extensions
```

This is because I use a webpack loader that transforms the SVG if I provide the right query string, however I would expect eslint-plugin-import to find that I have an `.svg` in there.

I think that this bug was always there bug got revealed due to #1521 fixing #1518.

pcorpet:
Fixes #1567.

### DIFF
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -132,10 +132,12 @@ module.exports = {
       // bail if the declaration doesn't have a source, e.g. "export { foo };"
       if (!source) return
 
-      const importPath = source.value
+      const importPathWithQueryString = source.value
 
       // don't enforce anything on builtins
-      if (isBuiltIn(importPath, context.settings)) return
+      if (isBuiltIn(importPathWithQueryString, context.settings)) return
+
+      const importPath = importPathWithQueryString.replace(/\?(.*)$/, '')
 
       const resolvedPath = resolve(importPath, context)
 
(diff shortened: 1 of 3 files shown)


src/rules/extensions.js
```diff
@@ -154,14 +156,14 @@ module.exports = {
           context.report({
             node: source,
             message:
-              `Missing file extension ${extension ? `"${extension}" ` : ''}for "${importPath}"`,
+              `Missing file extension ${extension ? `"${extension}" ` : ''}for "${source.value}"`,
```
ljharb:
```suggestion
              `Missing file extension ${extension ? `"${extension}" ` : ''}for "${importPathWithQueryString}"`,
```

src/rules/extensions.js
```diff
@@ -154,14 +156,14 @@ module.exports = {
...
           })
         }
       } else if (extension) {
         if (isUseOfExtensionForbidden(extension) && isResolvableWithoutExtension(importPath)) {
           context.report({
             node: source,
-            message: `Unexpected use of file extension "${extension}" for "${importPath}"`,
+            message: `Unexpected use of file extension "${extension}" for "${source.value}"`,
```
ljharb:
```suggestion
            message: `Unexpected use of file extension "${extension}" for "${importPathWithQueryString}"`,
```

PR #1586, opened 2019-12-27
Enhancement alphabetize case insensitive

### LINKED ISSUE #1581: Option to make "alphabetize" sorting case insensitive
The new "alphabetize" option is great! For those of us migrating from tslint to eslint, it would be nice if we could have an extra option to make this alphabetical sorting case insensitive.

dbrewer5:
Fixes #1581

### DIFF
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -189,16 +189,18 @@ import index from './';
 import sibling from './foo';
 ```
 
-### `alphabetize: {order: asc|desc|ignore}`:
+### `alphabetize: {order: asc|desc|ignore, caseInsensitive: true|false}`:
 
 Sort the order within each group in alphabetical manner based on **import path**:
 
 - `order`: use `asc` to sort in ascending order, and `desc` to sort in descending order (default: `ignore`).
+- `caseInsensitive`: use `true` to ignore case, and `false` to consider case (default: `false`).
 
 Example setting:
 ```js
 alphabetize: {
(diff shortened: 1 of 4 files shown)


docs/rules/order.md
```diff
@@ -210,12 +212,14 @@ import React, { PureComponent } from 'react';
 import aTypes from 'prop-types';
 import { compose, apply } from 'xcompose';
 import * as classnames from 'classnames';
+import blist from 'BList';
```
ljharb: perhaps a relative path would be better here, since npm package names can't be capitalized (for many years now)

src/rules/order.js
```diff
@@ -278,10 +278,14 @@ function mutateRanksToAlphabetize(imported, order) {
 
   const groupRanks = Object.keys(groupedByRanks)
 
-  const sorterFn = order === 'asc' ? importsSorterAsc : importsSorterDesc
+  const sorterFn = alphabetizeOptions.order === 'asc' ? importsSorterAsc : importsSorterDesc
   // sort imports locally within their group
   groupRanks.forEach(function(groupRank) {
-    groupedByRanks[groupRank].sort(sorterFn)
+    groupedByRanks[groupRank].sort((a, b) => {
+      return sorterFn.apply(null, alphabetizeOptions.caseInsensitive
```
ljharb: rather than using apply here, i'm going to hoist the comparator out of the iteration

PR #1587, opened 2019-12-30
Add debug ^4.1.1 to dependencies

artursvonda:
Adding newer debug version while keeping same node support

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "array-includes": "^3.0.3",
     "array.prototype.flat": "^1.2.1",
     "contains-path": "^0.1.0",
-    "debug": "^2.6.9",
+    "debug": "^2.6.9 || ^4.1.1",
     "doctrine": "1.5.0",
     "eslint-import-resolver-node": "^0.3.2",
     "eslint-module-utils": "^2.4.1",
--- a/resolvers/node/package.json
+++ b/resolvers/node/package.json
@@ -28,7 +28,7 @@
   },
   "homepage": "https://github.com/benmosher/eslint-plugin-import",
   "dependencies": {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -83,7 +83,7 @@
     "array-includes": "^3.0.3",
     "array.prototype.flat": "^1.2.1",
     "contains-path": "^0.1.0",
-    "debug": "^2.6.9",
+    "debug": "^2.6.9 || ^4.1.1",
```
ljharb: this doesn't actually keep the same node support; `npm install` will install debug v4, which means that it will force users to install a top-level v2 and rely on deduping to be able to get back their support. thus, this is a breaking change.

artursvonda: The recommendation by NPM is to enable `engine-strict` option which will force v2 be installed on older node versions since debug v4 requires node v8+.
I thought this is a nice way to allow newer versions of the dependency while still having solution for older node versions.

ljharb: That's only something the top-level app can do - meaning, if the consumers of this package have to do *any work whatsoever* to keep the plugin working, then it's a breaking change, and that includes setting `engine-strict` to true (which would be a terrible idea, since "engines" is purely advisory, and should never block installs).

What's the point of a newer version tho? What do we get in v4 or v3 that we don't get in v2?

artursvonda: Fair enough.

The reasoning is just keeping up with latest dependencies and allow deduping.

ljharb: Linting is a dev dep, deduping doesn’t matter; as for keeping up with the latest, that only matters if there’s something that we’re missing out on. Otherwise, it’s just capriciously breaking consumers for no good reason.

PR #1591, opened 2020-01-02
utils: Uses createRequireFromPath to resolve loaders

arcanis:
The current logic uses `require.resolve` in order to locate the resolver modules. The problem is that `require.resolve` semantically operates based on the currently evaluated module. That makes the resolution incorrect, as the loaders aren't dependencies of `eslint-module-utils`. It probably works thanks to the hoisting, but hoisting is on its way out and such problems will start to reveal themselves.

To fix that, the resolution simply needs to use `createRequireFromPath` so that the loaders are located based on the dependencies of the package that contains the source file being linted - which I think was the intent.

### DIFF
--- a/tests/files/node_modules/eslint-import-resolver-foo/index.js
+++ b/tests/files/node_modules/eslint-import-resolver-foo/index.js
@@ -1,0 +1,1 @@
+../../foo-bar-resolver-v2.js
--- a/tests/src/core/resolve.js
+++ b/tests/src/core/resolve.js
@@ -94,6 +94,14 @@ describe('resolve', function () {
                       )).to.equal(utils.testFilePath('./bar.jsx'))
   })
 
+  it('finds resolvers from the source files rather than eslint-module-utils', function () {
+    const testContext = utils.testContext({ 'import/resolver': { 'foo': {} } })
+
+    expect(resolve( '../files/foo'
+                      , Object.assign({}, testContext, { getFilename: function () { return utils.getFilename('foo.js') } }),
+                      )).to.equal(utils.testFilePath('./bar.jsx'))
+  })
+
   it('reports invalid import/resolver config', function () {
     const testContext = utils.testContext({ 'import/resolver': 123.456 })
     const testContextReports = []
--- a/tests/src/rules/no-unresolved.js
+++ b/tests/src/rules/no-unresolved.js
@@ -343,9 +343,9 @@ ruleTester.run('no-unresolved unknown resolver', rule, {
     // logs resolver load error
     test({
       code: 'import "./malformed.js"',
-      settings: { 'import/resolver': 'foo' },
+      settings: { 'import/resolver': 'doesnt-exist' },
       errors: [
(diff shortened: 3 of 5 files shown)


ljharb:
Is there a way to have a test for this?

arcanis:
Added a test that fails on master and passes on this branch 👍

johanneswilm:
This seems to fail whenever one doesn't follow the standard directory structure of eslint6 and because this was released as a patch version it's also pulled in by users of eslint5.

johanneswilm:
I wonder what the need was to merge this on. Was ot broken for anyone before this was merged? If not, I'd suggest undoing this as there is then no benefit to anyone and it only creates problems for some users.
In general there should be an eslint option to specify where to resolve modules, resolvers, parsers from. Otherwise it just doesn't work for the workflows of many.

arcanis:
> Was ot broken for anyone before this was merged? If not, I'd suggest undoing this as there is then no benefit to anyone and it only creates problems for some users.

Well, as you can guess it wouldn't have been merged unless it had been fixing something.

Without going into the specific, the loaders were loaded in a rule-breaking way that would have caused a lot of pain in a few weeks. We certainly can find a smoother migration path, but reverting this diff won't go into our users' best interests.

johanneswilm:
Could this be applied in a way so that it doesn't get installed for users of eslint5 and current versions of eslint6? For example by not just marking it a patch? Or what the benefit be for them (me)?

DavideDaniel:
As yarn 2 gains adoption this is becoming a real issue at our company. We are currently debating removing this plugin which would be unfortunate as it's provided immense value. What can we do to help address this?

ljharb:
@DavideDaniel have you fully updated eslint-plugin-import and eslint-module-utils? As far as I'm aware, this issue is fixed for yarn v2.

webuniverseio:
I came here because I also tried to get yarn 2 working with for eslint-config-airbnb-base. It was not to clear why it was not working and fortunately after I added `eslint-module-utils` I got rid of resolver errors. It is quite possible that I overlooked a place where it was advised to have it installed... anyway, here is a list of dependencies I needed:
```
    "eslint": "~6.8.0",
    "eslint-config-airbnb-base": "^14.1.0",
    "eslint-module-utils": "^2.6.0",
    "eslint-plugin-import": "^2.20.2",
```

DavideDaniel:
> @DavideDaniel have you fully updated eslint-plugin-import and eslint-module-utils? As far as I'm aware, this issue is fixed for yarn v2.

Super delayed but we fixed it via this: https://github.com/yarnpkg/berry/issues/1427#issuecomment-637538879

SirWindfield:
It still doesn't work correctly without the workaround posted above. The above one is just a workaround for packages that forgot to declare a dependency. It just tells yarn to use the package from your own devDependencies instead.

I have fully updated the packages and still experience the issue. However, I am not entirely sure how this would have to be solved. My guess would be to add a `peerDependency` that matches the version from the locally committed version inside the repository (so `0.3.4`).

PR #1595, opened 2020-01-06
chore: add LICENSE file to all npm packages

opichals:
Together with a new npm release of the modified fodlers this fixes https://github.com/benmosher/eslint-plugin-import/issues/1225


memo-parser/LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License (MIT)
```
ljharb: let's make these gitignored and copied in via prepublish from the root, rather than duplicating the content.

opichals: or better yet, what about making this a proper monorepo a use lerna@3 to deal with this.

ljharb: absolutely not; for one, we'd have to move all of our packages into some kind of "packages" dir, which is too much git churn; but also in my experience lerna simply does not work well beyond `bootstrap` and `exec`, and i have no interest in using it on additional projects.

opichals:
> let's make these gitignored and copied in via prepublish from the root, rather than duplicating the content.

@ljharb addressed

opichals:
@ljharb Thank you!

ljharb:
Every package should have been released at this point with the license, ftr.

PR #1602, opened 2020-01-11
Use createRequire instead of createRequireFromPath if available

iamnapo:
In order to avoid: `DeprecationWarning: Module.createRequireFromPath() is deprecated. Use Module.createRequire() instead.` when using Node >= `v12.2.0`.


ljharb:
cc @arcanis as a heads up

PR #1605, opened 2020-01-12
Fix import type determination for monorepo setup w/ webpack resolver

### LINKED ISSUE #1597: Incorrect behavior of import/order in a monorepo setup
Minimal reproduction of the bug and a detailed analysis of the cause: https://github.com/skozin/eslint-plugin-import-bug-repro.

## The bug

The `import/order` rule gives a false positive in a setup with the following conditions met:

1. The project uses [Yarn workspaces].
2. The packages inside the monorepo are scoped.
3. Webpack resolver is used.

## A quick example

Consider this code:

```js
import lodash from 'lodash';
import {sharedFn} from '@my-scope/package-a';

import {someFn} from './util';

export default {};
```

and this config:

```json
{
  "extends": [
    "plugin:import/errors",
    "plugin:import/warnings"
  ],
  "rules": {
    "import/order": ["error", {"newlines-between": "always"}]
  },
  "settings": {
    "import/resolver": "webpack",
    "import/external-module-folders": ["node_modules", "packages"]
  },
  "parserOptions": {
    "ecmaVersion": 2017,
    "sourceType": "module"
  }
}
```

The expected result is no errors being reported. However, this comes up instead:

```text
eslint-plugin-import-bug-repro/packages/package-b/index.js
  1:1  error  There should be at least one empty line between import groups        import/order
  4:1  error  `./util` import should occur before import of `@my-scope/package-a`  import/order
```

I've prepared a repo with a minimal reproduction and a detailed analysis of the cause: https://github.com/skozin/eslint-plugin-import-bug-repro.

## What happens

See [the repo with a reproduction](https://github.com/skozin/eslint-plugin-import-bug-repro#minimal-reproduction-of-a-bug-in-the-eslint-plugin-import-package) for a detailed explanation.

In short, the problem is that paths to scoped packages get considered as internal ones by the [`isExternalPath`] function even though the directory where these packages reside is added to the `import/external-module-folders` list in the config. Before #1294, this was not a problem since scoped packages were always considered as external ones, but after merging that PR the behavior changed and now scoped packages are considered external only if their resolved path is determined as external.

In turn, seemingly incorrect behavior of [`isExternalPath`] is caused by the logic inside it that doesn't only check that the resolved path to a package contains one of the `import/external-module-folders` as a subpath, but also that it includes the package name as a subpath. This is not true in the case of a monorepo with scoped packages since a path to a Yarn workspace containing the package contents doesn't need to include the package scope or even the package name as a subpath.

For example, in the described case the resolved path to `@my-scope/package-a` is `/path/to/the/project/packages/package-a/index.js` and it doesn't contain `@my-scope` part at all.

[Yarn workspaces]: https://yarnpkg.com/lang/en/docs/workspaces
[`isExternalPath`]: https://github.com/benmosher/eslint-plugin-import/blob/cd25d9d/src/core/importType.js#L27

skozin:
Fixes #1597.

This PR changes the logic of [`isExternalPath`] function so that it doesn't consider the import name. For a path to be external, it's now sufficient that it matches one of the items in the `import/external-module-folders` array specified in settings.

This is to support a setup where some of the modules are symlinked from `node_modules` into directories elsewhere in the project (called ["workspaces"] in Yarn terminology). For example, it's common to have a monorepo setup like this:

```
node_modules
  @my-scope
    core -> ../../packages/core
    utils -> ../../packages/utils
packages
  core
    package.json
    index.js
  utils
    package.json
    index.js
package.json
```

Usually, one wants these packages to be treated as external since they're normally installed from a registry. Intuitively, this should be achieved by adding `packages` to `import/external-module-folders` array. However, this didn't work in a project that uses the `webpack` resolver.

The reason is that `webpack` resolver returns a fully resolved path to the module's main file, e.g. for `@my-scope/core` it would be something like `/home/me/project/packages/core/index.js`. This path doesn't contain the package name, `@my-scope/core`, as a substring, so [`isExternalPath`] would consider this path as internal prior to the changes proposed in this PR. This led to the import itself being considered as internal, as after merging #1294 a scoped import is determined as external only if its path is external too.

Apart from removing the import name check from [`isExternalPath`], this PR also makes `import/external-module-folders` more strict. It won't match incomplete path segments anymore, so `some/path` won't match `/my/awesome/path` but will still match `/my/some/path` and `/my/some/path/index.js`. Also, if an item starts with a forward slash, it will be treated as an absolute path prefix, so `/home/me/project/packages` will only match the directory itself and everything inside it.

This PR doesn't break any existing tests, and I believe that it shouldn't break any existing setups either.

[`isExternalPath`]: https://github.com/benmosher/eslint-plugin-import/blob/cd25d9d/src/core/importType.js#L27
["workspaces"]: https://yarnpkg.com/lang/en/docs/workspaces/

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -1,5 +1,4 @@
 import coreModules from 'resolve/lib/core'
-import { join } from 'path'
 
 import resolve from 'eslint-module-utils/resolve'
 
@@ -26,11 +25,19 @@ export function isBuiltIn(name, settings, path) {
 
 function isExternalPath(path, name, settings) {
   const folders = (settings && settings['import/external-module-folders']) || ['node_modules']
+  return !path || folders.some(folder => isSubpath(folder, path))
+}
 
-  // extract the part before the first / (redux-saga/effects => redux-saga)
-  const packageName = name.match(/([^/]+)/)[0]
-
-  return !path || folders.some(folder => -1 < path.indexOf(join(folder, packageName)))
+function isSubpath(subpath, path) {
+  const normSubpath = subpath.replace(/[/]$/, '')
+  if (normSubpath.length === 0) {
+    return false
+  }
+  const left = path.indexOf(normSubpath)
+  const right = left + normSubpath.length
+  return left !== -1 &&
+        (left === 0 || normSubpath[0] !== '/' && path[left - 1] === '/') &&
+        (right >= path.length || path[right] === '/')
 }
 
 const externalModuleRegExp = /^\w/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Fixed
+- [`import/external-module-folders` setting] now correctly works with directories containing modules symlinked from `node_modules` ([#1605], thanks [@skozin])
+
+### Changed
+- [`import/external-module-folders` setting] behavior is more strict now: it will only match complete path segments ([#1605], thanks [@skozin])
 
 ## [2.20.0] - 2020-01-10
 ### Added
@@ -636,6 +641,7 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#1605]: https://github.com/benmosher/eslint-plugin-import/pull/1605
 [#1589]: https://github.com/benmosher/eslint-plugin-import/issues/1589
 [#1586]: https://github.com/benmosher/eslint-plugin-import/pull/1586
 [#1572]: https://github.com/benmosher/eslint-plugin-import/pull/1572
@@ -1069,3 +1075,4 @@ for info on changes for earlier releases.
 [@rsolomon]: https://github.com/rsolomon
 [@joaovieira]: https://github.com/joaovieira
 [@ivo-stefchev]: https://github.com/ivo-stefchev
+[@skozin]: https://github.com/skozin
--- a/README.md
+++ b/README.md
@@ -339,7 +339,19 @@ Contribution of more such shared configs for other platforms are welcome!
 
 #### `import/external-module-folders`
 
-An array of folders. Resolved modules only from those folders will be considered as "external". By default - `["node_modules"]`. Makes sense if you have configured your path or webpack to handle your internal paths differently and want to considered modules from some folders, for example `bower_components` or `jspm_modules`, as "external".
+An array of folders. Resolved modules only from those folders will be considered as "external". By default - `["node_modules"]`. Makes sense if you have configured your path or webpack to handle your internal paths differently and want to consider modules from some folders, for example `bower_components` or `jspm_modules`, as "external".
+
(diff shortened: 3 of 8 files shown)


skozin:
CI build is probably failing due to the failing ESLint checks of the plugin source code and tests, but these checks were as well failing before this PR.

skozin:
Hmm, not sure why Coveralls reports that coverage of `resolvers/node/index.js` decreased, I didn't change or remove any tests or code related to that resolver.

src/core/importType.js
```diff
@@ -26,11 +25,17 @@ export function isBuiltIn(name, settings, path) {
 
 function isExternalPath(path, name, settings) {
   const folders = (settings && settings['import/external-module-folders']) || ['node_modules']
+  return !path || folders.some(folder => isSubpath(folder, path))
+}
 
-  // extract the part before the first / (redux-saga/effects => redux-saga)
-  const packageName = name.match(/([^/]+)/)[0]
+function isSubpath(subpath, path) {
+  const subpathRegExpSrc = escapeRegExp(subpath.replace(/[/]$/, ''))
```
ljharb: i'm pretty concerned here about trying to manually regex escape and make a dynamic regex. is there a more reliable means of doing this test?

skozin: There definitely is — I can do manual parsing and checking in a loop. I still don't think that the current way is unreliable, but I see how the more imperative solution might get better readability, so I'll go ahead with it.

Just in case, how the current way works:

1. Remove the trailing slash in case it's presented in an item.
2. In the case the item contains special regex symbols, [escape them](https://stackoverflow.com/a/3561711/804678).
3. Make a regex that checks if a path contains the item as a complete subpath:
3.a. If the item starts with a slash, make a regex that checks if a path contains the item between the start of the string and a path segment boundary (either end of the string or a slash).
3.b. Otherwise, make a regex that checks if a path contains the item between path segment boundaries (either start/end of the string or a slash).

But I see how this can be pretty unreadable from the code, so I'll re-write this using a loop instead.

src/core/importType.js
```diff
@@ -26,11 +25,19 @@ export function isBuiltIn(name, settings, path) {
...
+function isSubpath(subpath, path) {
+  const normSubpath = subpath.replace(/[/]$/, '')
+  if (normSubpath.length === 0) {
+    return false
+  }
+  const left = path.indexOf(normSubpath)
+  const right = left + normSubpath.length
+  return left !== -1 &&
+        (left === 0 || normSubpath[0] !== '/' && path[left - 1] === '/') &&
+        (right >= path.length || path[right] === '/')
```
skozin: @ljharb, how about this one? I did without any loops or dynamic regexes, so now it should be more clear what the thing does. If that's ok, I'll force-push overwriting the original commit.

ljharb: looks great

skozin: Awesome, I'll push in couple of hours then

skozin: @ljharb, done! Please let me know if there's anything else that's needed in order for this to be merged. Thank you!

ljharb:
@skozin i've rebased this and tweaked the test descriptions a bit, but tests are failing. Can you take a look?

skozin:
@ljharb, thank you so much for assisting me and merging this!

PR #1606, opened 2020-01-12
Makes the loader resolution more tolerant

### LINKED ISSUE #1603: Using `import/resolver` in a global config file
I use a global ESLint config file. It contains most `import` rules, plus:

```js
module.exports = {
  settings: {
    'import/resolver': {
      node: {
        extensions: ['.js', '.jsx', '.ts', '.tsx'],
      },
    },
  },
}
```

I lint with it using the command `eslint . --resolve-plugins-relative-to /Users/gary/.dotfiles/` and a `.eslintrc.js` file of the following, in my projects:

```js
module.exports = {
  extends: '/Users/gary/.eslintrc.js',
};
```

Every `import` rule reports the error: `Resolve error: unable to load resolver "node"`.

This error only started appearing after upgrading from v2.18.2 to v2.20.0.

### LINKED ISSUE #1604: eslint-module-utils 2.5.1 broken
Hey, when using `eslint-module-utils` 2.5.1, my eslint stops being able to resolve most imports. I receive lots of errors of this type: 

```
1:1   error    Resolve error: unable to load resolver "node"    import/namespace
```

Pinning it at 2.5.0 makes all the errors go away. I won't ass lots of logs here as I assume it's something obvious. If it turns out to be more complicated, let me know and I'll give a full description of my setup. This was tested both with eslint 5.16.0 and 6.8.0.

arcanis:
(Sorry @iamnapo, it was very late at night and I initially opened this PR against your repo ... 😅)

Assuming that the repro provided by https://github.com/benmosher/eslint-plugin-import/issues/1604#issuecomment-573364743 will be valid, a simple solution would be to be too laxist for now and just fallback to the hoisting resolution if the file-relative one doesn't find the loader - while still using the expected resolution as primary code path.

And then the next time eslint-import-utils receives a major bump, revert the try/catch block and go back to using exclusively createRequire.

cc @ljharb, fixes #1604, fixes #1603

### DIFF
--- a/utils/resolve.js
+++ b/utils/resolve.js
@@ -34,7 +34,11 @@ function tryRequire(target, sourceFile) {
   try {
     // Check if the target exists
     if (sourceFile != null) {
-      resolved = createRequire(path.resolve(sourceFile)).resolve(target)
+      try {
+        resolved = createRequire(path.resolve(sourceFile)).resolve(target)
+      } catch (e) {
+        resolved = require.resolve(target)
+      }
     } else {
       resolved = require.resolve(target)
     }
--- a/utils/CHANGELOG.md
+++ b/utils/CHANGELOG.md
@@ -5,6 +5,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## Unreleased
 
+- Makes the loader resolution more tolerant ([#1606], thanks [@arcanis])
 - Use `createRequire` instead of `createRequireFromPath` if available ([#1602], thanks [@iamnapo])
 
 ## v2.5.1 - 2020-01-11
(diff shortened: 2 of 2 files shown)


utils/resolve.js
```diff
@@ -34,7 +34,11 @@ function tryRequire(target, sourceFile) {
   try {
     // Check if the target exists
     if (sourceFile != null) {
-      resolved = createRequire(path.resolve(sourceFile)).resolve(target)
+      try {
+        resolved = createRequire(path.resolve(sourceFile)).resolve(target)
+      } catch {
```
ljharb: bindingless catch isn't covered by our babel settings atm
```suggestion
      } catch (e) {
```

Issue #1538, opened 2019-11-12
no-duplicates: add an allow-namespace-imports option


sveyret:
I would like to migrate from `tslint` `no-duplicate-imports` rule, but `import/no-duplicates` is missing one important option: `allow-namespace-imports`. This option is required, especially with `React` because we often need to do such things:

    import * as React from 'react';
    import { FunctionComponent } from 'react;

In this case, it is a syntax error to put both imports on the same line, but according to the doc, the rule will prevent us from doing that. The `allow-namespace-imports` allow the import to be done twice on the same module if one of the two imports is a namespace import.

See https://palantir.github.io/tslint/rules/no-duplicate-imports/

ljharb:
That doesn't seem like it should need to be an option; if it's a syntax error to combine them, then it should just allow this case.

ljharb:
a PR with test cases and optionally a fix would be most helpful!

sveyret:
Hi, I'm currently working on it.

PR #1612, opened 2020-01-15
no-duplicates: allow duplicate if one is namespace and other not

sveyret:
It is a syntax error to put both namespace and non namespace import on the same line, so allow it.
Fix issue #1538

### DIFF
--- a/tests/src/rules/no-duplicates.js
+++ b/tests/src/rules/no-duplicates.js
@@ -31,12 +31,20 @@ ruleTester.run('no-duplicates', rule, {
       code: "import x from './bar?optionX'; import y from './bar?optionY';",
       options: [{'considerQueryString': true}],
       settings: { 'import/resolver': 'webpack' },
-     }),
+    }),
     test({
       code: "import x from './foo'; import y from './bar';",
       options: [{'considerQueryString': true}],
       settings: { 'import/resolver': 'webpack' },
-     }),
+    }),
+
+    // #1538: It is impossible to import namespace and other in one line, so allow this.
+    test({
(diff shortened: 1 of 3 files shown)


sveyret:
Note that this is a simple implementation which will not check:

    import * as ns from './foo'
    import './foo'

tests/src/rules/no-duplicates.js
```diff
@@ -179,17 +187,24 @@ ruleTester.run('no-duplicates', rule, {
     }),
 
     test({
-      code: "import * as ns from './foo'; import {y} from './foo'",
-      // Autofix bail because first import is a namespace import.
```
ljharb: i'd prefer not to modify existing tests, to help ensure it's not a breaking change.

sveyret: Well, for this specific test, it IS a breaking change, because it currently is an error, which we don't want it to be anymore.

ljharb: turning an error into a non-error isn't breaking, it's a patch :-) i see that you've actually just moved this test up to "valid", so we're good here

SCasarotto:
> Note that this is a simple implementation which will not check:
> 
> ```
> import * as ns from './foo'
> import './foo'
> ```
> 
> I'm currently working on something better…

@sveyret Did you ever circle back on this? I am seeing this issue come up with the following: 
```
import firebase from 'firebase/app';
import 'firebase/auth';
import 'firebase/database';
```
and was trying to figure out if this was a mistake on my side or if I should create an issue.

sveyret:
I don’t really remember, but I don’t think I did anything since this PR. It means an evolution is needed.

PR #1620, opened 2020-01-20
Add eslint-plugin-eslint-plugin internally and fix violations

bmish:
This PR adds [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin) (a popular plugin for linting eslint plugins), enables relevant rules from it, and fixes violations.

The primary changes included are:
1. Adds missing rule schemas ([require-meta-schema](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-schema.md)).
    1a. Note that `schema: []` is added to enforce when a rule should have no schema.
    1b. Removed the `es6-only` option from a test case for the `named` rule since that option was removed a long time ago.
2. Adds missing rule `type` property to `no-unused-modules` rule ([require-meta-type](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-type.md)).
3. Removes duplicate test cases (autofixed by [no-identical-tests](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-identical-tests.md)).


bmish:
It looks like there are some Node 4 + eslint 2/3 failures due to incompatibility with eslint-plugin-eslint-plugin. As a path forward, I could still fix all the violations without actually adding the new dev dependency on eslint-plugin-eslint-plugin?

src/rules/no-nodejs-modules.js
```diff
@@ -14,6 +14,17 @@ module.exports = {
     docs: {
       url: docsUrl('no-nodejs-modules'),
     },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          allow: {
+            type: 'array',
```
ljharb: what is the type of the items? must they be unique?

tests/src/rules/no-commonjs.js
```diff
@@ -30,15 +29,15 @@ ruleTester.run('no-commonjs', require('rules/no-commonjs'), {
     },
 
     // allowed requires
-    { code: 'function a() { var x = require("y"); }' }, // nested requires allowed
```
ljharb: i prefer the object form in all cases; it reduces diff churn when later adding options.

bmish: Sounds good, I reverted this change.

src/rules/no-nodejs-modules.js
bmish: Good catch, added those checks.

ljharb:
Thanks!

PR #1629, opened 2020-01-26
Set `eslint-plugin/consistent-output` lint rule to always require test case output assertions (internal only)

bmish:
Use the more strict option for this internal [eslint-plugin/consistent-output](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/consistent-output.md) lint rule. This will ensure that developers always assert what the autofixer output should be.

If a rule has no autofix, we should assert that there is no output. All of the test cases I fixed have no autofix, so the output matches the input. Once eslint v2 support is dropped, we can switch to `output: null` to represent this, and re-enable the [eslint-plugin/prefer-output-null](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-output-null.md) internal lint rule.

Since this change is internal-only, I have not added it to the changelog.

Follow-up to #1620.


bmish:
Unfortunately, it looks like this change fails in eslint v2. I believe that's because eslint v2 does not support `output: null`.

ljharb:
@bmish does it support repeating the input as the output?

bmish:
@ljharb good point, I switched to repeating the input as the output.

PR #1637, opened 2020-01-30
Add import/extensions ignorePackages rule to the TypeScript plugin

thewilkybarkid:
Should prevent https://github.com/benmosher/eslint-plugin-import/issues/1615 when using `plugin:import/typescript` and others (like `airbnb-base`) which aren't TypeScript aware.

Refs https://github.com/benmosher/eslint-plugin-import/issues/1615#issuecomment-579936931.

### DIFF
--- a/config/typescript.js
+++ b/config/typescript.js
@@ -6,6 +6,15 @@ var allExtensions = ['.ts', '.tsx', '.d.ts', '.js', '.jsx']
 
 module.exports = {
 
+  rules: {
+    'import/extensions': ['error', 'ignorePackages', {
+      ts: 'never',
+      tsx: 'never',
+      js: 'never',
+      jsx: 'never',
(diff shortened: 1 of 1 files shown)


thewilkybarkid:
That sounds like the best solution.

In the meantime, found [eslint-config-airbnb-typescript](https://github.com/iamturns/eslint-config-airbnb-typescript) which resolves my usage for now.

rfgamaral:
> 
> 
> That sounds like the best solution.
> 
> In the meantime, found [eslint-config-airbnb-typescript](https://github.com/iamturns/eslint-config-airbnb-typescript) which resolves my usage for now.

Sorry about hijacking this but if you also use Prettier you might be interested on this config: https://github.com/rfgamaral/eslint-config-typescript-unified

Issue #1643, opened 2020-02-03
[import/order] Regression in 2.20.1


South-Paw:
Installed `2.20.1`, started getting ordering errors like the following;

![image](https://user-images.githubusercontent.com/2154764/73621398-c7402880-469a-11ea-8330-9bf933cf8437.png)

Our rule for `import/order` is

```
'import/order': [
  'error',
  {
    groups: ['builtin', 'external', 'parent', 'sibling', 'index'],
    'newlines-between': 'never',
  },
],
```

Rolling back `eslint-plugin-import` to `2.20.0` resolves the issue.

yoyo837:
Only in Windows.

ljharb:
That makes testing for it difficult; appveyor tests are currently failing for unrelated reasons.

targos:
What I found by debugging:
- The import type of an external module is not determined as such
- The reason is that the `path` passed to `isSubpath` contains Windows separators (backslashes) and the function expects slashes.

ljharb:
`path.normalize` might help us here?

kryops:
This also seems to affect `import/no-extraneous-dependencies`: As the import type is reported as `internal` instead of `external` due to the path containing backslashes, the rule never reports any errors on Windows.

kryops:
I tried to create a minimal reproduction for the `import/order` and `import/no-extraneous-dependencies` rules: https://github.com/kryops/eslint-plugin-import-issue1643

ljharb:
A PR with a failing test case would be most welcome; I can add a fix to that (altho a PR that has both is even better!)

targos:
I think the existing tests should already fail on Windows. I was unable to run the tests on that platform when I tried today, though.

raphinesse:
I had to roll back to `2.19.1` for this issue to go away. `2.20.0` complained about the order of the following imports:
```js
const nearley = require('nearley');
const grammar = require('./grammar');
```

ljharb:
@South-Paw can you check on latest master of this plugin and see if it's fixed?

ljharb:
I'll close this for now, but am happy to reopen if the issue isn't solved on latest master.

flymans:
Still isn't working for me

thyde1:
If this issue has been resolved in master, is there a reason that it has not yet been released?

ljharb:
Because maintainers are humans who aren’t being paid for their time, and can’t always drop everything to cut a release :-)

[...]

PR #1648, opened 2020-02-07
Fix build on Windows

raphinesse:
This fixes the package installation and test running on Windows to expose failing tests related to #1643.

More details in the commit messages.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,7 +8,7 @@ environment:
   # - nodejs_version: "4"
 
 matrix:
-  fast_finish: true
+  fast_finish: false
 
   # allow_failures:
   #   - nodejs_version: "4" # for eslint 5
@@ -26,6 +26,9 @@ install:
   - ps: >-
       if ($env:nodejs_version -eq "4") {
         npm install -g npm@3;
+      }
+      if ($env:nodejs_version -in @("8", "10", "12")) {
+        npm install -g npm@6.10.3;
       }
   - npm install
 


raphinesse:
Apparently using `nyc@15` does not support `node@<8`. Given that those Node.js versions are EOL, this does not seem to be a big issue to me. However, that fact would have to be addressed with further changes.

ljharb:
EOL is irrelevant; we'll support what our supported eslint versions work on.

package.json
```diff
@@ -31,6 +30,7 @@
     "test-compiled": "npm run prepublish && NODE_PATH=./lib mocha --compilers js:babel-register --recursive tests/src",
     "test-all": "npm test && for resolver in ./resolvers/*; do cd $resolver && npm test && cd ../..; done",
     "prepublish": "npm run build",
+    "prepublishOnly": "npm run copy-metafiles",
```
ljharb: why is this changed? we want it running after the build step, not just on prepublish.

package.json
```diff
@@ -76,7 +76,7 @@
     "eslint-plugin-import": "2.x",
     "linklocal": "^2.8.2",
     "mocha": "^3.5.3",
-    "nyc": "^11.9.0",
+    "nyc": "^15.0.0",
```
ljharb: I'd prefer not to upgrade this; i've asked on https://github.com/istanbuljs/nyc/issues/1205 about a workaround.

package.json
raphinesse: Why do you want it to run after the build step? From what it does, it seems to be only necessary or at all meaningful if done before packaging.

The reason for limiting  the `copy-metafiles` script to `prepublishOnly` is that it throws a syntax error if run on Windows (because it's a `sh` script), one of the two reasons the tests don't run on Appveyor CI right now.

This is a quick fix that only prevents publishing from Windows versus preventing to publish or even to do a clean `npm install`.

package.json
raphinesse: Another option would be to decouple testing and coverage collection and either

- update to `nyc@15` and stop collecting coverage on node@<8
- stop collecting coverage on windows

raphinesse:
Playing around with the node or npm version in the CI config _might_ serve to work around the issue of tests not running on Appveyor but it won't resolve the issue for someone cloning the repo to their local Windows machine.

ljharb:
That is true. I'll keep playing with alternatives.

ljharb:
k, looks like adding in-publish, and downgrading npm on node 12, has exposed 50 failing tests, that seem legitimate.

package.json
ljharb: I fixed this by adding `in-publish` rather than using the hacky `prepublishOnly` or by breaking `npm run build`.

raphinesse:
Regarding your changes to the `prepublish` script: Is it intended that the `build` script is neither run on local `npm install` without any arguments nor on `npm pack`? You can achieve that by using the built-in `prepublishOnly` w/out using any additional dependencies. I would consider that less hacky.

ljharb:
`prepublishOnly` doesn't work on all npm versions, nor does `prepare`; I'll only ever use `prepublish` on any project.

raphinesse:
> `prepublishOnly` doesn't work on all npm versions, nor does `prepare`; I'll only ever use `prepublish` on any project.

Ah, I see. against the backdrop of trying to support as many `npm` versions as possible, it makes sense, of course.


What about the other part of my question?
> Is it intended that the `build` script is neither run on local `npm install` without any arguments nor on `npm pack`?

armano2:
there seems to be more issues on windows

https://github.com/armano2/test-eslint-monorpo/actions/runs/36409848

ljharb:
@raphinesse no; that's what i added `not-in-publish`; that way the build script only runs with `npm publish`, and not with `npm install`.

@armano2 the current appveyor failures in this PR seem to be legitimate issues on windows, yes.

fisker:
I think #1651 is the reason those tests fails

ljharb:
@fisker i rebased #1651 on top of this, and its tests are still failing (albeit with 1 failure instead of 50). Almost there :-)

appveyor.yml
```diff
@@ -22,11 +22,10 @@ install:
   # Get the latest stable version of Node.js or io.js
   - ps: Install-Product node $env:nodejs_version
 
+  # Use npm@6.10.3 to work-around issues w/ nyc@<15 on Windows
+  - npm install -g npm@6.10.3
```
ljharb: please prefer the if-based check; this won't work on node 4, for example.

raphinesse: Node 4 is commented out in the Appveyor config. I usually only consider live code. But this branch is exploratory anyhow. I do not find the `npm` version pinning done here to be a satisfactory solution to the `nyc` problem (See https://github.com/benmosher/eslint-plugin-import/pull/1648#issuecomment-583602524).

ljharb: yes, but the intention is to uncomment it.

Fair that it's exploratory, but we were down to 1 failing test before you both started force pushing everywhere :-)

ljharb: As for local dev, the first step is to get CI working - all local dev by maintainers is done on a Mac.

Once CI is working, we can try to fix local Windows dev, for the rare contributor that uses it.

raphinesse: > yes, but the intention is to uncomment it.

Understood 


> Fair that it's exploratory, but we were down to 1 failing test before you both started force pushing everywhere :-)

But test not running at all on Node 8 and 10.

ljharb: Again, first step is to get CI passing *somehow*.

We can uncomment node 10, 8, 6, 4, etc on appveyor in a future PR.

raphinesse: > As for local dev, the first step is to get CI working - all local dev by maintainers is done on a Mac.

Understood

raphinesse:
@ljharb I'm done force-pushing here. Can confirm that `npm` version pinning resolves the `nyc` problem on all Node versions currently running on AppVeyor.

ljharb:
It does still have to stay open until it’s merged tho :-)

PR #1651, opened 2020-02-11
[Fix] `order`: fix `isExternalModule` detect on windows

fisker:
`isExternalModule` is not right, when use `\` as separator.

This cause external module detected as `internal`.

Seems we are not running test on windows. I add a test on `isExternalModule`

### DIFF
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -29,15 +29,16 @@ function isExternalPath(path, name, settings) {
 }
 
 function isSubpath(subpath, path) {
-  const normSubpath = subpath.replace(/[/]$/, '')
+  const normPath = path.replace(/\\/g, '/')
+  const normSubpath = subpath.replace(/\\/g, '/').replace(/\/$/, '')
   if (normSubpath.length === 0) {
     return false
   }
-  const left = path.indexOf(normSubpath)
+  const left = normPath.indexOf(normSubpath)
   const right = left + normSubpath.length
   return left !== -1 &&
-        (left === 0 || normSubpath[0] !== '/' && path[left - 1] === '/') &&
-        (right >= path.length || path[right] === '/')
+        (left === 0 || normSubpath[0] !== '/' && normPath[left - 1] === '/') &&
+        (right >= normPath.length || normPath[right] === '/')
 }
 
 const externalModuleRegExp = /^\w/
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -1,7 +1,7 @@
 import { expect } from 'chai'
 import * as path from 'path'
 
-import importType from 'core/importType'
+import importType, {isExternalModule} from 'core/importType'
 
 import { testContext, testFilePath } from '../utils'
 
@@ -180,6 +180,12 @@ describe('importType(name)', function () {
   })
 
   it('returns "external" for a scoped module from a symlinked directory which partial path is contained in "external-module-folders" (webpack resolver)', function() {
+    const originalFoldersContext = testContext({
+      'import/resolver': 'webpack',
+      'import/external-module-folders': [],
+    })
+    expect(importType('@test-scope/some-module', originalFoldersContext)).to.equal('internal')
+
     const foldersContext = testContext({
       'import/resolver': 'webpack',
       'import/external-module-folders': ['files/symlinked-module'],
@@ -223,5 +229,12 @@ describe('importType(name)', function () {
       'import/external-module-folders': [testFilePath('symlinked-module')],
     })
     expect(importType('@test-scope/some-module', foldersContext)).to.equal('external')
+  })
+
+  it('`isExternalModule` works with windows directory separator', function() {
+    expect(isExternalModule('foo', {}, 'E:\\path\\to\\node_modules\\foo')).to.equal(true)
+    expect(isExternalModule('foo', {
+      'import/external-module-folders': ['E:\\path\\to\\node_modules'],
+    }, 'E:\\path\\to\\node_modules\\foo')).to.equal(true)
   })
 })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Fixed
+- [`order`]: fix `isExternalModule` detect on windows ([#1651], thanks [@fisker])
 
 ## [2.20.1] - 2020-02-01
 ### Fixed
@@ -652,6 +654,7 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#1651]: https://github.com/benmosher/eslint-plugin-import/pull/1651
 [#1635]: https://github.com/benmosher/eslint-plugin-import/issues/1635
 [#1625]: https://github.com/benmosher/eslint-plugin-import/pull/1625
 [#1620]: https://github.com/benmosher/eslint-plugin-import/pull/1620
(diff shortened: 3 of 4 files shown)


src/core/importType.js
```diff
@@ -29,10 +29,11 @@ function isExternalPath(path, name, settings) {
 }
 
 function isSubpath(subpath, path) {
-  const normSubpath = subpath.replace(/[/]$/, '')
+  const normSubpath = subpath.replace(/\/$/, '')
   if (normSubpath.length === 0) {
     return false
   }
+  path = path.replace(/\\/g, '/')
```
ljharb: perhaps instead of handling both cases, we should use `path.sep`, so that it handles file paths according to the filesystem you're on?

ljharb:
(windows tests are broken pending #1648)

src/core/importType.js
fisker: I'm not sure that if `path.sep` works, but I guess no harm we just simply normalize the path?

ljharb: in that case, would node's `path.normalize` work?

fisker: I don't think `path.normalize` work

```text
require('path').normalize('E:/path\\to/')
'E:\\path\\to\\'
```

ljharb:
Rebased this on top of #1648; still 1 test failing.

fisker:
Should be good now.

fisker:
@ljharb I fixed the failed test. But seems still breaking on Node.js@10

raphinesse:
@fisker The failures on Node 10 are due to the same problem with `nyc` that we did work around for in #1648 for Node 12 only.

fisker:
@raphinesse 

The error info shows require error

If we use that code for node.js 10, will it work?

raphinesse:
@fisker I'm trying right now: https://github.com/benmosher/eslint-plugin-import/pull/1648/commits/9e8960581f302c78465bb2ba285d957be6a1043a

fisker:
@raphinesse If that's a bug from `nyc`, why not disable coverage on windows? Should be simplier

raphinesse:
> @raphinesse If that's a bug from `nyc`, why not disable coverage on windows? Should be simplier

@fisker I suggested that in https://github.com/benmosher/eslint-plugin-import/pull/1648#discussion_r376596783 but that comment got no reply from @ljharb. So I cannot tell you why.

fisker:
@raphinesse You're right, problem solved. Thanks

raphinesse:
@fisker The problem remains on Node 8. I've updated #1648 to use `npm@6.10.3` for all Node versions. We'll see if that works for Node 8 too.

ljharb:
I have no interest in disabling coverage anywhere; anything that disables coverage on windows is a nonstarter.

fisker:
@raphinesse Have you try `c8` on windows? Maybe that works.

raphinesse:
@fisker Sorry, I don't have time to do so right now. Plus I fear it might not support enough Node/npm versions. That's just speculation though.

fisker:
@raphinesse @ljharb Seems windows tests fixed now.

tests/src/core/importType.js
```diff
@@ -224,4 +224,11 @@ describe('importType(name)', function () {
     })
     expect(importType('@test-scope/some-module', foldersContext)).to.equal('external')
   })
+
+  it('`isExternalModule` works with windows directory separator', function() {
+    expect(isExternalModule('foo', {}, 'E:\\path\\to\\node_modules\\foo')).to.equal(true)
+    expect(isExternalModule('foo', {
+      'import/external-module-folders': ['E:\\path\\to\\node_modules'],
```
ljharb: it doesn't seem like we're testing `external-module-folders` here if it was true before? meaning, maybe we need to add a test that returns `false` without `external-module-folders`, and `true` with it?

fisker: No, without `external-module-folders` still true, because `subpath` defualts to `node_modules`, the test make sure we need replace `subpath`, remove that line will cause error

ljharb: Sorry, to clarify; I'm glad these tests fail without your fix! However, don't we want to also test that a windows path would *not* be external without the right external-module-folders setting?

tests/src/core/importType.js
```diff
@@ -227,8 +227,13 @@ describe('importType(name)', function () {
...
   it('`isExternalModule` works with windows directory separator', function() {
     expect(isExternalModule('foo', {}, 'E:\\path\\to\\node_modules\\foo')).to.equal(true)
+    expect(isExternalModule('foo', {}, '.\\foo')).to.equal(false)
+
     expect(isExternalModule('foo', {
       'import/external-module-folders': ['E:\\path\\to\\node_modules'],
     }, 'E:\\path\\to\\node_modules\\foo')).to.equal(true)
+    expect(isExternalModule('foo', {
+      'import/external-module-folders': ['E:\\path\\to\\node_modules'],
+    }, '.\\foo')).to.equal(false)
```
ljharb: i'm hoping for "false before, true after", not "same before and after". Can we come up with an example where the external-module-folders setting *changes* the result?

fisker: I'm not sure I understand this.

Setting `import/external-module-folders` did not change the result.

```js
    expect(isExternalModule('foo', {
      'import/external-module-folders': ['E:\\path\\to\\node_modules'],
    }, 'E:\\path\\to\\node_modules\\foo')).to.equal(true)
```

```js
    expect(isExternalModule('foo', {
    }, 'E:\\path\\to\\node_modules\\foo')).to.equal(true)
```

There are all true, and should be true, but before I add `subpath.replace(/\\/g, '/')`, this is false on windows, because it's bug.  How do I test that, if you want reproduction, you just remove this fix, and

```js
    expect(isExternalModule('foo', {
      'import/external-module-folders': ['E:\\path\\to\\node_modules'],
    }, 'E:\\path\\to\\node_modules\\foo')).to.equal(true)
```

this will fail.

ljharb: sure - i appreciate that the bug is that you get `true, false` instead of `true, true`.

However, to prevent different regressions later, i'm hoping to have a test that correctly ensures `false, true`.

ljharb: I went ahead and added something like that.

ljharb:
@fisker i'm sorry to hear this was a bad contribution experience for you; I'd love to understand in what way that was the case.

fisker:
This supposed to be a simple fix, but I have to wait/help fix the failed/unrelated tests. If I remember correctly, the failed tests is about coverage on old version of Node.js(at least not something I care at all), and I think it's already failed before I send this. I agree coverage is important, but I don't think it something that important to block. If I'm the maintainer, I'd merge first and try to fix later, especially #1648 already working on it at that time.

I was patient only because I need this fix get released.

It's good to see this plugin still support very old version of ESLint and Node.js, I like write code can run anywhere.
For example, I always use `document.{getElements,getElementById}`  instead of `document.querySelector` even my target browsers support it, I do this because it's easy to get the same result in one line, but I don't write 10 lines of code when I can do it in one line, I write tools to do the job.

Anyway the contribute experience is really bad to make tests pass on old stuff, because the contributors have to deal with things that they don't care at all.

I stopped look into this repository because I'm an ESLint rule author of many rules, I can simply give up and make my own one. I remember I wrote first rule for prettier internal also because bug in this plugin, and I didn't try to fix it at all..

Forgive me to bring this up, no hard feelings, this is just my feeling.

ljharb:
@fisker not at all; i appreciate your willingness to give the feedback even if you've already decided to not contribute in the future. I'll see what I can do in the future to alleviate the "tests for old versions" work for contributors.

(the extensions rule should work with `require`, fwiw; the config may need `commonjs: true` though, but if not it's a bug i'd want to fix)

fisker:
I don't remember the details, but I think I speed time to figure but didn't work, so I have to write one. I didn't try because we don't have one at first, now we have [a lot](https://github.com/prettier/prettier/tree/main/scripts/tools/eslint-plugin-prettier-internal-rules), it's much easier than debug existing rules..

fisker:
Reminds me that we don't need [import/extensions](https://github.com/prettier/prettier/pull/15961) anymore. 😄

PR #1657, opened 2020-02-15
[importType] Fix core modules can be typed as `internal` when using webpack resolver.

keik:
with .eslintrc as below

```js
{
  ...,
  settings: { 'import/resolver': 'webpack' },
  ...,
  rules: {
    'import/order': [
      'error',
      {
        'newlines-between': 'always'
      }
    ],
  },
  ...
}
```

then occur the error as below

```js
import fs from 'fs'; // <- error  There should be at least one empty line between import groups  import/order
import path from 'path';
```
I think the error is false positive because of both `fs` and `path` are `builtin` modules.

But actually `path` is typed as `external` module because it seems that webpack resolve some of core modules to [node-libs-browser](https://github.com/webpack/node-libs-browser). Then resolved path end up being found and the module is not typed as `builtin`. ([code](https://github.com/benmosher/eslint-plugin-import/blob/v2.20.1/src/core/importType.js#L18-L20)). 

I found that 
* node resolver return path of core modules as `null`.
  https://github.com/benmosher/eslint-plugin-import/blob/v2.20.1/resolvers/node/index.js#L12-L15
* resolved path is referred whether `undefined` or not, but not `null`. 
https://github.com/benmosher/eslint-plugin-import/blob/v2.20.1/src/rules/no-unresolved.js#L31-L34

My modify is aligned to above design.

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -133,7 +133,7 @@ exports.resolve = function (source, file, settings) {
     return { found: true, path: resolveSync(path.dirname(file), source) }
   } catch (err) {
     if (source in coreLibs) {
-      return { found: true, path: coreLibs[source] }
+      return { found: true, path: null }
     }
 
     log('Error during module resolution:', err)
--- a/tests/src/core/importType.js
+++ b/tests/src/core/importType.js
@@ -20,6 +20,12 @@ describe('importType(name)', function () {
     expect(importType('path', context)).to.equal('builtin')
   })
 
+  it("should return 'builtin' for node.js modules (webpack resolver)", function() {
+    const webpackContext = testContext({ 'import/resolver': 'webpack' })
+    expect(importType('fs', webpackContext)).to.equal('builtin')
+    expect(importType('path', webpackContext)).to.equal('builtin')
+  })
+
   it("should return 'external' for non-builtin modules without a relative path", function() {
     expect(importType('lodash', context)).to.equal('external')
     expect(importType('async', context)).to.equal('external')


ljharb:
You need to specify the node resolver as well, especially since there's no `fs` in the browser.

ljharb:
(Specifically, it makes no sense to run webpack on code that relies on `fs`, since it can't work in the browser)

keik:
Oh sorry, my example was bad.

The subject is that `path` typed as `builtin`, not `fs`. 

Other example:

with .eslintrc
```js
{
  ...,
  settings: { 'import/resolver': 'webpack' },
  ...,
  rules: {
    'import/order': [
      'error',
      {
        'newlines-between': 'always'
     {
        alphabetize: { order: 'asc' },
        'newlines-between': 'always'
      }
    ],
  },
  ...
}
```

then occur the error as below

```
import path from 'path';  // <- error  There should be no empty line within import group        import/order

import _ from 'lodash-es'; // <- error  `lodash-es` import should occur before import of `path`  import/order
```

to pass errors, change to below

```
import _ from 'lodash-es';
import path from 'path'; // path is `builtin` module in my mental, but the rule force me to place here.
```

ljharb:
@keik thanks, that makes sense. What happens if you add `node: {}` to your `import/resolver` settings object (alongside `webpack: {}`)?

keik:
with eslintrc
```js
'import/resolver': ['node', 'webpack']
// same as 'import/resolver': { node: {}, webpack: {} },
```
then
```js
import path from 'path';

import _ from 'lodash-es';
```
has no errors.

---

with eslintrc
```js
'import/resolver': ['webpack', 'node'],
// same as 'import/resolver': { webpack: {}, node: {} },
```
then
```js
import path from 'path'; // <- error  There should be no empty line within import group        import/order

import _ from 'lodash-es'; // <- error  `lodash-es` import should occur before import of `path`  import/order
```
has errors like above.

---

These results indicate that it seems that `import/resolver` only affect first resolver.



### Other example

against the code
```js
// redux not export `a`.
// redux package.json has no `jsnext:main` field.
// redux package.json has `module` field.
import { a } from 'redux';
```

So `import/named` violation should be occur with `webpack` resolver, but results are

| resolver | result |
|---|---|
| `'node'` | no error (expected) |
| `'webpack'` | error (expected) |
| `['node', 'webpack']` | no error (???) |
| `['webpack', 'node']` | error (expected) |

ljharb:
Not that it only is affected by the first one; it's that the first one to resolve wins, and for node core modules, you want node, not webpack, to be the first to resolve.

In other words, all 4 results of your table are expected to me. There's not really a reason to have the webpack resolver ordered before the node resolver.

keik:
I see, sorry my misunderstanding!

So the point is that some core modeles can be typed as ´external´ when using webpack resolver as first resolver.

ljharb:
Yes, because webpack maps some node core modules to things in node_modules.

keik:
Yes, `node-libs-browser` is that!

Go back to the beginning, this is problem or not?

In my mental, `path` is builtin module even if using webpack.

I can set config `import/resolver: ['node', 'webpack']` to avoid `import/order` error, but then other problem that `import/named` does not respect `module` field in package.json is occured.

I want to resolve both of them.
So my propose is changing webpack resolver to treat core modules in `node-libs-browser` as `builtin` instead of `external`.

Of course I can resolve by configuration like below, but not like out-of-the-box.

```
settings: {
  'import/core-modules': ['path'],
  'import/resolver': 'webpack',
},
```

ljharb:
I would suggest not relying on the "module" field, and fixing that by getting that package to use jsnext:main, or to use the official node solution and use "exports" with an `.mjs` file.

ljharb:
I'm reasonably sure this is correct behavior, per the above discussion. Please close it if you agree; if it's still relevant, please rebase and mark it as ready for review.

keik:
OK. Thank you for discussion!

PR #1658, opened 2020-02-17
[fix] `order`: recognize ".." as a "parent" path

### LINKED ISSUE #1405: order: '..' is not recognized
https://github.com/benmosher/eslint-plugin-import/pull/1375 added support for "unknown" imports. Imports from `..` are classified as *unknown*. I think it ought to be *parent*.

```js
import foo from '../foo'
import { bar } from '..'
import baz from '.'
```

golopot:
fixes #1405

### DIFF
--- a/tests/src/rules/order.js
+++ b/tests/src/rules/order.js
@@ -19,6 +19,7 @@ ruleTester.run('order', rule, {
         var relParent1 = require('../foo');
         var relParent2 = require('../foo/bar');
         var relParent3 = require('../');
+        var relParent4 = require('..');
         var sibling = require('./foo');
         var index = require('./');`,
       }),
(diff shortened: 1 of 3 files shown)


tests/src/rules/order.js
```diff
@@ -3,7 +3,7 @@ import { test, getTSParsers } from '../utils'
 import { RuleTester } from 'eslint'
 
 const ruleTester = new RuleTester()
-    , rule = require('rules/order')
+    , rule = require('../../../src/rules/order')
```
ljharb: not sure why this is here; `npm run mocha path/to/file` will set up the NODE_PATH for you to point to `src`.
```suggestion
    , rule = require('rules/order')
```

golopot: This is how I run tests:
```sh
npx mocha --compilers js:babel-register tests/src/rules/order.js
```
Without that change the `require` fails.

golopot:
> Per #1405 (comment), should this go through the resolver to consider "parent"-ness?

I believe it is better to order the import statements by the apparent path rather the resolved path. What I wanted for this rule is a consistent ordering that looks nice. Ordering by resolved path does not provide a consistent ordering that looks nice.

tests/src/rules/order.js
ljharb: Right, that's not a valid way to run tests in this repo - you need to use `npm run mocha path/to/file` instead.

ljharb:
True, that makes sense.

PR #1660, opened 2020-02-17
[New] `no-unused-modules`: support dynamic imports 

### LINKED ISSUE #1951: [no-unused-modules] Dynamic import doesn't mark exports as used
Repro:

```ts
// foo.ts
export const foo = "foo"

// index.ts
async function test() {
  const { foo } = await import("./foo")
  const foo2 = await import("./foo").then(mod => mod.foo)
  const mod = await import("./foo") // should mark all non-type exports as used
}
```

My specific use case is:

```tsx
import { lazy } from "react"
const Foo = lazy(() => import("./Foo"))
```

> exported declaration 'foo' not used within other modules

Config:

```json5
{
  "extends": [
    "plugin:import/recommended",
    "plugin:import/typescript",
    // and more ...
  ],
  "rules": {
    "import/no-unused-modules": ["error", { "unusedExports": true }],
  },
}
```

Versions:

```
@typescript-eslint/parser: 4.8.0
eslint: 7.13.0
eslint-plugin-import: 2.22.1
```

maxkomarychev:
All occurences of `import('...')` are treated as namespace imports
(`import * as X from '...'`)

Fixes #1951.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -7,6 +7,7 @@ import debug from 'debug';
 import { SourceCode } from 'eslint';
 
 import parse from 'eslint-module-utils/parse';
+import visit from 'eslint-module-utils/visit';
 import resolve from 'eslint-module-utils/resolve';
 import isIgnored, { hasValidExtension } from 'eslint-module-utils/ignore';
 
@@ -354,15 +355,57 @@ ExportMap.parse = function (path, content, context) {
   const isEsModuleInteropTrue = isEsModuleInterop();
 
   let ast;
+  let visitorKeys;
   try {
-    ast = parse(path, content, context);
+    const result = parse(path, content, context);
+    ast = result.ast;
+    visitorKeys = result.visitorKeys;
   } catch (err) {
-    log('parse error:', path, err);
     m.errors.push(err);
     return m; // can't continue
   }
 
-  if (!unambiguous.isModule(ast)) return null;
+  m.visitorKeys = visitorKeys;
+
+  let hasDynamicImports = false;
+
+  function processDynamicImport(source) {
+    hasDynamicImports = true;
+    if (source.type !== 'Literal') {
+      return null;
+    }
+    const p = remotePath(source.value);
+    if (p == null) {
+      return null;
+    }
+    const importedSpecifiers = new Set();
+    importedSpecifiers.add('ImportNamespaceSpecifier');
+    const getter = thunkFor(p, context);
+    m.imports.set(p, {
+      getter,
+      declarations: new Set([{
+        source: {
+        // capturing actual node reference holds full AST in memory!
+          value: source.value,
+          loc: source.loc,
+        },
+        importedSpecifiers,
+      }]),
+    });
+  }
+
+  visit(ast, visitorKeys, {
+    ImportExpression(node) {
+      processDynamicImport(node.source);
+    },
+    CallExpression(node) {
+      if (node.callee.type === 'Import') {
+        processDynamicImport(node.arguments[0]);
+      }
+    },
+  });
+
+  if (!unambiguous.isModule(ast) && !hasDynamicImports) return null;
 
   const docstyle = (context.settings && context.settings['import/docstyle']) || ['jsdoc'];
   const docStyleParsers = {};
--- a/tests/src/rules/no-unused-modules.js
+++ b/tests/src/rules/no-unused-modules.js
@@ -112,41 +112,49 @@ ruleTester.run('no-unused-modules', rule, {
       options: unusedExportsOptions,
       code: 'import { o2 } from "./file-o";export default () => 12',
       filename: testFilePath('./no-unused-modules/file-a.js'),
+      parser: require.resolve('babel-eslint'),
     }),
     test({
       options: unusedExportsOptions,
       code: 'export const b = 2',
       filename: testFilePath('./no-unused-modules/file-b.js'),
+      parser: require.resolve('babel-eslint'),
     }),
     test({
       options: unusedExportsOptions,
       code: 'const c1 = 3; function c2() { return 3 }; export { c1, c2 }',
       filename: testFilePath('./no-unused-modules/file-c.js'),
+      parser: require.resolve('babel-eslint'),
     }),
     test({
       options: unusedExportsOptions,
       code: 'export function d() { return 4 }',
       filename: testFilePath('./no-unused-modules/file-d.js'),
+      parser: require.resolve('babel-eslint'),
     }),
     test({
       options: unusedExportsOptions,
       code: 'export class q { q0() {} }',
       filename: testFilePath('./no-unused-modules/file-q.js'),
+      parser: require.resolve('babel-eslint'),
     }),
     test({
       options: unusedExportsOptions,
       code: 'const e0 = 5; export { e0 as e }',
       filename: testFilePath('./no-unused-modules/file-e.js'),
+      parser: require.resolve('babel-eslint'),
     }),
     test({
       options: unusedExportsOptions,
       code: 'const l0 = 5; const l = 10; export { l0 as l1, l }; export default () => {}',
       filename: testFilePath('./no-unused-modules/file-l.js'),
+      parser: require.resolve('babel-eslint'),
     }),
     test({
       options: unusedExportsOptions,
       code: 'const o0 = 0; const o1 = 1; export { o0, o1 as o2 }; export default () => {}',
       filename: testFilePath('./no-unused-modules/file-o.js'),
+      parser: require.resolve('babel-eslint'),
     }),
   ],
   invalid: [
@@ -234,7 +242,72 @@ ruleTester.run('no-unused-modules', rule, {
   ],
 });
 
-// test for export from
+
+describe('dynamic imports', () => {
+  if (semver.satisfies(eslintPkg.version, '< 6')) {
+    beforeEach(function () {
+      this.skip();
+    });
+    return;
+  }
+
+  // test for unused exports with `import()`
+  ruleTester.run('no-unused-modules', rule, {
+    valid: [
+      test({
+        options: unusedExportsOptions,
+        code: `
+            export const a = 10
+            export const b = 20
+            export const c = 30
+            const d = 40
+            export default d
+            `,
+        parser: require.resolve('babel-eslint'),
+        filename: testFilePath('./no-unused-modules/exports-for-dynamic-js.js'),
+      }),
+    ],
+    invalid: [
+      test({
+        options: unusedExportsOptions,
+        code: `
+        export const a = 10
+        export const b = 20
+        export const c = 30
+        const d = 40
+        export default d
+        `,
+        parser: require.resolve('babel-eslint'),
+        filename: testFilePath('./no-unused-modules/exports-for-dynamic-js-2.js'),
+        errors: [
+          error(`exported declaration 'a' not used within other modules`),
+          error(`exported declaration 'b' not used within other modules`),
+          error(`exported declaration 'c' not used within other modules`),
+          error(`exported declaration 'default' not used within other modules`),
+        ] }),
+    ],
+  });
+  typescriptRuleTester.run('no-unused-modules', rule, {
+    valid: [
+      test({
+        options: unusedExportsTypescriptOptions,
+        code: `
+            export const ts_a = 10
+            export const ts_b = 20
+            export const ts_c = 30
+            const ts_d = 40
+            export default ts_d
+            `,
+        parser: require.resolve('@typescript-eslint/parser'),
+        filename: testFilePath('./no-unused-modules/typescript/exports-for-dynamic-ts.ts'),
+      }),
+    ],
+    invalid: [
+    ],
+  });
+});
+
+// // test for export from
 ruleTester.run('no-unused-modules', rule, {
   valid: [
     test({
--- a/utils/parse.js
+++ b/utils/parse.js
@@ -3,8 +3,41 @@ exports.__esModule = true;
 
 const moduleRequire = require('./module-require').default;
 const extname = require('path').extname;
+const fs = require('fs');
 
 const log = require('debug')('eslint-plugin-import:parse');
+
+function getBabelVisitorKeys(parserPath) {
+  if (parserPath.endsWith('index.js')) {
+    const hypotheticalLocation = parserPath.replace('index.js', 'visitor-keys.js');
+    if (fs.existsSync(hypotheticalLocation)) {
+      const keys = moduleRequire(hypotheticalLocation);
+      return keys.default || keys;
+    }
+  } else if (parserPath.endsWith('index.cjs')) {
+    const hypotheticalLocation = parserPath.replace('index.cjs', 'worker/ast-info.cjs');
+    if (fs.existsSync(hypotheticalLocation)) {
+      const astInfo = moduleRequire(hypotheticalLocation);
+      return astInfo.getVisitorKeys();
+    }
+  }
+  return null;
+}
+
+function keysFromParser(parserPath, parserInstance, parsedResult) {
+  if (/.*espree.*/.test(parserPath)) {
+    return parserInstance.VisitorKeys;
+  }
+  if (/.*(babel-eslint|@babel\/eslint-parser).*/.test(parserPath)) {
+    return getBabelVisitorKeys(parserPath);
+  }
+  if (/.*@typescript-eslint\/parser/.test(parserPath)) {
+    if (parsedResult) {
+      return parsedResult.visitorKeys;
+    }
+  }
+  return null;
+}
 
 exports.default = function parse(path, content, context) {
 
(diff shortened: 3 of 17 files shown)


ljharb:
`const a = await import('a')` is not equivalent to `import a from 'a'`, it's equivalent to getting a promise for `import * as mod from 'a'`.

`await` is largely irrelevant here, i think.

Because `import()` produces a promise, I don't think there's any value in checking that specific imports/exports are used or not, because it's not very reliably to detect that.

maxkomarychev:
hey, thanks for feedback!

> const a = await import('a') is not equivalent to import a from 'a', it's equivalent to getting a promise for import * as mod from 'a'. You might be thinking of const { default: a } = await import('a')

good point, you're right - I should be treating this as namespace :)

> await is largely irrelevant here, i think.
> Because import() produces a promise, I don't think there's any value in checking that specific imports/exports are used or not, because it's not very reliably to detect that.

Not sure I entirely follow here. I was not going to check if any specific name imported from the module has ever been referenced (there are other eslint rules for this matter) but when import is assigned to destructured object I want to mark specific names that were imported, isn't this good for this rule?

ljharb:
Right, that's what I'm saying - it's not going to be safe in enough cases to mark specific names, so any path that's dynamically imported should be treated as potentially using every export.

maxkomarychev:
> Right, that's what I'm saying - it's not going to be safe in enough cases to mark specific names, so any path that's dynamically imported should be treated as potentially using every export.

Even though handling all kinds of expressions would be very complex I believe is possible to handle some finite amount of patterns and it would increase accuracy of the lib, i.e.

|import base|with await|with then|
|-|-|-|
|`import * as a from 'a'` | `const a = await import('a')` | `import('a').then(a =>{})` or `import('a').then(function(a){})`
|`import { a } from 'a' `|`const { a } = await import('a')` | `import('a').then(({a}) =>{})` or `import('a').then(function(({a}){})`

all other calls to `import('a')` will simply be treated as namespace import

is that something you would consider to review?

ljharb:
In your first row, the `a` could be passed elsewhere. I'm not sure how often the second one happens to be useful.

`import()` is also often used behind an abstraction, or in `Promise.all` like `Promise.all([import('a'), import('b')])`, which is then either awaited or `.then`d. Do you propose attempting to handle those cases as well?

maxkomarychev:
alright, this is too much :) I'll treat `import('a')` as namespace no matter what

tests/src/rules/no-unused-modules.js
```diff
@@ -163,6 +173,40 @@ ruleTester.run('no-unused-modules', rule, {
...
+           const d = 40
+           export default d
+           `,
+           parser: require.resolve('babel-eslint'),
+           filename: testFilePath('./no-unused-modules/exports-for-dynamic-js.js')}),
+  ],
+  invalid: [
+  ],
+})
+typescriptRuleTester.run('no-unused-modules', rule, {
```
maxkomarychev: this test always passes however corresponding ts file is empty, what am I doing wrong?

maxkomarychev:
@ljharb I almost cleaned it up although I am stuck at typescript files - I put a dedicated test with typescript parser and it is always green - how do I configure it to fail?

tests/src/rules/no-unused-modules.js
ljharb: i believe the `filename` doesn't have to exist, it's just to set the context for rules

maxkomarychev: hmm. how does it check if specific file is imported? I thought filename here is provided specifically to check that this code belongs to specificed file and then track imports of such file. 

Which in case of ts does not work and tests always green.

maxkomarychev:
ok I believe I figured this out, although one last bit that puzzles me - dynamic imports are supported by espree since 6.1.0 (https://github.com/eslint/espree/commit/f5e58cc5e9030793baca3426366b8d7286ef5b89) but tests fail even if `ecmaVersion` is set to 11. so I can only got it working with `babel-eslint` or `@typescript-eslint/parser`. any suggestion how to make espree to handle this? or would it instead be acceptable to just specify this in docs for specific rule to use custom parser?

ljharb:
It wasn't stage 4 until ES2020, so eslint wouldn't support it until that ecmaVersion.

If it's still not working with the right ecmaVersion, it might not work in eslint until v7; for now, it's fine just to support those two parsers.

maxkomarychev:
should I modify readme with this?
otherwise - ready for review 🎆

maxkomarychev:
it fails tests on eslint 2 and 3, what's the best way to deal with these?

maxkomarychev:
bump?

src/ExportMap.js
```diff
@@ -344,14 +345,43 @@ ExportMap.parse = function (path, content, context) {
...
-    log('parse error:', path, err)
     m.errors.push(err)
     return m // can't continue
   }
 
-  if (!unambiguous.isModule(ast)) return null
+  let hasDynamicImports = false
+
+  visit(ast, visitorKeys, {
+    CallExpression(node) {
```
Hypnosphi, line 357: It's `ImportExpression` in the latest `@babel/parser` which is a dependency of `babel-eslint`
https://github.com/babel/babel/pull/10828

src/ExportMap.js
```diff
@@ -344,14 +345,43 @@ ExportMap.parse = function (path, content, context) {
...
+      if (node.callee.type === 'Import') {
+        hasDynamicImports = true
+        const firstArgument = node.arguments[0]
```
Hypnosphi, line 360:
```suggestion
    ImportExpression(node) {
      hasDynamicImports = true
      const source = node.source
        
      // replace `firstArgument` with `source` below
```

ljharb: We'd need to keep both; `ImportExpression` is the newer form, but older parsers still have a `CallExpression` of callee type Import.

Hypnosphi:
@maxkomarychev looks like you also have to update https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-unused-modules.js#L610

maxkomarychev:
@Hypnosphi can you elaborate please? this pr puts dynamic imports with as they are namespace imports, given that all other code is using `ExportMap` then it should be transparent to them with no special handling, unless I'm missing something. 

Thanks.

Hypnosphi:
`updateImportUsage` doesn't use `ExportMap`, it does its own iteration instead

ljharb:
@maxkomarychev there's still tests failing, the PR needs a rebase, plus https://github.com/benmosher/eslint-plugin-import/pull/1660#issuecomment-631161376. are you planning to update the PR?

maxkomarychev:
@ljharb yes. although I have a question: I see tests have been failing in lower versions of eslint, do lower versions of eslint support newer syntaxes or maybe you can point to a place where I can see how old versions should be handled in rules? thanks.

ljharb:
It's entirely likely that `import()` won't work in eslint < 6; you can use an approach like this: https://github.com/benmosher/eslint-plugin-import/blob/0547c7efa0a1becd2cdf35ae03f6fe74472398d9/tests/src/cli.js#L32

utils/visit.js
```diff
@@ -0,0 +1,29 @@
...
+  for (const fieldName of childFields) {
+    const field = node[fieldName]
+    if (Array.isArray(field)) {
+      for (const item of field) {
+        visit(item, keys, visitorSpec)
+      }
+    } else {
+      visit(field, keys, visitorSpec)
+    }
+  }
```
ljharb, line 25:
```suggestion
  childFields.forEach((fieldName) => {
    [].concat(node[fieldName]).forEach((item) => {
      visit(item, keys, visitorSpec)
    })
  })
```

utils/visit.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
+exports.__esModule = true
```
ljharb, line 2:
```suggestion
```
this should only ever be added by babel.

utils/visit.js
```diff
@@ -0,0 +1,29 @@
...
+
+exports.default = function visit(node, keys, visitorSpec) {
```
ljharb, line 4:
```suggestion
module.exports = function visit(node, keys, visitorSpec) {
```

utils/parse.js
```diff
@@ -55,16 +84,25 @@ exports.default = function parse(path, content, context) {
       console.warn(
         '`parseForESLint` from parser `' +
           parserPath +
-          '` is invalid and will just be ignored'
+          '` is invalid and will just be ignored',
+        path
```
ljharb, line 88:
```suggestion
          '` is invalid and will just be ignored'
```
if the path is needed, let's put it in the string

utils/parse.js
```diff
@@ -3,9 +3,33 @@ exports.__esModule = true
...
 const log = require('debug')('eslint-plugin-import:parse')
 
+function getBabelVisitorKeys(parserPath) {
+  const hypotheticalLocation = parserPath.replace('index.js', 'visitor-keys.js')
+  if (fs.existsSync(hypotheticalLocation)) {
+    const keys = moduleRequire(parserPath.replace('index.js', 'visitor-keys.js'))
+    return keys
+  } else {
+    return null
+  }
```
ljharb, line 17:
```suggestion
  }
  return null
```

utils/parse.js
```diff
@@ -3,9 +3,33 @@ exports.__esModule = true
...
 const moduleRequire = require('./module-require').default
 const extname = require('path').extname
+const fs = require('fs')
 
 const log = require('debug')('eslint-plugin-import:parse')
 
+function getBabelVisitorKeys(parserPath) {
+  const hypotheticalLocation = parserPath.replace('index.js', 'visitor-keys.js')
+  if (fs.existsSync(hypotheticalLocation)) {
+    const keys = moduleRequire(parserPath.replace('index.js', 'visitor-keys.js'))
```
ljharb, line 13:
```suggestion
    const keys = moduleRequire(hypotheticalLocation)
```

utils/parse.js
```diff
@@ -3,9 +3,33 @@ exports.__esModule = true
...
+    return null
+  }
+}
+
+function keysFromParser(parserPath, parserInstance, parsedResult) {
+  if (/.*espree.*/.test(parserPath)) {
+    return parserInstance.VisitorKeys
+  } else if (/.*babel-eslint.*/.test(parserPath)) {
+    return getBabelVisitorKeys(parserPath)
+  } else if (/.*@typescript-eslint\/parser/.test(parserPath)) {
```
ljharb, line 25:
```suggestion
  }
  if (/.*babel-eslint.*/.test(parserPath)) {
    return getBabelVisitorKeys(parserPath)
  }
  if (/.*@typescript-eslint\/parser/.test(parserPath)) {
```

ljharb:
@maxkomarychev we need to keep both node names - we have to retain compatibility with old version. See https://github.com/benmosher/eslint-plugin-import/pull/1660#discussion_r427657128

Hypnosphi:
@maxkomarychev do you need any help with that? If you're not interested in it anymore, I can open a new PR

ljharb:
@Hypnosphi please don't open a new PR to this repo - either make a PR to the OP's repo (and ping me in it) or comment here with a link to a branch, and i'll update this PR. Thanks!

maxkomarychev:
Hey, I don't have capacity to push this to completion in near future.
I'm sorry.

@Hypnosphi feel free to submit your own version.

Hypnosphi:
@ljharb https://github.com/Hypnosphi/eslint-plugin-import/tree/no-unused-modules-dynamic-imports

aladdin-add:
any news?

ljharb:
@Hypnosphi unfortunately due to the intervening time, there are now merge conflicts between your branch and master. Any chance you could rebase your branch?

Hypnosphi:
I will, but I have now a concern about this implementation though. We heavily [rely](https://github.com/benmosher/eslint-plugin-import/blob/c10e2172c533fc29f7404e6cd4cc628ac3df5f8e/utils/parse.js#L10-L17) on internal file structure of parser packages which can change in minor releases. It actually changed in some of the `@babel/eslint-parser` releases. I figured out a [workaround](https://github.com/Hypnosphi/eslint-plugin-import/commit/513960d9f559fc98bc15a8215e4a63814e6f63d7), but maybe there's a more generic way to find all dynamic imports?

Hypnosphi:
Done

ljharb:
@Hypnosphi arguably anything exposed is part of the api, and can't be moved within a major without violating semver. If the parser refuses to comply with that, then the only other alternative is to convince them to provide the API we need as something that's officially supported.

aladdin-add:
generally, it's not encouraged to do so. But for `babel-eslint`, I think it's fine to do so (`babel-eslint` is no longer maintained, in favor of the new `@babel/eslint-parser`).

ljharb:
lol properties with "_" in front are not encouraged to be relied upon either, but they're public and part of the API anyways ¯\\\_(ツ)\_/¯ reachable is public.

i agree that for babel-eslint the risk is effectively zero; i'm comfortable relying on that.

aladdin-add:
@Hypnosphi Did you forget to push the latest changes? it does not work well on my computer.

1. lint error, I made a PR: https://github.com/Hypnosphi/eslint-plugin-import/pull/1
2. after fixed 1, I got another error when run `npm test`:
```bash
TypeError: Error while loading rule 'no-unused-modules': Cannot read property 'forEach' of undefined
Occurred while linting /Users/weiran/repo/dian/eslint-plugin-import/tests/files/no-unused-modules/typescript/file-ts-e-unused.ts
      at src/rules/no-unused-modules.js:240:28
      at Map.forEach (<anonymous>)
      at src/rules/no-unused-modules.js:235:23
      at Set.forEach (<anonymous>)
      at prepareImportsAndExports (src/rules/no-unused-modules.js:184:12)
      at doPreparation (src/rules/no-unused-modules.js:333:3)
      at create (src/rules/no-unused-modules.js:473:7)
      at Object.create (node_modules/eslint/lib/rule-tester/rule-tester.js:455:73)
      at createRuleListeners (node_modules/eslint/lib/linter/linter.js:761:21)
      at node_modules/eslint/lib/linter/linter.js:931:31
      at runRules (node_modules/eslint/lib/linter/linter.js:876:34)
      at Linter._verifyWithoutProcessors (node_modules/eslint/lib/linter/linter.js:1175:31)
      at Linter.verify (node_modules/eslint/lib/linter/linter.js:1238:21)
      at runRuleForItem (node_modules/eslint/lib/rule-tester/rule-tester.js:556:37)
      at testInvalidTemplate (node_modules/eslint/lib/rule-tester/rule-tester.js:654:28)
      at Context.<anonymous> (node_modules/eslint/lib/rule-tester/rule-tester.js:893:25)
      at processImmediate (node:internal/timers:464:21)
```

hope it can be landed soon! :)

Hypnosphi:
@aladdin-add good catch, fixed

aladdin-add:
@Hypnosphi I've tested it on a large repo, and got an error:

```bash
➜  honor git:(4d146d797f) ✗ npx eslint /Users/weiran/repo/dian/honor/src/common/components-new/navigator/index.tsx

ESLint: 7.27.0

TypeError: Error while loading rule 'import/no-unused-modules': Cannot read property 'get' of undefined
Occurred while linting /Users/weiran/repo/dian/honor/src/common/components-new/navigator/index.tsx
    at /Users/weiran/repo/dian/honor/node_modules/eslint-plugin-import/lib/rules/no-unused-modules.js:266:44
    at Set.forEach (<anonymous>)
    at /Users/weiran/repo/dian/honor/node_modules/eslint-plugin-import/lib/rules/no-unused-modules.js:264:11
    at Map.forEach (<anonymous>)
    at prepareImportsAndExports (/Users/weiran/repo/dian/honor/node_modules/eslint-plugin-import/lib/rules/no-unused-modules.js:263:13)
    at doPreparation (/Users/weiran/repo/dian/honor/node_modules/eslint-plugin-import/lib/rules/no-unused-modules.js:333:3)
    at Object.create (/Users/weiran/repo/dian/honor/node_modules/eslint-plugin-import/lib/rules/no-unused-modules.js:473:7)
    at createRuleListeners (/Users/weiran/repo/dian/honor/node_modules/eslint/lib/linter/linter.js:761:21)
    at /Users/weiran/repo/dian/honor/node_modules/eslint/lib/linter/linter.js:931:31
```

I am using:
* node: v16.1.0
* eslint: v7.27.0
* @typescript-eslint/parser: v4.24.0

the code is something like:
```tsx
import React, { MouseEvent, ReactNode } from 'react'
import './styles.less'

export interface actionItem {
  icon?: string
  node?: ReactNode
  action?: (e: MouseEvent<any>) => void
}

export interface NavigatorProps {
  user: { icon?: string; title?: string }
  onClickLeft?: (e: MouseEvent<any>) => void
  actions?: actionItem[]
}

const Navigator: React.FC<NavigatorProps> = ({
  user = { icon: HEAD_PORTRAIT, title: '' },
  onClickLeft,
  actions,
}) => {
  return (
    <div className="dian-navigator-wraper">
      <div className="dian-navigator-top" />
      <div className="dian-navigator">
        <div
          className="icon-wrapper"
          onClick={e => onClickLeft && onClickLeft(e)}
        >
          <img className="icon" src={user.icon} />
        </div>
        <div className="icon-wrapper search-wrapper">
          <span className="title">{user.title}</span>
        </div>
        {actions?.map((action: actionItem) => (
          <div className="icon-wrapper" onClick={action.action}>
            {action.icon && <img className="icon small" src={action.icon} />}
            {!action.icon && action.node && action.node}
          </div>
        ))}
      </div>
    </div>
  )
}

export default Navigator

```

Hypnosphi:
@aladdin-add just to be clear, did you apply both `eslint-plugin-import` and `eslint-module-utils` from my branch?

aladdin-add:
yes! used yarn's resolutions.

Hypnosphi:
@aladdin-add can you please either send a PR with a failing test or create a reproduction repo? I can't reproduce the issue with the code you provided

aladdin-add:
sure, will create a repo later. :)

ljharb:
@Hypnosphi any update?

Hypnosphi:
@ljharb I tried adding `export * from '@org/package'` to various test files but still wasn't able to reproduce the issue reported by @aladdin-add

ljharb:
@aladdin-add any chance you could come up with a failing test case?

aladdin-add:
I was not able to repro it either - it happened in a large and private repo. /_ \

it seems a very edge case. we can try to fix it when someone else hit the issue?

ljharb:
Sure! This PR then would need a rebase, and then I can land it.

aladdin-add:
friendly ping @ljharb

ljharb:
I'll take a look at this in the next 24 hours or so.

ljharb:
@Hypnosphi perhaps i messed up the rebase, but a bunch of tests are still failing. could you make sure to fetch the latest branch, and then get the tests passing?

Hypnosphi:
Sorry, it's hard to track what's changed here comparing to my branch. I think I'll open a separate PR

ljharb:
@Hypnosphi i really, really wish you had not done that; PRs once opened are *eternally* present in a repo's refs. Now I just have to keep both of these PRs in sync until they're landed.

ljharb (approved):

Evertt:
Hey, so I'm using this plugin at version 2.26.0, which I believe should include this merged PR already, right? Well I still get a false positive with this.

image

image

So I'm guessing there's still a little bug left in this feature? Can I do anything to help you track down the bug?

aladdin-add:
yes, please file a new issue if you still encounter it.

PR #1663, opened 2020-02-18
`resolvers/node`: [New] add `.node` extension

### LINKED ISSUE #802: Fail to resolve Node module sha3
Hello,

I get the error `Unable to resolve path to module 'sha3'.` when I install and import module sha3 with `import sha3 from 'sha3';.`.
When I run my script with Node (Windows 7) everything is ok. Everything is ok with other modules.

sha3 package don't have an `index.js` and instead has `"main": "build/Release/sha3"` in its `package.json` and a `sha3.node` file in its `build/Release/` subdirectory.

Is it a bug with `eslint-plugin-import-node` ?

Thanks for your help.

ljharb:
Fixes #802

I can't decide if this is semver-minor or breaking for the node resolver - ie, whether it'd be included in v0.3.4 or v0.4.0.

### DIFF
--- a/resolvers/node/CHANGELOG.md
+++ b/resolvers/node/CHANGELOG.md
@@ -4,10 +4,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## Unreleased
+### Added
+- add `.node` extension ([#1663])
 
 ## v0.3.2 - 2018-01-05
 ### Added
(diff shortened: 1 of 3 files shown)


benmosher:
I think it's simpler to call it minor. Gut check on whether the set of users exists that both has .node files and also didn't want to lint them? Feels like it is an empty set to me. So calling it breaking probably hurts more than helps. But go with your instincts.

ljharb:
Appveyor should be working now, so i'll rebase and get it passing before merging. Minor sounds fine to me, thanks.

PR #1664, opened 2020-02-19
Add a way to disable absolute-first option explicitly

### LINKED ISSUE #1661: [import/first] Impossible to turn off absolute-import from dervied config
The base config we use sets
`'import/first': ['error', 'absolute-first'],`
However, while we want the imports to be at the top, we don't need them to be sorted beyond that. As such, we don't need 'absolute-first' set.
Before, we redefined the rule through
`"import/first": ["error", ""],`
However, since 2.20.1 the config only allows 'absolute-first' string in options. Since there is only one option to redefine, and severity changes don't override options, it is impossible to unset 'absolute-first' from derived config right now.

This seems like unintended behavior.
Could you perhaps add an option to explicitly turn off absolute-first?

TheCrueltySage:
Fixes #1661 .
Not sure if the alternative should be just an empty string or an explicit disable option. Went for the second for now.

### DIFF
--- a/tests/src/rules/first.js
+++ b/tests/src/rules/first.js
@@ -11,6 +11,9 @@ ruleTester.run('first', rule, {
                   export { x, y }" })
   , test({ code: "import { x } from 'foo'; import { y } from './bar'" })
   , test({ code: "import { x } from './foo'; import { y } from 'bar'" })
+  , test({ code: "import { x } from './foo'; import { y } from 'bar'"
+         , options: ['disable-absolute-first'],
+         })
   , test({ code: "'use directive';\
                   import { x } from 'foo';" })
   ,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`order`]: recognize ".." as a "parent" path ([#1658], thanks [@golopot])
 - [`no-duplicates`]: fix fixer on cases with default import ([#1666], thanks [@golopot])
 - [`no-unused-modules`]: Handle `export { default } from` syntax ([#1631], thanks [@richardxia])
+- [`first`]: Add a way to disable `absolute-first` explicitly ([#1664], thanks [@TheCrueltySage])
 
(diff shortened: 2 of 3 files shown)


TheCrueltySage:
Any news on this? It's literally a one-liner change.

TheCrueltySage:
I don't have much experience in writing tests.
What do you mean by a test case here? As in, checking that absolute-first doesn't trigger with 'disable-absolute-first' option?

TheCrueltySage:
Something like this?

tests/src/rules/first.js
```diff
@@ -11,6 +11,9 @@ ruleTester.run('first', rule, {
                   export { x, y }" })
   , test({ code: "import { x } from 'foo'; import { y } from './bar'" })
   , test({ code: "import { x } from './foo'; import { y } from 'bar'" })
+  , test({ code: "import { x } from './foo'; import { y } from 'bar'"
```
ljharb: Close! This is the same code as the previous line. the goal is to have a test that *fails* unless `'disable-absolute-first'` is passed, but passes with it.

TheCrueltySage:
Such a test cannot exist.
This option doesn't actually do anything by itself, it's purely structural, intended for reset to default behavior. In particular, if a config you source from sets 'absolute-first', this options allows you to actually unset it. Before, we did it by passing through an empty string to options, but the new consistency checks on this options prevent you from doing that, so I set out to fix that.
I honestly think it's a bug, not an enhancement, since it's the default validation that is overly restrictive, there's no new functionality here.
It could have as well have been an empty string, I just figured explicitly writing you want to turn it off might be nicer to look at.  Perhaps this only confuses people, and I should have gone with an empty string instead.

Bottom line is, 'disable-absolute-first' doesn't have any behavior that differs from how things work by default, it's only intended to reset options if they aren't default already.

ljharb:
Thanks, I think i get it now :-)

TheCrueltySage:
So, should I merge explicit 'disable-absolute-first', or is an empty string option better? I'm kinda in doubt here. Or are both fine?

ljharb:
This one's good; an empty string is super confusing.

PR #1679, opened 2020-03-06
Added allow and forbid options to no-namespace rule

el-angel:
I've added `allow` and `forbid` options, so this rule can be used for certain namespaces, instead for all

### DIFF
--- a/docs/rules/no-namespace.md
+++ b/docs/rules/no-namespace.md
@@ -13,6 +13,12 @@ Valid:
 import defaultExport from './foo'
 import { a, b }  from './bar'
 import defaultExport, { a, b }  from './foobar'
+
+// { allow: ['foo'] }
+import * as foo from 'foo';
+
+// { forbid: ['foo'] }
+import * as bar from 'bar';
 ```
 
 Invalid:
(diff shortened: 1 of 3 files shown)


docs/rules/no-namespace.md
```diff
@@ -25,6 +31,22 @@ import * as foo from 'foo';
...
+
+// { forbid: ['foo'] }
+import * as foo from 'foo';
+```
+
+### Options
+
+This rule takes the following options:
+
+`allow`: an array of namespaces the rule allows
```
ljharb: i assume this should cover *specifiers*, not the binding names?

(having the binding name and the specifier be the same in most of the examples make it unclear to which this applies)

ljharb:
Also, can you elaborate on the use case here?

el-angel:
Use cases would be there you want to force certain namespaces, and forbid others. This rule would make it very easy to force consistency and allow:

`
// { allow: ['constants'] }
import * as constants from '../constants.js;`

while it would disallow:
`import * as C from '../constants.js` 
for example.

Also you could make it possible to only disallow:
`
// { forbid: ['React'] }
import * as React from 'react';`
while allowing all other namespace imports

docs/rules/no-namespace.md
el-angel: the allow option works as follows: 
```
// { allow: ['myModule', 'React']
import * as React from 'react';
import * as myModule from '../module/randomFile.js';
```

ljharb:
I'm not clear tho on why you'd want *any* path to be imported as `constants`?

It seems more like what you want is to point to a specific file, and allow that one to be imported as a namespace with a specific name?

PR #1681, opened 2020-03-11
[New] `no-cycle`: add ignoreExternal option

### LINKED ISSUE #1647: [import/no-cycle] do not check node_modules
I've got projects with only a very few small files, which I have hard time working on because of no-cycle being too slow (99% of the full `eslint` time). I wanted to speed-up this time, but I did not want to use `maxDepth` because I may have some more complex projects and those are the ones where I'd mostly create an unwanted cycle.

While benchmarking the code, I found that there doesn't seem to be a cache, or the cache is not working. For my project, most of my files import the same very big module (`typescript`) and this module is analyzed at each import, instead of only at the first time.

But an easy correction which could also improve the speed would be to ignore external modules: there is very, very little chance that a dependency of my project references a file in my project.

For the moment, this should not be a big problem (except for my big `typescript` dependency), because most dependencies are transpiled into _CommonJS_ and use `require` instead of `import`. But if one day, issue #1056 is corrected, or because external dependencies use more and more `import`, things will become worse.

I made a quick test by directly modifying the code and the time spent by no-cycle fell from more than 2mn to less than 30s.

I can submit a PR if you want. I don't know if it could be considered an issue (the correction would be a breaking change) or if it the behavior should be controlled by an option of no-cycle.

sveyret:
Add an `ignoreExternal` option to `no-cycle` which prevent cycling detection to look into external modules.

Fixes #1647

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Added
 - [`import/default`]: support default export in TSExportAssignment ([#1528], thanks [@joaovieira])
+- [`no-cycle`]: add `ignoreExternal` option ([#1681], thanks [@sveyret])
 
 ### Fixed
 - [`group-exports`]: Flow type export awareness ([#1702], thanks [@ernestostifano])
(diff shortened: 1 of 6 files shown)


sveyret:
CI error is a timeout. Is it possible to re-execute the failing builds?

golopot:
> CI error is a timeout. Is it possible to re-execute the failing builds?

You can do a no-code-change `git commit --amend` and a force push. That will restart CI.

ljharb:
@sveyret since you made this PR from a fork under "not your own username" (ie, slune-org instead of sveyret), I won't be able to force push to the PR branch. Can you give me write permissions on the fork so I can land this PR?

sveyret:
Tell me if it's OK now, @ljharb …

ljharb:
@sveyret hmm, it still doesn't seem like the "allow edits" box is checked - it's in the right side column of the PR.

sveyret:
Strange… I gave you Write role which allow to “read, clone and push” to the repo. Well, I upgraded to Manage role now. Does it work ?

ljharb:
appveyor tests passed, but the coveralls code crashed. merging anyways.

ljharb:
This was merged in 73211e80

PR #1689, opened 2020-03-19
[New] `import/default`: support default export in TSExportAssignment

### LINKED ISSUE #1527: Handle TSExportAssignment exports other than namespaces

Maxim-Mazurok:
Possible duplicate of #1528 (but uses `tsconfig.json`)
Aims to fix #1527

For example, I want to import React as `import React from "react"`. I can do this by using `esModuleInterop: true` in `tsconfig.json`, but eslint doesn't know about that.

This is a draft PR as I'm not really familiar with the codebase. I want to share my approach and receive feedback.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -13,6 +13,12 @@ import isIgnored, { hasValidExtension } from 'eslint-module-utils/ignore'
 import { hashObject } from 'eslint-module-utils/hash'
 import * as unambiguous from 'eslint-module-utils/unambiguous'
 
+import { tsConfigLoader } from 'tsconfig-paths/lib/tsconfig-loader'
+
+import includes from 'array-includes'
+
+import {parseConfigFileTextToJson} from 'typescript'
+
 const log = debug('eslint-plugin-import:ExportMap')
 
 const exportCache = new Map()
@@ -445,8 +451,23 @@ ExportMap.parse = function (path, content, context) {
 
   const source = makeSourceCode(content, ast)
 
-  ast.body.forEach(function (n) {
+  function isEsModuleInterop() {
+    const tsConfigInfo = tsConfigLoader({
+      cwd: context.parserOptions && context.parserOptions.tsconfigRootDir || process.cwd(),
+      getEnv: (key) => process.env[key],
+    })
+    try {
+      if (tsConfigInfo.tsConfigPath !== undefined) {
+        const jsonText = fs.readFileSync(tsConfigInfo.tsConfigPath).toString()
+        const tsConfig = parseConfigFileTextToJson(tsConfigInfo.tsConfigPath, jsonText).config
+        return tsConfig.compilerOptions.esModuleInterop
+      }
+    } catch (e) {
+      return false
+    }
+  }
 
+  ast.body.forEach(function (n) {
     if (n.type === 'ExportDefaultDeclaration') {
       const exportMeta = captureDoc(source, docStyleParsers, n)
       if (n.declaration.type === 'Identifier') {
@@ -528,9 +549,14 @@ ExportMap.parse = function (path, content, context) {
       })
     }
 
+    const isEsModuleInteropTrue = isEsModuleInterop()
+
+    const exports = ['TSExportAssignment']
+    isEsModuleInteropTrue && exports.push('TSNamespaceExportDeclaration')
+
     // This doesn't declare anything, but changes what's being exported.
-    if (n.type === 'TSExportAssignment') {
-      const exportedName = n.expression.name
+    if (includes(exports, n.type)) {
+      const exportedName = n.expression && n.expression.name || n.id.name
       const declTypes = [
         'VariableDeclaration',
         'ClassDeclaration',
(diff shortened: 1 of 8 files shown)


ljharb:
@Maxim-Mazurok this seems fine, but is hard to review without tests. Could you add some?

Maxim-Mazurok:
Sure, @ljharb, I will first test if https://github.com/benmosher/eslint-plugin-import/issues/1527 was actually resolved and then will add some tests.
Also, as I've mentioned [here](https://github.com/benmosher/eslint-plugin-import/pull/1528#issuecomment-601020925) that we should consider the case when there are multiple `tsconfig.json` files. Anyway, supporting one root config would be a good starting point, I think.

Maxim-Mazurok:
Created reproduction repo: https://github.com/Maxim-Mazurok/eslint-plugin-import-1689-reproduction

The issue is still present in `2.20.2` and on `master` branch.

ljharb:
Thanks for confirming. I'd love to review and land this PR; could you modify your repro case to tests in this repo?

Maxim-Mazurok:
Sure, creating a repro was the first step :) Now I'm going to add tests.

Maxim-Mazurok:
@ljharb I fixed my code and added tests, please, take a look.
I also had to fix `import ReactDOM from 'react-dom';` case in my last commit to make it work on real-world app, where you import both `React` and `ReactDOM`.

Maxim-Mazurok:
Tests are failing on native because of this: https://github.com/nickmerwin/node-coveralls/issues/80

But pass on WSL

src/ExportMap.js
```diff
@@ -445,6 +449,21 @@ ExportMap.parse = function (path, content, context) {
...
 
+  function isEsModuleInterop() {
+    const tsConfig = tsConfigLoader({
+      cwd: context.parserOptions && context.parserOptions.tsconfigRootDir || process.cwd(),
+      getEnv: (key) => process.env[key],
+    })
+    try {
+      if (tsConfig.tsConfigPath !== undefined) {
+        const json = fs.readFileSync(tsConfig.tsConfigPath)
+        return JSON.parse(json).compilerOptions.esModuleInterop
```
ljharb, line 460: i don't believe that tsconfig is valid JSON, since it allows comments. I think we have to use whetever parsing package typescript itself uses to be able to parse this?

Maxim-Mazurok: Yep, valid point, they use their own parser, so I added commit to address this

Maxim-Mazurok:
@ljharb thanks for your comments and fixes. I added fix for `JSON.parse`. Please, don't force-push your commits that overwrite my commits, because I have no visibility as to what was changed by you. I understand that I did more commits than necessary, but you can "squash and merge" at the end to fix this problem. Thanks :)

ljharb:
@Maxim-Mazurok "squash and merge" is not viable - it leaves the PR ref cluttered with the old commits, and orphaned from the master branch.

I'm going to force push this one again before landing it; while you can uncheck the "allow edits" box to prevent me from doing that, that has the unfortunate consequence of meaning I'll never merge the PR on any repo I manage, so I hope you won't do that :-)

Maxim-Mazurok:
Oh, right, I totally forgot that, now it makes sense :)

Thanks!

ljharb:
Thank you for the PR!

PR #1691, opened 2020-03-22
Update no internal modules with exports

adjerbetian:
Resolves https://github.com/benmosher/eslint-plugin-import/issues/1481

## About the breaking change

In the ticket it was suggested to add an option for this behavior. Do you still want this rule behind an option? This behavior with the `export` feels so natural to me that I would feel strange if it wasn't the default behavior. 

## Changelog

I update the documentation, but I didn't know how to update the changelog. I wouldn't be against some guidance or help here :blush:.

## Step down rule

I hesitated to clean the code in [`/src/rules/no-internal-modules.js`](https://github.com/benmosher/eslint-plugin-import/blob/5101b73effbf3706495f62121a8719f0ea0e2c68/src/rules/no-internal-modules.js) to use the step down rule, but I don't want to impose my tastes on this. Do you want me to do it?


adjerbetian:
I don't understand why the build failed. It's some kind of job error not linked to the code:

> `An error occurred while generating the build script.`

adjerbetian:
@ljharb Can you give me your opinion on this?

ljharb:
I agree; this feels like a bugfix.

PR #1703, opened 2020-03-28
Release 2.20.2

ljharb:
Includes an already-published release of eslint-module-utils.


ljharb:
Tests failing are only eslint 3 in node 4, because esquery had a breaking change in a minor version (https://github.com/estools/esquery/issues/95).

PR #1708, opened 2020-03-29
glob pattern in `except` option of the `no-restricted-paths` rule

### LINKED ISSUE #1701: no-restricted-paths improve `except`
Hey I have a suggestion on how to improve the `except` option of the _import/no-restricted-paths_ rule (and I'm happy to implement it if you think it is a good idea).

Let me start with my **use case**:

```
my-project
├── client
│   └── foo.js
│   └── baz.js
└── server
    └── bar
        └── index.js
        └── types.js
```

I can use this rule to forbid importing server-only files in my client code. That is nice. But at the same time I also still want to allow to **import from type files** even though they are in the server directory and this is where I'm limited, because I can currently only pass a path to `except`.

Instead I would love to either pass a glob pattern like this

``` js
// solution proposal 1
{ "zones": [ {
    "target": "./client",
    "from": "./server",
    "except": ["./server/**/types"]
} ] }
```

...or a regex would also work:

``` js
// solution proposal 2
{ "zones": [ {
    "target": "./client",
    "from": "./server",
    "except": [/types/]
} ] }
```

Let me know what you think.
I know that #1662 might want to try to solve a similar issue, but I think it is not the same, but could still solve what @vladimiry wanted to achieve. I don't know his use case though.

KevinHerklotz:
Allow to also pass a glob pattern to the `except` option of the `no-restricted-paths` rule.
Before you could only add a relative path. This is still working and the new glob pattern is also only working in the relative path.

I also improved the documentation in general.

Fixes #1701


KevinHerklotz:
The CI is failing and I think the error message is not due to my changes. So I close the PR and reopen it again to trigger a rebuild.
This was the error:
```
$ nvm install lts/*
Downloading and installing node v12.16.1...
Downloading https://nodejs.org/dist/v12.16.1/node-v12.16.1-linux-x64.tar.xz...
curl: (56) GnuTLS recv error (-110): The TLS connection was non-properly terminated.
Binary download from https://nodejs.org/dist/v12.16.1/node-v12.16.1-linux-x64.tar.xz failed, trying source.
grep: /home/travis/.nvm/.cache/bin/node-v12.16.1-linux-x64/node-v12.16.1-linux-x64.tar.xz: No such file or directory
Provided file to checksum does not exist.
Binary download failed, trying source.
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Added
+- [`no-restricted-paths`]: Now you can also add glob patterns to the `except` option ([#1701], thanks [@KevinHerklotz])
```
ljharb, line 9:
```suggestion
- [`no-restricted-paths`]: Now you can also add glob patterns to the `except` option ([#1708], thanks [@KevinHerklotz])
```

CHANGELOG.md
```diff
@@ -680,6 +682,7 @@ for info on changes for earlier releases.
 [#1722]: https://github.com/benmosher/eslint-plugin-import/issues/1722
 [#1719]: https://github.com/benmosher/eslint-plugin-import/issues/1719
 [#1702]: https://github.com/benmosher/eslint-plugin-import/issues/1702
+[#1701]: https://github.com/benmosher/eslint-plugin-import/issues/1701
```
ljharb, line 685:
```suggestion
[#1708]: https://github.com/benmosher/eslint-plugin-import/pull/1708
```

docs/rules/no-restricted-paths.md
```diff
@@ -9,71 +9,86 @@ In order to prevent such scenarios this rule allows you to define restricted zon
...
 
 ```js
-import bar from '../server/bar';
+{ "zones": [ {
+    "target": "./client",
+    "from": "./server"
+} ] }
 ```
 
-The following patterns are not considered problems when configuration set to `{ "zones": [ { "target": "./client", "from": "./server" } ] }`:
```
ljharb, line 35: The "The following patterns" wording follows eslint's own docs; please restore it in both places.

src/rules/no-restricted-paths.js
```diff
@@ -75,13 +77,35 @@ module.exports = {
...
 
         matchingZones.forEach((zone) => {
-          const exceptionPaths = zone.except || []
+          const exceptions = zone.except || []
           const absoluteFrom = path.resolve(basePath, zone.from)
+          let exceptionPaths = []
+          let exceptionGlobs = []
+
+          exceptions.forEach(exception => {
+            if (isGlob(exception)) {
```
ljharb, line 86: is this not something minimatch can validate?

KevinHerklotz: I just looked into it, but unfortunately `minimatch` is not enough here. It would just tell me that it does not match, but it doesn't tell me if it is because the glob is not matching or because it cannot handle relative paths. If I don't know that I would also not know when to call `reportInvalidExceptionPath`.
So I guess we have to use isGlob here.

ljharb: what do you mean by "cannot handle relative paths"? is there a need for a different error message between globs and paths?

KevinHerklotz: Sorry, maybe I wasn't clear enough.
`isGlob` checks if it is a glob and thus a relative path like "./server" will be marked as non-glob ("exceptionPaths"). Otherwise if I would check for example "./server" with minimatch it will tell me this path doesn't match, no matter if this path would actually match, because minimatch only works with globs.

Also the check if you have an invalid exception path is done by `path.relative()`, which also does not work with globs. That's why you will currently not see the _"Restricted path exceptions must be descendants of the configured `from` path for that zone."_ error message for any glob you provide.
And I think it is totally ok, because you would use globs for different reasons anyways. For example if you want to exclude all "**/types.js", you cannot really check if this is a descendant (you could check if there is any file that matches, but this would take a lot of resources and wouldn't make sense to me).

KevinHerklotz:
To be honest I'm a bit lost now and don't really understand why the tests are failing. The error messages are not really helpful for me. Help from someone else is much appreciated.

ljharb:
@KevinHerklotz sorry for the delay. i've partially rebased this, and tests seem to be passing locally :-) but it'll need another rebase onto latest master.

Isn't a current "except" value already a glob, just without a `*`? Why can't we pass the entire array to minimatch without checking isGlob?

PR #1713, opened 2020-04-01
add test case for webpack alias

githoniel:
1. fix babelrc error
2. add webpack alias tests case for no-extraneous-dependencies

here is also a reproduce project for this issue

https://github.com/githoniel/eslint-import-webpack-alias-issue


ljharb:
I've rebased this; it fails on master, and v2.20.1, but indeed passes on v2.20.0.

Bisecting points at b4d5fd38a29073e273523e262d7d59a7c48e30fe / #1605. cc @skozin for thoughts?

ljharb:
ping @skozin; any ideas?

PR #1719, opened 2020-04-04
[Fix] `order`: Recognize pathGroup config for first group

forivall:
With the config
```js
  'import/order': [
    'warn',
    {
      groups: ['builtin', 'external', 'internal', 'parent', 'sibling', 'index'],
      alphabetize: {
        order: 'asc',
      },
      pathGroups: [{ pattern: 'react', group: 'builtin' }],
      pathGroupsExcludedImportTypes: ['builtin'],
    },
  ],
```
, react wasn't being recognized as a part of the "builtin" group, since the rank would end up being `0`, which is a falsy value. The fix here, is to check for `undefined` instead.

My current workaround is to define
```
      pathGroups: [{ pattern: 'react', group: 'builtin', position: 'before' }],
```
instead, (which results in a rank `-0.1`, so it's not falsy).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Fixed
 - [`group-exports`]: Flow type export awareness ([#1702], thanks [@ernestostifano])
+- [`order`]: Recognize pathGroup config for first group ([#1719], [#1724], thanks [@forivall], [@xpl])
 
 ## [2.20.2] - 2020-03-28
 ### Fixed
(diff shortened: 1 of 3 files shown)


forivall:
An alternative fix, in which i was playing around with javascript numbers: https://github.com/benmosher/eslint-plugin-import/commit/ac14b7b4bfa4b06cbbd8782e6f5cbec0bea00287

ljharb:
Because I pushed this to two PRs, two appveyor builds ran. The second failed (https://ci.appveyor.com/project/benmosher/eslint-plugin-import/builds/32116513), but the first succeeded (https://ci.appveyor.com/project/benmosher/eslint-plugin-import/builds/32116512), so I'm going to merge through the failure.

PR #1720, opened 2020-04-05
Run tests in Windows

mohsen1:
I made a PR instead of creating an issue because I'm offering to take on this work. This changelsit is simply using GitHub UI to add a few rows in travis.yml.  I'm pretty sure this will blow up. 

if you're open to GitHub actions, it offers a better experience running things in Windows with Powershell

We noticed an [issue in Redwood.js](https://github.com/redwoodjs/redwood/pull/363) that was rooted in this plugin not running its tests in Windows. I would like to fix the bugs and run the test in Windows to prevent future bugs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,12 @@ matrix:
   - os: osx
     env: ESLINT_VERSION=5
     node_js: 10
+  - os: windows
+    env: ESLINT_VERSION=5
+    node_js: 12
+  - os: windows
+    env: ESLINT_VERSION=5
+    node_js: 10
   - os: osx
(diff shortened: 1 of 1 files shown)


mohsen1:
Oh I completely missed Appveyor test. Never mind then. 

I think those tests are failing in master though: 

https://ci.appveyor.com/project/benmosher/eslint-plugin-import/builds/31956493/job/somxwgidmq6kafk5#L227

ljharb:
That link shows everything passing, just some eslint warnings.

ljharb:
Regarding the redwood.js issue, please feel free to open a PR with a failing test case :-)

PR #1724, opened 2020-04-09
Bug in order.js preventing from assigning `external` to patterns using `pathGroups`

xpl:
Here, if `group` is `external` or `builtin` then the returned rank will be `0`:

```javascript
if ((0, _minimatch2.default)(path, pattern, patternOptions || { nocomment: true })) {
      return ranks[group] + position / maxPosition;
}
```

...and then, there we would erroneously throw it away, as if `0` was not a valid rank (which isn't true):

```javascript
if (!rank) {
    rank = ranks.groups[impType];
}
```

Making the comparison more strict fixes the bug:

```javascript
if (rank === undefined) {
    rank = ranks.groups[impType];
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Fixed
 - [`group-exports`]: Flow type export awareness ([#1702], thanks [@ernestostifano])
+- [`order`]: Recognize pathGroup config for first group ([#1719], [#1724], thanks [@forivall], [@xpl])
 
 ## [2.20.2] - 2020-03-28
 ### Fixed
@@ -663,6 +664,8 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#1724]: https://github.com/benmosher/eslint-plugin-import/issues/1724
+[#1719]: https://github.com/benmosher/eslint-plugin-import/issues/1719
 [#1702]: https://github.com/benmosher/eslint-plugin-import/issues/1702
 [#1666]: https://github.com/benmosher/eslint-plugin-import/pull/1666
 [#1664]: https://github.com/benmosher/eslint-plugin-import/pull/1664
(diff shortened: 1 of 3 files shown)


golopot:
Can you add test cases in https://github.com/benmosher/eslint-plugin-import/blob/master/tests/src/rules/order.js ?

xpl:
Sure. Will add the case a bit later.

xpl:
@golopot I've added the case. Also, sorry about some confusion in my initial message — in my case, `builtin` and `external` were merged to a single group, having the same rank (`0`), hence the title.

I should've probably titled the PR "preventing from assigning `builtin`", but that doesn't really matter as the bug is still there.

So here is the code, it assigns `./make-me-external` to `external` group, and it fails without the fix, and succeeds with the fix:

```javascript
    test({
      code: `
        import fs from 'fs';
        import external from 'external';
        import externalTooPlease from './make-me-external';

        import sibling from './sibling';`,
      options: [{
        'newlines-between': 'always',
        pathGroupsExcludedImportTypes: [],
        pathGroups: [
          { pattern: './make-me-external', group: 'external' },
        ],
        groups: [['builtin', 'external'], 'internal', 'parent', 'sibling', 'index'],
      }],
    }),
```

Under the hood, this happens:

```
{
    ranks: {
        builtin: 0,   // <----- having rank 0
        external: 0,  // <----- having rank 0
        internal: 1,
        parent: 2,
        sibling: 3,
        index: 4,
        unknown: 5
    },
    pathGroups: [{
        pattern: './make-me-external',
        group: 'external',
        position: 0  // because it's 0, the expression evaluates to 0:
                     //     ranks[group] + (position / maxPosition)
    }]
}
```

xpl:
Oh my, here is a PR with the same fix (and with a bit more proper title)! :)

https://github.com/benmosher/eslint-plugin-import/pull/1719

ljharb:
Thanks, I'll handle getting both landed.

ljharb:
Because I pushed this to two PRs, two appveyor builds ran. The second failed (https://ci.appveyor.com/project/benmosher/eslint-plugin-import/builds/32116513), but the first succeeded (https://ci.appveyor.com/project/benmosher/eslint-plugin-import/builds/32116512), so I'm going to merge through the failure.

PR #1726, opened 2020-04-13
Disable import/named in the TypeScript configuration

astorije:
As I was linting our projects, I noticed a few violations across them:

```ts
import { IconName } from '@fortawesome/free-solid-svg-icons';
import { GetTrackProps } from 'react-compound-slider';
import { SliderItem } from 'react-compound-slider';
import { InjectedIntl } from 'react-intl';
```

Looking around, it seems the recommendations are to either use `import type` (I haven't tried) or disabling the rule locally.
I was wondering what this rule brings when used in a TypeScript setup and realized [I am not the only one](https://github.com/typescript-eslint/typescript-eslint/issues/154#issuecomment-517037192).

Since it makes sense to disable the rule when using TypeScript, and this rule is enabled by the `recommended` preset, would it be a ridiculous idea to disable it in the `typescript` preset of this repo, and save consumers the inconvenience of doing it locally?

_(I'm sure [some of these](https://github.com/benmosher/eslint-plugin-import/blob/master/config/recommended.js#L10-L19) could also be disabled, but I figured I would ask your opinions on this one first before looking at others)_

### DIFF
--- a/config/typescript.js
+++ b/config/typescript.js
@@ -19,4 +19,10 @@ module.exports = {
     },
   },
 
+  rules: {
+    // analysis/correctness
+
+    // TypeScript compilation already ensures that named imports exist in the referenced module
+    'import/named': 'off',
+  },
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`group-exports`]: Flow type export awareness ([#1702], thanks [@ernestostifano])
 - [`order`]: Recognize pathGroup config for first group ([#1719], [#1724], thanks [@forivall], [@xpl])
 
+### Changed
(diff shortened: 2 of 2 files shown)


config/typescript.js
```diff
@@ -19,4 +19,8 @@ module.exports = {
     },
   },
 
+  rules: {
+    // analysis/correctness
```
ljharb, line 23: the comment here should explain why it's disabled, i think

astorije: Good point, I added this as thee "category" header same as https://github.com/benmosher/eslint-plugin-import/blob/master/config/recommended.js#L9 but I can also do that!

astorije: @ljharb, done, what do you think?

astorije:
@ljharb, do you have any rough estimate as to when you are planning to release the next versionn of this plugin? Thanks! 🙏

PR #1727, opened 2020-04-13
Fix counting re-export as usage when used in combination with import

### LINKED ISSUE #1722: [no-unused-modules] Imports overriding reexports
I noticed that `no-unused-modules` with unusedExports trips up on this syntax:

a.js:
```js
export function a() {}

// The following line will wrongly error as unused:
export function aa() {}
```

b.js:
```js
import { a } from './a';

export { aa } from './a';
```

I've tracked this down to [this line](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-unused-modules.js#L199) in the src, which overwrites any previously detected reexports. The proposed fix is to do something like what is already being done in [these lines](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-unused-modules.js#L187-L191) to keep both reexported entries and imports.

I can give a shot at throwing together a PR if this looks reasonable?

Ephem:
Closes #1722 

This PR makes sure `aa` below is correctly counted as usage:

```js
import { a } from './a';
export { aa } from './a';
```

In addition to the attached test, this fix was tested against a previous false negative in a private project which now worked.

I found the existing tests quite hard to read and understand, so I went a bit off the beaten path and tried to isolate this test in new files to avoid coupling them too much to the existing ones. I also named the files descriptively as opposed to following the existing standard of letters. This was just best effort after finally grokking what was going on, so I'm happy to take any pointers on how to improve or tweak!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Fixed
 - [`group-exports`]: Flow type export awareness ([#1702], thanks [@ernestostifano])
 - [`order`]: Recognize pathGroup config for first group ([#1719], [#1724], thanks [@forivall], [@xpl])
+- [`no-unused-modules`]: Fix re-export not counting as usage when used in combination with import ([#1722], thanks [@Ephem])
 
 ### Changed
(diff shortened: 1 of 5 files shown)


ljharb:
New fixture files are great, thanks :-)

PR #1735, opened 2020-04-26
[Refactor] `no-extraneous-dependencies`: use moduleVisitor

adamborowski:
Hi.
I start with simple refactoring but I already have few suggestions with code changes and tests that will make no-extraneous-dependencies more flexible for monorepo projects.

This PR is a warmup. I'd like to see if our interaction goes well.

I hope you will enjoy my contribution.
Cheers,
Adam

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`no-unused-modules`]: avoid order-dependence ([#1744], thanks [@darkartur])
 
 ### Changed
+- [Refactor] `no-extraneous-dependencies`: use moduleVisitor ([#1735], thanks [@adamborowski])
 - TypeScript config: Disable [`named`][] ([#1726], thanks [@astorije])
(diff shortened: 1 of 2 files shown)


adamborowski:
I don't think so, this is just refactoring. I will add several PRs with tests from our forked rule that we really miss.

ljharb:
Thanks! I can merge it now without them, since existing tests pass, but more test cases are always appreciated, so I'll wait for those if you're OK with that :-)

adamborowski:
I think for this particular PR there is no more cases.

ljharb:
@adamborowski can you please check "allow edits" on the RHS of the PR? also, there's a github bug where PRs from forks that aren't your own (CloudinaryLtd vs adamborowski) won't grant the permission properly, so I'll need you to add me to that fork so I can force push to the PR branch. Thanks!

adamborowski:
Added. There is no option to allow edit on the right side

ljharb:
Interesting, maybe they "fixed" their bug by hiding the checkbox in those cases :-p Thanks, I should be able to land this shortly.

Issue #1667, opened 2020-02-23
"import type" from TypeScript 3.8 not recognized


juergenzimmermann:
After upgrading to TypeScript 3.8.2 I also use the new "import type" feature https://www.typescriptlang.org/docs/handbook/modules.html#importing-types. Now I'm getting a warning from _import/no-duplicates_.

My affected code is e.g.:
```
import type { Options } from 'express-rate-limit';
import rateLimit from 'express-rate-limit';
```

ljharb:
3.8 is brand new, so we definitely need updates to work with it, probably on many rules.

kmui2:
I think this PR may fix this specific issue with type imports: #1666

noahbrenner:
Just so we're all on the same page, I thought I'd confirm where we're all coming from. If I understand correctly:

* The PR @kmui2' referred to (#1666) includes a fixer that can combine a default export and a named export into one line.
* @juergenzimmermann's issue is that that `import type` and `import` lines *should* be separate since they behave differently -- they shouldn't be combined and they shouldn't be reported as an error for the `import/no-duplicates` rule.

Please correct me if I misunderstood either of you.

kmui2:
> Just so we're all on the same page, I thought I'd confirm where we're all coming from. If I understand correctly:
> 
> * The PR @kmui2' referred to (#1666) includes a fixer that can combine a default export and a named export into one line.
> * @juergenzimmermann's issue is that that `import type` and `import` lines _should_ be separate since they behave differently -- they shouldn't be combined and they shouldn't be reported as an error for the `import/no-duplicates` rule.
> 
> Please correct me if I misunderstood either of you.

Sorry, I was only speculating because I saw the plugin had support for `import type` coming from Flow types and thought it was something related to a generic default and named export problem.

noahbrenner:
No worries! I'm not seeing the string `"type"` anywhere in that particular PR, could that have just been a typo in the issue number, and there's another PR you found that *does* cover this? It would be great to see it if so!

kmui2:
> No worries! I'm not seeing the string `"type"` anywhere in that particular PR, could that have just been a typo in the issue number, and there's another PR you found that _does_ cover this? It would be great to see it if so!

I was actually looking into this myself for a bit, but I don't see any other related PR.

ljharb:
https://github.com/benmosher/eslint-plugin-import/issues/1667#issuecomment-594243654 matches my understanding.

[...]

PR #1736, opened 2020-04-27
[Tests] `order`: Add TS import type tests

kmui2:
Add TypeScript `import type` tests for `import/order` rule in #1667.


WolfgangKluge:
Btw. `import type * as name from "...";` is perfectly valid and thus should be in the test's, too.
This syntax is not mentioned in the typescript docs, but the compiler has no problem with it...

ljharb:
Figured out the node type name in eslint v2 :-)

PR #1740, opened 2020-04-29
[New] `order`: add `caseInsensitive: 'invert'` option

forivall:
Adds the "invert" option to "caseInsensitive"
- `caseInsensitive`: use `true` to ignore case, `false` to consider case, and `'invert'` to sort lowercase before uppercase (default: `false`).

bikeshed: I'd rename `caseInsensitive` to `caseSensitive` with default `true`. other naming options are viable.


forivall:
My current workaround which works for ~80% of the instances in the target codebase is
```js
const rules = {
  ...
  'import/order': [
    'warn',
    {
      groups: ['builtin', 'external', 'internal', 'parent', 'sibling', 'index'],
      alphabetize: {
        order: 'asc',
        caseInsensitive: true,
      },
      pathGroups: [
        { pattern: './[A-Z]*{,/**}', group: 'sibling', position: 'after' },
      ],
    },
  ],
  ...
}
```

forivall:
Sure: the use case is that in the frontend code at $job, data/core modules start with a lower case, and the Components start with UpperCase, and so, when i'm sorting imports, i want the business logic imports to be first, and then the imports of Presentation Logic.

(sorry for the late reply, got lost in my notifications)

docs/rules/order.md
```diff
@@ -211,12 +211,12 @@ import index from './';
 import sibling from './foo';
 ```
 
-### `alphabetize: {order: asc|desc|ignore, caseInsensitive: true|false}`:
+### `alphabetize: {order: asc|desc|ignore, caseInsensitive: true|false|'invert'}`:
```
ljharb, line 214: bikeshed: maybe in addition to true/false, we could support (and recommend) a three-option string enum? iow, this can be `'ignore' | 'upperFirst' | 'lowerFirst'`, with `false` being an alias for `ignore` and `true` being an alias for `upperFirst`?

src/rules/order.js
```diff
@@ -277,7 +286,10 @@ function mutateRanksToAlphabetize(imported, alphabetizeOptions) {
   const groupRanks = Object.keys(groupedByRanks)
 
   const sorterFn = getSorter(alphabetizeOptions.order === 'asc')
-  const comparator = alphabetizeOptions.caseInsensitive ? (a, b) => sorterFn(String(a).toLowerCase(), String(b).toLowerCase()) : (a, b) => sorterFn(a, b)
+  const comparator =
+    alphabetizeOptions.caseInsensitive === 'invert' ? (a, b) => sorterFn(swapCase(String(a)), swapCase(String(b)))
+    : alphabetizeOptions.caseInsensitive ? (a, b) => sorterFn(String(a).toLowerCase(), String(b).toLowerCase())
```
ljharb, line 291: seems like this should be handled inside `getSorter` rather than here?

ljharb (changes requested):
LGTM overall, just a few comments

docs/rules/order.md
forivall: Hmm, perhaps the "caseInsensitive" should be deprecated and make the 3 option string enum under an option called "caseSensitive" or "caseSensitivity"?

i wonder what other things call their options?
* GNU coreutils `sort` calls its option `--ignore-case`/`-f` (the f is for "fold lower case to upper case" - maybe too obscure of a term to use here)
  * although python has `str.casefold()`, algorithm defined in unicode spec)
* JS [`Intl.locale.caseFirst`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/caseFirst) and [`sensitivity` (under options)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#parameters)
* JS [`regExp.ignoreCase`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/ignoreCase)
* @typescript-eslint/eslint-plugin has [`sort-type-union-intersection-members`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/sort-type-union-intersection-members.md) (non-configurable)
* eslint core has [`sort-vars`](https://github.com/eslint/eslint/blob/master/docs/rules/sort-vars.md) has `ignoreCase`
* eslint [`sort-vars`](https://github.com/eslint/eslint/blob/master/docs/rules/sort-vars.md) and [`sort-imports`](https://github.com/eslint/eslint/blob/master/docs/rules/sort-imports.md) have `ignoreCase`
* eslint [`sort-keys`](https://github.com/eslint/eslint/blob/02fcc055710e8d69d986f1e682cae2014ad881e2/docs/rules/sort-keys.md) has `caseSensitive` and `natural`

With that all said, i'd like to suggest keeping `caseInsensitive` as a boolean, and for this option, introduce `caseFirst: 'upper' | 'lower'`, based on the `Intl.Collator` options

src/rules/order.js
forivall: Resolved with my new refactor commit.

docs/rules/order.md
forivall: It would probably also be a good idea to just use `Intl.Collator` like [what typescript-eslint does](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/rules/sort-type-union-intersection-members.ts#L164) instead of doing the case-munging manually

forivall: alternative diff, using `caseFirst` is here: https://github.com/benmosher/eslint-plugin-import/compare/master...forivall:feat/case-first

PR #1744, opened 2020-05-02
[Fix] Fixing typo in IDE section

### LINKED ISSUE #1743: [no-unused-modules] Typo in code
I've spent whole day to find this mistake) This line brakes lint-staged combined with no-unused-modules rule
https://github.com/benmosher/eslint-plugin-import/blob/92caa3594e0f8d7bf143dedba0c7c2b47b541f34/src/rules/no-unused-modules.js#L660

darkartur:
Fixes #1743

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`no-unused-modules`]: Revert "[flow] `no-unused-modules`: add flow type support" ([#1770], thanks [@Hypnosphi])
 - TypeScript: Add nested namespace handling ([#1763], thanks [@julien1619])
 - [`namespace`]/`ExportMap`: Fix interface declarations for TypeScript ([#1764], thanks [@julien1619])
+- [`no-unused-modules`]: avoid order-dependence ([#1744], thanks [@darkartur])
 
 ### Changed
 - TypeScript config: Disable [`named`][] ([#1726], thanks [@astorije])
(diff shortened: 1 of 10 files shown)


ljharb:
Can you provide a test case, that would have prevented this from happening?

darkartur:
> Can you provide a test case, that would have prevented this from happening?

I wanted but I haven't found any similar case in tests. It requires a big work to achieve this

ljharb:
Maybe I'm a bit confused; #1743 doesn't have much background. It shouldn't be possible for an IDE integration to do *anything* different than is done on the command line; if something works on the CLI but breaks in an IDE, the IDE itself is broken, full stop.

darkartur:
> Maybe I'm a bit confused; #1743 doesn't have much background. It shouldn't be possible for an IDE integration to do _anything_ different than is done on the command line; if something works on the CLI but breaks in an IDE, the IDE itself is broken, full stop.

Please take a look on update my previous message

ljharb:
Interesting, thanks - I'd bet you could add that setup inside https://github.com/benmosher/eslint-plugin-import/tree/master/tests/files, and that might lead to reproducing it?

darkartur:
> Interesting, thanks - I'd bet you could add that setup inside https://github.com/benmosher/eslint-plugin-import/tree/master/tests/files, and that might lead to reproducing it?

I can try, but how I can run two files in row in test?

darkartur:
> Interesting, thanks - I'd bet you could add that setup inside https://github.com/benmosher/eslint-plugin-import/tree/master/tests/files, and that might lead to reproducing it?

It was easier then I thought. My example was wrong a bit. This is right example:

```
// Component/components.ts
export { default as Component } from './Component.tsx'; // this used somewhere else
```
```
// Component/Component.tsx
export default function Component() {
}
```
```
eslint ./Component/components.ts ./Component/Component.tsx // fails
eslint ./Component/Component.tsx ./Component/components.ts // pass
```

darkartur:
I have found another case and fixed it too but I think this function have even more issues.

darkartur:
@ljharb what I can do to merge this?

Issue #1761, opened 2020-05-13
compatibility with eslint 7.0.0


akkumar:
eslint 7.0.0 got released recently. 

Using the latest version of eslint-plugin-import gives the following warning - 

`
npm WARN eslint-plugin-import@2.20.2 requires a peer of eslint@2.x - 6.x but none is installed. You must install peer dependencies yourself.
`

some relevant links - 

https://eslint.org/blog/2020/05/eslint-v7.0.0-released
https://eslint.org/docs/user-guide/migrating-to-7.0.0 . 

it would be useful to get eslint 7.0.0 supported .  tracking issue for the same

akkumar:
PR #1762

akkumar:
Apparently - the changes are already in the latest master for 7.0.0 compatibility. 

Can we just have a later release 2.20.3 or similar, to address that compatibility issue ?

ljharb:
Indeed, this is done and only pending a release.

PR #1762, opened 2020-05-13
eslint 7.x

akkumar:
eslint 7.x

Enforce necessary node versions as required by eslint 7.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "2.20.2",
   "description": "Import with sanity.",
   "engines": {
-    "node": "^10.12.0 || >=12.0.0"
-  },  
+    "node": ">=4"
+  },
   "main": "lib/index.js",
   "directories": {
     "test": "tests"
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "2.20.2",
   "description": "Import with sanity.",
   "engines": {
-    "node": ">=4"
-  },
+    "node": "^10.12.0 || >=12.0.0"
+  },  
   "main": "lib/index.js",
   "directories": {
     "test": "tests"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -3,8 +3,8 @@
   "version": "2.20.2",
   "description": "Import with sanity.",
   "engines": {
-    "node": ">=4"
-  },
+    "node": "^10.12.0 || >=12.0.0"
```
silviuburceadev, line 6: This shouldn't be here. Whilst [ESLint 7.0.0 requires it](https://eslint.org/docs/user-guide/migrating-to-7.0.0#drop-node-8), eslint-plugin-import shouldn't require Node 10+ as it can still be used with ESLint 6.x.

akkumar: Thanks for the review. Reverting changes to the old node engine requirement - >=4 ( although I think that is fairly old ! )

package.json
```diff
@@ -88,7 +88,7 @@
     "typescript-eslint-parser": "^22.0.0"
   },
   "peerDependencies": {
-    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.0.0-0"
```
ljharb, line 91: this already covers v7, as well as prereleases. There’s no need to change it.

akkumar:
@ljharb - I see this change (v7.0.0-0) has been added since the release - 2.20.2 . 

Right now - with the latest release (2.20.2 ) we get - 

`
npm WARN eslint-plugin-import@2.20.2 requires a peer of eslint@2.x - 6.x but none is installed. You must install peer dependencies yourself
` 
[ #1761 has more details ]. 

So given that the change is already in, may be - can we have another release 2.20.3 , or similar to address that issue ?

ljharb:
Yes, a new release is needed before eslint 7 is supported.

PR #1779, opened 2020-05-27
Remove wrong .babelrc file

nicolashenry:
`.babelrc` files should contain [json](https://babeljs.io/docs/en/6.26.3/babelrc) (JSON5) and not a path unless I miss something.


nicolashenry:
Oh ok this seems to not work on windows (this is not the only problem), I will try to contribute with modified files in my workspace then or I will try to use a docker image.

ljharb:
Indeed, it probably doesn't (unless you're using WSL).

PR #1783, opened 2020-05-28
Update some dev dependencies

nicolashenry:
This seems to fix some issues.
- nyc update seems to fix this error:
```
> nyc -s mocha "tests/src"

internal/modules/cjs/loader.js:969
  throw err;
  ^

Error: Cannot find module 'X:\NoBackup\git\eslint-plugin-import\node'
←[90m    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:966:15)←[39m
←[90m    at Function.Module._load (internal/modules/cjs/loader.js:842:27)←[39m
←[90m    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)←[39m
←[90m    at internal/main/run_main_module.js:17:47←[39m {
  code: ←[32m'MODULE_NOT_FOUND'←[39m,
  requireStack: []
}
```
- mocha update implies test/mocha.opts is replaced by .mocharc.* files
- some "no-extraneous-dependencies" tests were failing, I had to move the "require check" after the "dependencies check" to fix those:
```
  1843 passing (18s)
  2 failing

  1) no-extraneous-dependencies
       invalid
         import "lodash.isarray":

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Should have 1 error but had 0: []
      + expected - actual

      -0
      +1

      at testInvalidTemplate (node_modules\eslint\lib\rule-tester\rule-tester.js:657:24)
      at Context.<anonymous> (node_modules\eslint\lib\rule-tester\rule-tester.js:874:25)
      at processImmediate (internal/timers.js:456:21)

  2) no-extraneous-dependencies
       invalid
         var eslint = require("lodash.isarray"):

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Should have 1 error but had 0: []
      + expected - actual

      -0
      +1

      at testInvalidTemplate (node_modules\eslint\lib\rule-tester\rule-tester.js:657:24)
      at Context.<anonymous> (node_modules\eslint\lib\rule-tester\rule-tester.js:874:25)
      at processImmediate (internal/timers.js:456:21)
```


.mocharc.jsonc
```diff
@@ -0,0 +1,5 @@
+{
+	"reporter": "dot",
+	"recursive": true,
+	"timeout": 5000
+}
```
ljharb, line 5: this file needs a trailing  ., and i prefer rc files not to have extensions.

package.json
```diff
@@ -67,24 +67,24 @@
     "babylon": "^6.18.0",
     "chai": "^4.2.0",
     "coveralls": "^3.0.6",
-    "cross-env": "^4.0.0",
+    "cross-env": "^7.0.2",
```
ljharb, line 70: cross-eng v7 only supports node 10+, so we can't use it.

package.json
```diff
@@ -67,24 +67,24 @@
...
     "eslint-import-resolver-webpack": "file:./resolvers/webpack",
     "eslint-import-test-order-redirect": "file:./tests/files/order-redirect",
     "eslint-module-utils": "file:./utils",
     "eslint-plugin-eslint-plugin": "^2.2.1",
     "eslint-plugin-import": "2.x",
     "in-publish": "^2.0.0",
     "linklocal": "^2.8.2",
-    "mocha": "^3.5.3",
-    "nyc": "^11.9.0",
+    "mocha": "^7.2.0",
```
ljharb, line 81: mocha > 3 drops support for older engines, so we'll never be able to upgrade beyond mocha v3.

package.json
```diff
@@ -67,24 +67,24 @@
...
+    "nyc": "^15.0.1",
     "redux": "^3.7.2",
-    "rimraf": "^2.7.1",
-    "semver": "^6.3.0",
-    "sinon": "^2.4.1",
-    "typescript": "~3.8.3",
+    "rimraf": "^3.0.2",
+    "semver": "^7.3.2",
```
ljharb, line 85: same with semver > 6

package.json
```diff
@@ -67,24 +67,24 @@
...
     "eslint-import-test-order-redirect": "file:./tests/files/order-redirect",
     "eslint-module-utils": "file:./utils",
     "eslint-plugin-eslint-plugin": "^2.2.1",
     "eslint-plugin-import": "2.x",
     "in-publish": "^2.0.0",
     "linklocal": "^2.8.2",
-    "mocha": "^3.5.3",
-    "nyc": "^11.9.0",
+    "mocha": "^7.2.0",
+    "nyc": "^15.0.1",
```
ljharb, line 82: same with nyc > 11

package.json
nicolashenry: It's maybe the time to drop old engines? I mean at least node <= 8. I pretty sure that if people does not update node, they are not updating eslint-plugin-import either.

ljharb: Nope, as far as I'm concerned it'll never be the time. Plenty of people update their deps while stuck on older nodes.

.mocharc.jsonc
nicolashenry: Ok but it needs an extension in mocha following the [documentation](https://mochajs.org/#priorities):
```
.mocharc.js
.mocharc.yaml
.mocharc.yml
.mocharc.jsonc
.mocharc.json
```
Pick the one you prefer. :)

ljharb: since we can't upgrade mocha past mocha 3, and mocha 3 supports `test/mocha.opts`, which we already [have](https://github.com/benmosher/eslint-plugin-import/blob/master/test/mocha.opts), i'm not sure we have to pick?

package.json
nicolashenry: You mean that they update `eslint-plugin-import` but not `eslint` (which now require node >= 10.12)? I seems really unlikely to me. :/

I will close this issue then, you don't seem to be ready to update before some decades.

ljharb: yep, that's indeed likely; i've worked in codebases that have done it.

nicolashenry: Can I ask the reason to have this strange codebase update policy? I am curious :) I don't see any reason to not being able to update node, there is almost only non breaking changes between versions, almost only additions and bug/security fixes.

ljharb: Maximum back compact helps the ecosystem. Dropping support for old platforms makes it harder for those users to upgrade - and updating node is *very* difficult when you have a sizable app, for many reasons (binary deps, performance concerns, organizational fear)

Why would we need to update any of these deps? mocha and nyc and semver, for example, work perfectly for us. Why fix what isn’t broken?

nicolashenry: Not broken? It is for potential contributors like me because of what I said in description if you have read it and also in your own [AppVeyor results](https://ci.appveyor.com/project/benmosher/eslint-plugin-import/builds/32988451). 
What the point of putting this CI in place if you don't care of the result?

nyc seems to be the reason of the error so I updated it and I updated some others which seems good to me if the CI results are good too.

My only objective with this pull request was to be able to help for other issues but as it is today it is impossible (probably because I am on Windows and I can't use WSL sadly). 

Just to explain my opinion on this, updating libraries is not just a fancy thing, it helps to obtain help when searching on documentation/issues/etc when an error happens and also in development dependencies. From my experience, updated libraries have always been one of the best indicator of an application/library health.

Thank you for your response :)

ljharb: Thanks, that is a fair point - although these versions used to work on appveyor, so something else changed, and presumably we could fix whatever that is without changing deps.

I very much appreciate that a non-WSL Windows user is in a tough spot.

If mocha 3 isn't viable moving forward, then I'd prefer to switch the entire test suite to the only test framework that maintains back compat, tape - but I'm not sure that would address nyc's issue.

If there's a way to get Windows tests passing without updating anything, I would be quite happy about it :-)

PR #1786, opened 2020-05-29
Add fix for Windows Subsystem for Linux

manuth:
This PR will add support for Windows Subsystem for Linux.

Windows Subsystem for Linux seems to have quite a weird treatment with file and directory-names.
While directories are case-sensitive (directory `./tests` exists while `./TeSts` doesn't), files don't seem to be (both `reSOLVE.js` and `resolve.js` exist).

This PR adds WSL-support by checking whether both directories and files are case-sensitive before setting `CASE_SENSITIVE_FS`.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,7 @@
+configuration:
+  - Native
+  - WSL
+
 # Test against this version of Node.js
 environment:
   matrix:
@@ -8,8 +12,12 @@ environment:
   # - nodejs_version: "6"
   # - nodejs_version: "4"
 
+image: Visual Studio 2019
 matrix:
   fast_finish: false
+  exclude:
+    - configuration: WSL
+      nodejs_version: "8"
 
   # allow_failures:
   #   - nodejs_version: "4" # for eslint 5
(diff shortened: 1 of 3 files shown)


manuth:
https://github.com/benmosher/eslint-plugin-import/blob/3f46ccfe203fa5aad5767699df339a9b2f0294e3/tests/src/core/resolve.js#L165-L186

ljharb:
ahhh so the existing tests do cover it, but we're not actually running tests on WSL - just on normal windows.

If you already know how to add that to appveyor or travis, then that would be *amazing*, but if not, that is also OK and we can punt it to later.

appveyor.yml
```diff
@@ -18,45 +26,98 @@ matrix:
...
+  matrix:
+    only:
+      - configuration: WSL
+  # Install scripts. (runs after repo cloning)
+  install:
+    # Get the latest stable version of Node.js or io.js
+    - ps: $env:WSL_ENV += ":nodejs_version"
+    - ps: >-
+      wsl curl -sL 'https://deb.nodesource.com/setup_${nodejs_version$}.x' `| sudo -E bash -
+      wsl sudo apt install -y nodejs
```
ljharb, line 83: this line seems to trigger a parsing error?

manuth: Yeah - I indented the code incorrectly.
I just fixed it... no idea whether this WSL thing actually will work but we'll see 😅

manuth:
Ouch... looks like `apt` just invoked a form which requires user-interaction... might take a while until all the jobs run into a timeout :/

manuth:
@ljharb is it possible to run `console.log(require("./utils/resolve").CASE_SENSITIVE_FS)` on appveyor for test-purposes?

Though WSL works now, the npm-tests still fail which is very strange.
An error about the case-sensitive tests is reported even if they actually should be skipped (cause of `CASE_SENSITIVE_FS` being set to true)

manuth:
@ljharb as expected, "CASE_SENSITIVE_FS" is `true`, but still the case sensitive tests aren't skipped.

Any idea what might be buggy here?
Probably a function binding problem?

manuth:
@ljharb I can't help it - `CASE_SENSITIVE_FS` definitely is `true` and still the tests aren't skipped...
Anyways the WSL jobs are up and running now.

Any other idea on how to fix it?

manuth:
@ljharb sorry for spaming you with notifications
I found out what's causing the problem.

When only running `npm run tests-only`, `pretest` isn't executed, which actually would symlink `./utils` as module `eslint-module-utils`. The fact that `pretest` isn't executed causes npm to use `eslint-module-utils` from the npm-registry instead of `./utils`.

Should I change the `.travis.yml` and `appveyor.yml` files to execute `pretest`

ljharb:
Yes, that sounds like a great idea.

manuth:
@ljharb Alright, made it!
The WSL-jobs now should pass.
After merging this PR I could create another one for this here, if you want:
https://github.com/benmosher/eslint-plugin-import/blob/3f46ccfe203fa5aad5767699df339a9b2f0294e3/appveyor.yml#L40-L42

After doing so, this project's fully Windows compatible (using WSL, at least)

ljharb:
Yes, that would be amazing, thanks. I don't have any Windows machines to test, so Appveyor is the only validation this project really has.

ljharb:
Looks like node 8 and 14 "Native" tests are still failing. I've rebased this (and squashed and added a changelog entry), so let's see what happens.

manuth:
Looks like Babel is causing trouble
I'll try to update it and create a PR

ljharb:
hmm, now all the native tests are failing - but on the coveralls step, all the tests are passing on node 10 and 12 (node 14 still fails with "Cannot find module 'C:\projects\eslint-plugin-import\node", and node 8 still fails with a ton of "Path must be a string. Received undefined" errors)

Issue #1781, opened 2020-05-28
The development-environment doesn't work on windows


manuth:
When trying to test the module on a windows platform, weird errors appear as seen in AppVeyor:
https://ci.appveyor.com/project/benmosher/eslint-plugin-import/builds/33096831/job/gg7jvh5ehfb4s9ag

This also might be because the dependencies weren't updated for a long time and don't support `node 14.x`.

ljharb:
Do they pass in node 12? Which dependencies are you thinking of?

Typically non-compiled deps that work in node 0.6 work in every version of node since :-)

manuth:
I'm talking 'bout these deprecation-warnings:
```
npm WARN deprecated typescript-eslint-parser@22.0.0: Deprecated: Use @typescript-eslint/parser instead
npm WARN deprecated chokidar@1.7.0: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated core-js@2.6.11: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3.
npm WARN deprecated mkdirp@0.5.1: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
npm WARN deprecated json3@3.3.2: Please use the native JSON object instead of JSON 3
npm WARN deprecated typescript-estree@18.0.0: This package was moved to @typescript-eslint/typescript-estree, please install the latest version from there instead
npm WARN deprecated formatio@1.2.0: This package is unmaintained. Use @sinonjs/formatio instead
npm WARN deprecated samsam@1.3.0: This package has been deprecated in favour of @sinonjs/samsam
npm WARN deprecated fsevents@1.2.13: fsevents 1 will break on node v14+ and could be using insecure binaries. Upgrade to fsevents 2.
npm WARN deprecated text-encoding@0.6.4: no longer maintained
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
```

Most importantly `chokidar@1.7.0` and `fsevents@1.2.13` as they have something like "will break on node 14+" in their messages.

[...]

PR #1789, opened 2020-05-30
Make development-environment platform-independent

manuth:
This PR will make the `copy-metafiles` script alongside with some other scripts platform independent by using `node` rather than a command.
This script could be exported to an actual `.js`-file and invoked using `node ./copyMetafiles.js`.

Related to #1781


ljharb:
Oops, turns out `fs.copyFileSync` doesn't exist on node <= 6.

manuth:
Personally I wouldn't care that much about node <= 6 as the current LTS version is 12
I don't think there are any OSes and PCs out there you want to develop nodejs on which natively install nodejs v6.

But if not dropping there seems to be a polyfill-module on npm for exactly this purpose:
https://www.npmjs.com/package/fs-copy-file-sync

ljharb:
If we need to use a dep, then we might as well move it to a script file.

manuth:
@ljharb alright
I created separate scripts for all non-cross platform commands.
Please keep in mind that windows-tests still will fail as `windows-fix` isn't merged yet.

package.json
```diff
@@ -21,15 +21,15 @@
     "prebuild": "rimraf lib",
     "build": "babel --quiet --out-dir lib src",
     "postbuild": "npm run copy-metafiles",
-    "copy-metafiles": "for DIR in memo-parser resolvers/node resolvers/webpack utils; do cp LICENSE .npmrc \"${DIR}/\"; done",
+    "copy-metafiles": "node -e \"require('babel-register'); require('./scripts/copyMetafiles');\"",
```
ljharb, line 24:
```suggestion
    "copy-metafiles": "node --require=babel-register ./scripts/copyMetafiles",
```

package.json
```diff
@@ -21,15 +21,15 @@
...
     "watch": "npm run tests-only -- -- --watch",
     "pretest": "linklocal",
     "posttest": "eslint .",
     "mocha": "cross-env BABEL_ENV=test nyc -s mocha",
     "tests-only": "npm run mocha tests/src",
     "test": "npm run tests-only",
     "test-compiled": "npm run prepublish && BABEL_ENV=testCompiled mocha --compilers js:babel-register tests/src",
-    "test-all": "npm test && for resolver in ./resolvers/*; do cd $resolver && npm test && cd ../..; done",
+    "test-all": "node -e \"require('babel-register'); require('./scripts/testAll');\"",
```
ljharb, line 32:
```suggestion
    "test-all": "node --require=babel-register ./scripts/testAll",
```

scripts/resolverDirectories.js
```diff
@@ -0,0 +1,3 @@
+import glob from 'glob'
+
+export default glob.sync('./resolvers/*/')
```
ljharb, line 3: trailing newline

scripts/testAll.js
```diff
@@ -0,0 +1,20 @@
...
+
+let npmPath = npmWhich(__dirname).sync('npm')
+let spawnOptions = {
+    stdio: [process.stdin, process.stdout, process.stderr],
+}
+
+spawnSync(
+    npmPath,
+    ['test'],
+    Object.assign({ cwd: __dirname }, spawnOptions))
```
ljharb, line 13:
```suggestion
    { cwd: __dirname ...spawnOptions }
```
since we're using babel anyways

scripts/testAll.js
```diff
@@ -0,0 +1,20 @@
...
+
+for (let resolverDir of resolverDirectories) {
+    spawnSync(
+        npmPath,
+        ['test'],
+        Object.assign({ cwd: resolverDir }, spawnOptions))
```
ljharb, line 19:
```suggestion
        { cwd: resolverDir, ...spawnOptions }
```

scripts/testAll.js
```diff
@@ -0,0 +1,20 @@
+import { spawnSync } from 'child_process'
+import npmWhich from 'npm-which'
+import resolverDirectories from './resolverDirectories'
+
+let npmPath = npmWhich(__dirname).sync('npm')
+let spawnOptions = {
+    stdio: [process.stdin, process.stdout, process.stderr],
```
ljharb, line 7:
```suggestion
    stdio: 'inherit',
```
i think?

package.json
manuth: Oh - awesome, didn't know this is possible 😅

manuth:
@ljharb object spreading doesn't seem to be supported by this babel-setup

ljharb:
Had to tweak the CI script; the metafiles script was previously ran prior to running `npm install`.

PR #1791, opened 2020-05-31
Fix development-environment to work for Windows

### LINKED ISSUE #1780: 62 tests are failing
As mentioned in issue #1244 I just tried to create a PR for fixing said issue.
Sadly in the current state (without touching anything but just running `npm install` and `npm test`) there are 62 unit-tests failing.

That way it's a bit hard for me to tell whether changes made to the project "would" or "would not" cause any harm to the project.

### LINKED ISSUE #1781: The development-environment doesn't work on windows

manuth:
~Babel~ `nyc` caused some errors on its currently installed version as seen in the AppVeyor history logs.
Some key modules of the development-environment have been updated:
- `nyc` ("on the fly"-update in AppVeyor-script)
- ~`babel`-modules~
- ~`mocha`~

~Though, applying this PR won't cause any issues when using with older versions of `nodeJS` (unless babel is configured correctly (see below)), after applying this PR, a nodeJS `>=8` is required to develop this project.
As mentioned this should have no impact to the usage of the module, but to the development of the module which then will be performed with more modern module-versions.~

~To ensure, the built module also works for NodeJS `argon`, setting [this option (`targets.node`)](https://babeljs.io/docs/en/babel-preset-env#targetsnode) to the appropriate version-number might be necessary.~

This PR fixes #1780 and fixes #1781

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "test-compiled": "npm run prepublish && BABEL_ENV=testCompiled mocha --compilers js:babel-register tests/src",
     "test-all": "node --require babel-register ./scripts/testAll",
     "prepublish": "not-in-publish || npm run build",
-    "coveralls": "nyc report --reporter lcovonly && cat ./coverage/lcov.info | coveralls"
+    "coveralls": "nyc report --reporter lcovonly && coveralls < ./coverage/lcov.info"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "glob": "^7.1.6",
     "in-publish": "^2.0.0",
     "linklocal": "^2.8.2",
+    "lodash.isarray": "^4.0.0",
     "mocha": "^3.5.3",
     "npm-which": "^3.0.1",
     "nyc": "^11.9.0",
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -39,12 +39,17 @@ for:
     # install modules
     - ps: >-
         if ($env:nodejs_version -eq "4") {
-          npm install -g npm@3;
+          cmd /c npm install -g npm@3;
         }
         if ($env:nodejs_version -in @("8", "10", "12")) {
-          npm install -g npm@6.10.3;
+          cmd /c npm install -g npm@6.14.5;
         }
     - npm install
+
+    - ps: >-
+        if ([int]$env:nodejs_version -le 8) {
+          cmd /c npm i eslint@6 2`>`&1;
+        }
 
     # fix symlinks
     - git config core.symlinks true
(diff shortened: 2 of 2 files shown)


manuth:
@ljharb All's set up now, the tests for Windows as well as the tests for WSL run flawlessly now.
Only the `coveralls` command at the very end doesn't seem to work.
Probably you want to remove this from `appveyor` as there is a separate `coverage/coveralls` check anyway

package.json
```diff
@@ -79,8 +79,9 @@
     "eslint-plugin-import": "2.x",
     "in-publish": "^2.0.0",
     "linklocal": "^2.8.2",
-    "mocha": "^3.5.3",
-    "nyc": "^11.9.0",
+    "lodash.isarray": "^4.0.0",
+    "mocha": "^7.2.0",
+    "nyc": "^15.0.1",
```
ljharb, line 84: Unfortunately this is a change we can't ever really do; tests have to always be able to run in all the node versions we support, and we won't be dropping any node versions any time soon.

We can make the *build* require a newer node, and then run tests on the build output, as long as CI can be configured so it runs the build in newest node and then runs the tests in the appropriate node version.

manuth: Neither the `build`-command, nor the `test`-command will work on windows if `nyc` is left on v11, like it is currently.

manuth: What we could do instead would be to allow failures in appveyor tests.
Using `Windows Subsystem for Linux` we've got windows covered anyways.

manuth: @ljharb what about the `dep-time-travel.sh`-script here?
https://github.com/benmosher/eslint-plugin-import/blob/b280acdce76cb209564451805687eb244ef9799d/tests/dep-time-travel.sh

Is there a chance we could use this to downgrade `nyc` and `mocha` before running tests?

ljharb: I'd be concerned about incpompatibilies across the versions; mocha and nyc config, for example, or the way coverage is calculated.

Unfortunately, WSL doesn't cover an actual windows filesystem, which is the point of the appveyor tests :-/

Again, in travis, this is solved with stages - I'm not sure how else to solve it for windows.

Perhaps it's worth moving Windows/WSL to Github Actions or to Travis?

manuth: Currently I'm waiting for some AppVeyor tests to run. If the latest commit, which upgrades any existing `nyc` version to a predefined version, doesn't work, then we should rather move to GitHub Actions or Travis, yeah

manuth: @ljharb alright, that's it 😄 
The "on the fly"-update of `nyc` works flawlessly now, all jobs passed! 🥳 
I'll just get rid of some errors (like peer-dependency stuff) now and then we're good to go, I think ^^

manuth: Nvm - couldn't get rid of the peer-dependency error 😂
But still it's passing now ^^

manuth:
@ljharb alright I'm done here, too.
Would be awesome if you could merge this PR before #1792, so that I can rebase and refractor #1792 accordingly.

PR #1802, opened 2020-06-06
[no-restricted-path] Custom error message

malykhinvi:
**Main change**
Adds a `message` configuration param for every `zone`, that will be displayed in case of the rule violation. 
The logic is pretty similar to eslint's [no-restricted-imports](https://eslint.org/docs/rules/no-restricted-imports#options)

**Background**
Sometimes, it is not very clear, why you can not import certain modules in certain scenarios. This is especially the case for newcomers. To be more user friendly, we could display additional info, right in the error message. It would look something like this:

```javascript
//eslint config
'import/no-restricted-paths': [
  'error',
  {
    basePath: './src',
    zones: [ {
      target: './core',
      from: './features',
      message: '' +
        '\n\nCore should be completely isolated from features.' +
        '\nFeatures may use core but not vice versa',
    }],
  }
],
```
Then VSCode would display it like that:
Screenshot 2020-06-06 at 13 27 11

### DIFF
--- a/src/rules/no-restricted-paths.js
+++ b/src/rules/no-restricted-paths.js
@@ -32,6 +32,7 @@ module.exports = {
                   },
                   uniqueItems: true,
                 },
+                message: { type: 'string' },
               },
               additionalProperties: false,
             },
@@ -102,7 +103,7 @@ module.exports = {
 
           context.report({
             node,
-            message: `Unexpected path "{{importPath}}" imported in restricted zone.`,
+            message: `Unexpected path "{{importPath}}" imported in restricted zone.${zone.message ? ` ${zone.message}` : ''}`,
             data: { importPath },
           })
         })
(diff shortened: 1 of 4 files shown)


src/rules/no-restricted-paths.js
```diff
@@ -102,7 +103,8 @@ module.exports = {
 
           context.report({
             node,
-            message: `Unexpected path "{{importPath}}" imported in restricted zone.`,
+            message: `Unexpected path "{{importPath}}" imported in restricted zone.` +
+              (zone.message ? ` ${zone.message}` : ''),
```
ljharb, line 107:
```suggestion
            message: `Unexpected path "{{importPath}}" imported in restricted zone.${(zone.message ? ` ${zone.message}` : ''}`),
```

tests/src/rules/no-restricted-paths.js
```diff
@@ -145,6 +145,24 @@ ruleTester.run('no-restricted-paths', rule, {
...
+        zones: [ {
+          target: './tests/files/restricted-paths/server/one',
+          from: './tests/files/restricted-paths/server',
+          except: ['./one'],
+          message: 'Custom message',
+        } ],
+      } ],
+      errors: [ {
+        message: 'Unexpected path "../two/a.js" imported in restricted zone. ' +
+          'Custom message',
```
ljharb, line 161:
```suggestion
        message: 'Unexpected path "../two/a.js" imported in restricted zone. Custom message',
```

PR #1820, opened 2020-06-12
[Fix] `no-extraneous-dependencies`/TypeScript: do not error when importing type from dev dependencies

### LINKED ISSUE #1618: [import/no-extraneous-dependencies] Problems when only using types
We've started to see failures when moving from 2.18.2 to 2.20.0 regarding importing _only_ types from `devDependencies` in TypeScript source code. Previously the rule wasn't triggering for these (which I would expect), but it now is.

Refs https://github.com/libero/article-store/pull/174.

fernandopasik:
fixes #1618 

This PR is to avoid error `should be listed in the project's dependencies, not devDependencies.` when only importing typescript types

### DIFF
--- a/tests/src/rules/no-extraneous-dependencies.js
+++ b/tests/src/rules/no-extraneous-dependencies.js
@@ -1,4 +1,4 @@
-import { test } from '../utils'
+import { getTSParsers, test } from '../utils'
 import * as path from 'path'
 import * as fs from 'fs'
 
@@ -17,6 +17,7 @@ const packageFileWithSyntaxErrorMessage = (() => {
   }
 })()
 const packageDirWithFlowTyped = path.join(__dirname, '../../files/with-flow-typed')
+const packageDirWithTypescriptDevDependencies = path.join(__dirname, '../../files/with-typescript-dev-dependencies')
 const packageDirMonoRepoRoot = path.join(__dirname, '../../files/monorepo')
 const packageDirMonoRepoWithNested = path.join(__dirname, '../../files/monorepo/packages/nested-package')
 const packageDirWithEmpty = path.join(__dirname, '../../files/empty')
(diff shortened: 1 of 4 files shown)


fernandopasik:
@ljharb please can you review? :)

fernandopasik:
The tests are not passing in the old versions of eslint because they use an old version of `typescript-eslint-parser` that does not support `import type`. I was wondering if there is a way to not run these new tests in the old eslint versions, as this feature is not supported anyway.

fernandopasik:
I've filtered the unsupported parser but still having an issue with eslint 5 and the new ts parser

tests/src/rules/no-extraneous-dependencies.js
```diff
@@ -312,3 +313,35 @@ ruleTester.run('no-extraneous-dependencies', rule, {
...
+            code: 'import { JSONSchema7Type } from "@types/json-schema";',
+            options: [{packageDir: packageDirWithTypescriptDevDependencies, devDependencies: false }],
+            errors: [{
+              message: "'@types/json-schema' should be listed in the project's dependencies, not devDependencies.",
+            }],
+          }, parserConfig)),
+        ],
+      })
+    })
+})
```
ljharb, line 347: trailing newline

tests/src/rules/no-extraneous-dependencies.js
```diff
@@ -312,3 +313,35 @@ ruleTester.run('no-extraneous-dependencies', rule, {
     }),
   ],
 })
+
+describe('TypeScript', function () {
+  getTSParsers()
+    .filter((parser) => parser !== require.resolve('typescript-eslint-parser'))
```
ljharb, line 319: why filter this one out?

fernandopasik: `typescript-eslint-parser` is the old parser and doesn't support the `importKind` property. That parser was not done at the time of typescript 3.8 which is the one that supports this feature

I've added it back with the cases as all invalid.

tests/src/rules/no-extraneous-dependencies.js
fernandopasik: sorry about that one.

Would it be a good idea to add https://eslint.org/docs/rules/eol-last to the eslint config of the project?

ljharb: Yes, it would, thanks; i'll do that.

bradennapier:
Awesome, was just coming to ask if this was potentially possible - thanks for your work on this @fernandopasik !  Looking forward to this being released.

AndyOGo:
@fernandopasik 
This PR is great.
Though it doesn't support inline type imports.

```ts
// works
import type { Foo } from "types-package";

// doesn't work yet
import { type Foo } from "types-package".
```

fernandopasik:
Hey @AndyOGo ! Thanks, I think typescript introduced those after this PR

AndyOGo:
@fernandopasik 
@ljharb 
Thanks for your quick reply.
Right.

Anyway, I admire your work.
And I provided a patch to support inline types.
https://github.com/import-js/eslint-plugin-import/pull/2735

fernandopasik:
😳 You must mean you only admire @ljharb work, he's the only legend there

thank you! 🙇‍♂️

AndyOGo:
Of course, I admire @ljharb bombastic work a lot.

In general I admire people who seek to understand semantics, to help and make contributions.
Maybe my opinion is cynical, though to my observation only a few rare developers take the time to understand e.g. AST trees, take part in a project for free, help making it better.

So, I admire both of you.

PR #1785, opened 2020-05-29
[New] `order`: Add support for TypeScript's "import equals"-expressions

manuth:
This PR will add support for TypeScript's `import =`-syntax.

TypeScript not only allows you to import modules using `import path = require("path")`, but also to import entities for easier access:
```ts
import { Rule } from "eslint";
import RuleContext = Rule.RuleContext;
```

As the 2nd sort of import doesn't refer to a module, the `name` of this import will be set to `null`.
The sorting-functions for alphabetized imports has been updated according to this change.

Object-imports (which have their `name` set to `null`) will be forced to be put at the end of the group.

This PR fixes #1244

PR #1823, opened 2020-06-12
Fix Object-Import Errors

### LINKED ISSUE #1808: 'order ' rule breaks v2.21.1 - TypeError: Cannot read property 'indexOf' of null
I'm currently on the latest version of `eslint-plugin-import` (`2.21.1`), and it seems to be breaking my linting stage for some reason. 

```
TypeError: Cannot read property 'indexOf' of null
Occurred while linting/index.ts:40
    at isAbsolute (/node_modules/eslint-plugin-import/lib/core/importType.js:48:15)
    at typeTest (/node_modules/eslint-plugin-import/lib/core/importType.js:117:7)
    at resolveImportType (/node_modules/eslint-plugin-import/lib/core/importType.js:149:10)
    at computeRank (/node_modules/eslint-plugin-import/lib/rules/order.js:315:44)
    at registerNode (/node_modules/eslint-plugin-import/lib/rules/order.js:331:16)
    at handleImports (/node_modules/eslint-plugin-import/lib/rules/order.js:598:9)
    at /node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Object.emit (/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/node_modules/eslint/lib/linter/node-event-generator.js:254:26)

```

Can't see how this particular line will go wrong, unless there are any breaking changes in that rule:

`  export import CreateSomething = _CreateSomething;
`

I didn't face any of such issues on the previous minor version. If there are any new changes to the rule, it should be reflected as part of the linting result, rather than throwing an error halfway.

### LINKED ISSUE #1821: "`null` import" should occur before relative imports with TS `import =` syntax
I am encountering the following issue when trying to import individual members from a TS namespace:

image

This was last working without issue on v2.20.2. It first became an error in v2.21.0, but I saw there were some other issues opened around the same time about `import =` syntax, so I wanted to wait and see if it would be fixed in any of the subsequent releases.

[Here is a repo](https://github.com/acdibble/null-import) with a recreation of the issue.

manuth:
The changes I've made in PR #1785 caused object-imports to fail.
This happened because the `order`-rule is not able to handle `null` as an import-name and I totally forgot to add tests for object-imports.

This PR will fix the support of object-imports by introducing a new import-type in addition to `require` and `import`: The `import:object`-type.

In order to match the `order`-rule's behavior perfectly, the object-literal being imported acts as the name which means following:
```ts
import log = console.log; // name: "console.log"
import debug = console.debug; // name: "console.debug"
```

Additionally there is a new import-`group` you can set for the `order`-rule called `"object"`, which includes all object-imports.

This new `"object"` group is in the `pathGroupsExcludedImportTypes`-array by default.

This PR will fix #1821 and will fix #1808 
Sorry for causing trouble 😅

### DIFF
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -22,6 +22,8 @@ import bar from './bar';
 import baz from './bar/baz';
 // 6. "index" of the current directory
 import main from './';
+// 7. "object"-imports (only available in TypeScript)
+import log = console.log;
 ```
 
 Unassigned imports are ignored, as the order they are imported in may be important.
@@ -77,12 +79,15 @@ This rule supports the following options:
 
 ### `groups: [array]`:
 
-How groups are defined, and the order to respect. `groups` must be an array of `string` or [`string`]. The only allowed `string`s are: `"builtin"`, `"external"`, `"internal"`, `"unknown"`, `"parent"`, `"sibling"`, `"index"`. The enforced order is the same as the order of each element in a group. Omitted types are implicitly grouped together as the last element. Example:
+How groups are defined, and the order to respect. `groups` must be an array of `string` or [`string`]. The only allowed `string`s are:
+`"builtin"`, `"external"`, `"internal"`, `"unknown"`, `"parent"`, `"sibling"`, `"index"`, `"object"`.
+The enforced order is the same as the order of each element in a group. Omitted types are implicitly grouped together as the last element. Example:
 ```js
 [
   'builtin', // Built-in types are first
   ['sibling', 'parent'], // Then sibling and parent types. They can be mingled together
   'index', // Then the index file
+  'object',
   // Then the rest: internal and external type
 ]
 ```
@@ -91,7 +96,7 @@ The default value is `["builtin", "external", "parent", "sibling", "index"]`.
 You can set the options like this:
 
 ```js
-"import/order": ["error", {"groups": ["index", "sibling", "parent", "internal", "external", "builtin"]}]
+"import/order": ["error", {"groups": ["index", "sibling", "parent", "internal", "external", "builtin", "object"]}]
 ```
 
 ### `pathGroups: [array of objects]`:
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -248,14 +248,14 @@ function makeOutOfOrderReport(context, imported) {
 }
 
 function getSorter(ascending) {
-  let multiplier = (ascending ? 1 : -1)
+  const multiplier = ascending ? 1 : -1
 
   return function importsSorter(importA, importB) {
     let result
 
-    if ((importA < importB) || importB === null) {
+    if (importA < importB) {
       result = -1
-    } else if ((importA > importB) || importA === null) {
+    } else if (importA > importB) {
       result = 1
     } else {
       result = 0
@@ -310,8 +310,13 @@ function computePathRank(ranks, pathGroups, path, maxPosition) {
   }
 }
 
-function computeRank(context, ranks, name, type, excludedImportTypes) {
-  const impType = importType(name, context)
+function computeRank(context, node, ranks, name, type, excludedImportTypes) {
+  let impType
+  if (type === 'import:object') {
+    impType = 'object'
+  } else {
+    impType = importType(name, context)
+  }
   let rank
   if (!excludedImportTypes.has(impType)) {
     rank = computePathRank(ranks.groups, ranks.pathGroups, name, ranks.maxPosition)
@@ -319,7 +324,7 @@ function computeRank(context, ranks, name, type, excludedImportTypes) {
   if (typeof rank === 'undefined') {
     rank = ranks.groups[impType]
   }
-  if (type !== 'import') {
+  if (type !== 'import' && !type.startsWith('import:')) {
     rank += 100
   }
 
@@ -327,7 +332,7 @@ function computeRank(context, ranks, name, type, excludedImportTypes) {
 }
 
 function registerNode(context, node, name, type, ranks, imported, excludedImportTypes) {
-  const rank = computeRank(context, ranks, name, type, excludedImportTypes)
+  const rank = computeRank(context, node, ranks, name, type, excludedImportTypes)
   if (rank !== -1) {
     imported.push({name, rank, node})
   }
@@ -338,7 +343,7 @@ function isInVariableDeclarator(node) {
     (node.type === 'VariableDeclarator' || isInVariableDeclarator(node.parent))
 }
 
-const types = ['builtin', 'external', 'internal', 'unknown', 'parent', 'sibling', 'index']
+const types = ['builtin', 'external', 'internal', 'unknown', 'parent', 'sibling', 'index', 'object']
 
 // Creates an object with type-rank pairs.
 // Example: { index: 0, sibling: 1, parent: 1, external: 1, builtin: 2, internal: 2 }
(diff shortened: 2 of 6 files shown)


docs/rules/order.md
```diff
@@ -77,21 +79,24 @@ This rule supports the following options:
...
 [
   'builtin', // Built-in types are first
   ['sibling', 'parent'], // Then sibling and parent types. They can be mingled together
   'index', // Then the index file
+  'object',
   // Then the rest: internal and external type
 ]
 ```
-The default value is `["builtin", "external", "parent", "sibling", "index"]`.
+The default value is `["builtin", "external", "parent", "sibling", "index", "object"]`.
```
ljharb, line 94: generally it's a breaking change to alter the default value here. can you help me understand why this wouldn't be breaking?

docs/rules/order.md
```diff
@@ -100,11 +105,11 @@ To be able to group by paths mostly needed with aliases pathGroups can be define
...
-| patternOptions |          | object | options for minimatch, default: { nocomment: true } |
-| group          |     x    | string | one of the allowed groups, the pathGroup will be positioned relative to this group |
+| property       | required | type   | description                                                                                                                                              |
+| -------------- | :------: | ------ | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| pattern        |    x     | string | minimatch pattern for the paths to be in this group (will not be used for builtins or externals)                                                         |
+| patternOptions |          | object | options for minimatch, default: { nocomment: true }                                                                                                      |
+| group          |    x     | string | one of the allowed groups, the pathGroup will be positioned relative to this group                                                                       |
```
ljharb, line 112: i'm not sure why this table was changed at all (perhaps prettier running on a project that doesn't use it?) but please revert it.

src/rules/order.js
```diff
@@ -310,24 +310,29 @@ function computePathRank(ranks, pathGroups, path, maxPosition) {
...
+  }
   let rank
   if (!excludedImportTypes.has(impType)) {
     rank = computePathRank(ranks.groups, ranks.pathGroups, name, ranks.maxPosition)
   }
   if (typeof rank === 'undefined') {
     rank = ranks.groups[impType]
   }
-  if (type !== 'import') {
+  if (!/import(:.*)?/.test(type)) {
```
ljharb, line 327:
```suggestion
  if (!type.startsWith('import')) {
```

src/rules/order.js
```diff
@@ -609,18 +614,18 @@ module.exports = {
...
         let name
+        let type = 'import'
         if (node.moduleReference.type === 'TSExternalModuleReference') {
           name = node.moduleReference.expression.value
-        } else if (node.isExport) {
-          name = node.moduleReference.name
         } else {
-          name = null
+          name = context.getSourceCode().getText(node.moduleReference)
+          type += ':object'
```
ljharb, line 622:
```suggestion
          type = 'import:object'
```
this makes it easier to grep for

tests/src/rules/order.js
```diff
@@ -711,6 +711,51 @@ ruleTester.run('order', rule, {
...
+        parser,
+        options: [
+          {
+            alphabetize: {
+              order: 'asc',
+            },
+          },
+        ],
+      })
+    )),
```
ljharb, line 758:
```suggestion
    ...flatMap(getTSParsers(), parser => [
      test({
        code: `
          import blah = require('./blah');
          import { hello } from './hello';`,
        parser,
        options: [
          {
            alphabetize: {
              order: 'asc',
            },
          },
        ],
      }),

      // Order of object-imports
      test({
        code: `
          import blah = require('./blah');
          import log = console.log;`,
        parser,
        options: [
          {
            alphabetize: {
              order: 'asc',
            },
          },
        ],
      }),
      test({
        code: `
          import debug = console.debug;
          import log = console.log;`,
        parser,
        options: [
          {
            alphabetize: {
              order: 'asc',
            },
          },
        ],
      }),
    ]),
```

docs/rules/order.md
manuth: Oh... Sorry, my interpretation of "breaking change" was a change of the API 😅

If changing the default value of a setting is a breaking change, then this indeed is one.

docs/rules/order.md
manuth: Will do

src/rules/order.js
manuth: When using "startsWith" it's not ensured, that it's either just "import" or "import:something". Is this change meant for performance gain?

src/rules/order.js
manuth: Will do

tests/src/rules/order.js
manuth: Will do

src/rules/order.js
manuth: I'd recommend to change this to `^import(:.*)?$/.test(type)` to be sure, that `type` either matches `import` or `import:something`.

ljharb: I'd prefer a non-regex check - even multiple `===` would be better.

docs/rules/order.md
ljharb: What happens now with the ordering of object imports, versus what happens after this change?

manuth: Before #1785, `import [...] = Object.literal;`s were ignored as well as `import [...] = require("module-name");`.
After #1785 `import [...] = require([...]);`s were included into the sorting while object-imports caused an error cause of this piece of code:
https://github.com/benmosher/eslint-plugin-import/blob/62b554bc5f5024695b40d8701fbd8267fb4eb73f/src/rules/order.js#L614-L616

After this PR, #1823, object-imports are included into the sorting, having their own group called `object`. Additionally, object-imports are then sorted according to their object-literal:

```ts
/* eslint import/order: ["error", {"alphabetize": {"order": "asc"}}] */
import path = require("path");
import Options = ESLint.Options; // This will fail, because "ESLint.Options" must be imported after "ESLint.LintResult"
import LintResult = ESLint.LintResult;
```

ljharb: That means that valid code before would suddenly start being invalid, and an autofix would suddenly move code.

This means that we can't add this category to the defaults at this time.

manuth: But wouldn't removing this from the `groups` cause it to be forced to be at the end anyway?
At least if I understand this correctly:
https://github.com/benmosher/eslint-plugin-import/blame/13d2e10fa42e5a978fa00a52a1b10553913d3c16/docs/rules/order.md#L84

manuth: Should I remove `object` from the default value of the `groups` setting?

ljharb: I would expect that with it not in "groups" they should be ignored entirely.

ljharb: The current behavior is that all of the missing groups are treated as the same - meaning, not sorted within themselves, just mushed together. By adding it, you're forcing `object` to go *before* all the other missing ones.

so yes, it should be removed.

manuth: Aaaah got it - thanks for the elaboration :D

manuth: There you go, I just reverted the mutation of the default settings 😄

manuth:
@ljharb do you still have a fair share of PRs to work on before the next minor or build version is released?
Having this change implemented would help me a lot working on my projects 😄

ljharb:
Not many; I’ll look into a release soon.

Issue #1829, opened 2020-06-19
TypeScript's "export import" is treated as import


be5invis:
TypeScript has a special syntax to re-export an namespace looks like `export import <name> = <binding>`. Currently eslint-plugin-import is treating them as imports and applies incorrect rules.

```ts
// with import/order, the "export import" line will cause an error
import { Point } from "./point";
export namespace SomeNamespace {
    export import Pt = Point;
}
```

ljharb:
If you could provide a PR with a failing test case, that'd be most helpful.

ljharb:
Although, this might have been fixed in #1823 - can you check with latest master?

be5invis:
@ljharb It looks like that #1823 fixed this issue. I created a PR to add a test about this.

be5invis:
@ljharb Tests in #1823 shows `newline-after-import` doesn't recognize TypeScript's `export import` and treated it as an import instead of an export.

ljharb:
Fixed in #1830.

PR #1830, opened 2020-06-19
`order`/`newline-after-import`: Add support for TypeScript's "export import"

be5invis:
This PR adds support for TypeScript's "export import", typically used to re-export things.
In current `eslint-plugin-import` they are treated as imports rather than exports.
Related: #1829.
cc. @ljharb

### DIFF
--- a/src/rules/newline-after-import.js
+++ b/src/rules/newline-after-import.js
@@ -119,8 +119,13 @@ after ${type} statement not followed by another ${type}.`,
         const { parent } = node
         const nodePosition = parent.body.indexOf(node)
         const nextNode = parent.body[nodePosition + 1]
+        
+        // skip "export import"s
+        if (node.type === 'TSImportEqualsDeclaration' && node.isExport) {
+          return
+        }
 
-        if (nextNode && nextNode.type !== 'ImportDeclaration' && nextNode.type !== 'TSImportEqualsDeclaration') {
+        if (nextNode && nextNode.type !== 'ImportDeclaration' && (nextNode.type !== 'TSImportEqualsDeclaration' || nextNode.isExport)) {
           checkForNewLine(node, nextNode, 'import')
         }
     }
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -615,6 +615,10 @@ module.exports = {
       TSImportEqualsDeclaration: function handleImports(node) {
         let name
         let type
+        // skip "export import"s
+        if (node.isExport) {
+          return
+        }
         if (node.moduleReference.type === 'TSExternalModuleReference') {
           name = node.moduleReference.expression.value
           type = 'import'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`order`]/TypeScript: properly support `import = object` expressions ([#1823], thanks [@manuth])
 - [`no-extraneous-dependencies`]/TypeScript: do not error when importing type from dev dependencies ([#1820], thanks [@fernandopasik])
 - [`default`]: avoid crash with `export =` ([#1822], thanks [@AndrewLeedham])
+- [`order`]/[`newline-after-import`]: ignore TypeScript's "export import object" ([#1830], thanks [@be5invis])
 
 ### Changed
 - [`no-extraneous-dependencies`]: add tests for importing types ([#1824], thanks [@taye])
@@ -712,6 +713,7 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#1830]: https://github.com/benmosher/eslint-plugin-import/pull/1830
 [#1824]: https://github.com/benmosher/eslint-plugin-import/pull/1824
 [#1823]: https://github.com/benmosher/eslint-plugin-import/pull/1823
 [#1822]: https://github.com/benmosher/eslint-plugin-import/pull/1822
@@ -1235,3 +1237,4 @@ for info on changes for earlier releases.
 [@fernandopasik]: https://github.com/fernandopasik
 [@taye]: https://github.com/taye
 [@AndrewLeedham]: https://github.com/AndrewLeedham
+[@be5invis]: https://github.com/be5invis
--- a/tests/src/rules/newline-after-import.js
+++ b/tests/src/rules/newline-after-import.js
@@ -213,6 +213,24 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
         parser: parser,
         parserOptions: { ecmaVersion: 2015, sourceType: 'module' },
       },
+      {
+        code: `
+          export import a = obj;\nf(a);
+        `,
+        parser: parser,
+        parserOptions: { ecmaVersion: 2015, sourceType: 'module' },
+      },
+      {
+        code: `
+          import { a } from "./a";
+
+          export namespace SomeNamespace {
+              export import a2 = a;
+              f(a);
+          }`,
+        parser: parser,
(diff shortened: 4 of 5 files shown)


be5invis:
@ljharb Hah, these tests reveal a bug in `newline-after-import`. I will file another issue.

ljharb:
lol, thanks :-) we can just turn this PR into the fix for it.

be5invis:
Add @manuth 

@ljharb I may not able to create a fix because I cannot get `eslint-plugin-import`'s testing work on my machine, it always ends with these errors:

```
npm run test

> eslint-plugin-import@2.21.2 pretest Z:\eslint-plugin-import
> linklocal

tests\files\order-redirect
resolvers\webpack
tests\files\symlinked-module
tests\files\order-redirect-scoped
utils
resolvers\node

Linked 6 dependencies

> eslint-plugin-import@2.21.2 test Z:\eslint-plugin-import
> npm run tests-only


> eslint-plugin-import@2.21.2 tests-only Z:\eslint-plugin-import
> npm run mocha tests/src


> eslint-plugin-import@2.21.2 mocha Z:\eslint-plugin-import
> cross-env BABEL_ENV=test nyc -s mocha "tests/src"

internal/modules/cjs/loader.js:985
  throw err;
  ^

Error: Cannot find module 'Z:\eslint-plugin-import\node'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:982:15)
    at Function.Module._load (internal/modules/cjs/loader.js:864:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12)
    at internal/main/run_main_module.js:18:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
events.js:288
      throw er; // Unhandled 'error' event
      ^

Error: spawn nyc ENOENT
    at notFoundError (Z:\eslint-plugin-import\node_modules\cross-spawn\lib\enoent.js:11:11)
    at verifyENOENT (Z:\eslint-plugin-import\node_modules\cross-spawn\lib\enoent.js:46:16)
    at ChildProcess.cp.emit (Z:\eslint-plugin-import\node_modules\cross-spawn\lib\enoent.js:33:19)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:275:12)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess.cp.emit (Z:\eslint-plugin-import\node_modules\cross-spawn\lib\enoent.js:36:37)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:275:12) {
  errno: 'ENOENT',
  code: 'ENOENT',
  syscall: 'spawn nyc'
}
```

manuth:
@be5invis for getting this repo to work on a windows machine I recommend you to use "Windows Subsystem for Linux" aka. WSL (VSCode provides extensions for developing projects inside WSL.

If you don't want to use WSL, you can upgrade the nyc dependency:
```
npm install - - no-save nyc@latest
```
After this, tests should work fine on windows.

manuth:
@be5invis do I understand correctly, that your goal is, that `export import` statements basically are ignored by the `import/order` rule?

be5invis:
@manuth 
I think `import/order` ignores "object imports" so I added similar ignoring to `import/newline-after-import`.
Upgrading `nyc` fixed the test runner so I can now make a fix 🥂.

manuth:
Recent changes in #1823 cause them to not be ignored by `import/order` 😅
Looks like there's more work to be done for supporting `TSImportEqualsDeclaration`...

manuth:
Shall we discuss `TSImportEqualsDeclaration`s right here or should I create a separate PR?

be5invis:
@manuth Ah I understand what you mean...
There's also a `isExport` property for `TSImportEqualsDeclaration`s. We can use that to ignore `export import`s.

manuth:
Exactly... it's not just this, sadly...
As we have learned, there are two different kinds of `TSImportEqualsDeclaration`s: Object-literal imports and module-imports.
Each of them can either be imported or imported and exported both at the same time:

```ts
// Modules
import fs = require("fs");
// or
export import fs = require("fs");

// Object-literals
import LintResult = ESLint.LintResult;
// or
export import LintResult = ESLint.LintResult;
```

### Problems About Exports
As you have brought up in this issue, `export import`s can occur anywhere (that is in namespaces, in nested contexts etc.).
Imo, this can only be solved by letting `import/order` ignore `export import`s.

### Problems About Object-Literals
My last changes in mentioned PR caused object-imports to be forced to be alphabetized as well.
I've noticed, that this is a bad behavior, too, as object-imports might depend on each other:
```ts
import A = ESLint;
import LintResult = A.LintResult; // Reports an error because "ESLint" > "A.LintResult"
```
I don't really know what to do about it... either we could set the `name` of each object-import to `""`, which causes any order of object-imports to be allowed.
Otherwise we could just let `import/order` ignore object-imports as well.

Also there's more to be done about `TSImportEqualsDeclaration`s. I think we should add support for this sort of import-expressions for all the other rules as well.

be5invis:
@manuth Agree. We could file other work items for further support of these.

manuth:
Alright! What about fixing object-imports and `export import`s?
Shall we just ignore them in the `import/order`-rule?

src/rules/order.js
```diff
@@ -615,6 +615,10 @@ module.exports = {
       TSImportEqualsDeclaration: function handleImports(node) {
         let name
         let type
+        // skip "export import"s
+        if (node.isExport) {
+          return
+        }
         if (node.moduleReference.type === 'TSExternalModuleReference') {
           name = node.moduleReference.expression.value
           type = 'import'
```
manuth, line 624: Sadly I can't create a suggestion because the `registerNode` is too far appart from your actual change... however:
I'd recommend you to change this piece of code to...

```ts
    if (!node.isExport && moduleReference.type === 'TSExternalModuleReference') {
        registerNode(
            context,
            node,
            node.moduleReference.expression.value,
            'import',
            raunks,
            imported,
            pathGroupdExcludedImportTypes
        )
    }

Comment (manuth) [reply src/rules/order.js:624]: Also please notice, that these pieces of code:
https://github.com/benmosher/eslint-plugin-import/blob/c36429bf6cb80e6ca75696de031a58927bba8db5/src/rules/order.js#L315-L319
https://github.com/benmosher/eslint-plugin-import/blob/c36429bf6cb80e6ca75696de031a58927bba8db5/src/rules/order.js#L327-L329
Is not necessary anymore anymore after applying this change, because `import:object` won't exist anymore in this rule.

File: src/rules/newline-after-import.js [hunk 2]
```diff
@@ -119,6 +119,11 @@ after ${type} statement not followed by another ${type}.`,
         const { parent } = node
         const nodePosition = parent.body.indexOf(node)
         const nextNode = parent.body[nodePosition + 1]
+        
+        // skip "export import"s
+        if (node.type === 'TSImportEqualsDeclaration' && node.isExport) {
+          return
+        }
 
         if (nextNode && nextNode.type !== 'ImportDeclaration' && nextNode.type !== 'TSImportEqualsDeclaration') {
```
manuth, line 128: Make sure to ignore `export import`s here, too:
```suggestion
        if (nextNode && nextNode.type !== 'ImportDeclaration' && (nextNode.type !== 'TSImportEqualsDeclaration' || nextNode.isExport)) {
```

src/rules/order.js
be5invis: There may still be `import x = object`, so this needs to be kept.

manuth: Though, that's true, but forcing object-literals to be alphabetized might cause errors as elaborated sooner.
I honestly don't really know what to do about object-imports 😅

be5invis: @manuth Maybe we should simply set their rank to a same, super-large value...
Anyway this should be in another PR.

PR #1831, opened 2020-06-19
[patch] `order`/TypeScript: ignore ordering of object imports

manuth:
Forcing object-imports to be alphabetized might lead to issues as they might depend on each other.

***Example:***
```ts
// Object-imports
import A = ESLint;
import OutputFixes = A.outputFixes;

OutputFixes([]);
```

Currently, this piece of code will report an error because `A.LintResult` has a lower rank than `ESLint`.
Alphabetizing these imports would cause this code to fail:
```ts
// Object-imports
import OutputFixes = A.outputFixes;
import A = ESLint;

OutputFixes([]);
```
```
Cannot read property 'outputFixes' of undefined
```

I see three different approaches to fix this:
* ~~Sort object-imports with following conditions:~~
  1. ~~Number of dots (e.g. `ts` comes before `ts.protocol` and `ts.protocol` comes before `ts.protocol.Request`)~~
  2. ~~Import-Name~~
* Allow any sorting by setting names of all object-imports to `''`
* Ignore object-imports

Though I'd like the first approach the most as it definitely would work and even would fix code like the one above, the second approach is ways easier to implement which is why I just went for it.

### Remarks
The first approach won't work as, for example, this code still would cause an error:
```ts
import B = ESLint;
import A = B;
```

Therefore I'd recommend to go for the second approach as done in this PR.
Feedback is welcome - do you guys think it's possible to implement the first mentioned approach at some point?

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -11,11 +11,7 @@ const defaultGroups = ['builtin', 'external', 'parent', 'sibling', 'index']
 
 function reverse(array) {
   return array.map(function (v) {
-    return {
-      name: v.name,
-      rank: -v.rank,
-      node: v.node,
-    }
+    return Object.assign({}, v, { rank: -v.rank })
   }).reverse()
 }
 
@@ -197,8 +193,7 @@ function fixOutOfOrder(context, firstNode, secondNode, order) {
     newCode = newCode + '\n'
   }
 
-  const message = '`' + secondNode.name + '` import should occur ' + order +
-      ' import of `' + firstNode.name + '`'
+  const message = `\`${secondNode.displayName}\` import should occur ${order} import of \`${firstNode.displayName}\``
 
   if (order === 'before') {
     context.report({
@@ -270,7 +265,7 @@ function mutateRanksToAlphabetize(imported, alphabetizeOptions) {
     if (!Array.isArray(acc[importedItem.rank])) {
       acc[importedItem.rank] = []
     }
-    acc[importedItem.rank].push(importedItem.name)
+    acc[importedItem.rank].push(importedItem.value)
     return acc
   }, {})
 
@@ -295,7 +290,7 @@ function mutateRanksToAlphabetize(imported, alphabetizeOptions) {
 
   // mutate the original group-rank with alphabetized-rank
   imported.forEach(function(importedItem) {
-    importedItem.rank = alphabetizedRanks[importedItem.name]
+    importedItem.rank = alphabetizedRanks[importedItem.value]
   })
 }
 
@@ -310,31 +305,31 @@ function computePathRank(ranks, pathGroups, path, maxPosition) {
   }
 }
 
-function computeRank(context, node, ranks, name, type, excludedImportTypes) {
+function computeRank(context, ranks, importEntry, excludedImportTypes) {
   let impType
-  if (type === 'import:object') {
+  let rank
+  if (importEntry.type === 'import:object') {
     impType = 'object'
   } else {
-    impType = importType(name, context)
+    impType = importType(importEntry.value, context)
   }
-  let rank
   if (!excludedImportTypes.has(impType)) {
-    rank = computePathRank(ranks.groups, ranks.pathGroups, name, ranks.maxPosition)
+    rank = computePathRank(ranks.groups, ranks.pathGroups, importEntry.value, ranks.maxPosition)
   }
   if (typeof rank === 'undefined') {
     rank = ranks.groups[impType]
   }
-  if (type !== 'import' && !type.startsWith('import:')) {
+  if (importEntry.type !== 'import' && !importEntry.type.startsWith('import:')) {
     rank += 100
   }
 
   return rank
 }
 
-function registerNode(context, node, name, type, ranks, imported, excludedImportTypes) {
-  const rank = computeRank(context, node, ranks, name, type, excludedImportTypes)
+function registerNode(context, importEntry, ranks, imported, excludedImportTypes) {
+  const rank = computeRank(context, ranks, importEntry, excludedImportTypes)
   if (rank !== -1) {
-    imported.push({name, rank, node})
+    imported.push(Object.assign({}, importEntry, { rank }))
   }
 }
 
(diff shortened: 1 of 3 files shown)


manuth:
So I just noticed that the first approach wouldn't work out as something like
```ts
import B = ESLint;
import A = B;
```
Still would cause issues.
Sticking up with the solution of this PR (which is naming all object-imports `''`) seems to be the best way to go.

tests/src/rules/order.js
```diff
@@ -1285,25 +1299,8 @@ ruleTester.run('order', rule, {
...
-      }),
-      // Alphabetization of object-imports
-      test({
-        code: `
-          import log = console.log;
-          import debug = console.debug;`,
-        parser,
-        errors: [{
-          message: '`console.debug` import should occur before import of `console.log`',
+          message: '`./blah` import should occur before import of ``',
```
ljharb, line 1302: this isn't a very clear error message; is there a better way we can do this?

manuth: I gave it a go.
Do you think, this change is fine?

src/rules/order.js
```diff
@@ -197,8 +197,11 @@ function fixOutOfOrder(context, firstNode, secondNode, order) {
     newCode = newCode + '\n'
   }
 
-  const message = '`' + secondNode.name + '` import should occur ' + order +
-      ' import of `' + firstNode.name + '`'
+  const nameResolver = (importEntry) => importEntry.type === 'import:object' ?
+    context.getSourceCode().getText(importEntry.node.moduleReference) : importEntry.name
+
+  const message = '`' + nameResolver(secondNode) + '` import should occur ' + order +
+    ' import of `' + nameResolver(firstNode) + '`'
```
ljharb, line 204:
```suggestion
  const message = `\`${nameResolver(secondNode)}\` import should occur ${order} import of \`${nameResolver(firstNode))\``
```

src/rules/order.js
```diff
@@ -623,7 +626,7 @@ module.exports = {
           name = node.moduleReference.expression.value
           type = 'import'
         } else {
-          name = context.getSourceCode().getText(node.moduleReference)
+          name = '',
```
ljharb, line 629: should this use the `nameResolver` logic perhaps?

manuth: Not quite. The name is changed to `''` here in order to "disable" alphabetization for this kind of import.
Truth be told: This is a pretty dirty workaround, so I'm sorry 'bout this weird change 😅

Another approach would be to add a `value` and `displayName` property to the `imported`-entries in place of the `name` property.

This way we could set `value` to the... well... value to be sorted on and `displayName` the name to be shown in error-messages.

What do you think about this?

src/rules/order.js
manuth: Will be fixed after applying the change mentioned here:
https://github.com/benmosher/eslint-plugin-import/pull/1831#discussion_r443265731

src/rules/order.js
ljharb: I think that sounds more explicit and thus better :-)

src/rules/order.js
```diff
@@ -310,31 +306,32 @@ function computePathRank(ranks, pathGroups, path, maxPosition) {
...
     impType = 'object'
   } else {
-    impType = importType(name, context)
+    impType = importType(importEntry.value, context)
   }
-  let rank
+  importEntry.group = impType
   if (!excludedImportTypes.has(impType)) {
-    rank = computePathRank(ranks.groups, ranks.pathGroups, name, ranks.maxPosition)
+    rank = computePathRank(ranks.groups, ranks.pathGroups, importEntry.value, ranks.maxPosition)
```
manuth, line 319: @ljharb shal we rather pass `importEntry.displayName` to this function and allow to add path-groups for object-imports as well or leave it as it is (`importEntry.value`)?

I'd rather go for disallowing path-groups for object-imports by using `importEntry.value`.

src/rules/order.js
manuth: Alright, I fixed it now 😄

src/rules/order.js
```diff
@@ -310,31 +306,32 @@ function computePathRank(ranks, pathGroups, path, maxPosition) {
...
 }
 
-function registerNode(context, node, name, type, ranks, imported, excludedImportTypes) {
-  const rank = computeRank(context, node, ranks, name, type, excludedImportTypes)
-  if (rank !== -1) {
-    imported.push({name, rank, node})
+function registerNode(context, importEntry, ranks, imported, excludedImportTypes) {
+  importEntry.rank = computeRank(context, ranks, importEntry, excludedImportTypes)
+  if (importEntry.rank !== -1) {
+    imported.push(importEntry)
```
ljharb, line 334:
```suggestion
  const rank = computeRank(context, ranks, importEntry, excludedImportTypes)
  if (rank !== -1) {
    imported.push(Object.assign({}, importEntry, { rank }))
```

src/rules/order.js
ljharb: this seems fine as is

src/rules/order.js
```diff
@@ -310,31 +306,32 @@ function computePathRank(ranks, pathGroups, path, maxPosition) {
...
-  if (type === 'import:object') {
+  let rank
+  if (importEntry.type === 'import:object') {
     impType = 'object'
   } else {
-    impType = importType(name, context)
+    impType = importType(importEntry.value, context)
   }
-  let rank
+  importEntry.group = impType
```
ljharb, line 317: let's please avoid mutating `importEntry` here. if we need to return both rank and group, that's fine.

src/rules/order.js
```diff
@@ -197,8 +193,8 @@ function fixOutOfOrder(context, firstNode, secondNode, order) {
     newCode = newCode + '\n'
   }
 
-  const message = '`' + secondNode.name + '` import should occur ' + order +
-      ' import of `' + firstNode.name + '`'
+  const message = `\`${secondNode.displayName}\` import ` +
+    `should occur ${order} import of \`${firstNode.displayName}\``
```
ljharb, line 197:
```suggestion
  const message = `\`${secondNode.displayName}\` import should occur ${order} import of \`${firstNode.displayName}\``
```

Issue #1842, opened 2020-07-01
Add "forbid" option for import/no-internal-modules 


guillaumewuip:
I would like to use `import/no-internal-modules` to allow all imports except some rules.

For example, given

```js
import React from 'react';
import { Toolbar } from 'reakit/Toolbar';

import { thing } from '../dir2/subdir';
import { somethingElse } from '../dir2/subdir/file';
```

I would like to allow everything except `reakit/Toolbar` and `dir2/subdir/*`. 
Because this should work for the full codebase (lot of node_modules dependencies, complex directory tree), I prefer maintaining an exclusion list of what is forbidden than an inclusion list of what is allowed.

Would you accept a PR to add a `forbid` option to `import/no-internal-modules` to be used like below ? 

```
{
  ...
  "rules": {
    "import/no-internal-modules": [ "error", {
      "forbid": [ "reakit/Toolbar", "*/dir2/subdir/*" ]
    } ]
  }
}
```

ljharb:
As long as `forbid` and `allow` are mutually exclusive, that seems like a reasonable semver-minor PR.

PR #1846, opened 2020-07-04
[New] `no-internal-modules`: Add `forbid` option

guillaumewuip:
Here is the pull request that implements #1842. 

Without `forbid` option provided, the behavior stay the same. It should be a semver-minor PR.

With `forbid` option provided, the rule check if a given path is explicitly forbidden before testing if it's allowed.

### DIFF
--- a/docs/rules/no-internal-modules.md
+++ b/docs/rules/no-internal-modules.md
@@ -4,7 +4,10 @@ Use this rule to prevent importing the submodules of other modules.
 
 ## Rule Details
 
-This rule has one option, `allow` which is an array of [minimatch/glob patterns](https://github.com/isaacs/node-glob#glob-primer) patterns that whitelist paths and import statements that can be imported with reaching.
+This rule has two mutally exclusive options that are arrays of [minimatch/glob patterns](https://github.com/isaacs/node-glob#glob-primer) patterns:
+
+- `allow` that include paths and import statements that can be imported with reaching.
+- `forbid` that exclude paths and import statements that can be imported with reaching.
 
 ### Examples
 
@@ -33,7 +36,7 @@ And the .eslintrc file:
   ...
   "rules": {
     "import/no-internal-modules": [ "error", {
-      "allow": [ "**/actions/*", "source-map-support/*" ]
+      "allow": [ "**/actions/*", "source-map-support/*" ],
     } ]
   }
 }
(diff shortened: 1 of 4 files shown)


tests/src/rules/no-internal-modules.js
```diff
@@ -14,6 +14,13 @@ ruleTester.run('no-internal-modules', rule, {
       filename: testFilePath('./internal-modules/plugins/plugin.js'),
```
guillaumewuip, line 14: I've added tests, but maybe not enough. Don't hesitate to ask me to implement other tests if you see something I've missed :)

docs/rules/no-internal-modules.md
```diff
@@ -4,7 +4,10 @@ Use this rule to prevent importing the submodules of other modules.
 
 ## Rule Details
 
-This rule has one option, `allow` which is an array of [minimatch/glob patterns](https://github.com/isaacs/node-glob#glob-primer) patterns that whitelist paths and import statements that can be imported with reaching.
+This rule has two options that are arrays of [minimatch/glob patterns](https://github.com/isaacs/node-glob#glob-primer) patterns:
```
ljharb, line 7:
```suggestion
This rule has two mutually exclusive options that are arrays of [minimatch/glob](https://github.com/isaacs/node-glob#glob-primer) patterns:
```

docs/rules/no-internal-modules.md
```diff
@@ -33,7 +38,8 @@ And the .eslintrc file:
   ...
   "rules": {
     "import/no-internal-modules": [ "error", {
-      "allow": [ "**/actions/*", "source-map-support/*" ]
+      "allow": [ "**/actions/*", "source-map-support/*" ],
+      "forbid": [ "**/actions/special/*" ],
```
ljharb, line 42: should it be possible to both allow and forbid specific things? When "allow" is specified, it takes precedence over a default "forbid" of "**/*". When "forbid" is specified, it takes precedence over a default "allow" of "**/*".

I'd expect the schema to only allow one or the other, never both.

guillaumewuip: By 

> I'd expect the schema to only allow one or the other, never both.

You mean the user will either provide `allow` array or `forbid` array but never both ?

ljharb: Correct. If both are provided, we have to define a precedence between them - with `allow` of A and `forbid` of `B`, what's intended with `C`? Is it allowed, because it's not forbidden? Or is it forbidden, because it's not allowed?

guillaumewuip: Yep. I understand now! I will rework the PR to do that 👍

guillaumewuip:
@ljharb I took the time to work again on this PR. `allow` and `forbid` are now mutually exclusive. Same default behavior, should not be a breaking

guillaumewuip:
Hello @ljharb, do you need something else on this PR ?

ljharb (approved):

jenskuhrjorgensen:
Hi @ljharb

This feature seems to have been released in the docs, but it is not released in the newest [release (v.2.22.1)](https://github.com/benmosher/eslint-plugin-import/releases/tag/v2.22.1) which is pretty confusing. Is this a mistake? I was trying to use the feature as described in the docs, but it failed with the following error: 

```
Error: .eslintrc:
	Configuration for rule "import/no-internal-modules" is invalid:
	Value {"forbid":["**/actions/*","source-map-support/*"]} should NOT have additional properties.
```

Best regards
Jens

ljharb:
It's not a mistake at all. If you're looking at the docs on master, you'll see unreleased things.

I created the "github release" for v2.22.1 yesterday, but it was actually released in September (see the changelog link). This PR has not yet been released.

jenskuhrjorgensen:
Okay, thanks for clarifying. Never realized that I actually have to look in the release zip for the current documentation :)

ljharb:
Yes, on every project on github, you have to switch to the latest release _tag_ to see the current documentation.

richard-lopes-ifood:
When does it goes to a official release?

PR #1847, opened 2020-07-06
[New] `max-dependencies`: add option `ignoreTypeImports`

### LINKED ISSUE #1136: Ignoring Flow `import type` statements when using `import/max-dependencies`
Currently, Flow `import type` statements are included in the dependency count when using the `import/max-dependencies` rule. This makes the rule far less useful for identifying code smells in projects using Flow, especially if you tend to put type defintions in separate files for the rest of the code (which then need separate imports). 

I think it would be useful to have an option to ignore `import type` statements when counting depencencies. Thoughts?

### LINKED ISSUE #1843: [max-dependencies]: new option 'ignoreTypeImports'
This rule aims to notify about potential code smell or a module doing too much based on the number of imported dependencies.

It doesn't differentiate between regular imports and type imports though. Type imports are neither indicating code smell nor do they indicate that a module is doing too much.

Therefore it might be good to implement a new option allowing or disallowing the consideration of type imports.

Note: this would only influence users of Typescript and Flow.

rfermann:
Adds support for ignoring type imports.

Fixes #1843. Fixes #1136.

### DIFF
--- a/src/rules/max-dependencies.js
+++ b/src/rules/max-dependencies.js
@@ -2,15 +2,14 @@ import moduleVisitor from 'eslint-module-utils/moduleVisitor';
 import docsUrl from '../docsUrl';
 
 const DEFAULT_MAX = 10;
+const DEFAULT_IGNORE_TYPE_IMPORTS = false;
+const TYPE_IMPORT = 'type';
 
 const countDependencies = (dependencies, lastNode, context) => {
   const { max } = context.options[0] || { max: DEFAULT_MAX };
 
   if (dependencies.size > max) {
-    context.report(
-      lastNode,
-      `Maximum number of dependencies (${max}) exceeded.`
-    );
+    context.report(lastNode, `Maximum number of dependencies (${max}) exceeded.`);
   }
 };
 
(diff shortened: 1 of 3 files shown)


cherryblossom000:
This will also fix #1136, won't it?

rfermann:
@ljharb I added some tests using the TS parser. But as syntax for type import looks identical for flow and TS, I am not sure if that is what you were asking for.

src/rules/max-dependencies.js
```diff
@@ -36,9 +39,17 @@ module.exports = {
...
     return {
       ImportDeclaration(node) {
-        dependencies.add(node.source.value)
+        if (node.importKind === TYPE_IMPORT) {
+          if (!ignoreTypeImports) {
+            dependencies.add(node.source.value)
+          }
+        } else {
+          dependencies.add(node.source.value)
+        }
```
ljharb, line 52:
```suggestion
        if (node.importKind !== TYPE_IMPORT || !ignoreTypeImports) {
          dependencies.add(node.source.value)
        }
```

cherryblossom000:
Hi, I'm running into this issue a lot and keep having to put `// eslint-disable-next-line import/max-dependencies` for cases where the type imports make the number of imports exceed the configured `max` amount. It'll be nice if this PR could get a rebase and be merged.

ljharb (approved):

himynameisdave:
Would be nice to document this option [over here](https://github.com/import-js/eslint-plugin-import/blob/master/docs/rules/max-dependencies.md).

PR #1848, opened 2020-07-07
[Fix] `dynamic-import-chunkname`: allow single quotes to match Webpack support

### LINKED ISSUE #1130: Why force specific quotes in dynamic-import-chunkname?

straub:
Single quotes are supported by Webpack at least in v4.43.0, which is the
version I'm currently using. Since my eslint style prefers single 
quotes, I naturally wrote my Webpack magic comments to match, which 
worked w/o issue until I tried to enable this rule.

Implemented as a non-breaking relaxation of the existing rule to avoid 
complexity per
https://github.com/benmosher/eslint-plugin-import/issues/1130#issuecomment-464365363
but please let me know if you prefer a different approach!

Fixes #1130.

### DIFF
--- a/src/rules/dynamic-import-chunkname.js
+++ b/src/rules/dynamic-import-chunkname.js
@@ -30,8 +30,8 @@ module.exports = {
     const { webpackChunknameFormat = '[0-9a-zA-Z-_/.]+' } = config || {}
 
     const paddedCommentRegex = /^ (\S[\s\S]+\S) $/
-    const commentStyleRegex = /^( \w+: ("[^"]*"|\d+|false|true),?)+ $/
-    const chunkSubstrFormat = ` webpackChunkName: "${webpackChunknameFormat}",? `
+    const commentStyleRegex = /^( \w+: (["'][^"']*["']|\d+|false|true),?)+ $/
+    const chunkSubstrFormat = ` webpackChunkName: ["']${webpackChunknameFormat}["'],? `
     const chunkSubstrRegex = new RegExp(chunkSubstrFormat)
 
     function run(node, arg) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`default`]/TypeScript: avoid crash on `export =` with a MemberExpression ([#1841], thanks [@ljharb])
 - [`extensions`]/importType: Fix @/abc being treated as scoped module ([#1854], thanks [@3nuc])
 - allow using rest operator in named export ([#1878], thanks [@foray1010])
+- [`dynamic-import-chunkname`]: allow single quotes to match Webpack support ([#1848], thanks [@straub])
 
 ### Changed
 - [`export`]: add tests for a name collision with `export * from` ([#1704], thanks @tomprats)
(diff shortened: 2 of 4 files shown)


src/rules/dynamic-import-chunkname.js
```diff
@@ -30,8 +30,8 @@ module.exports = {
     const { webpackChunknameFormat = '[0-9a-zA-Z-_/.]+' } = config || {}
 
     const paddedCommentRegex = /^ (\S[\s\S]+\S) $/
-    const commentStyleRegex = /^( \w+: ("[^"]*"|\d+|false|true),?)+ $/
-    const chunkSubstrFormat = ` webpackChunkName: "${webpackChunknameFormat}",? `
+    const commentStyleRegex = /^( \w+: (["'][^"']*["']|\d+|false|true),?)+ $/
+    const chunkSubstrFormat = ` webpackChunkName: ["']${webpackChunknameFormat}["'],? `
```
ljharb, line 34: these don't use backreferences, so mismatched quotes would be supported. I see the test cases that suggest they *aren't* supported, but i'm not clear on how that works. can you help me understand?

straub:
> I'm concerned about webpack users prior to v4.43.0 getting false positives here.

Sorry for the confusion here, I only meant to confirm that single quotes are supported in 4.43.0 (their latest as of now), but I'm sure the support extends back farther than that since I've used single quotes for some time. It struck me as tangential to the purpose of this rule about the format of the chunkName to get too deep in the weeds of Webpack support, but I see what you're saying. If you think it's important, I can attempt to find out how far back Webpack support for single quotes goes.

src/rules/dynamic-import-chunkname.js
straub: I had the same thought at first! The gist is that mismatched quotes are invalid syntax, so they trigger the syntax error and don't need to be checked in the style regex. I can use backreferences in these patterns instead, if you prefer to have it covered both ways, just lmk :)

straub:
I ran a quick test, and the single quotes are supported in Webpack all the way back to v2.4.0, which as I understand it is the release that introduced magic comments. Given that, there wouldn't seem much value in a `"detect"` option.

Please let me know how you'd like to proceed, and thanks for your feedback so far!

ljharb:
I totally agree, thanks for doing the research.

src/rules/dynamic-import-chunkname.js
ljharb: It's inside a comment tho, so I'm not sure why it'd be invalid syntax?

straub: The comment value is parsed as the interior of a JS object literal, in order to validate the syntax: https://github.com/benmosher/eslint-plugin-import/blob/274b252138aa6c268db9084875f024b45c4eb9c5/src/rules/dynamic-import-chunkname.js#L72

straub:
@ljharb Did you still want me to pursue adding a `true`/`false` option, or any other changes?

ljharb:
@straub sorry for the delay :-) let me take a fresh look now.

src/rules/dynamic-import-chunkname.js
ljharb: I'd prefer the backreferences, but it doesn't need to block the PR.

PR #1854, opened 2020-07-12
Fix @/abc being treated as scoped module

### LINKED ISSUE #1851: Webpack alias not working
.eslintrc.js
```js
const path = require('path')

module.exports = {
  env: {
    node: true,
  },
  extends: ['plugin:vue/essential', 'eslint:recommended', '@vue/prettier'],
  parserOptions: {
    parser: 'babel-eslint',
  },
  plugins: ['import'],
  root: true,
  rules: {
    'import/exports-last': 'warn',
    'import/extensions': ['warn', 'ignorePackages'],
    'import/first': 'warn',
    'import/newline-after-import': 'warn',
    'import/no-duplicates': 'error',
    'import/no-relative-parent-imports': 'warn',
  },
  settings: {
    'import/resolver': {
      webpack: {
        config: {
          resolve: {
            alias: {
              '@': path.resolve('src'),
            },
          },
        },
      },
    },
  },
}
```
package.json
```
    "eslint-import-resolver-webpack": "^0.12.2",
    "eslint-plugin-import": "^2.22.0",
```
Result
image

3nuc:
Fixes #1851

Before this commit, @/abc was being wrongly detected as a scoped module.
This was a problem when somebody had a webpack alias `@`. (eg. @ -> ./src)

In general, scoped modules can't start with @/, I think they have to be like @abcd/sometext.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Fixed
 - [`default`]/TypeScript: avoid crash on `export =` with a MemberExpression ([#1841], thanks [@ljharb])
+- [`extensions`]/importType: Fix @/abc being treated as scoped module ([#1854], thanks [@3nuc])
 
 ## [2.22.0] - 2020-06-26
 ### Added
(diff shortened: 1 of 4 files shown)


3nuc:
This function is used only once in the whole codebase, but I guess it could be made less fragile by adding things like:
* isScopedModule('@abc/') should return false (ie. check if there's content after the slash)

wenfangdu:
@3nuc You are awesome!

wenfangdu:
> Need a release with this fix

@benmosher

PR #1859, opened 2020-07-20
[New] match-default-export-name rule

dangreen:
[TSLint rule](https://palantir.github.io/tslint/rules/match-default-export-name/) implementation.

### DIFF
--- a/docs/rules/match-default-export-name.md
+++ b/docs/rules/match-default-export-name.md
@@ -1,0 +1,35 @@
+# import/match-default-export-name
+
+Reports use of import name what doesn't match the default export name. Does nothing for anonymous default exports.
+
+## Rule Details
+
+Given:
+```js
+// foo.js
+export default function myFunction() {}
+```
+
+...this would be valid:
+```js
+import myFunction from './foo.js';
+```
+
+...and this would be reported:
+```js
+// message: Expected import 'myFn' to match the default export 'myFunction'.
+import myFn from './foo.js';
+```
+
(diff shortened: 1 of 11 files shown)


dangreen:
@ljharb review please 🙏

dangreen:
@benmosher ping?

dangreen:
@ljharb @benmosher hellooo?

docs/rules/match-default-export-name.md
```diff
@@ -0,0 +1,77 @@
...
+```json
+{
+  "import/match-default-export-name": ["error", {
+    "overrides": [
+      {
+        "module": "react",
+        "name": "React"
+      },
+      {
+        "module": "/styles\\.css$/",
```
ljharb, line 66: i'm not comfortable supplying regex in configs; that leads to CVEs. Let's change this to use globs, instead.

docs/rules/match-default-export-name.md
```diff
@@ -0,0 +1,77 @@
...
+        "name": "styles"
+      },
+      {
+        "module": "/([\\w-]+)\\.module\\.css$/",
+        "name": "$1Styles",
+        "transform": "camelCase"
```
ljharb, line 72: can you elaborate on this? programmatic case conversions are typically very unreliable.

ljharb (changes requested):
What if someone wants to enforce this only on imports, or only on exports? I think we need an option that lets someone select `imports`, `exports`, or both.

PR #1860, opened 2020-07-23
[New]: add `no-relative-packages`

tapayne88:
Duplicating #966 and hopefully fixed it 🤞 (sorry if this isn't the process).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Added
 - [`no-commonjs`]: Also detect require calls with expressionless template literals: ``` require(`x`) ``` ([#1958], thanks [@FloEdelmann])
 - [`no-internal-modules`]: Add `forbid` option ([#1846], thanks [@guillaumewuip])
+- add [`no-relative-packages`] ([#1860], [#966], thanks [@tapayne88] [@panrafal])
 
 ### Fixed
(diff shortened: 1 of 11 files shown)


tapayne88:
@ljharb is there anything I can do to get this over the line?

wycats:
@tapayne88 I tried this out from your branch and it improved my workflow dramatically.

One recommendation: right now, the warning suggests importing from `package-name/index` instead of `package-name`. It seems like that ought not happen.

One wrinkle: in my case, I am not importing from the same location as the `main` entry point (which is `dist/index.js`) because the mistaken import is accidentally the `index.ts` in the root of the package. I have `types` pointing at `index.ts`, but that is probably not a perfect heuristic. In a situation where the import is pointing at an `index.ts` file, perhaps it's just fine to suggest `package-name`?

tapayne88:
hey @wycats, thanks for the feedback! To be honest, the bulk of the work was originally done by @panrafal but I'll try set aside some time to look into your recommendation - it does sound closer to what we want.

If I'm understanding your wrinkle correctly, I _think_ node module resolution should use the `"main"` field in the `package.json` first (if available) and resort to the `index.js`/`index.ts` second before heading up the filesystem in search of more `node_modules` folders so I think what you're seeing is the desired behaviour.

wycats:
@tapayne88 If it finds `index.ts` in the absence of a **types** field (even in the presence of **main** field), I think that would be ok.

wycats:
@ljharb what do you think of this PR?

ljharb (approved):
Looks great! Went ahead and rebased, and refactored to use the new `moduleVisitor` pattern.

Issue #1883, opened 2020-08-17
`export * as ns` false positive on import/named rule


mAAdhaTTah:
```js
// module1.js
export const hello = 'world';

// module2.js
export * as mod1 from './module1'

// module3.js
import { mod1 } from './module2'; // <-- getting "mod1 not found", triggered by `import/named` rule.
```

I can create a repro case if needed.

ljharb:
Thanks for the report! If you're not able to PR a fix, a PR with the failing test case would be stellar <3

mAAdhaTTah:
#1885 for failing tests. I could probs fix with some guidance.

rkulinski:
Still an issue :(

ljharb:
This `import/named` false positive was fixed back in v2.24.1 by 3ff4d772 (`[Fix] named, namespace: properly set reexports on export * as … from`) - the re-exported name is registered correctly now. Closing as resolved; please reopen with a repro if you still hit it on a current version.

PR #1885, opened 2020-08-18
Add failing test for export star as bug

mAAdhaTTah:
Related to #1883.

@ljharb I could probably dig into fixing this but first: I'm getting a parse error rather than the named import error I was expecting and it's entirely clear to me why or what I would need to adjust to fix. In the project where I experienced this, we have `"ecmaVersion": 2018`, which I think is the only relevant piece here? Not sure. Thoughts?


ljharb:
When I rebased and added babel-eslint as the parser (since it had a parsing error otherwise), it appears to pass.

mAAdhaTTah:
Hmm I'm definitely running the latest version of the plugin & v7 of ESLint & am still getting this error in my project. Not really sure what's going on.

ljharb:
@mAAdhaTTah do you have a lockfile? perhaps you have some older versions of transitive deps

mAAdhaTTah:
I do, I ran `yarn upgrade` and reran my linter and I'm still getting the error :/ . I'll try to get an isolated repro going. I will mention this is a re-export through a barrel, meaning I'm importing from an `index.js` file in a folder, although that doesn't seem to affect the plugin's ability to "see" the other exports so not sure that's relevant.

ljharb:
@mAAdhaTTah any update on a repro?

mAAdhaTTah:
@ljharb Actually, I think I was able to reproduce this on a side project. Let me look into that.

mAAdhaTTah:
@ljharb This PR on my personal site reproduces the issue: https://github.com/mAAdhaTTah/frontend/pull/201 I also included in the PR screenshots of the versions it's running when I see the issue.

ljharb:
well, at least one of the test cases is failing - altho oddly enough, it's the one using the default parser, and you're using babel-eslint. It's possible that if you update babel and/or babel-eslint that it will pass for you.

Either way, we now have failing tests, so now we can fix it :-) I suspect somewhere in ExportMap is the place, but I'm not 100% sure.

mAAdhaTTah:
Ok cool, I'll give that a shot and report back.

PR #1889, opened 2020-08-20
[Fix] `export`/TypeScript: properly detect export specifiers as children of a TS module block

### LINKED ISSUE #1773: [import/export] "Multiple default exports" in Typescript ambient module declarations
If a Typescript file (or, more likely, a `.d.ts` file) has several ambient module declarations with default exports, sometimes they're mistakenly reported as "multiple default exports".

```TS
declare module "a" {
  function a(): void;
  export {a as default};
}

declare module "b" {
  function b(): void;
  export {b as default};
}
```

```YAML
plugins:
  - import
rules:
  import/export: error
```

Here's a case where that doesn't happen, though:

```TS
declare module "a" {
  export default function a(): void;
}

declare module "b" {
  export default function b(): void;
}
```

andreubotella:
Fixes #1773

### DIFF
--- a/src/rules/export.js
+++ b/src/rules/export.js
@@ -87,7 +87,11 @@ module.exports = {
     return {
       'ExportDefaultDeclaration': (node) => addNamed('default', node, getParent(node)),
 
-      'ExportSpecifier': (node) => addNamed(node.exported.name, node.exported, getParent(node)),
+      'ExportSpecifier': (node) => addNamed(
+        node.exported.name,
+        node.exported,
+        getParent(node.parent)
+      ),
 
       'ExportNamedDeclaration': function (node) {
         if (node.declaration == null) return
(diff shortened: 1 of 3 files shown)


src/rules/export.js
```diff
@@ -87,7 +87,11 @@ module.exports = {
     return {
       'ExportDefaultDeclaration': (node) => addNamed('default', node, getParent(node)),
 
-      'ExportSpecifier': (node) => addNamed(node.exported.name, node.exported, getParent(node)),
+      'ExportSpecifier': (node) => addNamed(
+        node.exported.name,
+        node.exported,
+        getParent(node.parent)
+      ),
```
ljharb, line 95: i'm a bit confused; all this is doing is reformatting. do these tests pass on master?

andreubotella: It changes `getParent(node)` to `getParent(node.parent)`. That would've been more obvious if I hadn't reformatted.

andreubotella: Bump

ljharb (approved):

PR #1900, opened 2020-09-15
[no-restricted-paths] Add options to ignore certain import kinds

RebeccaStevens:
This adds a new option called `"allowedImportKinds"` to the rule `no-restricted-paths`.
This option defaults to an empty array and can be set to an array of strings.

This option allows imports of the specified kind(s) to be ignore by this rule.

Example use:

```jsonc
{
  // Allow type imports.
  "allowedImportKinds": ["type"],
  "zones": [ {
    "target": "./tests/files/restricted-paths/server/one",
    "from": "./tests/files/restricted-paths/server",
    // Override above `allowedImportKinds`, don't make an exception for any import kinds.
    "allowedImportKinds": [],
  } ]
}
```


RebeccaStevens:
@ljharb 

> I'm not sure if we need an enum - will there ever be another kind of import that's not value or type?

In this PR I haven't enforced the option being an enum, any strings can be given. With regard to whether there will be any other kinda of import, most likely not - but it's probably best to keep things flexible just in case.

RebeccaStevens:
@ljharb Think you can help me resolve the failing build? It's failing in older versions of eslint as `node.parent` is `undefined`.

src/rules/no-restricted-paths.js
```diff
@@ -6,6 +6,14 @@ import isStaticRequire from '../core/staticRequire'
...
 import importType from '../core/importType'
 
+const allowedImportKindsSchema = {
+  type: 'array',
+  items: {
+    type: 'string',
+  },
+  uniqueItems: true,
+}
+
```
ljharb, line 16: i do think we should restrict this for now to known import types. we can extend it in the future if needed, but we can't lock it down so easily.

ljharb (changes requested):
LGTM pending schema change.

PR #1906, opened 2020-09-22
`no-unused-modules`: ignore flow type exports

nicolashenry:
Fixes https://github.com/benmosher/eslint-plugin-import/issues/1564 :
- Ignore Flow type exports
- Add tests for Flow type imports/exports
- Remove Typescript tests using "babel-eslint" (because resulting types seems to be Flow types)
- Add some missing cases in Typescript tests
- Add some tests for Typescript type imports

### DIFF
--- a/tests/files/no-unused-modules/flow/flow-1.js
+++ b/tests/files/no-unused-modules/flow/flow-1.js
@@ -1,0 +1,3 @@
+// @flow strict
+export type Bar = number;
+export interface BarInterface {};
--- a/tests/src/rules/no-unused-modules.js
+++ b/tests/src/rules/no-unused-modules.js
@@ -744,57 +744,88 @@ describe('Avoid errors if re-export all from umd compiled library', () => {
   })
 })
 
-describe('correctly work with Typescript only files', () => {
-  typescriptRuleTester.run('no-unused-modules', rule, {
-    valid: [
-      test({
-        options: unusedExportsTypescriptOptions,
-        code: 'import a from "file-ts-a";',
-        parser: require.resolve('babel-eslint'),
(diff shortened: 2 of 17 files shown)


tests/src/rules/no-unused-modules.js
```diff
@@ -744,65 +744,96 @@ describe('Avoid errors if re-export all from umd compiled library', () => {
...
+          filename: testFilePath('./no-unused-modules/typescript/file-ts-e-3.ts'),
+        }),
       ],
       invalid: [
         test({
           options: unusedExportsTypescriptOptions,
           code: `export const b = 2;`,
           parser: parser,
-          filename: testFilePath('./no-unused-modules/typescript/file-ts-b.ts'),
+          filename: testFilePath('./no-unused-modules/typescript/file-ts-b-2.ts'),
```
ljharb, line 836: is there a way we could rework these tests so we don't have to change these existing ones, and so "b-2" could have a more meaningful name for the new tests?

tests/files/no-unused-modules/flow/flow-1.js
```diff
@@ -0,0 +1,3 @@
+// @flow strict
+export type Bar = number;
+export interface BarInterface {};
```
ljharb, line 3: please ensure every file has a trailing newline

ljharb (commented):
LGTM overall, just a few comments!

tests/src/rules/no-unused-modules.js
nicolashenry: I needed to split them into files exporting unused exports (the files ending with "-2") and the files exporting used exports (the files without "-2") because the existing tests was only actually testing the unused case. I will rename those with better names.

ljharb (approved):

PR #1916, opened 2020-10-03
[WIP] add allow/deny list options to no-namespace rule

### LINKED ISSUE #1903: [no-namespace] allow namespace imports for specific modules
I want to disallow using namespaces as a rule, but some third-party libraries don't use default exports. The specific one that we use quite a bit in our application is [yup](https://github.com/jquense/yup). Even if we use TypeScript's `allowSyntheticDefaultImports` rule, we still have to import the library using `import * as yup from 'yup';`.

I'd like to be able to do something like the following:
```
rules: {
    // ...
    'import/no-namespace': ['warn', {allow: ['yup', 'some-lib', '@scoped/lib']}],
    // ...
},
```

With the above config, the following would be considered correct:
```
import * as yup from 'yup';
import * as validation from 'yup';
import * as lib from 'some-lib';
```

And the following would be considered incorrect:
```
import * as nope from 'not-allowed';

// even though this contains the string "some-lib", it should not be valid; I think the allow list should be exact matches only
import * as stillNope from '@scoped/some-lib';
```

I came across pull request #1679 from earlier this year, but it appears it was closed without being merged.

chimericdream:
Closes #1903 

This is a starting point for adding this option. Unfortunately, I cannot get the tests to run on my local machine, so I am looking for help getting this tested so it can be merged.

Edit: it's worth noting that this is based on PR #1679 from earlier this year, with the primary difference being that this version of the option checks the namespace identifier rather than the variable name.

### DIFF
--- a/src/rules/no-namespace.js
+++ b/src/rules/no-namespace.js
@@ -17,19 +17,49 @@ module.exports = {
       url: docsUrl('no-namespace'),
     },
     fixable: 'code',
-    schema: [],
+    schema: [
+      {
+        oneOf: [
+          {
+            type: 'object',
+            properties: {
+              allowList: {
+                type: 'array',
+              },
+            },
+            additionalProperties: false,
+          },
+          {
+            type: 'object',
+            properties: {
+              denyList: {
+                type: 'array',
+              },
+            },
+            additionalProperties: false,
+          },
+        ],
+      },
+    ],
   },
 
   create: function (context) {
     return {
       'ImportNamespaceSpecifier': function (node) {
+        const options = context.options[0] || {};
+        const [mode, identifiers] = Array.isArray(options.denyList)
+          ? ['deny', options.denyList]
+          : ['allow', options.allowList || []];
+
         const scopeVariables = context.getScope().variables;
-        const namespaceVariable = scopeVariables.find((variable) =>
-          variable.defs[0].node === node
-        );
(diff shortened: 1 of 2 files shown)


src/rules/no-namespace.js
```diff
@@ -17,12 +17,40 @@ module.exports = {
...
     fixable: 'code',
-    schema: [],
+    schema: [
+      {
+        oneOf: [
+          {
+            type: 'object',
+            properties: {
+              allowList: {
+                type: 'array',
```
ljharb, line 27: let's make sure both of these arrays must have unique items.

src/rules/no-namespace.js
```diff
@@ -87,6 +119,20 @@ module.exports = {
...
+ * @param {string[]} list
+ * @param {TSESTree.Identifier[]} namespaceIdentifiers
+ */
+function shouldReport(mode, list, identifiers) {
+  for (const identifier of identifiers) {
+    if (list.includes(identifier)) {
+      return mode === 'deny'
+    }
+  }
+}
```
ljharb, line 134:
```suggestion
  return mode === 'deny' && identifiers.some((identifier) => list.includes(identifier));
}
```

ljharb (changes requested):

ljharb:
I rebased this; tests are still failing so it still needs some work.

Issue #1918, opened 2020-10-05
no-extraneous-dependencies does not report when importing from subfolder


pkuczynski:
My eslint settings:

```
'import/no-extraneous-dependencies': ['error', { devDependencies: ['**/*.test.js'] }],
```

and package.json:

```
    "devDependencies": {
        ...
        "uuid": "^8.3.0",
    }
```

The following import does not error:

```
const v5 = require('uuid/v5')
```

However it does error, when using exact package name:

```
const { v5 } = require('uuid')
```

Its quite common practice, so I believe it should complain also when referencing individual files from a package...

pkuczynski:
It looks like the issue is caused by `resolve(name, context)` returning `undefined` (https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-extraneous-dependencies.js#L122).

pkuczynski:
I created some tests illustrating the issue https://github.com/benmosher/eslint-plugin-import/pull/1919. Maybe they would help to solve it...

pierceray:
There is a similar failure for namespaced packages.  `import { MockedProvider } from '@apollo/client/testing';`

PR #1919, opened 2020-10-05
Tests example illustrating #1918

pkuczynski:
Added some tests illustrating #1918 issue...


ljharb:
@pkuczynski tests seem to be passing?

pkuczynski:
That's really strange. In my project [first case](https://github.com/benmosher/eslint-plugin-import/pull/1919/files#diff-05140578b7a943405919d8a058934d1dR326) was reported correctly. However the [second case](https://github.com/benmosher/eslint-plugin-import/pull/1919/files#diff-05140578b7a943405919d8a058934d1dR340) (with subfolder) was not. That leads to an error in production and that's why I fired-up #1918 and this example.

Is there anything wrong with this test? Should I fire up a sample project to better illustrate the issue?

ljharb:
@pkuczynski I've rebased this, but assuming tests still pass, then yes - a repro repo would be most helpful.

pkuczynski:
Sure! Here comes the repo which replicates my setup:

https://github.com/pkuczynski/eslint-plugin-import-1919-demo

I am not getting any eslint error for `const v5uuid = require('uuid/v5')`, while its correctly reporting for `const { v5 } = require('uuid')`

image

```
$ eslint --ext .ts --ext .js .

/eslint-plugin-import-1919-demo/src/index.js
  2:24  error  'uuid' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies

```

pkuczynski:
Is this helpful for you now?

pkuczynski:
@ljharb is that helpful?

ljharb:
Thanks, this is quite helpful. Tests are breaking right now due to npm 7's promotion to latest, but I'll take a look once that's resolved.

pkuczynski:
Cool, thanks! I noticed that tests are not failing due to the issue I reported. You can also have a look at my demo repo directly to observe the issue... Maybe that's due to the fact I am using `@typescript-eslint/parser` parser, and that's why the tests provided in this PR worked fine?

ljharb:
Typically we run each test case in all of the default parser, babel-eslint, the old TS parser, and the new TS parser, and that's a change I can make once i've got master passing again.

pkuczynski:
Perfect! Let me know if I can help any further...

ljharb:
@pkuczynski I've rebased this; tests continue to pass.

pkuczynski:
Dod you checked the example repo I provided? https://github.com/pkuczynski/eslint-plugin-import-1919-demo

ljharb:
I think this is perhaps because the uuid package doesn't actually have a `v5.js` file in it - in other words, it's using the `exports` field to point to `wrapper.mjs`, and `resolve`, and thus this package, doesn't yet support "exports" at all. Thus, it doesn't resolve properly inside the rule, and the rule fails to warn on it.

This would be fixed if `uuid`, for backwards compatibility, added a `v5.js` file in the root of the package (cc @bcoe).

I think, however, that there may in fact be two bugs related to this in this plugin:
1. `no-unresolved` should be warning on `uuid/v5` (i checked; it looks like the issue is that it's not warning on CJS at all)
1. `no-extraneous-dependencies` should be looking at the package name, `uuid`, regardless of whether there's subpaths there or not.

Fixing the first will immediately start warning on any CJS package required deeply that, like UUID, uses "exports" but doesn't provide pre-exports back compat entry points. That's the correct behavior, but will be quite noisy prior to "exports" support landing.

Fixing the second might be tricky for when it's not actually a node_modules path.

PR #1925, opened 2020-10-14
[Tests] `extensions`: add cases to verify `@/configs/chart` is treated as a package

### LINKED ISSUE #1851: Webpack alias not working

wenfangdu:
Fixes #1851

Before this fix, `'import/extensions': ['warn', 'ignorePackages']` wouldn't warn `import chart from '@/configs/chart'`, this is an improvement over PR #1854.

### DIFF
--- a/tests/src/rules/extensions.js
+++ b/tests/src/rules/extensions.js
@@ -339,6 +339,7 @@ ruleTester.run('extensions', rule, {
         import Component from './Component'
         import baz from 'foo/baz'
         import baw from '@scoped/baw/import'
+        import chart from '@/configs/chart'
         import express from 'express'
       `,
       options: [ 'always', { ignorePackages: true } ],
@@ -358,6 +359,7 @@ ruleTester.run('extensions', rule, {
         import Component from './Component'
         import baz from 'foo/baz'
         import baw from '@scoped/baw/import'
+        import chart from '@/configs/chart'
         import express from 'express'
       `,
       options: [ 'ignorePackages' ],
@@ -503,12 +505,19 @@ ruleTester.run('extensions', rule, {
       ],
     }),
     test({
-      code: 'import foo from "@/ImNotAScopedModule"',
+      code: `
+        import foo from "@/ImNotAScopedModule";
+        import chart from '@/configs/chart';
+      `,
       options: ['always'],
       errors: [
         {
           message: 'Missing file extension for "@/ImNotAScopedModule"',
-          line: 1,
+          line: 2,
+        },
+        {
+          message: 'Missing file extension for "@/configs/chart"',
+          line: 3,
         },
       ],
     }),


tests/src/rules/extensions.js
```diff
@@ -367,6 +368,11 @@ ruleTester.run('extensions', rule, {
           line: 4,
           column: 31,
         },
+        {
+          message: 'Missing file extension for "@/scoped/bav/import"',
+          line: 7,
+          column: 25,
+        },
```
ljharb, line 375: I'd expect `ignorePackages` to suppress this warning, since this is a package.

wenfangdu: I've edited the case, thanks.

tests/src/rules/extensions.js
```diff
@@ -367,6 +368,11 @@ ruleTester.run('extensions', rule, {
           line: 4,
           column: 31,
         },
+        {
+          message: 'Missing file extension for "@/configs/chart"',
```
ljharb, line 372: again tho, this should never have a warning when `ignorePackages` is set, because it's a bare idenfier - a package.

wenfangdu: According to the [scoped packages doc](https://docs.npmjs.com/using-npm/scope.html), `@somescope/somepackagename` is a scoped package, whereas `@/foo/bar` is not.
Publishing `@/foo/bar` as an unscoped package is not allowed, so it's neither a scoped package nor an unscoped package.
image

ljharb: I agree that it isn't "a scoped package", but since it doesn't start with `.` or `/`, it is a "bare identifier", which is always "a package" - packages aren't just "publishable packages", they're "anything with a bare identifier".

ljharb: Circling back on this - `@` is a package here (just not a publishable one), so I think this should be ignored when `ignorePackages` is enabled.

ljharb (approved):
I've rebased this to only include updated test cases, since this is already working as intended.

PR #1940, opened 2020-11-10
[Fix] `order`: ignore requires in template literal

### LINKED ISSUE #1936: import/order complains about inline requires
Rule config:

```js
    'import/order': [
      'error',
      {
        groups: ['builtin', 'external', 'parent', 'sibling', 'index'],
        'newlines-between': 'never',
      },
    ],
```

Code:

```js
import React from 'react';
import PropTypes from 'prop-types';
import { createGlobalStyle, ThemeProvider } from 'styled-components';

export const GlobalStyle = createGlobalStyle`
  @font-face {
    font-display: swap;
    font-family: 'font-name';
    src: url(${require('../fonts/font-name.eot')});
    src: url(${require('../fonts/font-name.woff2')}) format('woff2'),
      url(${require('../fonts/font-name.woff')}) format('woff'),
      url(${require('../fonts/font-name.ttf')}) format('truetype');
  }
`
```

It complains about the newline after the `styled-components` import as well as a newline after the last inline require. Not 100% sure this is a bug, admittedly, but I thought it would excuse inline requires that aren't assigned based on [this](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#pass). Happy to just disable this rule on this file if needed.

golopot:
Fixes #1936.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ### Fixed
 - [`export`]/TypeScript: properly detect export specifiers as children of a TS module block ([#1889], thanks [@andreubotella])
+- [`order`]: ignore non-module-level requires ([#1940], thanks [@golopot])
 
 ## [2.22.1] - 2020-09-27
 ### Fixed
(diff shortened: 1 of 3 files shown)


ljharb:
Thanks! It seems like a more general fix is in order; the only requires that this rule should warn on are module-level requires, whether in a template literal or not.

golopot:
It sounds great to only warn on module-level requires! Two cases are failing though:

```js
const bar = require('foo').bar
```
```js
const foo = require('foo')()
```

I add some handling for these two cases. If just for backward compatibility.

ljharb (approved):
Nice.

ljharb:
Those are module-level requires, even tho they're immediately destructured. Thanks for handling them!

mAAdhaTTah:
@golopot Thank you for fixing!

PR #1941, opened 2020-11-10
[New] Add no-indexes Rule

nickcherry:
As per the documentation:

This rule prevents files from being named `index`.

For example, with the `no-indexes` rule enabled, files named `index.js`, `index.jsx`, `index.ts`, and `index.tsx` would all be considered invalid.

## Ignore

The rule accepts an optional `ignore` setting, which can be used to ignore any file whose path matches one of the given regular expression patterns.

For example, the following would allow `index` files in the `allowed` directory:

```javascript
{
  rules: {
    'import/no-indexes': [
      'error',
      { ignore: ['\/allowed\/'] }
    ]
  }
}
```

### DIFF
--- a/docs/rules/no-indexes.md
+++ b/docs/rules/no-indexes.md
@@ -1,0 +1,22 @@
+# import/no-indexes
+
+This rule prevents files from being named `index`.
+
+For example, with the `no-indexes` rule enabled, files named `index.js`, `index.jsx`, `index.ts`, and `index.tsx` would all be considered invalid.
+
+## Ignore
+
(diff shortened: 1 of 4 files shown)


ljharb (changes requested):
Why would one want this?

Being able to transparently refactor between `foo.js` and `foo/index.js` is one of the best parts of node, and it's a best practice to use `index` files to represent what a directory "is".

ljharb:
Per discussion; this would be amazing as a "no manifest exports" rule, unrelated to the filename.

PR #1948, opened 2020-11-18
[patch] `no-extraneous-dependencies`: Add package.json cache

fa93hws:
This rule is super slow in our code base, and it seems it invoke one IO request for every single file..


fa93hws:
image
WSL failures seem unrelated?

ljharb (approved):
Seems reasonable, thanks!

PR #1954, opened 2020-11-24
[Improvement]  more compatible when user pass wrong paths

Edge00:
Scenario:
```javascript
  'error',
  {
    packageDir: [
      __dirname,
      'node_modules/module-exist',
      'node_modules/module-not-exist',
    ],
  },
]
```
If there is a module `module-not-exist` that's not installed in node_modules. Even if `module-exist` did add some dependencies. The lint won't apply dependencies in `module-exist`.  

I think this rule can be more compatible. If `packageContent` have add some dependencies. These dependencies can be applied.

Like our team is maintaining a shared `.eslintrc` file and some dependencies is not required in particular projects. We hope that users don't need to change the `.eslintrc` file when they add a new dependency.


src/rules/no-extraneous-dependencies.js
```diff
@@ -31,15 +31,24 @@ function extractDepFields(pkg) {
...
+    bundledDependencies: [],
+  }
+
+  const haveAddedAnyDependencies = () => [
+    packageContent.dependencies,
+    packageContent.devDependencies,
+    packageContent.optionalDependencies,
+    packageContent.peerDependencies,
+    packageContent.bundledDependencies,
+  ].some(hasKeys)
```
ljharb, line 49:
```suggestion
  const haveAddedAnyDependencies = [
    packageContent.dependencies,
    packageContent.devDependencies,
    packageContent.optionalDependencies,
    packageContent.peerDependencies,
    packageContent.bundledDependencies,
  ].some(hasKeys);
```
no need to make this a function, it'll be the same result both times.

src/rules/no-extraneous-dependencies.js
```diff
@@ -73,19 +82,18 @@ function getDependencies(context, packageDir) {
...
     }
 
-    if (![
-      packageContent.dependencies,
-      packageContent.devDependencies,
-      packageContent.optionalDependencies,
-      packageContent.peerDependencies,
-      packageContent.bundledDependencies,
-    ].some(hasKeys)) {
+    if (!haveAddedAnyDependencies()) {
```
ljharb, line 85:
```suggestion
    if (!haveAddedAnyDependencies) {
```

src/rules/no-extraneous-dependencies.js
```diff
@@ -73,19 +82,18 @@ function getDependencies(context, packageDir) {
...
       return null
     }
 
     return packageContent
   } catch (e) {
     if (paths.length > 0 && e.code === 'ENOENT') {
+      if (haveAddedAnyDependencies()) {
```
ljharb, line 92:
```suggestion
      if (haveAddedAnyDependencies) {
```

src/rules/no-extraneous-dependencies.js
```diff
@@ -73,19 +82,18 @@ function getDependencies(context, packageDir) {
...
+        console.warn('[no-extraneous-dependencies] Some packageDir paths not exist.')
```
ljharb, line 93: what paths don't exist? can we tell them?

can we use `context.report` here, so it's a linting error?

ljharb (changes requested):

PR #1970, opened 2021-01-05
fix: fix wrong require resolver with relative path in monorepo

zthxxx:
- add support require relative resolver file in monorepo, for simplify local debugging
- add docs for cases why need to tryRequire

---

The wrong is when setting a relative path for resolver file, like this

```yaml
settings:
  import/resolver:
    './utils/simple-custom-resolver.js`:
      ... # config
```

then the files tree like:

```
/project/
├── src/
├── utils/
├── packages/
│   ├── monorepo-a/
│   └── monorepo-b/
└─ ...
```

but the `eslint-plugin-import` use `eslint-module-utils` to require resolver like

```js
/project/utils/simple-custom-resolver.js  // in project root files
/project/packages/monorepo-a/simple-custom-resolver.js  // in monorepo-a files
/project/packages/monorepo-b/simple-custom-resolver.js  // in monorepo-b files
```

and this solution fix this case by append require absolute path `/project/utils/simple-custom-resolver.js` in both all of monorepo


zthxxx:
@benmosher

ljharb:
There's no need to ping individual people; if someone doesn't notice the PR, they won't notice your ping either.

ljharb (changes requested):
First, this would need some tests.

However, it seems like this just papers over the problem - based on the location of an eslint config file, the intention of the resolver specifier should be pretty unambiguous. In a subfolder, the resolver specifier should still be resolved relative to the eslint config file that defines it, ideally.

zthxxx:
@ljharb

1. yes, I can add some tests

2. at the same time this is compatible your `ideally case`, because this is a `append fallback`, and the `ideally case` will hit the previous rule

this PR want to fix only this case, not ideally case:

image

ljharb:
Right - I’m saying the solution perhaps is that eslint-module-utils should be resolving “./simple-custom-resolver.js” from the eslint config’s dir, not from each individual project’s dir.

ljharb:
Yea, that’s a known problem. If you can make a separate PR to fix it, that’d be amazing.

zthxxx:
> Right - I’m saying the solution perhaps is that eslint-module-utils should be resolving “./simple-custom-resolver.js” from the eslint config’s dir, not from each individual project’s dir.

@ljharb  emmm, maybe only one case, I guess, use different version of the same name package in monorepo, like:

image

But anyway I can't fully understand the original intention here the code why need `getBaseDir()` method, but it obviously serves for monorepo, (at least it looks like)

so I supplemented some comments in code (as far as I guess)

zthxxx:
> Right - I’m saying the solution perhaps is that eslint-module-utils should be resolving “./simple-custom-resolver.js” from the eslint config’s dir, not from each individual project’s dir.

Back to this sentence, you means remove `getBaseDir(sourceFile)`, and only need `path.resolve(name)` ?

That's the first amendment to me, and I prefer it.

zthxxx:
Considering maybe some people use the original bug as feature, I keep the code of use `getBaseDir()`

zthxxx:
@ljharb   Considering maybe some people use the original bug as feature, I keep the code of use getBaseDir()

ljharb:
It's worth noting that the docs very clearly indicate that you have to use a JS config and `path.resolve` for a relative file: https://github.com/benmosher/eslint-plugin-import#resolvers

zthxxx:
> It's worth noting that the docs very clearly indicate that you have to use a JS config and `path.resolve` for a relative file: https://github.com/benmosher/eslint-plugin-import#resolvers

@ljharb this PR comment very clearly indicate that we use **yaml** config

zthxxx:
@ljharb  so you can interpret this as make `import/resolver` compatible with **yaml** config

zthxxx:
Should we DO NOT too arrogant to command users to use js config?

zthxxx:
@ljharb  Only relative file is not compatible with yaml, that's discrimination

ljharb:
There's no need to react so strongly. There are always some features that only work with a JS config, and it's possible this is one of them. It's not "discrimination", it's just the limits of static formats.

I think the difficulty is that eslint itself doesn't preserve the location of the eslint config file where the path was defined. So, I suspect there isn't any reliable solution for a JSON (the most common and recommended format) or a yaml config. What you've got here is a heuristics-based approach that I suspect will cause more bugs than it will solve.

Note that you can also make your resolver into a *package* - and then it will work in all three config formats.

zthxxx:
> Right - I’m saying the solution perhaps is that eslint-module-utils should be resolving “./simple-custom-resolver.js” from the eslint config’s dir, not from each individual project’s dir.

@ljharb  I remove `getBaseDir()` method, and updated

the currently tests contained `requireResolver`, so no need add more new tests

zthxxx:
@ljharb  What's next, dude?

zthxxx:
@ljharb

utils/resolve.js
```diff
@@ -174,14 +172,26 @@ function resolverReducer(resolvers, map) {
...
+     * normal abbreviation for third-party package
+     * like config `webpack` convert to `eslint-import-resolver-webpack`
+     */
+    tryRequire(`eslint-import-resolver-${name}`, sourceFile) ||
+    /**
+     * normal full-name for third-party package or absolute path
+     * like config `@xxx/webpack-resolver` or `/your/absolute/path/resolver.js`
+     */
     tryRequire(name, sourceFile) ||
-    tryRequire(path.resolve(getBaseDir(sourceFile), name))
```
ljharb, line 184: looks like this package has been used since 693be1788, the creation of this package.

I'm a bit concerned about its removal even though tests seem to be passing. @benmosher, any idea why this package is used here specifically?

ljharb (changes requested):
It'll be important to add some test cases that would have failed without this change, so that we don't unintentionally break this in the future.

zthxxx:
ok, I will do it next week

PR #1973, opened 2021-01-07
Add exemptFrom option to no-restricted-paths

skbogner:
Personally I want to use this for the following use case:

I have a directory like:
```
src/
  pages/
  features/
  ...many more directories
```

I want to disallow any files (outside of pages) to import from /src/pages. This is not currently possible because `from` (`src/pages`) is nested inside of `target` (`src`).

I would do that with the following rule:
```
{ 
  zones: [ {
    target: './src',
    from: '.src/pages',
    exemptFrom: true,
  } ],
}
```

A different approach would be to go with something like `exceptTarget` or `targetExcept` to mirror the existing `except` option. It might cover more use cases, but does not fit cleanly in the api.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+- [`no-restricted-paths`]: Add `exemptFrom` option to allow files within `from` to import eachother, even if `from` is inside `target`.
 
 ### Fixed
 - [`export`]/TypeScript: properly detect export specifiers as children of a TS module block ([#1889], thanks [@andreubotella])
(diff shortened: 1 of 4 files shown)


skbogner:
Now that I think about it more, I don't think it's actually that useful. Maybe it's simpler to just have the rule
```
{ 
  zones: [ {
    target: './src',
    from: '.src/pages',
  } ],
}
```

And then put a `.eslintrc` in `src/pages` to disable the rule.

ljharb:
You may also want https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-relative-parent-imports.md ? 

In general it’s best to discuss changes in an issue before making a PR, for future reference.

skbogner:
Thanks @ljharb. I would love that rule, but unfortunately that's going to take some work 😟

Yes, I apologise if it inconvenienced you. My way of thinking on the issue.

PR #1974, opened 2021-01-16
[Fix] `no-unused-modules`: make type imports mark a module as used

### LINKED ISSUE #1924: [import/no-unused-modules] False positive while importing TS interfaces
1. The use of `import/no-unused-modules` flags raises a false error: **exported declaration 'XYZ' not used within other modules** when using `import type { XYZ } from "file.ts"` notation. 🐛 

2. Source Code:

```ts
// file1.ts

export interface Foo {

}
```

```ts
// file2.ts

import type { Foo } from "./file1.ts";
```

3. Current Behavior: Lint error

```
exported declaration 'Foo' not used within other modules eslint(import/no-unused-modules)
```

4. Expected behavior: No error

---

Other notes:

I could use `import { Foo } from "./file1.ts"` but this coding style conflicts with the [@typescript-eslint/consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md) rule and I get a lint error:

```
Import "NumberFormatProps" is only used as types eslint(@typescript-eslint/consistent-type-imports)
```

Will be really great if `import/no-unused-modules` can be upgraded to handle `import type ...` syntax. 🎉 

---

Attn. https://github.com/benmosher/eslint-plugin-import/pull/1819
cc @nicolashenry @ljharb

cherryblossom000:
Fixes #1924

`ExportMap` now keeps track of each imported declaration (`imports.get(path).declarations`) and whether the declarations are only importing types (`isOnlyImportingTypes`). This is used by `no-cycle` to

- ignore only type imports
- report circular dependencies at the correct line

For example (from [this existing test](https://github.com/cherryblossom000/eslint-plugin-import/blob/1924/tests/src/rules/no-cycle.js#L159-L163)):

`depth-zero.js`
```js
import { bar } from './flow-types-depth-one'
```

`flow-types-depth-one.js`
```js
// @flow

// this import declaration doesn't trigger the rule because it's a type import
import type { FooType } from './flow-types-depth-two'
// this declaration does because of the 'bar' import, so the error message refers to this line
import { type BarType, bar } from './flow-types-depth-two'

export { bar }
```

`flow-types-depth-two.js`
```js
import { foo } from './depth-one'
```

`depth-one.js`
```js
import foo from './depth-zero'
```

The only significant change in `no-unused-modules` is [here](https://github.com/benmosher/eslint-plugin-import/pull/1974/files#diff-7d2908b010ec724c0f3a0d7615cb91f4673d765d52c6ba6607501e23470f2966R219-R222) where each declaration's `importedSpecifiers` is added to a file's import list. However, because the `imports` in a file's `ExportMap` now include type imports, type imports now mark an export as being used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ lib/
 yarn.lock
 package-lock.json
 npm-shrinkwrap.json
+
+# macOS
+.DS_Store
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`no-webpack-loader-syntax`]/TypeScript: avoid crash on missing name ([#1947], thanks @leonardodino)
 - [`no-extraneous-dependencies`]: Add package.json cache ([#1948], thanks @fa93hws)
 - [`prefer-default-export`]: handle empty array destructuring ([#1965], thanks @ljharb)
+- [`no-unused-modules`]: make type imports mark a module as used (fixes #1924) ([#1974], thanks [@cherryblossom000])
 
 ## [2.22.1] - 2020-09-27
 ### Fixed
@@ -739,6 +740,8 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#1974]: https://github.com/benmosher/eslint-plugin-import/pull/1974
+[#1924]: https://github.com/benmosher/eslint-plugin-import/issues/1924
 [#1965]: https://github.com/benmosher/eslint-plugin-import/issues/1965
 [#1948]: https://github.com/benmosher/eslint-plugin-import/pull/1948
 [#1947]: https://github.com/benmosher/eslint-plugin-import/pull/1947
@@ -1289,3 +1292,4 @@ for info on changes for earlier releases.
 [@tomprats]: https://github.com/tomprats
 [@straub]: https://github.com/straub
 [@andreubotella]: https://github.com/andreubotella
+[@cherryblossom000]: https://github.com/cherryblossom000
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -83,8 +83,8 @@ export default class ExportMap {
 
   /**
    * ensure that imported name fully resolves.
-   * @param  {[type]}  name [description]
-   * @return {Boolean}      [description]
+   * @param  {string} name
+   * @return {{ found: boolean, path: ExportMap[] }}
    */
   hasDeep(name) {
     if (this.namespace.has(name)) return { found: true, path: [this] };
@@ -241,8 +241,8 @@ const availableDocStyleParsers = {
 
 /**
  * parse JSDoc from leading comments
- * @param  {...[type]} comments [description]
- * @return {{doc: object}}
+ * @param {object[]} comments
+ * @return {{ doc: object }}
  */
 function captureJsDoc(comments) {
   let doc;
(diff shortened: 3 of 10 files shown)


ljharb (changes requested):
If you rebase, many of the styling changes will already be taken care of. I suggest running `npx eslint . --fix` after the rebase as well.

ljharb (approved):

tests/files/no-unused-modules/flow/flow-1.js
```diff
@@ -1,3 +1,3 @@
 // @flow strict
 export type Bar = number;
-export interface BarInterface {};
+export interface BarInterface {}
```
ljharb: why is this change needed? this should parse as an exported interface, *followed by* an empty statement, which should work fine.

cherryblossom000: The other test files didn't have semicolons after interfaces/enums, so it was more a consistency/stylistic thing.

ljharb: In that case let's revert this; it's desirable for text fixtures to vary stylistically.

cherryblossom000: I already reverted this; this is outdated.

saiichihashimoto:
@ljharb what's the timeline for releases in this repo? I was assuming CI deployment would just take it away but I'm seeing in the [CHANGELOG](https://github.com/benmosher/eslint-plugin-import/blob/6bb852836676a38d285c022c9ce03af7634db068/CHANGELOG.md) that this is unreleased.

ljharb:
Assuming automatic releases is not a safe assumption; I find that incredibly dangerous and don't do it in any package I maintain.

There is no concrete timeline for releases; I tend to find time every 3-6 months in this project.

saiichihashimoto:
Can I request that you release anytime soon? It's misleading to mark #1924 (and any other issues with unreleased fixes) as resolved considering it'll continue to be broken for dependents for up to 6 months.

ljharb:
@saiichihashimoto it is a decade-old standard practice that issues are closed when fixes are merged, entirely unrelated to the time of release. If that doesn't match your expectations, I'm afraid the easiest path is to adjust your expectations.

I am indeed planning a release soon, but it's blocked right now due to npm 7's release breaking tests on master.

saiichihashimoto:
That makes sense, no sass intended.

saiichihashimoto:
What's the progress on this? It's been a while...

ljharb:
It's still blocked on #1986.

saiichihashimoto:
Now that #1986 is resolved, is there a timeline of when we can expect a release?

ljharb:
There is no such timeline. As soon as I move through all the PRs that have been waiting, I plan to cut a release.

saiichihashimoto:
Is there a way you could cut a release and then go through those PRs? My understanding is that the changes that are waiting have already been tested so the wait for them to go out is arbitrary.

ljharb:
Every release that goes out increases maintenance burden for the project, and anything that's not in master yet, but would be in the next release if I wait, might add features or fix bugs, thus reducing that maintenance burden.

cherryblossom000:
@saiichihashimoto If you really want to use eslint-plugin-import with this bug fix without waiting for a release, you could use [GitHub dependencies in `package.json`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#github-urls):

```json
{
  "devDependencies": {
    "eslint-plugin-import": "benmosher/eslint-plugin-import#462b016"
  }
}
```

462b016cd70b2f192c3eab78a70adcd2284f100d is the commit that merged this PR, but you could also use a newer commit if you want.

cherryblossom000:
@saiichihashimoto [v2.23.0](https://github.com/benmosher/eslint-plugin-import/releases/tag/v2.23.0) has been released, which includes this bug fix.

saiichihashimoto:
Thanks for bearing with my persistence!

PR #1990, opened 2021-02-12
[New] `order`: Add `warnOnUnassignedImports` option to enable warnings for out of order unassigned imports

### LINKED ISSUE #1639: import/order doesn't sort assets that aren't assigned to a variable (e.g. `import './file.css'`)
When sorting imports, any assets that aren't assigned to a variable are ignored, and break/mess with sorting of other imports.

e.g.
```
import mapboxgl from 'mapbox-gl';

import './file.json';
import './file.css';
import './file.png';
import React from 'react';
import { ToastContainer } from 'react-toastify';

import { ErrorBoundary } from '@app/components';
import { useAccount, useUser } from 'src/state';

import mygif from './file.gif';
import Routes from './Routes';
```

If I change the above and assign assets to a variable, sorting is applied as expected.

```
import mapboxgl from 'mapbox-gl';
import React from 'react';
import { ToastContainer } from 'react-toastify';

import { ErrorBoundary } from '@app/components';
import { useAccount, useUser } from 'src/state';

import b from './file.css';
import mygif from './file.gif';
import a from './file.json';
import c from './file.png';
import Routes from './Routes';
```

hayes:
fixes #1639 

This seems like a pretty simple solution, but does have one weird side effect: The behavior of the newline option changes slightly, in that it can now remove newlines between groups correctly when there are unassigned imports.  This is only the case when the new option is enabled.  Updating this to make that behavior consistent (always allow new-line changes around unassigned imports) would be safe, and easy to implement, but is a breaking change, so leaving it as is seems better.

cc @Alphy11  @ljharb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - add [`no-import-module-exports`] rule: report import declarations with CommonJS exports ([#804], thanks [@kentcdodds] and [@ttmarek])
 - [`no-unused-modules`]: Support destructuring assignment for `export`. ([#1997], thanks [@s-h-a-d-o-w])
 - [`order`]: support type imports ([#2021], thanks [@grit96])
+- [`order`]: Add `warnOnUnassignedImports` option to enable warnings for out of order unassigned imports ([#1990], thanks [@hayes])
 
 ### Fixed
 - [`export`]/TypeScript: properly detect export specifiers as children of a TS module block ([#1889], thanks [@andreubotella])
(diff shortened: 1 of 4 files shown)


hayes:
@ljharb would having a configurable whitelist of extensions that are safe to autofix as unassigned imports be acceptable?  I know that side effects are generally not safe to re-order, but specifically for styles, there are use cases, where either ordering them can be known to be safe for a specific application.

Alphy11:
Specifically, css files are often safe to reorder (not always, but in many applications).  Plus opting-in makes it a much safer operation.

ljharb (approved):
Seems reasonable

ljharb:
@hayes let's explore that in a future issue.

PR #428, opened 2016-07-09
`no-extraneous-dependencies`: add support/option for `peerDependencies` (#423)

jfmengels:
`no-extraneous-dependencies`: add support/option for `peerDependencies` (#423)

PR #1993, opened 2021-02-24
[Docs] Correcting peerDependencies option default to `true`

dwardu:
Kindly confirm that the default really is `true`.


ljharb (approved):
Indeed, this seems to have been set in #428 but the docs not updated.

PR #1999, opened 2021-03-06
Fix: import/no-unresolved can not resolve sub-directory of built-in module

southerncross:
The rule [import/core-modules](https://github.com/benmosher/eslint-plugin-import#importcore-modules) allows us to add customized core modules, such as 'electron':

```js
import 'electron';
```

But it is unable to resolve sub-directory:

```js
import 'electron/foo'; // eslint error, unable to resolve "electron/foo"
```

However, there is another rule [no-extraneous-dependencies](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md) that supports sub-directory, for example:

```js
import 'lodash/get'; // no eslint error
```

So I think it should be better if we make those rules unified for subdirectory, so I opened this PR.

The modification is very simple: if a module is kind of built-in (even if it is customized core module), then we just skip the rest checks of rule `import/no-unresolved`.

Unit tests are also added :D

### DIFF
--- a/src/rules/no-unresolved.js
+++ b/src/rules/no-unresolved.js
@@ -7,6 +7,7 @@ import resolve, { CASE_SENSITIVE_FS, fileExistsWithCaseSync } from 'eslint-modul
 import ModuleCache from 'eslint-module-utils/ModuleCache';
 import moduleVisitor, { makeOptionsSchema } from 'eslint-module-utils/moduleVisitor';
 import docsUrl from '../docsUrl';
+import { isBuiltIn } from '../core/importType';
 
 module.exports = {
   meta: {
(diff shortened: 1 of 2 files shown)


ljharb (changes requested):
Core modules do not have subdirectories.

If you want `electron/foo`, that must be listed as a core module too. If there's any rule that *allows* a custom core module's subdirectory, that is a bug and it should be fixed.

southerncross:
> Core modules do not have subdirectories.
> 
> If you want `electron/foo`, that must be listed as a core module too. If there's any rule that _allows_ a custom core module's subdirectory, that is a bug and it should be fixed.

hmmmm, I got your point, it's meaningful.

PR #2009, opened 2021-03-18
[Fix]: no-unresolved: Support importing only types when the module isn't installed

EdenGottlieb:
- [ ] write tests
- [ ] implement feature/fix bug
- [ ] update docs

Don't trigger an error when only typings are imported and the actual module is not installed (only its typings are installed under (`node_modules/@types`).
This is useful when defining packages holding domain objects that only need to reference types, without using the actual module.

I chose memorystream for the test for several reasons:
1. I needed a module that won't be added to the project any time soon (adding it would break the tests!)
2. I needed to use a module with publicly available typings
3. It's what I used for the initial debugging :)

Please feel free to comment on code logic, test strategy and documentation phrasing, all feedback is welcome!

### DIFF
--- a/resolvers/node/index.js
+++ b/resolvers/node/index.js
@@ -30,7 +30,7 @@ function opts(file, config) {
   return Object.assign({
     // more closely matches Node (#333)
     // plus 'mjs' for native modules! (#939)
-    extensions: ['.mjs', '.js', '.json', '.node'],
+    extensions: ['.mjs', '.js', '.json', '.node', '.d.ts'],
   },
   config,
   {
--- a/tests/files/webpack.config.js
+++ b/tests/files/webpack.config.js
@@ -1,6 +1,6 @@
 module.exports = {
   resolve: {
-    extensions: ['', '.js', '.jsx'],
+    extensions: ['', '.js', '.jsx', '.d.ts'],
     root: __dirname,
     alias: {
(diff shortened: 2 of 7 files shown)


JounQin:
You should use `eslint-import-resolver-typescript`

ljharb:
Indeed; with the typescript eslint resolver (which is required when using typescript) this shouldn't be an issue.

@EdenGottlieb can you confirm that using it solves the issue for you?

resolvers/node/index.js
```diff
@@ -30,7 +30,7 @@ function opts(file, config) {
   return Object.assign({
     // more closely matches Node (#333)
     // plus 'mjs' for native modules! (#939)
-    extensions: ['.mjs', '.js', '.json', '.node'],
+    extensions: ['.mjs', '.js', '.json', '.node', '.d.ts'],
```
ljharb, line 33: this is incorrect; the node resolver must never know about typescript files, because node doesn't.

tests/files/webpack.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   resolve: {
-    extensions: ['', '.js', '.jsx'],
+    extensions: ['', '.js', '.jsx', '.d.ts'],
```
ljharb, line 3: webpack shouldn't be resolving d.ts files at all.

ljharb (changes requested):

Issue #1986, opened 2021-02-03
npm 7 breaks tests


ljharb:
As of npm 7's promotion yesterday, there's at least two problems breaking tests on master and every PR:
 - peer dep issues. eslint 4 is incompatible with `@typescript-eslint/parser`. this is solved with `--legacy-peer-deps` ([commit](https://github.com/ljharb/eslint-plugin-import/commit/b1aef78590f32b63d1539dac57141869ae039740)).
 - when node_modules is present in the root, and was installed with npm 7, both npm 6 and 7 [fail](https://github.com/ljharb/eslint-plugin-import/runs/1817778513?check_suite_focus=true) on a bunch of the resolver tests, and on a few of the root tests (presumably related). When node_modules is absent, or `chmod a-x node_modules`, or installed with npm 6, things are fine. ([lockfile diff](https://gist.github.com/ljharb/74527f6d042fb62dae2365bb1a9017ca))
 
It's being looked into. After a few days of this, I'll probably try to downgrade npm on affected builds, but it might reflect an issue that npm 7 users will encounter, so I'd prefer to find a proper fix.

exarus:
Plugin doesn't work with Yarn 2 PnP.

ESLint validation fails if there is any `import` present in code.
`error  Resolve error: unable to load resolver "node"  import/no-duplicates`

Explicitly installing `eslint-import-resolver-node` solves the issue.

ljharb:
@exarus indeed, but that's a bug in PnP, and not in the package. Specifically, https://github.com/benmosher/eslint-plugin-import/blob/master/package.json#L73 ensures it's available as a dev dep in the root package, just like https://github.com/benmosher/eslint-plugin-import/blob/master/package.json#L105 ensures it's available as a runtime dep for all consumers. The issue you're referring to is that consumers' `npm install` will hoist the `eslint-import-resolver-node` dep to the root, and yarn PnP incorrectly refuses to allow it to be imported unless the consuming package also explicitly lists it.

It's possible that listing it twice is exposing a bug in npm 7, though.

merceyz:
> indeed, but that's a bug in PnP [...] and yarn PnP incorrectly refuses to allow it to be imported unless the consuming package also explicitly lists it.

That's not a bug in PnP but in how the package is resolved, it's resolved relative to the file getting checked instead of the eslint plugin.

arcanis:
[...]

PR #2012, opened 2021-03-24
[Tests] `no-unresolved`: add tests for `import()`

davidbonnet:
Dynamic imports landed in EcmaScript 2020, and Eslint now parses it out of the box.
However, the AST `node` containing the dynamic import is an `ImportExpression`, which is ignored by the `no-unresolved` rule.

### DIFF
--- a/tests/src/rules/no-unresolved.js
+++ b/tests/src/rules/no-unresolved.js
@@ -22,7 +22,7 @@ function runResolverTests(resolver) {
   }
 
   ruleTester.run(`no-unresolved (${resolver})`, rule, {
-    valid: [
+    valid: [].concat(
       test({ code: 'import "./malformed.js"' }),
 
       rest({ code: 'import foo from "./bar";' }),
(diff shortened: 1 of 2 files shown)


davidbonnet:
I dedicate any and all copyright interest in this software to the
    public domain. I make this dedication for the benefit of the public at
    large and to the detriment of my heirs and successors. I intend this
    dedication to be an overt act of relinquishment in perpetuity of all
    present and future rights to this software under copyright law.

    Anyone is free to copy, modify, publish, use, compile, sell, or
    distribute this software, either in source code form or as a compiled
    binary, for any purpose, commercial or non-commercial, and by any
    means.

ljharb:
Thanks, the fix seems great. This is blocked for now on #1986, unfortunately.

ljharb (approved):

tests/src/rules/no-unresolved.js
```diff
@@ -85,7 +92,7 @@ function runResolverTests(resolver) {
         options: [{ commonjs: true }] }),
     ],
 
-    invalid: [
+    invalid: [].concat(
```
aladdin-add, line 95: just curious why the `[].concat()` is needed. :)

ljharb: because RuleTester requires an array of objects, so there's no way to conditionally provide an ignoreable value, so we have to use `[].concat()` and have the conditional helper return an empty array, since concat flatMaps over arrays.

ljharb:
@davidbonnet tests still seem to be failing here, so i think this one needs a bit more investigation. @aladdin-add if your PR can get the tests passing, that'd be great too.

Issue #1966, opened 2020-12-21
eslint-import-resolver-webpack is not compatible with webpack 5 - TypeError: callback is not a function


lavor:
eslint-import-resolver-webpack is not compatible with webpack 5 externals API of webpack config.

When I use [function for resolving externals](https://webpack.js.org/configuration/externals/#function), I am getting an error:
```
ESLint: 
Resolve error: TypeError: callback is not a function
    at D:\++projects\++datlowe\ui\datlowe-core\src\js\/externals.js:102:13
    at findExternal (D:\++projects\++datlowe\ui\datlowe-core\node_modules\eslint-import-resolver-webpack\index.js:330:15)
    at D:\++projects\++datlowe\ui\datlowe-core\node_modules\eslint-import-resolver-webpack\index.js:321:49
    at Array.some (<anonymous>)
    at findExternal (D:\++projects\++datlowe\ui\datlowe-core\node_modules\eslint-import-resolver-webpack\index.js:321:22)
    at Object.exports.resolve (D:\++projects\++datlowe\ui\datlowe-core\node_modules\eslint-import-resolver-webpack\index.js:135:7)
    at v2 (D:\++projects\++datlowe\ui\datlowe-core\node_modules\eslint-module-utils\resolve.js:117:23)
    at withResolver (D:\++projects\++datlowe\ui\datlowe-core\node_modules\eslint-module-utils\resolve.js:122:16)
    at fullResolve (D:\++projects\++datlowe\ui\datlowe-core\node_modules\eslint-module-utils\resolve.js:139:22)
    at relative (D:\++projects\++datlowe\ui\datlowe-core\node_modules\eslint-module-utils\resolve.js:84:10)
(import/namespace)
```
Issue happens because there was a change to that funtion API:
* Webpack 4: `function(context, request, callback) {}`
* Webpack 5: `function({ context, request }, callback) {}`
My function for resolving external is now looking like this:
```
// there was change of API in webpack 5
// webpack 4 version looks line this
// function(context, request, callback) {
function({context, request}, callback) {
    const modulePath = path.join(context || '', request || '');
    if (/node_modules[/\\]/.test(modulePath)) {
        const splittedPath = modulePath.split(/node_modules[/\\]/);
        if (splittedPath && splittedPath.length > 0) {
            // if request is not a relative path starting with dot (.), use request instead of the whole module path
            const externalModulePath = /^\./.test(request) ? splittedPath[splittedPath.length - 1] : request;
[...]

PR #1962, opened 2020-12-10
Add warning about async Webpack configs

ogonkov:
See discussion in #1961

PR #2023, opened 2021-04-03
[resolvers/webpack] [New] add support for webpack 5 'externals function'

jet2jet:
This PR allows functions for resolving externals for both webpack 5 and prior to webpack 5 (issue: #1966 )

- To handle configs for both webpack 5 and priors, this PR [checks `externals.length` as webpack does](https://github.com/webpack/webpack/blob/v5.30.0/lib/ExternalModuleFactoryPlugin.js#L166).
- In webpack 5.15, `getResolve` function is added to the object in the first argument, so this PR provides a very simple `getResolve` implementation (using the return value of `getResolveSync`).
- In this PR async 'external functions' are still not supported (as async webpack config not supported; related: #1962)
  - [synchronized-promise](https://www.npmjs.com/package/synchronized-promise) might be usable to resolve this, but I haven't tried yet. (If it is OK, I'll try to add using synchronized-promise into this PR or another.)

### DIFF
--- a/resolvers/webpack/.eslintrc
+++ b/resolvers/webpack/.eslintrc
@@ -3,4 +3,7 @@
     "import/no-extraneous-dependencies": 1,
     "no-console": 1,
   },
+  "env": {
+    "es6": true,
+  },
 }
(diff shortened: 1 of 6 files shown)


ljharb (approved):
nice work!

PR #2025, opened 2021-04-09
Add footnote to rules list for rules enabled in default configs

Pezmc:
Adds footnotes to the rules enabled in [recommended](https://github.com/benmosher/eslint-plugin-import/blob/master/config/recommended.js), [errors](https://github.com/benmosher/eslint-plugin-import/blob/master/config/errors.js), and [warnings](https://github.com/benmosher/eslint-plugin-import/blob/master/config/warnings.js).

I find myself having to manually cross-reference this all the time, eslint's default rules use a table to indicate their defaults, some footnote symbols get's us similar functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,12 +12,14 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 
 ## Rules
 
+\* = enabled in recommended config as error, † = enabled in recommended config as warning
+
 ### Static analysis
 
-* Ensure imports point to a file/module that can be resolved. ([`no-unresolved`])
-* Ensure named imports correspond to a named export in the remote file. ([`named`])
-* Ensure a default export is present, given a default import. ([`default`])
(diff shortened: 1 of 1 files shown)


ljharb:
This seems like something that could quickly go out of date, and isn’t actually useful unless you’re using the recommended config, which most users aren’t (most users use this via other shared configs, like airbnb or create-react-app, neither of which match the recommended config).

Pezmc:
That's fair, though it's only out of date if it's not updated 🤷. Eslint does manage to keep the readme up to date, could we perhaps add a comment to the config files? 

Totally understood re those using shared config, though both `plugin:import/errors` and `plugin:import/warnings` are explicitly recommended by the readme, so I'd imagine those coming directly are using them.

ljharb:
I think this would be a good change to add if we could automate it with a script (we can add a CI check to ensure the script is ran).

PR #2033, opened 2021-04-25
Allow mapping of expected extensions in import/extension rule

pwhissell:
Because all of the following requirements are true:
- [nodejs requires file extensions when using relative paths](https://nodejs.org/api/esm.html#esm_mandatory_file_extensions)
- [typescript only supports "top-level await" when transpiling to ESM](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#top-level-await)
- [nodejs modules (CJS) don't support "top-level await"](https://github.com/nodejs/node/issues/21267)

Then if someone wants to use Typescript to gereate nodejs apps that use "top-level await", it makes sens to start usiung nodejs's
newly supported ESM. (this is also probably going to apply to more and more module functionalities). To do so, while developping, people have to make sure that all relative imports comply with nodejs's requirement to specify the trailing ".js" file extension.

Typescript has no functionality to enforce ".js" file extensions. Because tslint was deprecated in favor of eslint, I feel
this eslint plugin would make a lot of people happy in filling out the gap.

__Note this is currently a draft which more work__
- [ ] write tests
- [ ] update docs

fixes https://github.com/benmosher/eslint-plugin-import/issues/2030

### DIFF
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -24,6 +24,7 @@ function buildProperties(context) {
     defaultConfig: 'never',
     pattern: {},
     ignorePackages: false,
+    extensionMap: [],
   };
 
   context.options.forEach(obj => {
@@ -31,6 +32,11 @@ function buildProperties(context) {
     // If this is a string, set defaultConfig to its value
     if (typeof obj === 'string') {
       result.defaultConfig = obj;
+      return;
+    }
+
+    if (obj.extensionMap !== undefined) {
+      result.extensionMap = obj.extensionMap;
       return;
     }
 
@@ -49,6 +55,7 @@ function buildProperties(context) {
     if (obj.ignorePackages !== undefined) {
       result.ignorePackages = obj.ignorePackages;
     }
+
   });
 
   if (result.defaultConfig === 'ignorePackages') {
@@ -135,6 +142,10 @@ module.exports = {
       return false;
     }
 
+    function mapExtension(extension, extensionMap) {
+      return extensionMap && extensionMap[extension] ? extensionMap[extension] : extension;
+    }
+
     function checkFileExtension(source) {
       // bail if the declaration doesn't have a source, e.g. "export { foo };"
       if (!source) return;
@@ -154,7 +165,9 @@ module.exports = {
 
       // get extension from resolved path, if possible.
       // for unresolved, use source value.
-      const extension = path.extname(resolvedPath || importPath).substring(1);
+      let extension = path.extname(resolvedPath || importPath).substring(1);
+
+      extension = mapExtension(extension, props.extensionMap);
 
       // determine if this is a module
       const isPackage = isExternalModule(
--- a/tests/files/typescript-import-js-extension/file.ts
+++ b/tests/files/typescript-import-js-extension/file.ts
@@ -1,0 +1,1 @@
+export const helloWorld = 'helloWorld';
--- a/tests/files/typescript-import-js-extension/tsconfig.json
+++ b/tests/files/typescript-import-js-extension/tsconfig.json
@@ -1,0 +1,5 @@
+{
+    "compilerOptions": {
+        "baseUrl": "."
+    }
+}
--- a/tests/src/rules/extensions.js
+++ b/tests/src/rules/extensions.js
@@ -1,6 +1,7 @@
 import { RuleTester } from 'eslint';
+import path from 'path';
 import rule from 'rules/extensions';
-import { test, testFilePath } from '../utils';
+import { getTSParsers, test, testFilePath } from '../utils';
 
 const ruleTester = new RuleTester();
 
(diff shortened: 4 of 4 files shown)


ljharb:
I'm still pretty confused here.

If you want to require an extension, configure the rule to do so.

If you want to only require an extension in `.ts` files, use `overrides`, set `files` to `**/*.ts`, and configure the rule to do so there.

What does this PR allow you to do that you can't already do with an eslint config?

pwhissell:
I'm new to eslint so I wouldn't be suprised that I misunderstand something, but here is how I understand "overrides":
```
"overrides": [
    {
      "files": ["**/*.ts"],
      "rules": {
        "import/extensions": "#Some Config That Would Force All (or relative) Imports To Have .js extension#"
      }
    }
  ]
```
I havn't found any configuration that would enforce a specific (".js" in my case) extension for all imports

pwhissell:
Given the following files:
```
tsconfig.json
file1.ts
file2.ts
```

if the content of file1.ts is:
```typescript
import {something} from "./file2"
import {other} from "./file2.js"
```

the typescript module resolver resolves both "./file2" and "file2.js" to the following file: "./file2.ts" which is right because the resolution happens before transpilation.
When using the rule: import/extensions, the expected extension is chosen from the file name resolved by the module resolver. In this case because the resolved module is defined in the "./file.ts" file, the plugin expects the extension to be ".ts". 
The linting error disapears if you right the import statement like so:
`import {other} from "./file2.ts"`
This import statement is invalid and flagged by Typescript as such.

This PR allows the developper to configure the "import/extensions" in a way that would rather make the linter expect ".js" extension in all import statements even if the resolved module is a file with another extension. 
In my example, the linter would now expect and accept nothing other than:
```typescript
import {other} from "./file2.js"
```

ljharb:
You’re trying to force an extension appear in code that doesn’t actually exist on the filesystem?

pwhissell:
yes exactly. It isvreally awkward but typescript  ensures you that it will. Otherwise typescript raises an exception if it cant make that garantee

pwhissell:
typescript should definately implement such a functionality but this plugin is 80 lines away from doing it

pwhissell:
its a non-intrusive solution that would please any nodejs+typescript enthousiasts
even more so now that tslint is deprecated in favor of eslint.

perhaps the proposed solution could be extended to other transpilers or contexts in its flexibility. Or should I change the implementation to focuse on ts transpilation?

pwhissell:
im linting the ts file and not the transpiled es file

ljharb:
The proper solution here would be to implement this as an option in the typescript resolver, not in this plugin itself.

pwhissell:
I think your opinion is very defendable, but I would argue that the resolver is correct in that it resolves the module to the right file. Concerning separation of concerns, I also argue that enforcing a specific extension fits very well with this module's responsibility.
If this were an invasive solution, I wouldn't even have considered the option of a PR.

ljharb:
It’s that the extension doesn’t already exist that’s the problem. This plugin is only attempting to combine node’s CJS resolution algorithm with a style preference.

pwhissell:
Hahah now I guess I'm confused :S
>  This plugin is only attempting to combine node’s CJS resolution algorithm
When reading the readme from the root of this repo, I found that typescript the resolver is supported. I also appreciated got inspired by existing unit tests, when specifying validating my functionality for all supported typescript versions.

I may have missinterpreted the readme when reading that this repo was "trying to support linting of ES2015+ (ES6+) import/export syntax, and prevent any issues with misspelling of file paths and import names." 

I also think that linters are not ownly good for validating stylistic errors, but also (and probably more importantly) programmatic errors.

pwhissell:
I think the un-introssiveness of the implementation makes this at least worth a second and third opinion
@benmosher, @jfmengels, @rfermann what do you guys think

ljharb:
It was only very recently that node changed the plan and decided not to auto-add extensions for ESM. As such, either we enforce what babel/ts does (no extensions, primarily) or what node does (.mjs or, sometimes, .js, but only when the files exist).

It seems like you’re trying to work around typescript’s poor integration with node’s native ESM support, rather than accepting that the two shouldn’t be used together yet.

pwhissell:
Typescript should indeed improve esm support, but they wont, because they point to here saying that linting should be the solution.

Given the popularity of the MEAN stack I think having typescript accross the entire stack is goingg to be a trend for the future. Before this can happen, I think features like tgese need to be implemented.

In the context where the imported module exists (in a version not yet transpiled),
I honestly don't grasp the importance of the idea that the imported file needs to exist before linting

pwhissell:
> It seems like you’re trying to work around typescript’s poor integration with node’s native ESM support, rather than accepting that the two shouldn’t be used together yet.

I wouldn't call it working around, I woulf call it "improve"

> As such, either we enforce what babel/ts does (no extensions, primarily)
Typescript does support imports with extensions: 
given a file called "y.ts"
`import x from y.js` would 
- at transpile time resolve to the y.ts
- at runtime time resolve to the y.js (which is the transpiled y.ts)

ljharb:
And prior to transpile time, would explicitly point to nothing - which is when the linter runs. It would likely work better to transpile your TS with babel, and transform the specifiers to add .js in the build process, and continue to omit them in your source code - then you don’t need to do somersaults in your linter config to support this case.

pwhissell:
prior to transpile time, it does not point to nothing. When you use a typescript resolver, it points to the ts file

pwhissell:
You suggest I don't transpile my ts files with Typescrtipt but rather do it with Babel?

ljharb:
Always, yes. Babel is much better at transpiling TS than tsc is.

> When you use a typescript resolver, it points to the ts file

if that is true, then i'm not sure why any change is needed - the `.js` extension should Just Work already.

pwhissell:
Exactly the js extension does work perfectly
typescript supports both "no extension" and ".js" extension.
The problem arizes when you want your project to always use one or the other but not both.
This is when eslint-plugin-import comes in.

Currently the plugin **takes for granted that the path to file path to the resolved module has the module's correct extension**.
In the case of typescript modules, the extension of the file that defines the module (.ts) doesn't correspond to the the extension of the module itself (.js)
this applies for any transpilers that require different extensions for pre-transpiled files (if there are any out there)

> Always, yes. Babel is much better at transpiling TS than tsc is.

I think that this is a valid opinion which I share in some regards (it seems more and more so as time moves on). However until [typed-objects](https://github.com/tschneidereit/proposal-typed-objects) become more stable I will be sticking to typescript (I don't think I will be the sole developper to do so out there)

pwhissell:
transpiling with babel doesn't solve the problem. The workaround is to rename typescript files so that the extension is ".js" which is even more awkward 

there really is no way to get this to pass is there? I'm not tied to the implementation at all, but allowing the linter to expect module extensions other than the extension of the file it is defined in seems perfectly aligned with the scope & goal of this repo.

pwhissell:
this would fix https://github.com/microsoft/TypeScript/issues/16577

pwhissell:
Just found [this](https://nodejs.org/api/esm.html#esm_customizing_esm_specifier_resolution_algorithm) . You can basically configure how nodejs resolves esm modules. running nodejs with the following flag solves the initial problem behind this PR "node --experimental-specifier-resolution=node"

evelynhathaway:
I am a little hesitant to add my voice in this thread with the hostility in it, but I am looking for a similar solution. I am sorry this has become a tough issue for those involved.

If this pull request is stuck between a rock and a hard place, is there a possibility of allowing resolver plugins to implement this mapping while we are in this uncanny valley of TypeScript + Node.js ESM? I imagine the rule primarily being used for TypeScript and gradually being faded out depending on whether the module specifier resolution aligns more with node, or TypeScript implements a solution for full import paths besides manually adding them in the TypeScript code with `.js` extensions.

From my testing, [`eslint-import-resolver-typescript`](https://github.com/alexgorbatchev/eslint-import-resolver-typescript) supports resolving imports using `.js` extensions in TypeScript imports, even when they don't exist on the file system. I can imagine that plugin wanting to add support for the extension rule as well.

ljharb:
@evelynhathaway that's an interesting idea - I'm not sure yet how we'd rework things so that was done cleanly, but allowing a custom resolver to do this stuff is a lot safer/simpler than building it into the core plugin.

PR #2036, opened 2021-04-26
[Deps] update `contains-path`, `doctrine`, `pkg-up`, `read-pkg-up`

Semigradsky:


ljharb (changes requested):
These are all breaking changes, and almost all of them drop node versions, which means this plugin can effectively _never_ update to them.

Why should these be updated - what motivates this PR?

Semigradsky:
Maybe I missed something. As I see all these versions still support node.js 4.

ljharb:
Yes, and we support node 0.10 and above (via eslint 2), so that would be a breaking change.

Semigradsky:
Sorry, I don't know about this... Shouldn't there be `0.10` in `engines` field in `package.json`?

ljharb:
There could be, but there shouldn't be because then our package would be claiming compatibility with node 0.10 even if the user uses eslint 7, for example. In this case, the engines requirements of our peer deps apply to us, but must be implicit.

Semigradsky:
Eslint 7 officially doesn't support node 4, 6, even 8, but you do. Is this not the same case?

ljharb:
Yes - if we put node 4 in out "engines" field, then it would be inaccurate when eslint 7 is installed, but it is accurate with older eslint versions. Thus, we don't have an "engines" field, but we still need to remain compatible with eslint 2 - 7 (and all the node versions those support).

Semigradsky:
Maybe I smth don't understand but it looks the same for me.

If we put node **0.10** in "engines" field, then it would be inaccurate when eslint 7 is installed, but it is accurate with older eslint versions (eslint 2).

But there is `">=4"` in engines:
https://github.com/benmosher/eslint-plugin-import/blob/e871a9a2e18fd4d67a5a237789131b44cc705b5b/package.json#L6

ljharb:
ha, fair point, i didn't realize we'd already required that.

In that case, perhaps we could update to these versions - can you confirm that you've checked the changelogs for each, and that the only breaking changes are dropping node < 4 support?

Semigradsky:
Yep.
- `"contains-path": "^1.0.0"` - "node": ">=0.10.0"
- `"doctrine": "^2.1.0"` - "node": ">=0.10.0"
- `"is-core-module": "^2.3.0"` - no `engines` field, but no specific in code and there is only one dependency which has "node": ">= 0.4.0"
- `"pkg-up": "^2.0.0"` - "node": ">=4"
- `"read-pkg-up": "^3.0.0"` - "node": ">=4"

ljharb:
Master's currently broken (see #1986), so this will have to wait until that's resolved.

ljharb (approved):

PR #1997, opened 2021-02-28
[New] `no-unused-modules`: Support destructuring

s-h-a-d-o-w:
Currently, when using a destructuring assignment with `export`, errors like this are thrown:
```
error  exported declaration 'undefined' not used within other modules  import/no-unused-modules
```

This PR fixes that issue.
Since the reasoning here seemed quite straightforward to me, I figured I'd jump straight to a PR. :)

(Minor question out of sheer curiosity: In tests, why the code duplication with code-to-be-linted in files but then also as strings within the tests?)

PR #2038, opened 2021-04-27
Destructured exports no unused modules

KristjanTammekivi:
Fixes https://github.com/benmosher/eslint-plugin-import/issues/2037

### DIFF
--- a/tests/src/rules/no-unused-modules.js
+++ b/tests/src/rules/no-unused-modules.js
@@ -500,7 +500,7 @@ describe('renameDefault', () => {
   });
 });
 
-describe('test behaviour for new file', () => {
+describe('test behavior for new file', () => {
   before(() => {
     fs.writeFileSync(testFilePath('./no-unused-modules/file-added-0.js'), '', { encoding: 'utf8' });
   });
@@ -588,7 +588,7 @@ describe('test behaviour for new file', () => {
   });
 
 
-  describe('test behaviour for new file', () => {
+  describe('test behavior for new file', () => {
     before(() => {
       fs.writeFileSync(testFilePath('./no-unused-modules/file-added-1.js'), '', { encoding: 'utf8' });
     });
@@ -619,7 +619,7 @@ describe('test behaviour for new file', () => {
   });
 });
 
-describe('test behaviour for new file', () => {
+describe('test behavior for new file', () => {
   before(() => {
     fs.writeFileSync(testFilePath('./no-unused-modules/file-added-2.js'), '', { encoding: 'utf8' });
   });
@@ -641,7 +641,7 @@ describe('test behaviour for new file', () => {
   });
 });
 
-describe('test behaviour for new file', () => {
+describe('test behavior for new file', () => {
   before(() => {
     fs.writeFileSync(testFilePath('./no-unused-modules/file-added-3.js'), '', { encoding: 'utf8' });
   });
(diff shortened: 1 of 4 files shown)


tests/src/rules/no-unused-modules.js
```diff
@@ -663,6 +663,33 @@ describe('test behaviour for new file', () => {
   });
 });
 
+describe('test behaviour for destructured exports', () => {
+  before(() => {
```
ljharb, line 667: Please always use beforeEach/afterEach; before/after is “all” semantics, which leads to test pollution.

tests/src/rules/no-unused-modules.js
```diff
@@ -663,6 +663,33 @@ describe('test behaviour for new file', () => {
   });
 });
 
+describe('test behaviour for destructured exports', () => {
```
ljharb, line 666:
```suggestion
describe('test behavior for destructured exports', () => {
```

ljharb (changes requested):
Named exports aren’t destructuring; please update the terminology :-)

KristjanTammekivi:
I don't see how `export const { a } = { a: 5 };` is not destructuring? In MDN there's an [example](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) of destructured assignments
```js
// Exporting destructured assignments with renaming
export const { name1, name2: bar } = o;
```

ljharb:
Ohh, i misunderstood! Thanks.

tests/src/rules/no-unused-modules.js
```diff
@@ -663,7 +663,34 @@ describe('test behaviour for new file', () => {
   });
 });
 
-describe('test behaviour for new file', () => {
+describe('test behavior for destructured exports', () => {
+  beforeEach(() => {
+    fs.writeFileSync(testFilePath('./no-unused-modules/file-added-3.js'), '', { encoding: 'utf8' });
```
ljharb, line 668: Is there a reason not to make this just be a checked-in empty file? We have a number of fixtures already.

KristjanTammekivi: There's the reason that I was just hacking it together copying and pasting as a proof of concept :)
I sort of fixed it now but I couldn't figure out why, even though I reset the empty_file in afterEach, I had to change the invalid case to a different variable name to make it fail.

tests/src/rules/no-unused-modules.js
```diff
@@ -663,7 +663,35 @@ describe('test behaviour for new file', () => {
   });
 });
 
-describe('test behaviour for new file', () => {
+describe('test behavior for destructured exports', () => {
+  beforeEach(() => {
+    fs.writeFileSync(testFilePath('./no-unused-modules/file-destructured.js'), '', { encoding: 'utf8' });
+  });
```
ljharb, line 669: my hope is to avoid any fs operations here - can these not be committed, with their contents?

KristjanTammekivi: Uhh, sure just a moment, not sure why You prefer committed files though

ljharb: It's more deterministic, doesn't require any sort of permissions to run tests, and is much easier to debug.

ljharb (approved):
Thanks, seems good

ljharb:
(Note that merging this is blocked on #1986)

ljharb:
This was actually fixed by #1997, but I've rebased this to keep the test.

emreboga:
Hey @ljharb, I don't think this was fully fixed by #1997 as it only addresses object type destucturing, not including array types. You can see it in the code:
https://github.com/import-js/eslint-plugin-import/pull/1997/files#diff-7d2908b010ec724c0f3a0d7615cb91f4673d765d52c6ba6607501e23470f2966R85

I'm seeing this in the code with the version 2.28.1, which has the fix with #1997. Here is a code chunk I have:
```
export const [authUserLoading, authUserError] = createLoadingAndErrorSelectors(
  AUTH_USER
);
```
and here is the error I get:
```
error  exported declaration 'undefined' not used within other modules  import/no-unused-modules
```

ljharb:
@emreboga interesting - would you mind filing a new issue about that one?

emreboga:
@ljharb Sure! Please see https://github.com/import-js/eslint-plugin-import/issues/2930

Issue #903, opened 2017-07-26
Ignore rules based on module name for no-extraneous-dependencies


lukechilds:
It would be great if the `import/no-extraneous-dependencies` rule could be ignored project wide but only for certain dependencies.

`import/no-unresolved` already has this functionality with it's `ignore` regex setting.

This could also be a potential solution for https://github.com/benmosher/eslint-plugin-import/issues/496

klimashkin:
Right.
It is really needed in two cases:

1. In case of Flow.
For instance in typing react component. 
```javascript
import type {AnyReactElement, ReactChildren} from 'react-flow-types';
```
`react-flow-types` are obviously in devDependencies, it never goes to production

2. In case of `__DEV__` environment variable (or `process.env.development` or whatever) when we dynamcally require packages for development (or vice versa)
```javascript
if (process.env.NODE_ENV !== 'production' && __WHY_UPDATE__ === true) {
  const {whyDidYouUpdate} = require('why-did-you-update'); 
  const RedBox = require('redbox-react');

  require('expose-loader?Perf!react-addons-perf');
}
```

Without having `ignore` option in `no-extraneous-dependencies` rule, I have to add `// eslint-disable-line import/no-extraneous-dependencies` on such lines. But in case of Flow it hardly possible, because requires too many such lines, in all react components files

ljharb:
Linting should be done in dev, not production, so all dev dependencies would always be present.

klimashkin:
@ljharb Not sure I got what you meant.
Of course linting happens in dev.

But one of the purpose of `no-extraneous-dependencies` rule is to restrict accidental usage of packages from devDependencies in your application that is built with webpack for browser.
In previous message I describe why it should allow exceptions - when developer knows that he uses package from `devDependencies` in built application but only in development mode. Otherwise developer needs to put, for instance, `why-did-you-update` into `dependencies`, what is wrong.

* I'm talking about application development process, not library

ljharb:
Ah, I think I understand.

Would it perhaps make more sense to have an option that skips checking dev deps inside NODE_ENV === production blocks, or similar? An ignore list is just as hard to maintain and get right as individual ignore comments are (which is what I'd suggest in the meantime)

klimashkin:
But variable name can be absolutely random, not everybody uses NODE_ENV.
And it would not solve imports of flow types, they are not conditional.

I don't think ignore list for that rule is harder to maintain than for other. We have about 5-6 such packages

[...]

PR #2039, opened 2021-04-27
feat: add ignore option to extraneous deps

Aghassi:
We are doing a migration with 1st party dependencies in a monorepo that requires this ability, otherwise we get false positives. This option allows the user to pass a regex so that if a package matches the regex it won't report as an offender.

### DIFF
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -126,7 +126,7 @@ function getModuleRealName(resolved) {
   return getFilePackageName(resolved);
 }
 
-function reportIfMissing(context, deps, depsOptions, node, name) {
+function reportIfMissing(context, deps, depsOptions, node, name, ignore) {
   // Do not report when importing types
   if (node.importKind === 'type' || (node.parent && node.parent.importKind === 'type') || node.importKind === 'typeof') {
     return;
@@ -144,6 +144,10 @@ function reportIfMissing(context, deps, depsOptions, node, name) {
   // fallback on original name in case no package.json found
   const packageName = getModuleRealName(resolved) || getModuleOriginalName(name);
 
+  if (packageName.match(ignore)) {
+    return;
+  }
+  
   const isInDeps = deps.dependencies[packageName] !== undefined;
   const isInDevDeps = deps.devDependencies[packageName] !== undefined;
   const isInOptDeps = deps.optionalDependencies[packageName] !== undefined;
@@ -200,6 +204,7 @@ module.exports = {
           'peerDependencies': { 'type': ['boolean', 'array'] },
           'bundledDependencies': { 'type': ['boolean', 'array'] },
           'packageDir': { 'type': ['string', 'array'] },
+          'ignore': { 'type': 'regexp' },
         },
         'additionalProperties': false,
       },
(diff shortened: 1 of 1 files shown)


src/rules/no-extraneous-dependencies.js
```diff
@@ -200,6 +204,7 @@ module.exports = {
           'peerDependencies': { 'type': ['boolean', 'array'] },
           'bundledDependencies': { 'type': ['boolean', 'array'] },
           'packageDir': { 'type': ['string', 'array'] },
+          'ignore': { 'type': 'regexp' },
```
ljharb, line 207: it's a very bad idea to allow regex strings in eslint configs; that's how you get CVEs. this should be a glob string instead.

i'd also expect it to be allowed to be an array of glob strings.

ljharb (changes requested):
Can you elaborate on why this is needed?

ljharb:
Also, see #903.

Aghassi:
@ljharb Sure thing. We are migrating our codebase to bazel, and as an interim step for users we use a preinstall script to generate the bazel built versions of 1st party dependencies. We then specify the built artifacts as part of a `workspace` to avoid dealing with symlinking oddities under yarn. However, we are not using the `workspace` in the normal sense since we just have a single package.json that manages the projects while we are in this migration state. As such, we never declare the packages in the dependencies object, but they are guaranteed to be there. The `workspace` field is currently ignored by this rule (and also is not structure in a standard way that we can make assumptions around it IMO), so I added the `ignore` option as a way to eject for first party things using our companies `@` scope. Maybe this isn't a use case for everyone, but I wanted to at least start discussion of why we needed this. I'm using `patch-package` internally to do this for now, so if it isn't taken in upstream that's fine. I wanted to open a draft PR to at least show what we had done and leave a paper trail 😄 . I'll read through those issues.

src/rules/no-extraneous-dependencies.js
Aghassi: Thanks, that's a good idea!

ljharb:
I'm confused; bazel has workspaces support (using links). What oddities?

Aghassi:
> I'm confused; bazel has workspaces support (using links). What oddities?

I think we are crisscrossing the word `workspaces`.  Just to confirm, we are talking about `yarn workspaces`. And in the above, I should clarify that we are not yet running the main projects under bazel, but we are generating dependencies that these projects rely on under bazel. So essentially we are giving developers a local experience that is still based in the NPM ecosystem, but linking first party artifacts that bazel generated. The issues were hit were when you use `link:` yarn fails to install because of missing node_modules if I recall. Using `file:` sort of solves the problem, but then it doesn't get updated when you reinstall over existing `node_modules` folder. It's quite possible we are going about this the "wrong" way for now, but using `yarn workspaces` has removed all caching invalidation issues and linking issues of the bazel built artifacts to these projects that aren't under bazel yet.

Does that make sense? I'm probably doing a poor job of explaining it :/

shellscape:
I dropped a quick comment on a very common ecosystem use case that demonstrates the need for this feature here: https://github.com/import-js/eslint-plugin-import/issues/903#issuecomment-893588446

Aghassi:
We solved this issue using https://github.com/import-js/eslint-plugin-import#importinternal-regex

PR #2041, opened 2021-05-02
Proposal: (no-unused-modules) Restrict report to the identifier on named exports.

edupsousa:
First of all thanks for the tool. I would like to propose a change in the reporting behavior for the rule no-unused-modules:

Currently the rule "flags" all the ExportNamedDeclaration body for unused exports, but when using tools like vscode-eslint the editor gets pretty convoluted with all lines being painted when you are working on a new exported function. This behavior also creates a misleading situation when exporting multiple variables within a unique export keyword, if one of the variables is unused, all the variables in the block gets flagged.

The new behavior is consistent with rules like "@typescript-eslint/no-unused-vars" where only the identifier is flagged if the variable is unused. I attached 2 images showing the current and the new behavior on my IDE.

no_unused_modules_current_behavior

no_unused_modules_new_behavior

### DIFF
--- a/src/rules/no-unused-modules.js
+++ b/src/rules/no-unused-modules.js
@@ -77,10 +77,10 @@ function forEachDeclarationIdentifier(declaration, cb) {
       declaration.type === TS_TYPE_ALIAS_DECLARATION ||
       declaration.type === TS_ENUM_DECLARATION
     ) {
-      cb(declaration.id.name);
+      cb(declaration.id.name, declaration.id.loc);
     } else if (declaration.type === VARIABLE_DECLARATION) {
       declaration.declarations.forEach(({ id }) => {
-        cb(id.name);
+        cb(id.name, id.loc);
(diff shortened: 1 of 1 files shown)


ljharb:
I think the challenge here is that it's not warning on the *value* being unused, it's warning on the export itself being unused.

In other words, it might still be used in the same file - just not outside the file. Can we target the `export` as well as the identifier, or can you only target a single node?

edupsousa:
@ljharb i agree with you that targeting the `export` would show better what the rule means, but the context.report() method expects a single `loc` parameter continuous from start to end, so if you have something in between the export and the identifier (like the lines 6-7 of the screenshots) this behavior will not work. :-(

ljharb:
I'm all for making the targeted node more useful! I'm not sure this actually achieves that, though - it just makes it less noisy, which isn't always better.

Issue #2046, opened 2021-05-06
Npm Audit fails


zabibabar:
Moderate        Regular Expression Deinal of Service                          

  Package         hosted-git-info                                               

  Patched in      >=3.0.8                                                       

  Dependency of   eslint-plugin-import [dev]

  Path            eslint-plugin-import > read-pkg-up > read-pkg >
                  normalize-package-data > hosted-git-info

  More info       https://npmjs.com/advisories/1677

ljharb:
Like most CVEs, this is a false positive, since the content comes from your own package.json.

Either way, the place to file this issue is in `normalize-package-date` or `hosted-git-info`, not anywhere else.

mgwalker:
This issue is **actually** in `eslint-plug-import` because it relies on `read-pkg-up@^2.0.0`:

https://github.com/benmosher/eslint-plugin-import/blob/master/package.json#L113

Here's the dependency tree:
```
eslint-plugin-import@2.22.1
|__ read-pkg-up@2.0.0
    |__ read-pkg@2.0.0
        |__ normalize-package-data@2.5.0
            |__ hosted-git-info@2.8.9
```

That dependency on `read-pkg-up@2.0.0` prevents users of this library from getting a newer version of `hosted-git-info`. The entire dependency chain could be updated at once by using a newer version of `read-pkg-up`, which is up to 8.0.0 now. Until `eslint-plugin-import` updates its dependency on `read-pkg-up`, everyone who uses it will get npm audit and GitHub security alerts (along with other static analysis tools) because they will necessarily depend on a known-vulnerable version of `hosted-git-info`.

ljharb:
@mgwalker that's not how the ecosystem works. The affected package is an implementation detail of normalize-package-data, which is an implementation detail of read-pkg, which is an implementation detail of read-pkg-up.

Unfortunately, vulnerability warnings are breaking the abstraction here, but that still doesn't make it something we can fix. `read-pkg-up` can unfortunately never be upgraded here, because higher versions drop support for node versions we continue to support.

Most of these audit/security alerts are false positives; this is a sad reality of the incentives in the security ecosystem. In this case, the "vulnerability" would be that you'd have to have a package.json on your own machine, in your own code, that contained a malicious URL, and the damage caused would be no broader than "your linter would run forever, and CI would time out". It's a non-problem.

[...]

PR #2047, opened 2021-05-06
[Refactor] switch to an internal replacement for `pkg-up` and `read-pkg-up`

mgwalker:
Inspired by #2046, I dug into what [`read-pkg-up`](https://github.com/sindresorhus/read-pkg-up/tree/v2.0.0) provides and found it to be very simple, for this particular case. Its downstream dependencies build in some safety and options that aren't necessary when you already know you're working with a `package.json`, and in this library's case, when you know you don't want to "normalize" it. Since `read-pkg-up` drags several libraries along with it, at least two of which don't actually add any functionality to `eslint-plugin-import`, and its use here is pretty straightforward, it seemed moderately reasonable to me to reproduce the functionality internally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,14 +104,11 @@
     "doctrine": "^2.1.0",
     "eslint-import-resolver-node": "^0.3.6",
     "eslint-module-utils": "^2.6.2",
-    "find-up": "^2.0.0",
     "has": "^1.0.3",
     "is-core-module": "^2.6.0",
     "is-glob": "^4.0.1",
     "minimatch": "^3.0.4",
     "object.values": "^1.1.4",
-    "pkg-up": "^2.0.0",
-    "read-pkg-up": "^3.0.0",
     "resolve": "^1.20.0",
     "tsconfig-paths": "^3.11.0"
   }
--- a/utils/readPkgUp.js
+++ b/utils/readPkgUp.js
@@ -35,7 +35,7 @@ function stripBOM(str) {
  * THE SOFTWARE.
  */
 exports.default = function readPkgUp(opts) {
-  const fp = pkgUp(opts.cwd);
+  const fp = pkgUp(opts);
 
   if (!fp) {
     return {};
--- a/utils/readPkgUp.js
+++ b/utils/readPkgUp.js
@@ -1,0 +1,52 @@
+'use strict';
+exports.__esModule = true;
+
+const fs = require('fs');
+const pkgUp = require('./pkgUp').default;
+
+function stripBOM(str) {
+  return str.replace(/^\uFEFF/, '');
+}
+
+/**
+ * Derived significantly from read-pkg-up@2.0.0. See license below.
+ * 
+ * @copyright Sindre Sorhus
+ * MIT License
+ *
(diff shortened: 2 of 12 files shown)


utils/readPkgUp.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+exports.__esModule = true;
```
ljharb, line 2: this shouldn't be here; babel adds it.

utils/readPkgUp.js
```diff
@@ -0,0 +1,55 @@
...
+
+const fs = require('fs');
+const findUp = require('find-up');
+
+function stripBOM(str) {
+  if(str.charCodeAt(0) === 0xFEFF) {
+    return str.slice(1);
+  }
+  return str;
```
ljharb, line 11:
```suggestion
  return str.replace(/^\uFEFF/, '');
```

ljharb (changes requested):
I'm not super stoked on copy-pasting, but removing this dep is valuable.

utils/readPkgUp.js
mgwalker: When does babel add it? I was just following the pattern from everything else in `utils/`. If I remove these two lines, the tests fail on some `import` statements.

silverwind:
The replacement `find-up@2.1.0` is very out of date and contains security vulnerabilites to boot. I'd say drop the dependency completely and replace it with a small function:

````js
const {dirname} = require('path');
const {accessSync} = require('fs');

function findUpSync(dir, file) {
  const path = join(dir, file);
  try {
    accessSync(path);
    return path;
  } catch {}
  const parent = dirname(dir);
  return parent === dir ? null : findSync(parent, file);
}
````

Use via `findUpSync(filePath, 'package.json')`.

File: utils/readPkgUp.js (hunk 1 above)
Comment (ljharb) [reply utils/readPkgUp.js:2]: huh, i guess we'll leave it in for now.

**Comment by ljharb** [comment]
There isn’t; you could make the changes in two PRs, and we could land and publish the utils one first?

altho i wouldn’t expect that to be needed - we use linklocal in this repo so everything should be using the latest local code.

**Comment by mgwalker** [comment]
Oh, well in that case, maybe just switching the `dependency` to the local package will do the job. Pushed a new commit.

**Comment by mgwalker** [comment]
Also as you mentioned, the [CVE got updated](https://www.npmjs.com/advisories/1677/versions), so I guess closing this PR as moot would also make sense. 🤷🏻‍♂️

File: package.json [hunk 3]
```diff
@@ -103,14 +103,13 @@
     "debug": "^2.6.9",
     "doctrine": "1.5.0",
     "eslint-import-resolver-node": "^0.3.4",
-    "eslint-module-utils": "^2.6.0",
+    "eslint-module-utils": "file:./utils",
```
ljharb, line 106: we definitely can't do this; it will break for downstream consumers.

ljharb (changes requested):

ljharb:
Another alternative from a bit ago is #1017.

ljharb (approved):
went ahead and replaced `pkg-up` as well, and rebased. this should be good to go.

PR #2051, opened 2021-05-07
[New] add `sort-by-length` option to order rule

azat-io:
Add option to `import/order` rule for sorting imports by length.

For example:

```js
// { groups: ['external', 'index'], 'sort-by-length': { order: 'desc' } }

import { foo , bar } from 'module-name/path/to/specific/un-exported/file';
import { export1 as alias } from 'module-name-1';
import { export1, export2 } from 'module-name-2';
import defaultExport from 'module-name-3';
import { export3 } from 'module-4';
import * as name from 'module-5';

import index from './';
import './module';
```

### DIFF
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -260,7 +260,7 @@ Sort the order within each group in import string length based on **import path*
 Example setting:
 ```js
 'sort-by-length': {
-  order: 'desc', /* sort in ascending order. Options: ['ignore', 'asc', 'desc'] */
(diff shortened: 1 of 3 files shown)


ljharb:
Is the intention here to sort by length of the module specifier only, or by the length of the whole line? if the latter, what happens to multiline named imports?

azat-io:
Currently, I use ESLint Perfectionist for sorting.
https://perfectionist.dev/rules/sort-imports

ljharb:
I'm confused why you spent the last 4.5 years building a replacement rather than replying to my question, but enjoy, i guess.

PR #2071, opened 2021-05-14
[Fix] `order`: fix alphabetical sorting

### LINKED ISSUE #2064: v2.23 broke alphabetical sorting
[This](https://github.com/benmosher/eslint-plugin-import/commit/1c10e79dc4761d69c5ce26736541f8d024ce7339#diff-2825fcc427c846fe18d501930e1a3d83be722027f1f643801afdcc2c255dd4a8R268) and [this](https://github.com/benmosher/eslint-plugin-import/commit/1c10e79dc4761d69c5ce26736541f8d024ce7339#diff-2825fcc427c846fe18d501930e1a3d83be722027f1f643801afdcc2c255dd4a8R293) changes cause the following weird effect.

With imports like this:
```js
import React from 'react'
import { BrowserRouter } from 'react-router-dom'
...
```
func `mutateRanksToAlphabetize` produces following groups:
```js
{
  '0.8': [ 'react-value', 'react-router-dom-value' ],
  ...
}
```
Then this group gets sorted alphabetically, and obviously turns into `[ 'react-router-dom-value', 'react-value' ]`. While normally, `'react' < 'react-router-dom' === true`.

So after updating to 2.23 I suddenly have warnings about wrong imports order all over the project . 

Probably, you may just use `|` instead of `-` to separate importKind from importName?

### LINKED ISSUE #2070: `import/order`: treat `type` imports like regular imports?
How can we treat type imports like a regular import (i.e. the behavior before v2.23.0)? I would expect the following import order to be fine:
```ts
import type { Dictionary } from 'lodash';
import { isEmpty } from 'lodash';

import { something } from './relative';
```
... but the first line gets highlighted:

> There should be no empty line within import group

Also, eslint is now running forever, so I can't automatically fix this.

Rule config:
```json
{
  "import/order": [
    "warn",
    {
      "alphabetize": { "order": "asc" },
      "groups": [
        ["builtin", "external"],
        "internal",
        ["parent", "sibling", "index"],
        "object"
      ],
      "newlines-between": "always",
      "pathGroups": [
        {
          "pattern": "*.json",
          "group": "object",
          "patternOptions": { "matchBase": true }
        }
      ]
    }
  ]
}

```

grit96:
Fixes #2064

Added a new test for the broken scenario.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Fixed
 - [`newline-after-import`]: fix crash with `export {}` syntax ([#2063], [#2056], thanks [@ljharb])
 - `ExportMap`: do not crash when tsconfig lacks `.compilerOptions` ([#2067], thanks [@ljharb])
+- [`order`]: fix alphabetical sorting ([#2071], thanks [@grit96])
 
 ## [2.23.0] - 2021-05-13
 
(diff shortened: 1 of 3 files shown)


ljharb (approved):
Nice! I added another test case as well.

PR #2083, opened 2021-05-17
Fix a couple of `no-cycle` bugs

cherryblossom000:
This fixes these two bugs:

```ts
// a.ts
// false positive: reports a cycle even though b.ts is only importing types
import { foo } from './b'

// b.ts
import type { Bar } from './a'
```

```js
// a.js
// @flow
// false negative: doesn't report a cycle even though b.js is importing the value bar
import { foo } from './b'

// b.js
// @flow
import { bar, type Baz } from './a'
```

There isn't an open issue for these, but these issues were relatively quick and easy for me to fix so I don't really mind if this PR gets rejected. Also, these bugs were introduced by me in #1974 (oops), so I wanted to fix them.

If you would like me to split these two bug fixes into two PRs, I'm more than happy to do that as well.

### DIFF
--- a/src/rules/no-cycle.js
+++ b/src/rules/no-cycle.js
@@ -84,18 +84,26 @@ module.exports = {
         traversed.add(m.path);
 
         for (const [path, { getter, declarations }] of m.imports) {
-          if (path === myPath) return true;
           if (traversed.has(path)) continue;
-          for (const { source, isOnlyImportingTypes } of declarations) {
-            if (ignoreModule(source.value)) continue;
+          const toTraverse = [...declarations].filter(({ source, isOnlyImportingTypes }) =>
+            !ignoreModule(source.value) &&
             // Ignore only type imports
-            if (isOnlyImportingTypes) continue;
+            !isOnlyImportingTypes
+          );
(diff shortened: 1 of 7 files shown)


src/rules/no-cycle.js
```diff
@@ -84,18 +84,26 @@ module.exports = {
...
-                mget: getter,
-                route: route.concat(source),
-              });
+          a.ts:
+          import { foo } from './b' // should not be reported as a cycle
+
+          b.ts:
+          import type { Bar } from './a'
+          */
+          if (path === myPath && toTraverse.length) return true;
```
ljharb, line 103:
```suggestion
          if (path === myPath && toTraverse.length > 0) return true;
```

ljharb (approved):
Seems good, thanks!

Nokel81:
I am wondering when a new patch release will happen. I would like to use `no-cycle` in my codebase but we would rely heavily on this bug fix. Thanks.

ljharb:
v2.23.3 is released.

Nokel81:
Thank you very much

PR #2086, opened 2021-05-17
Match against "pattern*" as RegExp

anthonyp-usafacts:
RE: https://github.com/benmosher/eslint-plugin-import/issues/1281#issuecomment-461200205

We want to allow people to use a wildcard string in their core-modules settings.

This will convert "pattern*" into `RegExp('^pattern.*$')` while also still enforcing a direct match on core-module strings that do not have the wildcard.

Per the issue thread, the allowable scenario was that the string ends in a wildcard, and nothing else.

### DIFF
--- a/utils/resolve.js
+++ b/utils/resolve.js
@@ -87,7 +87,12 @@ function relative(modulePath, sourceFile, settings) {
 function fullResolve(modulePath, sourceFile, settings) {
   // check if this is a bonus core module
   const coreSet = new Set(settings['import/core-modules']);
-  if (coreSet.has(modulePath)) return { found: true, path: null };
+  if (coreSet.some(c => {
+    const regex = new RegExp(`^${c.slice(-1) === '*' ? c.replaceFirst('.$', '.*') : c}$`);
+    return regex.test(modulePath);
+  })) {
+    return { found: true, path: null };
+  }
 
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
Allowing regexes in eslint configs is a fast path to ReDOS CVEs. We can use minimatch to support globs, however.

anthonyp-usafacts:
> Allowing regexes in eslint configs is a fast path to ReDOS CVEs. We can use minimatch to support globs, however.

Very true. I'm kinda hoping the issue author jumps in here because I like people solving their own issues in community-driven repos. If I don't hear back I'll revise this later.

ljharb:
After 2 years it's unlikely they will do so; any help from you is appreciated.

We'll also need a test covering this behavior.

penx:
> Very true. I'm kinda hoping the issue author jumps in here because I like people solving their own issues in community-driven repos.

I understand, but I've not written an eslint rule. When I looked 2 years ago I didn't know where to begin and didn't know anyone who could help me with it. Now you have a draft PR I have some context of what needs doing - although I still find it pretty confusing 😄

If someone wants to help walk me through it I'd be happy to give it a go.

Is there an existing test I could point a debugger at to step through this code and figure out what it's doing?

Issue #2084, opened 2021-05-17
`import/order` throws error in every order


vtaits:
Input:

```typescript
/* eslint-disable @typescript-eslint/no-explicit-any */

import {
  Form,
} from 'react-final-form';
import type {
  FormProps as FinalFormProps,
} from 'react-final-form';
import {
  useMemo,
  useCallback,
} from 'react';
import type {
  ReactElement,
  ReactNode,
} from 'react';
import {
  serialize as formSchemaSerialize,
  parse as formSchemaParse,
  mapFieldErrors as formSchemaMapFieldErrors,
} from '@vtaits/form-schema';
import type {
  GetFieldSchema,
} from '@vtaits/form-schema';
```

Result:

```
   9:1  error  `react` import should occur before import of `react-final-form`                import/order
  17:1  error  `@vtaits/form-schema` import should occur before import of `react-final-form`  import/order
```

Trying to fix:

```typescript
/* eslint-disable @typescript-eslint/no-explicit-any */

import {
  useMemo,
  useCallback,
} from 'react';
import type {
  ReactElement,
  ReactNode,
} from 'react';
import {
  serialize as formSchemaSerialize,
  parse as formSchemaParse,
  mapFieldErrors as formSchemaMapFieldErrors,
} from '@vtaits/form-schema';
import type {
  GetFieldSchema,
} from '@vtaits/form-schema';
import {
  Form,
} from 'react-final-form';
import type {
  FormProps as FinalFormProps,
} from 'react-final-form';
```

Result:

```
  11:1  error  `@vtaits/form-schema` import should occur before import of `react`  import/order
  19:1  error  `react-final-form` import should occur before import of `react`     import/order
```

2nd try:

```typescript
/* eslint-disable @typescript-eslint/no-explicit-any */

import {
  serialize as formSchemaSerialize,
  parse as formSchemaParse,
  mapFieldErrors as formSchemaMapFieldErrors,
} from '@vtaits/form-schema';
import type {
  GetFieldSchema,
} from '@vtaits/form-schema';
import {
  Form,
} from 'react-final-form';
import type {
  FormProps as FinalFormProps,
} from 'react-final-form';
import {
  useMemo,
  useCallback,
} from 'react';
import type {
  ReactElement,
  ReactNode,
} from 'react';
```

Result:

```
  11:1  error  `react-final-form` import should occur before import of `@vtaits/form-schema`  import/order
[...]

PR #2087, opened 2021-05-17
[Fix] `order`: restore default behaviour unless `type` is in groups

grit96:
#2070 and #2084

### DIFF
--- a/tests/src/rules/order.js
+++ b/tests/src/rules/order.js
@@ -2247,13 +2247,12 @@ context('TypeScript', function () {
             {
               code: `
                 import c from 'Bar';
+                import type { C } from 'Bar';
                 import b from 'bar';
                 import a from 'foo';
+                import type { A } from 'foo';
 
                 import index from './';
-
-                import type { C } from 'Bar';
-                import type { A } from 'foo';
               `,
               parser,
               options: [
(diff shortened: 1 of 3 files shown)


tests/src/rules/order.js
```diff
@@ -2258,14 +2302,14 @@ context('TypeScript', function () {
               parser,
               options: [
                 {
-                  groups: ['external', 'index'],
+                  groups: ['external', 'index', 'type'],
```
ljharb, line 2305: confirming: this was the test you added, thus, the only impact is on the feature from that PR?

tests/src/rules/order.js
```diff
@@ -2281,7 +2325,7 @@ context('TypeScript', function () {
...
```
ljharb, line 2328: same here?

tests/src/rules/order.js
```diff
@@ -2303,35 +2347,130 @@ context('TypeScript', function () {
...
                 {
                   groups: ['external', 'index'],
+                  alphabetize: { order: 'desc' },
+                },
+              ],
+              errors: [
+                {
+                  message: semver.satisfies(eslintPkg.version, '< 3')
+                    ? '`bar` import should occur before import of `Bar`'
+                    : /(`bar` import should occur before import of `Bar`)|(`Bar` import should occur after import of `bar`)/,
```
ljharb, line 2441: why both, in eslint >= 4?

ljharb (approved):
Can you confirm that all changed tests were ones you added in your original PR?

grit96:
@ljharb yes, I reverted the test changes from the previous commit and added new ones for the type group

tests/src/rules/order.js
grit96: yes, I reverted the test changes from the previous commit and added new ones for the type group

tests/src/rules/order.js
grit96: yes, I reverted the test changes from the previous commit and added new ones for the type group

tests/src/rules/order.js
grit96: I think I just copied this from the other tests, I can change it if required.

mmiszy:
Could you, please, stop introducing breaking changes in `patch` releases? Thank you :P

ljharb:
@mmiszy the word you’re looking for is “bugs”. Feel free to request a refund though.

mmiszy:
@ljharb there's no need to be sarcastic and condescending.

ljharb:
@mmiszy yes, i agree that your initial comment was both rude and unnecessary.

mmiszy:
@ljharb I'm sorry that you feel that way about my feedback.

ljharb:
@mmiszy it's not feedback. nobody intentionally introduces breaking changes in a patch release - it's a bug that it's happened. Bugs are unavoidable, and can never be fully stopped, so asking for that can't possibly ever achieve anything except to make someone feel bad that they introduced a bug (which in this case, wasn't even me).

Please stop pretending like your comment was made in good faith. It was clearly snark born out of frustration, and that should be kept to oneself.

mmiszy:
@ljharb I'm sorry you see it this way 😥

PR #2090, opened 2021-05-18
[Fix] `no-restricted-paths`: false-positive matches

### LINKED ISSUE #2089: no-restricted-path: incorrect zone.target match
In this PR https://github.com/benmosher/eslint-plugin-import/pull/2036 `contains-path` was updated to version 1.0.0. Looks like, there is a regression related to false-positive match of folders like `/foo-bar` to `zone.target: '/foo'`

**Folder structure:** 
Screenshot 2021-05-18 at 14 58 43

**Steps:**
1. Declare no-restricted-paths so imports from `'/server'` are forbidden inside `/server/two`. (Pay attention, zone target is `/two`, not `/two-new`. 
2. Import `/server/one/a.js` from `/server/two-new/a.js`

**Actual results:**
Import is forbidden

**Expected result:**
Import is allowed

See a corresponding issue: https://github.com/jonschlinkert/contains-path/issues/5
See a PR, with test case, reproducing the issue: https://github.com/benmosher/eslint-plugin-import/pull/2090

malykhinvi:
Fixes #2089

In this PR I removed `contains-paths` dependency and used node's `path.relative` to identify if a file is inside the `target`. I used an approach similar to the one used in `src/core/importType`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## [Unreleased]
 
+### Fixed
+- [`no-restricted-paths`]: fix false positive matches ([#2090], thanks [@malykhinvi])
+
 ### Changed
 - [Docs] Add `no-relative-packages` to list of to the list of rules ([#2075], thanks [@arvigeus])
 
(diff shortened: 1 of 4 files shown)


ljharb (approved):
Thanks, this is great. Normally I firmly believe that more deps is better, but I'm stoked to remove this particular dep.

PR #2094, opened 2021-05-24
no-unused-modules: support dynamic imports

aladdin-add:
All occurences of `import('...')` are treated as namespace imports
(`import * as X from '...'`)

Skip tests for lower versions of eslint

Use strict

Address some review comments

Straighten if-else

Replace loops with foreach

Support newer babel parser

Update updateImportUsage

Support @babel/eslint-parser


ljharb:
Is this intended to replace #1660? Please don't create duplicate PRs; instead, you can post a link to the branch on the original PR and i can update that.

Either way, this one isn't fully rebased.

(at this point, please keep this PR open; i'll now need to keep both PRs in sync forever, since opened PRs leave a ref behind on a repo forever)

aladdin-add:
oh, my bad!  I just mean to open a PR to my fork.

PR #2097, opened 2021-05-25
[Fix] `no-extraneous-dependencies`: fix package name algorithm

### LINKED ISSUE #2058: "peerDependencies" option has no effect in "import/no-extraneous-dependencies" rule since 2.23.0
With version `2.23.0` eslint fails due to `peerDependency` imports in our code:

`1:1  error  '{PACKAGE_NAME}' should be listed in the project's dependencies. Run 'npm i -S {PACKAGE_NAME}' to add it  import/no-extraneous-dependencies`

Our package is scoped (`@org-name/package-name`), not sure if it makes any difference.

Version `2.22.1` works fine.

Our config looks like this:

```
      'error',
      {
        devDependencies: [
          //
        ],
        peerDependencies: true,
        packageDir: './',
      },
    ],
```

### LINKED ISSUE #2065: `import/no-extraneous-dependencies` and `import/order` false postives since version 2.23
Since the update to version 2.23, we have a lot of false positives in our project being part of two categories :
 
* `error  'rxjs/operators' should be listed in the project's dependencies. Run 'npm i -S rxjs/operators' to add it  import/no-extraneous-dependencies` : `rxjs/operators` is a sub-module of `rxjs` so I don't have to add it to project dependencies.
*```error  `@models/partner/partner-id` import should occur before import of `@models/partner/partner`               import/order``` : I can fix these issues, but it seems that the order has changed between two versions and I don not understand why and if it is on purpose ? I can not find mention to this in the changelogs.
* 
Cheers 🤙

### LINKED ISSUE #2066: Allow to require from internal project file with `import/no-extraneous-dependencies`
When importing or requireing an internal file instead of the raw project export, `import/no-extraneous-dependencies` fails. This is a valid use case, so for example requires to `@apollo/client/link/error` should be successful if the `@apollo/client` package is included in `package.json` file, but right now it's failing since it checks for the full string.

### LINKED ISSUE #2078: v2.23 causes `no-extraneous-dependencies` errors for related `@types/` packages that aren't imported
Since upgrading to v2.23 I'm seeing lot's of errors in my TypeScript codebase related to `@types` dependencies, even though they are not directly being imported. for example:

```ts
import React from "react";
import * as THREE from "three";
import { OrbitControls } from "three/examples/jsm/controls/OrbitControls";
```

Is throwing:

```shell
  3:1  error  '@types/react' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
  4:1  error  '@types/three' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
  5:1  error  '@types/three' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
```

paztis:
- resolve nested package.json problems (a/b/c import will check a, a/b and a/b/c)
- resolve renamed dependencies: checks the import name and the resolve package name

Fixes #2066. Fixes #2065. Fixes #2058. Fixes #2078.

### DIFF
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -126,6 +126,36 @@ function getModuleRealName(resolved) {
   return getFilePackageName(resolved);
 }
 
+function checkDependencyDeclaration(deps, packageName) {
+  // in case of sub package.json inside a module
+  // check the dependencies on all hierarchy
+  const packageHierarchy = [];
+  const packageNameParts = packageName.split('/');
+  packageNameParts.forEach((namePart, index) => {
+    if (!namePart.startsWith('@')) {
+      const ancestor = packageNameParts.slice(0, index + 1).join('/');
+      packageHierarchy.push(ancestor);
+    }
+  });
+
+  return packageHierarchy.reduce((result, ancestorName) => {
+    return {
+      isInDeps: result.isInDeps || deps.dependencies[ancestorName] !== undefined,
+      isInDevDeps: result.isInDevDeps || deps.devDependencies[ancestorName] !== undefined,
+      isInOptDeps: result.isInOptDeps || deps.optionalDependencies[ancestorName] !== undefined,
+      isInPeerDeps: result.isInPeerDeps || deps.peerDependencies[ancestorName] !== undefined,
+      isInBundledDeps:
+        result.isInBundledDeps || deps.bundledDependencies.indexOf(ancestorName) !== -1,
+    };
+  }, {
+    isInDeps: false,
+    isInDevDeps: false,
+    isInOptDeps: false,
+    isInPeerDeps: false,
+    isInBundledDeps: false,
+  });
+}
+
 function reportIfMissing(context, deps, depsOptions, node, name) {
   // Do not report when importing types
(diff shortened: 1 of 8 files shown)


src/rules/no-extraneous-dependencies.js
```diff
@@ -126,6 +126,36 @@ function getModuleRealName(resolved) {
   return getFilePackageName(resolved);
 }
 
+function checkDependencyDeclaration(deps, packageName) {
+  // in case of sub package.json inside a module
+  // check the dependencies on all hierarchy
+  const packageHierarchy = [];
+  const packageNameParts = packageName.split('/');
```
ljharb, line 133: on windows, this might need `.split(path.sep)`?

ljharb (changes requested):
The fix looks good, but we'll need to add some regression tests, so this can't happen again.

src/rules/no-extraneous-dependencies.js
paztis: PackageName variable is @types/react or lodash/get

It is always / independently of the platform

paztis:
unit tests added

ljharb (approved):

PR #2112, opened 2021-06-02
[New] `no-namespace`: Add `ignore` option

### LINKED ISSUE #1903: [no-namespace] allow namespace imports for specific modules

aberezkin:
Closes #1916
Closes #1903

Hi, we want to use the rule `no-namespace` in our project but we have a `i18n` codegen tool that forces us to use wildcard imports for certain files that follow some pattern (`\S+\.i18n`, to be precise). Also I've seen that there's people with similar issue.

I thought that we can implement something similar to `argsIgnorePattern/varsIgnorePattern` in [core eslint rules](https://github.com/eslint/eslint/blob/master/lib/rules/no-unused-vars.js#L107). 

I made a test for that case and updated the docs as well. Please let me know if there's something else I should do to get this merged.

### DIFF
--- a/docs/rules/no-namespace.md
+++ b/docs/rules/no-namespace.md
@@ -5,6 +5,12 @@ Enforce a convention of not using namespace (a.k.a. "wildcard" `*`) imports.
 +(fixable) The `--fix` option on the [command line] automatically fixes problems reported by this rule, provided that the namespace object is only used for direct member access, e.g. `namespace.a`.
 The `--fix` functionality for this rule requires ESLint 5 or newer.
 
+### Options
+
+This rule supports the following options:
+
+- `ignore`: array of glob strings for modules that should be ignored by the rule.
+
 ## Rule Details
 
 Valid:
@@ -13,6 +19,11 @@ Valid:
 import defaultExport from './foo'
 import { a, b }  from './bar'
 import defaultExport, { a, b }  from './foobar'
+```
+
+```js
+/* eslint import/no-namespace: ["error", {ignore: ['*.ext']] */
+import * as bar from './ignored-module.ext';
 ```
 
 Invalid:
--- a/src/rules/no-namespace.js
+++ b/src/rules/no-namespace.js
@@ -3,6 +3,7 @@
  * @author Radek Benkel
  */
 
+import minimatch from 'minimatch';
 import docsUrl from '../docsUrl';
 
 //------------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


aberezkin:
Hi @ljharb! Can you look at this PR? It's quite similar to #1916. But add a little more versatile option.

gairon:
Hi @aberezkin, any updates on that ? Have the same problem with this rule.

aberezkin:
Hi @gairon. I would be happy to provide some update but I'm can't merge it myself since I'm not a maintainer 😄

gairon:
Who else can help with verifying and pushing a really cool feature?
@ljharb @benmosher ?

src/rules/no-namespace.js
```diff
@@ -17,12 +17,30 @@ module.exports = {
...
+        },
+      },
+    }],
   },
 
   create: function (context) {
+    const firstOption = context.options[0] || {};
+    const ignoreFromPatternExpression =
+      firstOption.ignoreFromPattern &&
+      new RegExp(firstOption.ignoreFromPattern);
```
ljharb, line 34: eslint rules should never accept regex patterns; this is a common source of ReDOS CVEs. Globs are the only available choice here.

ljharb (changes requested):

src/rules/no-namespace.js
aberezkin: Hi! I rewrote it to use minimatch instead of regex.

docs/rules/no-namespace.md
```diff
@@ -5,6 +5,12 @@ Enforce a convention of not using namespace (a.k.a. "wildcard" `*`) imports.
 +(fixable) The `--fix` option on the [command line] automatically fixes problems reported by this rule, provided that the namespace object is only used for direct member access, e.g. `namespace.a`.
 The `--fix` functionality for this rule requires ESLint 5 or newer.
 
+### Options
+
+This rule supports the following options:
+
+- `ignoreFromPattern`: regular expression for modules that should be ignored by the rule.
```
ljharb, line 12: I'd probably just name this "ignore", and let it be an array of glob strings (the description needs updating)

docs/rules/no-namespace.md
```diff
@@ -15,6 +21,11 @@ import { a, b }  from './bar'
 import defaultExport, { a, b }  from './foobar'
 ```
 
+```js
+/* eslint import/no-namespace: ["error", {ignoreFromPattern: '\\S+\\.ext'}] */
```
ljharb, line 25: the example also needs updating

src/rules/no-namespace.js
```diff
@@ -17,12 +18,34 @@ module.exports = {
...
     fixable: 'code',
-    schema: [],
+    schema: [{
+      type: 'object',
+      properties: {
+        ignoreFromPatterns: {
+          type: 'array',
+          items: {
+            type: 'string',
+          },
```
ljharb, line 28:
```suggestion
          type: 'array',
          items: {
            type: 'string',
          },
          uniqueItems: true,
```

aberezkin:
@ljharb Fixed! :)

ljharb (approved):

gairon:
Hi @aberezkin,
thanks for new option.
Found a typo in docs:
`/* eslint import/no-namespace: ["error", {ignore: ['*.ext']] */`
there missed `}` before last `]`.

ljharb:
@gairon thanks, fixed in 4f0f560

PR #2138, opened 2021-06-25
[Docs] extensions: removed incorrect cases

wenfangdu:
When using `'import/extensions': ['error', 'always']`, `import express from 'express'` is allowed.


ljharb (approved):
Thanks!

PR #2146, opened 2021-07-06
[Fix] `extensions`/`importType`: fix isScoped treating @/abc as scoped module

### LINKED ISSUE #2145: Set @/whatever import path as not scoped package
Actually imports like these are matching **scopedRegExp** in **isScope** function. In the example below is not a scoped package, is only an alias to **src** folder

```js
import UnitDevice from '@/components/UnitDevice'
```

So my proposal is to change from

```js
const scopedRegExp = /^@[^/]*\/?[^/]+/
```

to this:

```js
const scopedRegExp = /^@[^/]+\/?[^/]+/
```

which i have tested

rperello:
Closes #2145


.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
```
aladdin-add, line 1: why committing this?

aladdin-add (approved):

.vscode/settings.json
rperello: Apologize, that was a mistake. That should not be there

ljharb (approved):

PR #2150, opened 2021-07-10
[Fix] Process halts when body is undefined. 

senputra:
The change will do a null check on the `body` variable before calling `body.forEach()`

### DIFF
--- a/src/rules/namespace.js
+++ b/src/rules/namespace.js
@@ -78,7 +78,7 @@ module.exports = {
             }
           }
         }
-        body.forEach(processBodyStatement);
+        if (body) body.forEach(processBodyStatement);
       },
 
       // same as above, but does not add names to local map


ljharb (changes requested):
Please provide a regression test for this?

senputra:
Could you elaborate on what you meant by that? Did you mean that you want to see the result of all test cases? I surfed online and all mentioned that I have to test the changes against all test cases

Disclaimer: I am new to open source contributions.

ljharb:
@senputra sure! What i mean is, what is this fixing? Without a new test - one that would be failing without your fix - there's no way to verify that the fix is working, and the day after merging this, there's nothing to stop anyone from removing the truthiness check, since no tests would be failing.

Bug fixes in any software project, open source or otherwise, typically include regression tests like this to make sure the fix is verified and the problem can't occur in the future.

Issue #1665, opened 2020-02-20
[import/order] Force sort order of certain packages within "external" group


salimhamed:
We're using the [`import/order`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md) rule to enforce a convention in module imports. Here's how the `import/order` rule is currently configured in our project.

```javascript
// in .eslintrc.js

module.exports = {
  rules: {
      'import/order': ['error', {
          groups: ['builtin', 'external', 'internal'],
          'newlines-between': 'always',
          alphabetize: {
              order: 'asc',
              caseInsensitive: true,
          },
      }],
}
```

This will result in the following import conventions:

```javascript
// in a module with a React component 

// "external" modules are imported first
import { Button } from 'antd'
import PropTypes from 'prop-types'
import React from 'react'

// "internal" modules are imported second
import { BrandedIcon1 } from 'components/BrandedIcon1'
import { BrandedIcon2 } from 'components/BrandedIcon2'
```

All is well and good, except that our team's convention is to **import `React` first when a module has React components**. Therefore, the example above would become:

```javascript
import React from 'react' // force React to be imported first in the "external" group
import { Button } from 'antd'
import PropTypes from 'prop-types'

import { BrandedIcon1 } from 'components/BrandedIcon1'
import { BrandedIcon2 } from 'components/BrandedIcon2'
```

Is it possible to achieve this desired sort order using the `import/order` rule? I was a bit confused by the documentation of the `pathGroups` and `pathGroupsExcludedImportTypes` and I wasn't sure if these options could be used to achieve this result. Thanks in advance for the help and keep up the great work on this plugin!

kyranjamie:
At @blockstack we'd love this feature too.

ljharb:
You should be able to add `pathGroups`, marking React as going "before" the external group?

salimhamed:
@ljharb Thanks for the response. As far as I can tell, the `pathGroups` option doesn't work for `builtin` or `external` modules, as mentioned in the [documentation](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#pathgroups-array-of-objects).

![image](https://user-images.githubusercontent.com/1870760/75303218-f3ac1800-57f4-11ea-8905-4e3e5b1dbbcf.png)

The `react` module is `external`, so the following configuration options don't have any affect on the sort order.

```javascript
module.exports = {
  rules: {
      'import/order': ['error', {
[...]

PR #2156, opened 2021-07-18
[Docs] `order`:  improve the documentation for the `pathGroupsExcludedImportTypes` option

liby:
Motivation: #1665

> I think there's an opportunity to improve the documentation for the pathGroupsExcludedImportTypes option. The documentation made me think the allowed values for pathGroupsExcludedImportTypes were only groups (e.g., builtin, external, etc.) and not patterns (e.g., react, react-router-dom, etc).

I think this paragraph makes a lot of sense, and for a while, I couldn't figure out how to configure this option.

### DIFF
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -148,6 +148,27 @@ Example:
   }]
 }
 ```
+
+You can also use `patterns`(e.g., `react`, `react-router-dom`, etc).
+
+Example:
+```json
+{
+  "import/order": [
+    "error",
+    {
+      "pathGroups": [
+        {
+          "pattern": "react",
+          "group": "builtin",
+          "position": "before"
+        }
+      ],
+      "pathGroupsExcludedImportTypes": ["react"]
+    }
+  ]
+}
+```
 The default value is `["builtin", "external"]`.
 
 ### `newlines-between: [ignore|always|always-and-inside-groups|never]`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Changed
 - [Docs] `extensions`: removed incorrect cases ([#2138], thanks [@wenfangdu])
 - [Tests] `order`: add tests for `pathGroupsExcludedImportTypes: ['type']` ([#2158], thanks [@atav32])
+- [Docs] `order`:  improve the documentation for the `pathGroupsExcludedImportTypes` option ([#2156], thanks [@liby])
 
 ## [2.23.4] - 2021-05-29
 
@@ -812,6 +813,7 @@ for info on changes for earlier releases.
 
 [#2160]: https://github.com/benmosher/eslint-plugin-import/pull/2160
 [#2158]: https://github.com/benmosher/eslint-plugin-import/pull/2158
+[#2156]: https://github.com/benmosher/eslint-plugin-import/pull/2156
 [#2138]: https://github.com/benmosher/eslint-plugin-import/pull/2138
 [#2121]: https://github.com/benmosher/eslint-plugin-import/pull/2121
 [#2099]: https://github.com/benmosher/eslint-plugin-import/pull/2099
@@ -1347,6 +1349,7 @@ for info on changes for earlier releases.
 [@lencioni]: https://github.com/lencioni
 [@leonardodino]: https://github.com/leonardodino
 [@Librazy]: https://github.com/Librazy
+[@liby]: https://github.com/liby
 [@lilling]: https://github.com/lilling
 [@ljharb]: https://github.com/ljharb
 [@ljqx]: https://github.com/ljqx


liby:
@ljharb

Do you have any other questions? If you do not consider a merger, please let me know. Thank you.

docs/rules/order.md
```diff
@@ -132,6 +132,7 @@ Properties of the objects
 
 This defines import types that are not handled by configured pathGroups.
 This is mostly needed when you want to handle path groups that look like external imports.
+This option allows more than only `groups`(e.g., `builtin`, `external`, etc.), you can also use `patterns`(e.g., `react`, `react-router-dom`, etc).
```
ljharb, line 135: Maybe instead of the sentence, it'd be clearer to add a second example using `patterns`?

liby: It’s been added, please let me know if there is anything necessary to be revised.

liby:
@ljharb PTAL.

ljharb (approved):
Thanks!

yola-0316:
image

Hi, I see the impType is resolved as a fixed string in predefined set, it can't be a pattern.

ljharb:
@yola-0316 yes, do you have a question?

Akiq2016:
this PR add below demo
```json
{
  "import/order": [
    "error",
    {
      "pathGroups": [
        {
          "pattern": "react",
          "group": "builtin",
          "position": "before"
        }
      ],
      "pathGroupsExcludedImportTypes": ["react"]
    }
  ]
}
```

`"pathGroupsExcludedImportTypes": ["react"]` might be confused. As source code doesn't handle patterns like 'react', it just treats it as 'unknown' type.
Though this demo works, it's because pathGroupsExcludedImportTypes doesn't include external(default value) anymore, not because we set a 'react'

ljharb:
@Akiq2016 i'm not sure I understand, but I'd love to review a PR that either fixes the docs, or adds a failing test case :-)

Akiq2016:
#1665 I read this issue and knowing why we had this PR, but I think they use `pathGroupsExcludedImportTypes` in the wrong way. If we need to let pathGroup works, sometimes we have to add `pathGroupsExcludedImportTypes` as `[]`.

I think the origin docs describe clearly enough

> ### pathGroupsExcludedImportTypes: [array]:
> This defines import types that are not handled by configured pathGroups. This is mostly needed when you want to handle path groups that look like external imports.

As pathGroupsExcludedImportTypes was set to ['builtin', 'external'] as default. While we have some customize pattern in `pathGroups` which looks like an `external` pattern, we should ensure pathGroupsExcludedImportTypes without an 'external' value. In this way, the pathGroups will works.

This is also why #1665 's config doesn't work and after they add a `pattern pathGroupsExcludedImportTypes` works. But the value they set to pathGroupsExcludedImportTypes was treated as `unknown`

yola-0316:
Maybe we could improve the original description:
>### pathGroupsExcludedImportTypes: [array]:
>This defines import types that are not handled by configured pathGroups. This is mostly needed when you want to handle path groups that look like external imports.

In many times, the pathGroup you configured will work as your expected. but if the pathGroups pattern is same as `builtin` or `external`. it will ignored by default. we should override `pathGroupExcludedImportTypes" by set it to `[]` ensure it not include the `external`. the `["react"]` is a lucky hit, not the correct way.

Incorrect:
```js
{
  "import/order": [
    "error",
    {
      "pathGroups": [
        {
          "pattern": "react",
          "group": "builtin",
          "position": "before"
        }
      ],
      "pathGroupsExcludedImportTypes": ["react"]
    }
  ]
}
```

Correct:
```js
{
  "import/order": [
    "error",
    {
      "pathGroups": [
        {
          "pattern": "react",
          "group": "builtin",
          "position": "before"
        }
      ],
      "pathGroupsExcludedImportTypes": [] // not include external, because the react is belong to external form
    }
  ]
}
```

liby:
@Akiq2016
> #1665 I read this issue and knowing why we had this PR, but I think they use pathGroupsExcludedImportTypes in the wrong way. If we need to let pathGroup works, sometimes we have to add pathGroupsExcludedImportTypes as [].

I think you’re right. We were using `pathGroupsExcludedImportTypes` incorrectly.

> I think the origin docs describe clearly enough

I don’t think so. As can be seen from #1665, there are still a lot of people who are confused about it (including me).

> conclusion: this PR saying 'You can also use patterns' is wrong, the source code doesn't support this feature. and pathGroupsExcludedImportTypes: [react] doesn't make sense

I think we should update the document so that no one else uses it incorrectly.

@yola-0316
I very much agree with your proposal for improvement. The improved description will go a long way to avoid confusion or misuse by others.

Mairu:
Just to five my 2 cents. I introduced this option only because of backward compatibility reasons. I didn't want to change the default value of this exclusion list in a minor version.

And for the explanation, in my original documentation I added an example to make the usage clear. Because I don't know how to describe it clearly without adding too much sentences.

I can understand that this description could lead to confusions for people that are not into the internals.

I think something like

> This defines import types that are not handled by configured pathGroups.
> If you have added path groups with patterns that look like builtin or external imports, you have to remove this group ( builtin
> and/or external ) from the default exclusion list (`["builtin", "external"]`) to sort these path groups correctly.

would perhaps be better?

PR #2160, opened 2021-07-21
Use getPhysicalFilename() instead of getFilename() if available (ESLint 7.28+)

pmcelhaney:
Starting with 7.28.0, ESLint has a new feature [distinguishing the physical filename from the virtual filename](https://eslint.org/blog/2021/06/eslint-v7.28.0-released#contextgetphysicalfilename).

> ## context.getPhysicalFilename()
> Rules can now use the new method getPhysicalFilename() on the context object to get the full path of the file on disk without any code block information.
>
> The difference between getPhysicalFilename and getFilename is observable when ESLint is used with processors:
>
> ```js
> context.getPhysicalFilename() // "/project/example.md"        - original file
> context.getFilename()         // "/project/example.md/0_0.js" - virtual filename assigned to a code block
> ```

When there's a difference, this plugin needs the real, physical filename. (See 
https://github.com/eslint/eslint/issues/11989) 

I ran into the problem when I noticed `no-unresolved` was yielding false positives inside of code blocks in Storybook / Markdown files and helped push the necessary change through in ESLint. 

To maintain backwards compatibility, this change uses `context.getPhysicalFilename()` if available and otherwise falls back to `getFilename()`.

### DIFF
--- a/src/core/packagePath.js
+++ b/src/core/packagePath.js
@@ -4,7 +4,7 @@ import readPkgUp from 'read-pkg-up';
 
 
 export function getContextPackagePath(context) {
-  return getFilePackagePath(context.getFilename());
+  return getFilePackagePath(context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename());
 }
 
 export function getFilePackagePath(filePath) {
--- a/src/rules/named.js
+++ b/src/rules/named.js
@@ -43,7 +43,7 @@ module.exports = {
         if (!deepLookup.found) {
           if (deepLookup.path.length > 1) {
             const deepPath = deepLookup.path
-              .map(i => path.relative(path.dirname(context.getFilename()), i.path))
+              .map(i => path.relative(path.dirname(context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename()), i.path))
               .join(' -> ');
 
             context.report(im[key],
--- a/tests/src/core/resolve.js
+++ b/tests/src/core/resolve.js
@@ -1,4 +1,6 @@
 import { expect } from 'chai';
+import eslintPkg from 'eslint/package.json';
+import semver from 'semver';
 
 import resolve, { CASE_SENSITIVE_FS, fileExistsWithCaseSync } from 'eslint-module-utils/resolve';
 import ModuleCache from 'eslint-module-utils/ModuleCache';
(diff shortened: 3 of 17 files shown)


tests/src/core/resolve.js
```diff
@@ -21,34 +21,48 @@ describe('resolve', function () {
...
+    expect(resolve( '../files/foo'
+      , Object.assign({}, testContext, { getFilename: function () { return utils.getFilename('foo.js'); } }),
+    )).to.equal(utils.testFilePath('./bar.jsx'));
+
+    expect(resolve( '../files/foo'
+      , Object.assign({}, testContext, { getFilename: function () { throw new Error('Should call getPhysicalFilename() instead of getFilename()'); }, getPhysicalFilename: function () { return utils.getFilename('foo.js'); } }),
+    )).to.equal(utils.testFilePath('./bar.jsx'));
+
+
+  });
```
pmcelhaney, line 64: Added this test to make sure the fallback behavior is working. Otherwise the tests have been updated to use `getPhysicalFilename()` instead of `getFilename()`.

src/core/packagePath.js
```diff
@@ -4,7 +4,7 @@ import readPkgUp from 'read-pkg-up';
 
 
 export function getContextPackagePath(context) {
-  return getFilePackagePath(context.getFilename());
+  return getFilePackagePath((context.getPhysicalFilename && context.getPhysicalFilename()) || context.getFilename());
```
ljharb, line 7:
```suggestion
  return getFilePackagePath(context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename());
```

src/rules/named.js
```diff
@@ -43,7 +43,7 @@ module.exports = {
         if (!deepLookup.found) {
           if (deepLookup.path.length > 1) {
             const deepPath = deepLookup.path
-              .map(i => path.relative(path.dirname(context.getFilename()), i.path))
+              .map(i => path.relative(path.dirname((context.getPhysicalFilename && context.getPhysicalFilename()) || context.getFilename()), i.path))
```
ljharb, line 46:
```suggestion
              .map(i => path.relative(path.dirname(context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename()), i.path))
```

tests/src/core/resolve.js
```diff
@@ -21,34 +21,48 @@ describe('resolve', function () {
     const testContext = utils.testContext({ 'import/resolver': './foo-bar-resolver-v1' });
 
     expect(resolve( '../files/foo'
-      , Object.assign({}, testContext, { getFilename: function () { return utils.getFilename('foo.js'); } }),
+      , Object.assign({}, testContext, { getPhysicalFilename: function () { return utils.getFilename('foo.js'); } }),
```
ljharb, line 24: rather than modifying the existing tests, please leave these alone, so that the new `getPhysicalFilename` duplicated tests are clearly additions

ljharb (changes requested):
Thanks!

please swap all the `a && a() || b()` for `a ? a() : b()`; plus the tests comment

tests/src/core/resolve.js
pmcelhaney: Okay, which do you prefer?

1. Duplicate all of the existing tests so that the only difference is `getFilename()` vs. `getPhysicalFilename()`
2. Revert the changes so that we're testing all of the use cases for `getFilename()` and not exercising `getPhysicalFilename()` except for the couple of tests demonstrating that `getPhysicalFilename()` is the preferred API and `getFIlename()` is a fallback. 
3. Same as 1, but add `getFilename() { throw new Error(...) }` to the mock context on all the new tests rather than have a separate test for the fallback.

src/rules/named.js
pmcelhaney: Yes, that's better. Let me know how you want to proceed on the tests and I'll make both changes at the same time. Thanks for the feedback!

tests/src/core/resolve.js
ljharb: I was originally thinking option 1, but i really do like option 3.

pmcelhaney: @ljharb These changes are done.

ljharb (approved):

ljharb:
(Had to update the tests so the getPhysicalFilename stuff only applies in eslint v7.28+)

PR #2166, opened 2021-07-26
[New] support "module" field in dependencies

### LINKED ISSUE #2163: docs example for `import/named` errors in imports from third-party libraries doesn't work
`import/named` describes a few examples with 3rd party libraries, specifically with React and Redux. 

However, I attempted to reproduce the example in a [fresh create-react-app project](https://github.com/MustafaHaddara/eslint-plugin-import-repro/commit/6c371fa87cf04704c5c8ec462669597e5a360dd2), and it didn't flag anything wrong with my imports from redux. 

Redux doesn't have a `jsnext:main` key in its `package.json`, and the [link to the `rollup` wiki in the further reading section of the docs](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/named.md#further-reading) is dead. `jsnext:main` is [deprecated](https://github.com/jsforum/jsforum/issues/5) in favour of [`pkg.module`](https://github.com/rollup/rollup/wiki/pkg.module), but Redux ships with a `pkg.module` definition and yet `import/named` doesn't catch bad imports from redux.

What's the correct way to configure this plugin to report non-existent named imports from third party libraries?

MustafaHaddara:
Fixes #2163 

I still need to add unit tests (I plan on copying the existing `jsnext:main` tests in the webpack resolver)

I also wonder what's the best way for me to test this...it looks like it's multiple packages all in one repo, and I'm making changes to the webpack resolver. I can build it locally but then would I need to wire up my config for this plugin to use the local version of the resolver? Is that correct?

### DIFF
--- a/resolvers/node/index.js
+++ b/resolvers/node/index.js
@@ -42,8 +42,10 @@ function opts(file, config) {
 }
 
 function packageFilter(pkg) {
-  if (pkg['jsnext:main']) {
-    pkg['main'] = pkg['jsnext:main'];
+  if (pkg.module) {
+    pkg.main = pkg.module;
+  } else if (pkg['jsnext:main']) {
+    pkg.main = pkg['jsnext:main'];
   }
   return pkg;
 }
--- a/resolvers/node/test/packageMains.js
+++ b/resolvers/node/test/packageMains.js
@@ -1,0 +1,27 @@
+'use strict';
+
+const chai = require('chai');
+const expect = chai.expect;
+const path = require('path');
+
+const webpack = require('../');
+
+const file = path.join(__dirname, 'package-mains', 'dummy.js');
+
+
+describe('packageMains', function () {
+  it('captures module', function () {
+    expect(webpack.resolve('./module', file)).property('path')
+      .to.equal(path.join(__dirname, 'package-mains', 'module', 'src', 'index.js'));
+  });
+
+  it('captures jsnext', function () {
+    expect(webpack.resolve('./jsnext', file)).property('path')
+      .to.equal(path.join(__dirname, 'package-mains', 'jsnext', 'src', 'index.js'));
+  });
+
+  it('captures module instead of jsnext', function () {
+    expect(webpack.resolve('./module-and-jsnext', file)).property('path')
+      .to.equal(path.join(__dirname, 'package-mains', 'module-and-jsnext', 'src', 'index.js'));
+  });
+});
--- a/docs/rules/named.md
+++ b/docs/rules/named.md
@@ -5,7 +5,7 @@ Verifies that all named imports are part of the set of named exports in the refe
 For `export`, verifies that all named exports exist in the referenced module.
 
 Note: for packages, the plugin will find exported names
-from [`jsnext:main`], if present in `package.json`.
+from [`jsnext:main`] (deprecated) or `module`, if present in `package.json`.
 Redux's npm module includes this key, and thereby is lintable, for example.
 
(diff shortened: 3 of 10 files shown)


ljharb:
We have some test fixtures (fake projects in the repo) - I think the best approach is probably to create your own - one that has just module, one that has just jsnext:main (this probably already exists), and one that has both?

MustafaHaddara:
> one that has just jsnext:main (this probably already exists)

This does indeed exist, that was my plan. 

I was asking more about manual testing with an actual package.

ljharb:
I'd expect you'd be able to `npm link` this plugin into a project that's running into the issue (like in the linked issue) and then run the linter?

MustafaHaddara:
Yeah but my change is in the `node` resolver, which looks like it's treated as a separate package?

ljharb:
Yes, that's true. you'd need to `npm link` that package as well.

MustafaHaddara:
@ljharb Thanks for your help, I think this PR is ready to go now.

resolvers/node/index.js
```diff
@@ -42,7 +42,9 @@ function opts(file, config) {
 }
 
 function packageFilter(pkg) {
-  if (pkg['jsnext:main']) {
+  if (pkg['module']) {
+    pkg['main'] = pkg['module'];
+  } else if (pkg['jsnext:main']) {
     pkg['main'] = pkg['jsnext:main'];
```
ljharb, line 48:
```suggestion
  if (pkg.module) {
    pkg.main = pkg.module;
  } else if (pkg['jsnext:main']) {
    pkg.main = pkg['jsnext:main'];
```

resolvers/node/test/packageMains.js
```diff
@@ -0,0 +1,29 @@
...
+const expect = chai.expect;
+const path = require('path');
+
+const webpack = require('../');
+
+const file = path.join(__dirname, 'package-mains', 'dummy.js');
+
+
+describe('packageMains', function () {
+
```
ljharb, line 13:
```suggestion
```

resolvers/node/test/packageMains.js
```diff
@@ -0,0 +1,29 @@
...
+  it('captures jsnext', function () {
+    expect(webpack.resolve('./jsnext', file)).property('path')
+      .to.equal(path.join(__dirname, 'package-mains', 'jsnext', 'src', 'index.js'));
+  });
+
+  it('captures module instead of jsnext', function () {
+    expect(webpack.resolve('./module-and-jsnext', file)).property('path')
+      .to.equal(path.join(__dirname, 'package-mains', 'module-and-jsnext', 'src', 'index.js'));
+  });
+
```
ljharb, line 28:
```suggestion
```

ljharb (commented):
Looks good overall. I assume the webpack resolver calls into the node resolver, which is why you don't need any fix inside the webpack resolver?

MustafaHaddara:
@ljharb just wanted to follow up on this, is there anything else you need from me to get this merged? 

Thanks!

ljharb (approved):

MustafaHaddara:
@ljharb thanks for merging this! 

What's your normal release process? ie. when should I expect this to be in a new version on NPM?

No pressure, I just want to know what to expect.

Thanks again!

ljharb:
There’s no typical timeframe, but I’m trying to land some open PRs and then release it in the next week or two.

MustafaHaddara:
Great, thank you!

Issue #2169, opened 2021-07-29
Legacy dependencies causing warnings for vulnerability.


sandeshvijayraj:
There seem to be some legacy libraries getting used in the dependency definition. One such example is Debug <4.0 all debugger versions less than 4 are detected as vulnerable but are still being used in this project.

ljharb:
Duplicate of #1892.

The reason is because:
 a) debug 4 drops support for node versions we must support
 b) the vulnerabilities, like most in the JS ecosystem, are a false positive, and are non-issues for the manner in which this package uses `debug`.

Issue #1892, opened 2020-08-26
[deps][question] The reason of outdated packages


pahan35:
Hi there!

Is there any reason to keep some of your deps outdated?

For example, you have 
```
{
  "debug": "^2.6.9",
  "doctrine": "1.5.0",
  "read-pkg-up": "^2.0.0"
}
```

However, in my project, those packages are higher for most of the packages
```
{
  "debug": "^4.1.1",
  "doctrine": "^3.0.0",
  "read-pkg-up": "^7.0.1"
}
```

Because of it, such outdated deps are loaded inside of the package itself into `<package>/node_modules/`
![image](https://user-images.githubusercontent.com/13823215/91239367-05d92580-e748-11ea-996c-efc4edcc3242.png)

I prefer to avoid it, but it's possible only if you update your deps.

Are those deps kept outdated because of some reason or just because of a lack of time to update it?

ljharb:
A few thoughts:
1. there is never anything inherently wrong with having a dep that's "not the latest major", code that works will likely work forever.
1. `read-pkg-up` aggressively drops support for older node versions, so we *can't* use a newer major; I'd prefer to see that dep removed before it's ever updated.
1. `doctrine` is unmaintained, so unless there's a concrete benefit of v2 or v3 over v1, there's no reason to update it
1. `debug` has the same problem as `read-pkg-up`: v4 doesn't support node < 6, and iirc v3 doesn't support node < 4, so we *can't* upgrade to it.

pahan35:
@ljharb, thanks for the explanation!

As I see here, https://endoflife.date/nodejs Node v12 is the oldest LTS with active support.

Is there any reason to support the older node versions?

As I see, eslint itself actively drops support for old versions https://github.com/eslint/eslint/releases/tag/v7.0.0

Here https://github.com/TrySound/postcss-inline-svg/pull/68 is also a PR for dropping not supported node version, but it looks like that repo isn't actively maintained.

Maybe it's better to sync up the latest node support version with eslint itself, releasing drops for support with major releases?

I'm just curious about the reasons.

ljharb:
Yes - for one, platform support is irrelevant. People use older nodes, and thus they should be supported. Additionally, we’d have to have a semver-major to drop them, and doing that *just* to drop support for something is highly user-hostile.

[...]

Issue #868, opened 2017-06-09
Regular Expression Denial of Service (ReDoS)


schoenwaldnils:
I updated to v2.3.0 and low snyk.io returns what the issue title says.
Also this:
```
Vulnerable module: ms
Introduced through: eslint-plugin-import@2.3.0 › eslint-module-utils@2.0.0 › debug@2.2.0 › ms@0.7.1
```
More info to this vulnerability: https://snyk.io/vuln/npm:ms:20170412

Is there a quick way to fix this?

ljharb:
Just because there's a vulnerability, does not mean it necessarily affect you. In this case only your own eslint configuration can provide a regular expression, so the risk does not exist.

That said, it'd be ideal to update the dep chain.

benmosher:
fixed in v2.5.0 or so

PR #2177, opened 2021-08-06
Updated debug library version to avoid vulnerabilities scan error

sandeshvijayraj:
Debug Library < 4.1.1 in package.json was being detected as Vulnerable 

Updated Library to >4.1.3 that is considered to be non vulnerable

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
   "dependencies": {
     "array-includes": "^3.1.3",
     "array.prototype.flat": "^1.2.4",
-    "debug": "^2.6.9",
+    "debug": "^4.3.2",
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
No. This is a breaking change.

ljharb:
See #2169, #1892, #868.

Most vulnerabilities are false positives and this one is no exception. Just ignore the warning.

sandeshvijayraj:
Oh thanks for the info @ljharb I missed that comment earlier

ljharb:
@martinez-hugo see above why we can't.

Issue #2176, opened 2021-08-05
Mark package as internal not working


KostyaZgara:
I'm experiencing the issue when can't use the setting `import/internal-regex` to mark my packages as internal. After some manipulation and researching seems the reason is PnP mode which I'm using on my project.

My eslint config
```yaml
parser: '@typescript-eslint/parser'
parserOptions:
  sourceType: 'module'
  project: 'tsconfig.json'
plugins:
  - '@typescript-eslint/eslint-plugin'
extends:
  - 'plugin:@typescript-eslint/eslint-recommended'
  - 'plugin:@typescript-eslint/recommended'
  - 'prettier'
  - 'plugin:import/recommended'
  - 'plugin:import/typescript'
env:
  node: true
root: true
rules:
  import/order:
    - error
    - groups:
        - builtin
        - external
        - internal
        - parent
        - sibling
        - type
      newlines-between: never
settings:
  import/resolver: typescript
  import/internal-regex: ^@(lib|app)/
```

This configuration linting my code but doesn't recognize marked internal packages. When I remove `import/resolver: typescript` from `settings` it groups my imports in the right way, but throws an error `Resolve error: unable to load resolver "node"  import/order`. The same situation with the default `node` resolver. Seems like when resolver is working it ignores any settings.

My tools:
```
yarn -v
3.0.0

node -v
v14.17.0
```

```json
"eslint": "7.32.0",
"eslint-plugin-import": "2.23.4",
"@typescript-eslint/eslint-plugin": "4.28.5",
"@typescript-eslint/parser": "4.28.5",
"eslint-import-resolver-node": "0.3.4",
"eslint-import-resolver-typescript": "2.4.0",
```

Thanks in advance!

P.S. If you need to provide additional information, please inform me about it

ljharb:
If you’re using a nonstandard resolution (like pnp), then you need a custom resolver for it for this entire plugin to work properly. I’m not aware of any pnp resolver yet.

KostyaZgara:
Thanks for your response. Do you have some recommendations for customer resolvers that properly handle PnP mode? I checked this list https://github.com/import-js/eslint-plugin-import/wiki/Resolvers and didn't find any appropriate solution

ljharb:
None exist as far as I know.

KostyaZgara:
Okay, thanks! I close this issue then

KostyaZgara:
[...]

PR #2178, opened 2021-08-07
[Fix] correct recognize external modules in pnp mode

KostyaZgara:
PR should fix this issue #2176 

I created a repository https://github.com/KostyaZgara/eslint-plugin-import-berry-issue-example to reproduce an error.

The problem was `external` module was recognized only when `path` is not resolved (undefined) and the `isExternalPath` function tries to find a module in `node_modules` or provided external module folders, but not found cuz with PnP mode `node_modules` doesn't exist.

So, the issue was resolved by adding new setting `import/pnp` (check README to see usage) and searching cache entry in resolved path

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,6 +433,24 @@ settings:
   import/internal-regex: ^@scope/
 ```
 
+#### `import/pnp`
+
+Enable Plug'n'Play resolution. Set `true` if you are using `yarn berry` or `pnpm`.
+
+```yaml
+# .eslintrc.yml
+settings:
+  import/pnp: true
+```
+
+Also, you can provide custom cache folder if you are using non-standard path. Default value is `.yarn/cache`
+
+```yaml
+# .eslintrc.yml
+settings:
+  import/pnp:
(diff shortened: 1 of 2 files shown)


KostyaZgara:
It's a draft PR just to discuss the changes that fixing a bug. Unfortunately, I can't reproduce the error in the test now, cuz it requires creating `pnp` resolver to emulate PnP behaviour, but in production default `node` resolver works good (thanks to `resolve` package` and it resolves the right path.
Please, if my changes is on the right way, let me know it and I'll finalize the PR with adding tests and additional settings.

README.md
```diff
@@ -433,6 +433,24 @@ settings:
   import/internal-regex: ^@scope/
 ```
 
+#### `import/pnp`
```
ljharb, line 436: I don't think it's appropriate to add an option to the core rule for everyone, just for one package manager's nonstandard resolution.

This should be done via a separate package, like https://www.npmjs.com/package/eslint-import-resolver-node, that's specific to yarn pnp.

PR #2179, opened 2021-08-07
[Docs] add description how to use plugin with pnp/yarn berry

KostyaZgara:
Sorry in advance if you find mistakes. Feel free to point me where need to fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,6 +348,8 @@ An array of folders. Resolved modules only from those folders will be considered
 
 This option is also useful in a monorepo setup: list here all directories that contain monorepo's packages and they will be treated as external ones no matter which resolver is used.
 
+If you are using `yarn` PnP as your package manager, add the `.yarn` folder and all your installed dependencies will be considered as `external`, instead of `internal`.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -348,6 +348,8 @@ An array of folders. Resolved modules only from those folders will be considered
 
 This option is also useful in a monorepo setup: list here all directories that contain monorepo's packages and they will be treated as external ones no matter which resolver is used.
 
+Also, this option can be used with plug'n'play mode. For example, if you are using `yarn berry` you can set `.yarn` folder and all your installed dependencies will be considered as `external`, otherwise it will be `internal`.
```
ljharb, line 351: most users will have no idea what "plun'n'play mode" is, let alone yarn. What about:
```suggestion
If you are using `yarn` PnP as your package manager, add the `.yarn` folder and all your installed dependencies will be considered as `external`, instead of `internal`.
```

ljharb (changes requested):

README.md
KostyaZgara: Sounds good, fixed

ljharb (approved):

PR #2184, opened 2021-08-09
[Fix] ExportMap: Add default export when esModuleInterop is true and anything is exported

### LINKED ISSUE #2183: [import/default] - false negative for react-test-renderer
I'm doing this:
```
import reactTestRenderer from 'react-test-renderer';
```

Getting this:
```
No default export found in imported module "react-test-renderer". eslint (import/default)
```

Using these types: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-test-renderer/index.d.ts

And I have `esModuleInterop` in my tsconfig, so this should not throw error (thanks to https://github.com/import-js/eslint-plugin-import/pull/1689)

Maxim-Mazurok:
This PR aims to fix #2183 by adding default export if anything is exported and `esModuleInterop` is set to true.

This probably should've been done in #1689 that I originally opened to add `esModuleInterop` support.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -351,6 +351,7 @@ ExportMap.for = function (context) {
 
 ExportMap.parse = function (path, content, context) {
   const m = new ExportMap(path);
+  const isEsModuleInteropTrue = isEsModuleInterop();
 
   let ast;
   try {
@@ -576,8 +577,6 @@ ExportMap.parse = function (path, content, context) {
       });
     }
 
-    const isEsModuleInteropTrue = isEsModuleInterop();
-
     const exports = ['TSExportAssignment'];
     if (isEsModuleInteropTrue) {
       exports.push('TSNamespaceExportDeclaration');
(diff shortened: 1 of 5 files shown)


src/ExportMap.js
```diff
@@ -606,8 +605,11 @@ ExportMap.parse = function (path, content, context) {
         m.namespace.set('default', captureDoc(source, docStyleParsers, n));
         return;
       }
-      if (isEsModuleInteropTrue) {
-        m.namespace.set('default', {});
+      if (
+        isEsModuleInteropTrue && // esModuleInterop is on in tsconfig
+        !m.namespace.has('default') // and default isn't added already
```
ljharb, line 610:
```suggestion
        isEsModuleInteropTrue // esModuleInterop is on in tsconfig
        && !m.namespace.has('default') // and default isn't added already
```

src/ExportMap.js
```diff
@@ -645,6 +647,14 @@ ExportMap.parse = function (path, content, context) {
     }
   });
 
+  if (
+    isEsModuleInteropTrue // esModuleInterop is on in tsconfig
+     && m.namespace.size > 0 // anything is exported
+      && !m.namespace.has('default') // and default isn't added already
```
ljharb, line 653:
```suggestion
    && m.namespace.size > 0 // anything is exported
    && !m.namespace.has('default') // and default isn't added already
```

ljharb (approved):
Thanks, seems good.

PR #2191, opened 2021-08-16
Support for ESLint v8

ota-meshi:
This PR updates this plugin to support ESLint v8.


tests/dep-time-travel.sh
```diff
@@ -8,19 +8,23 @@ export NPM_CONFIG_LEGACY_PEER_DEPS=true
...
-# use these alternate TypeScript dependencies for ESLint < v4
-if [[ "$ESLINT_VERSION" -lt "4" ]]; then
-  echo "Downgrading babel-eslint..."
-  npm i --no-save babel-eslint@8.0.3
-
-  echo "Downgrading TypeScript dependencies..."
-  npm i --no-save typescript-eslint-parser@15 typescript@2.8.1
+re="^[0-9]+$"
+
+if [[ "$ESLINT_VERSION" =~ $re ]] ; then
```
ljharb, line 13: i'm not a fan of this approach; i'd rather hardcode the temporary 8 prerelease semver string, which we definitely aren't going to publish a release supporting.

tests/src/cli.js
```diff
@@ -3,82 +3,107 @@
...
-import { expect } from 'chai';
-import { CLIEngine } from 'eslint';
+import { expect, use } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import { CLIEngine, ESLint as ESLintClass } from 'eslint';
 import eslintPkg from 'eslint/package.json';
 import semver from 'semver';
+import * as importPlugin from '../../src/index';
+
+use(chaiAsPromised);
```
ljharb, line 13: why is this something we want?

tests/src/cli.js
```diff
@@ -3,82 +3,107 @@
...
+
+    for (const [name, plugin] of Object.entries(pluginsMap || {})) {
+      this.engine.addPlugin(name, plugin);
+    }
+  }
+  lintFiles(params) {
+    const result = this.engine.executeOnFiles(params);
+    return Promise.resolve(result.results);
+  }
+};
```
ljharb, line 34: why do we need a hardcoded class here? is this no longer exposed, even under legacy specifiers?

tests/src/rules/no-amd.js
```diff
@@ -2,7 +2,7 @@ import { RuleTester } from 'eslint';
 import eslintPkg from 'eslint/package.json';
 import semver from 'semver';
 
-const ruleTester = new RuleTester();
+const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2015, sourceType: 'module' } });
```
ljharb, line 5: why is this needed? did the default of RuleTester change?

ljharb (changes requested):
I'd prefer to find ways so the existing tests do not change at all, to increase confidence that we're not breaking eslint 2-7 support.

tests/dep-time-travel.sh
ota-meshi: I think you're right. But I wanted to check if CI works at this point.

tests/src/cli.js
ota-meshi: This plugin also supports ESLint <= v6, so I created a compatible class to see if it works for testing ESLint <= v6.

tests/src/rules/no-amd.js
ota-meshi: I've confirmed that previous tests seemed to rely on the scopeManager modified by the babel-eslint patch.

https://github.com/babel/babel-eslint/blob/v8.2.6/lib/patch-eslint-scope.js#L363

If we do not specify `sourceType:'module'` in a normal scopeManager, this rule will not always return an error because `scope.type ='module'` will not be created.

This change was necessary as the patch would fail when using eslint v8.

tests/src/cli.js
ota-meshi: I added it to use the `rejected` assertion.

https://github.com/import-js/eslint-plugin-import/pull/2191/files#diff-be7287d185556ab4df4ae56bd7f5d71cc67d1dec06fb5a9f7754470ca90b0277R53
https://www.chaijs.com/plugins/chai-as-promised/

ota-meshi:
Hmm. I'm not sure what caused this error.

> Error while loading rule 'import/no-unused-modules': Failed to load plugin 'import' declared in '.eslintrc': Package subpath './lib/util/glob-util' is not defined by "exports" in /home/runner/work/eslint-plugin-import/eslint-plugin-import/node_modules/eslint/package.json

I think I've fixed it so that this issue doesn't happen.

https://github.com/import-js/eslint-plugin-import/pull/2191/files#diff-7d2908b010ec724c0f3a0d7615cb91f4673d765d52c6ba6607501e23470f2966R21

I have removed the `import` plugin configuration from `./.eslintrc` in my local environment and most tests work fine with `npm run tests-only`.
I thought that in CI it works fine with `linklocal` without changing `./.eslintrc`, is that wrong?

tests/src/cli.js
ljharb: I'm strongly opposed to any noop assertions - ie, anything that's not a function call - because it's super easy to accidentally make a typo and get a silent failure.

If it's a promise and we want to be sure it's rejected, we can `return promise.then(() => { expect(false).toEqual(true); }, (e) => expect(e).toSomething(); });` instead of adding an entirely new library.

tests/src/cli.js
ljharb: I'd prefer not to use such a wrapper if possible; that would mean we're not directly testing eslint < 7, and our wrapper could hide bugs.

tests/src/rules/no-amd.js
ljharb: So, to be clear, this change is both correct and unnecessary in master? perhaps i could peel it off as a separate commit and land it first.

ljharb:
That message suggests that we're pulling `lib/util/glob-util` somewhere, and that's no longer a permitted path in eslint.

ota-meshi:
`lib/util/glob-util` appears to be used in the final failover.

https://github.com/import-js/eslint-plugin-import/pull/2191/files#diff-7d2908b010ec724c0f3a0d7615cb91f4673d765d52c6ba6607501e23470f2966R51

But the line number in the error seems to point to the previous module (line number 43, new line number is 51).
Is `eslint-plugin-import`(2.x) in `node_modules` used for testing?

tests/src/cli.js
ota-meshi: Thank you for telling me another method! I would like to change to use that method.

ljharb:
I think the way things work is that the previous version ends up in node_modules, so it's possible the try/catch stuff would need to be landed and published first - but hopefully there's another alternative.

ota-meshi:
FileEnumerator#iterateFiles loads the node_modules plugin from eslintrc.
I think we can test it correctly by linking myself eslint-plugin-import to a dependency eslint-plugin-import, is there a way to do this in CI?

tests/src/rules/no-amd.js
ota-meshi: If you run the following command in the repository before making this PR change, the test will fail.

`npx cross-env BABEL_ENV=test nyc -s mocha tests/src/rules/no-amd.js `

I will create another PR to fix this.

tests/src/cli.js
ota-meshi: I made change to this PR.
https://github.com/import-js/eslint-plugin-import/pull/2191/commits/184118366b6022dda011fa8eb170b598a6738b50

tests/src/cli.js
ota-meshi: I made change to this PR.
https://github.com/import-js/eslint-plugin-import/pull/2191/commits/184118366b6022dda011fa8eb170b598a6738b50

ota-meshi: I opened PR #2193.

tests/src/rules/no-amd.js
ota-meshi: I opened PR #2193.

tests/dep-time-travel.sh
```diff
@@ -33,3 +37,10 @@ if [[ -n "$TS_PARSER" ]]; then
   echo "Downgrading @typescript-eslint/parser..."
   npm i --no-save @typescript-eslint/parser@2
 fi
+
+if [[ "$ESLINT_VERSION" = "^8.0.0-0" ]]; then
+  echo "Installing self"
+  npm i --no-save eslint-plugin-import@'.' -f
+  echo "Build self"
+  npm run build
+fi
```
ota-meshi, line 46: I made it link as a workaround to get the test to work which fails due to eslint-plugin-import in node_modules.

tests/src/cli.js
```diff
@@ -4,82 +4,145 @@
...
         this.skip();
       } else {
-        cli = new CLIEngine({
-          useEslintrc: false,
-          configFile: './tests/files/just-json-files/.eslintrc.json',
-          rulePaths: ['./src/rules'],
-          ignore: false,
-        });
+        if (ESLint) {
+          this.timeout(10000); // In the ESLint v8 test, it takes a lot of time to generate the ESLint class because we linked eslint-plugin-import.
```
ota-meshi, line 59: I'm using timeout as a workaround because linking eslint-plugin-import has caused it to take so much time.

ota-meshi:
At this point, I think we've been able to confirm that at least the no-unused-modules.js changes work.
Please let me know if you have any advice on how to change the test cases and CI.

package.json
```diff
@@ -71,7 +71,8 @@
     "chai": "^4.3.4",
     "coveralls": "^3.1.0",
     "cross-env": "^4.0.0",
-    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0",
+    "escope": "^3.6.0",
```
ota-meshi, line 74: This change is a workaround to prevent the babel-eslint patch from crashing.

https://github.com/babel/babel-eslint/blob/v8.2.6/lib/patch-eslint-scope.js#L31

tests/src/cli.js
ota-meshi: Hmm. It seems that 10 seconds may not be enough.

image

tests/src/cli.js
```diff
@@ -4,82 +4,145 @@
...
         this.skip();
       } else {
-        cli = new CLIEngine({
-          useEslintrc: false,
-          configFile: './tests/files/just-json-files/.eslintrc.json',
-          rulePaths: ['./src/rules'],
-          ignore: false,
-        });
+        if (ESLint) {
+          this.timeout(15000); // In the ESLint v8 test, it takes a lot of time to generate the ESLint class because we linked eslint-plugin-import.
```
ota-meshi, line 59: It seems that 10 seconds is not enough, so I increased it to 15 seconds.

https://github.com/import-js/eslint-plugin-import/pull/2191#discussion_r689981103

ota-meshi: I investigated why link takes time.
It took a long time to try to put the coverage mark on "lib" by "nyc" because link no longer is node_modules.
I think it can be fixed by adding "lib" to "exclude" in ".nycrc".

tests/src/cli.js
```diff
@@ -4,82 +4,144 @@
...
+      } else {
+        cli = new CLIEngine({
+          useEslintrc: false,
+          configFile: './tests/files/issue210.config.js',
+          rulePaths: ['./src/rules'],
+          rules: {
+            'named': 2,
+          },
+        });
+        cli.addPlugin('eslint-plugin-import', importPlugin);
```
ljharb, line 38: why is this needed now, but wasn't needed before?

tests/src/cli.js
```diff
@@ -4,82 +4,144 @@
...
+      }
     });
     it("doesn't throw an error on gratuitous, erroneous self-reference", function () {
-      expect(() => cli.executeOnFiles(['./tests/files/issue210.js'])).not.to.throw();
+      if (eslint) {
+        return eslint.lintFiles(['./tests/files/issue210.js'])
+          .then(() => { }, () => expect.fail() );
```
ljharb, line 44: any assertion for the happy path here? if not:
```suggestion
          .catch() => expect.fail());
```
or actually, you could just return the promise, and mocha will fail if it rejects.

tests/src/cli.js
```diff
@@ -4,82 +4,144 @@
...
+            plugins: { 'eslint-plugin-import': importPlugin },
+          });
+        } else {
+          cli = new CLIEngine({
+            useEslintrc: false,
+            configFile: './tests/files/just-json-files/.eslintrc.json',
+            rulePaths: ['./src/rules'],
+            ignore: false,
+          });
+          cli.addPlugin('eslint-plugin-import', importPlugin);
```
ljharb, line 73: same question here

.nycrc
```diff
@@ -10,6 +10,7 @@
 	"exclude": [
 		"coverage",
 		"test",
-		"tests"
+		"tests",
+		"lib"
```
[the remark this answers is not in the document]
ota-meshi: This was necessary because it takes a long time to load eslint-plugin-import.

https://github.com/import-js/eslint-plugin-import/pull/2191#discussion_r690004072

src/rules/no-unused-modules.js
```diff
@@ -228,7 +236,7 @@ const prepareImportsAndExports = (srcFiles, context) => {
         }
         const localImport = imports.get(key) || new Set();
         value.declarations.forEach(({ importedSpecifiers }) =>
-          importedSpecifiers.forEach(specifier => localImport.add(specifier))
+          importedSpecifiers.forEach(specifier => localImport.add(specifier)),
```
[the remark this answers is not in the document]
ota-meshi: I think it was added by the following configuration.
https://github.com/import-js/eslint-plugin-import/blob/master/src/.eslintrc.yml

tests/src/cli.js
ljharb: gotcha

tests/src/cli.js
ota-meshi: If we don't specify this, I think eslint-plugin-import in node_modules will be used for linting.
I think I need this change to check if my PR change does not affect this test.

ljharb: ok, so it seems like this is a change we could/should add separately, since it will fix the issue of tests running with the last published version, instead of the current code?

tests/src/cli.js
ota-meshi: I didn't know that. Thank you for letting me know!

src/rules/no-unused-modules.js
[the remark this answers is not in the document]
ljharb: yes, but not on function arguments, which is the case here.

package.json
```diff
@@ -71,7 +71,8 @@
...
+    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8.0.0-0",
```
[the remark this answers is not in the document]
ota-meshi: > However, by not doing that, we could actually merge this now, so that tests are at least running on eslint 8.

:+1: I think it should be added to peerDeps after ESLint v8.0.0 is stable.

tests/src/cli.js
ota-meshi: I will open another PR later.

src/rules/no-unused-modules.js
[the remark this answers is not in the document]
ota-meshi: I didn't understand what you meant. (Maybe it's because I'm not familiar with English...)
Are you saying I shouldn't fix it even if ESLint warns me?

https://github.com/import-js/eslint-plugin-import/runs/3345983029?check_suite_focus=true

image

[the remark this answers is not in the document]
ljharb: yes, those are just warnings, not errors.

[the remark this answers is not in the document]
ota-meshi: Thank you! I understand. Revert comma changes.

tests/src/cli.js
ota-meshi: I opened PR #2197

src/rules/no-unused-modules.js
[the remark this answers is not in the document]
ota-meshi: I made change to this PR https://github.com/import-js/eslint-plugin-import/pull/2191/commits/b45d9ecb15d0dc28066176585c4fe88665e007d7

src/rules/no-unused-modules.js
```diff
@@ -15,9 +15,17 @@ import includes from 'array-includes';
...
 try {
-  const FileEnumerator = require('eslint/lib/cli-engine/file-enumerator').FileEnumerator;
+  // eslint/lib/cli-engine/file-enumerator has been moved to eslint/use-at-your-own-risk
+  // in version 8
+  let FileEnumerator;
+  try {
+    FileEnumerator = require('eslint/use-at-your-own-risk').FileEnumerator;
+  } catch (e) {
+    FileEnumerator = require('eslint/lib/cli-engine/file-enumerator').FileEnumerator;
+  }
```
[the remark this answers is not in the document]
ota-meshi: Let me know if you would like me to make changes to this file into another new PR.

ota-meshi:
I think I've done everything I can do with eslint v8 beta right now.
At this time, I understand that this PR will not be merged as is, as CI contains many workarounds.
If you want to merge only the changes in `src/rules/no-unused-modules.js` and `tests/src/cli.js`, please let me know as I will open another PR. We have confirmed that these changes work well with the CI of this PR.

ljharb:
The test one, sure - but for the rule, I think it's better if it fails in eslint 8 prior to us supporting it officially.

ota-meshi:
I will open a PR that changes `tests/src/cli.js`.

PR #2193, opened 2021-08-16
Fix test cases of no-amd and no-commonjs

ota-meshi:
This PR fixes the no-amd and no-commonjs rule test cases.
See https://github.com/import-js/eslint-plugin-import/pull/2191#discussion_r689245594.

Previous test cases depended on the babel-eslint v8 patch to work. The following command fails because the babel-eslint v8 patch is not used.

```bash
npx cross-env BABEL_ENV=test nyc -s mocha tests/src/rules/no-amd.js
npx cross-env BABEL_ENV=test nyc -s mocha tests/src/rules/no-commonjs.js
```

With this PR, the test cases will work even with the above command.

### DIFF
--- a/tests/src/rules/no-amd.js
+++ b/tests/src/rules/no-amd.js
@@ -2,7 +2,7 @@ import { RuleTester } from 'eslint';
 import eslintPkg from 'eslint/package.json';
 import semver from 'semver';
 
-const ruleTester = new RuleTester();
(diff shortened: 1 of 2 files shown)


ljharb (approved):
Thanks!

ljharb:
Looks like there's a few failing tests.

ota-meshi:
sorry. I ran the tests individually to make sure it worked.
When run the whole test, babel patch seems to ignore `sourceType:'script'` (It's in the test case I added.). I remove this test case.

PR #2194, opened 2021-08-16
feature: no-unused-modules: add support of eslint v8

coderaiser:
Add support of [ESLint v8](https://eslint.org/docs/8.0.0/user-guide/migrating-to-8.0.0), the `/lib` entry point has removed, and

```js
const { FileEnumerator } = require("eslint/lib/cli-engine/file-enumerator");
```

should be changed to
```js
const { FileEnumerator } = require("eslint/use-at-your-own-risk");
```

Here is how [it looks like in v8 right now](https://github.com/typescript-eslint/typescript-eslint/pull/3737/checks?check_run_id=3337101710#step:9:15)
image

Related to https://github.com/typescript-eslint/typescript-eslint/issues/3735 and https://github.com/typescript-eslint/typescript-eslint/pull/3737#issuecomment-899318887


src/rules/no-unused-modules.js
```diff
@@ -17,16 +17,19 @@ import includes from 'array-includes';
...
-    const e = new FileEnumerator({
-      extensions: extensions,
-    });
-    return Array.from(e.iterateFiles(src), ({ filePath, ignored }) => ({
-      ignored,
-      filename: filePath,
-    }));
-  };
+  let { FileEnumerator } = require("eslint/use-at-your-own-risk");
+} finally {
```
ljharb, line 21: this `finally` won't suppress the exception - you need to use `catch` here and nest everything else inside that catch.

ljharb (changes requested):
I believe this is a duplicate of #2191. Separately, we won't be supporting eslint 8 officially until it's out of prerelease.

However, this seems like it might be useful to pull in early.

src/rules/no-unused-modules.js
```diff
@@ -17,16 +17,20 @@ import includes from 'array-includes';
...
-  listFilesToProcess = function (src, extensions) {
-    const e = new FileEnumerator({
-      extensions: extensions,
-    });
-    return Array.from(e.iterateFiles(src), ({ filePath, ignored }) => ({
-      ignored,
-      filename: filePath,
-    }));
-  };
+  let { FileEnumerator } = require("eslint/use-at-your-own-risk");
```
ljharb, line 20: this let needs to be outside the try, since let is block scoped

src/rules/no-unused-modules.js
```diff
@@ -17,16 +17,20 @@ import includes from 'array-includes';
...
+} catch {}
+
+try {
+    FileEnumerator = FileEnumerator || require('eslint/lib/cli-engine/file-enumerator').FileEnumerator;
```
ljharb, line 24: Rather than the conditional, let’s nest this inside the catch.

coderaiser:
Let's use [try-catch](https://www.npmjs.com/package/try-catch) it is very popular, make codebase flat and similar concept with tuples has [Go](https://blog.golang.org/error-handling-and-go), [a lot people interested in such version](https://stackoverflow.com/questions/5126560/try-catch-oneliner-available), and it can became [standard](https://es.discourse.group/t/try-catch-oneliner/107) in the future. 

Nesting you suggest me to add is horrible and not maintainable.

src/rules/no-unused-modules.js
coderaiser: And build [pyramid of doom](https://en.wikipedia.org/wiki/Pyramid_of_doom_(programming))?

src/rules/no-unused-modules.js
```diff
@@ -12,46 +12,58 @@ import { dirname, join } from 'path';
 import readPkgUp from 'read-pkg-up';
 import values from 'object.values';
 import includes from 'array-includes';
+import tryCatch from 'try-catch';
```
ljharb, line 15: no thanks on adding this dep; i find deeply nested try-catch syntax highly superior to an arbitrary abstraction, even if it were a hundred levels deep.

ljharb (changes requested):

src/rules/no-unused-modules.js
ljharb: Yes, a thousand times over, since that's both correct and native language syntax.

SimenB:
Any chance of a release of this? While it might still give peer dep warnings etc until "proper" v8 support lands, I believe this should unblock modules using this plugin in their own migration work without resorting to patching the module 🙂

ljharb:
I'll try to make a pass through the open PRs and issues this weekend and see if I can cut a release.

SimenB:
v8 is released, a release of this PR would be awesome 🙂

https://github.com/eslint/eslint/releases/tag/v8.0.0

ljharb:
Yep, planning on it very soon.

PR #2196, opened 2021-08-16
Document `ignoreTypeImports` option for max-dependencies rule

himynameisdave:
Adds documentation for the `ignoreTypeImports` option which was introduced here: https://github.com/import-js/eslint-plugin-import/pull/1847


ljharb (approved):
Thanks!

PR #2197, opened 2021-08-17
[Tests] `cli`: test the current plugin, not an installed one

ota-meshi:
This PR fixes the `cli` test case to add the project eslint-plugin-import in CLIEngine and test.
See https://github.com/import-js/eslint-plugin-import/pull/2191#discussion_r690046415.

### DIFF
--- a/tests/src/cli.js
+++ b/tests/src/cli.js
@@ -7,6 +7,7 @@ import { expect } from 'chai';
 import { CLIEngine } from 'eslint';
 import eslintPkg from 'eslint/package.json';
 import semver from 'semver';
+import * as importPlugin from '../../src/index';
 
 describe('CLI regression tests', function () {
   describe('issue #210', function () {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -78,7 +78,6 @@
     "eslint-import-test-order-redirect": "file:./tests/files/order-redirect",
     "eslint-module-utils": "file:./utils",
     "eslint-plugin-eslint-plugin": "^2.3.0",
-    "eslint-plugin-import": "2.x",
```
ota-meshi, line 81: In the initialization of the cli engine, it is necessary to register it in the dependency in order to read the config. 
The plugin added to the cli engine is used for linting.

ota-meshi: This behavior may be a bug in eslint, but it may not be fixed prior to v8.

ljharb: yeah, i've been playing with it locally; removing this dev dep is much more complex.

ljharb (approved):

PR #2208, opened 2021-08-25
[Tests] use ESLint class in `cli` test

ota-meshi:
This PR change to use the ESLint class in `tests/src/cli.js` test case (when ESLint >= v7).

See https://github.com/import-js/eslint-plugin-import/pull/2191#issuecomment-902143316.


ljharb (approved):
Thanks!

Issue #2123, opened 2021-06-10
`no-restricted-paths` no longer supports regular expressions


RvanderLaan:
On version 2.22.1 I was able to use regular expressions in the paths for a zone, but this no longer works in 2.23.4: 
```json
    "import/no-restricted-paths": [
      "error",
      {
        "zones": [
          {
            "target": "*",
            "from": "src/generated/schema.ts",
            "except": ["src/.*/resolvers.ts"]
          }
        ]
      }
    ]
```
I suspect it's because of https://github.com/benmosher/eslint-plugin-import/pull/2090, where the `contains-paths` dependency was replaced by a custom function. 
The way I set up this zone might not really be the intended use case of this rule, but it might break more valid configurations too. Is this something that could be brought back?

Edit: Regex in the `except` field actually was not supported before, but that would be a nice addition too!

ljharb:
eslint rules should never support regexes; that's the primary source of CVEs (ReDOS, mostly).

https://www.npmjs.com/package/contains-path doesn't mention anything about asterisks being supported, but it's certainly possible they were unintentionally supported, or undocumented, and #2090 removed it.

I'd be happy to review a PR that restored/added the ability for *globs* (not regexes) to go in `zones`. It was certainly never intended for this plugin to accept regexes there.

ninofiliu:
Hey @ljharb I'd be happy to help there! That would greatly simplify an eslint config we have at our company.

I've never contributed here. [Is this section up to date](https://github.com/import-js/eslint-plugin-import/blob/main/CONTRIBUTING.md#prs)? If so, I'll implement such a glob feature in the next weeks or so

ljharb:
As far as i know it is. A PR would be great, thanks.

alexmi256:
This would be a nice feature to have back again.
My use case is that I want to allow imports from `/items` or but not its subfolder, i.e. `items/item1/foo`.
AFAICT using the current version requires that I have to add a zone for each different item type, where as with regex support I can use from: `items/[a-zA-Z0-9]+`

ljharb:
@alexmi256 regex support will not be returning; and should never have been permitted in the first place. However, globs will be supported once a PR comes in, and that will allow `items/*`

PR #2219, opened 2021-09-06
`no-restricted-paths` support glob patterns

stropho:
Connected to https://github.com/import-js/eslint-plugin-import/issues/2123

Added support for glob patterns. More details in `docs/rules/no-restricted-paths.md` in this PR

Although it would be easier to replace the rule so it works with globs only, I tried to don't make any breaking changes 🤞


ljharb (approved):
New tests only, nothing major jumps out at me in the implementation, nice!

PR #2220, opened 2021-09-07
[Fix] no-unresolved: ignore type-only imports

jablko:
I'm getting the following errors when linting TypeScript, without installing eslint-import-resolver-typescript:
```
  1:24  error  Unable to resolve path to module 'mdast'  import/no-unresolved
  3:24  error  Unable to resolve path to module 'unist'  import/no-unresolved
```

Can `import/no-unresolved: off` be added to the `plugin:import/typescript` preset, by the same logic as in #1726 (TypeScript compilation already confirms that imports are resolved)?

### DIFF
--- a/config/typescript.js
+++ b/config/typescript.js
@@ -2,7 +2,10 @@
  * Adds `.jsx`, `.ts` and `.tsx` as an extension, and enables JSX/TSX parsing.
  */
 
-const allExtensions = ['.ts', '.tsx', '.d.ts', '.js', '.jsx'];
+// Omit `.d.ts` because 1) TypeScript compilation already confirms that
+// types are resolved, and 2) it would mask an unresolved
+// `.ts`/`.tsx`/`.js`/`.jsx` implementation.
+const allExtensions = ['.ts', '.tsx', '.js', '.jsx'];
 
 module.exports = {
 
@@ -10,7 +13,7 @@ module.exports = {
     'import/extensions': allExtensions,
     'import/external-module-folders': ['node_modules', 'node_modules/@types'],
     'import/parsers': {
-      '@typescript-eslint/parser': ['.ts', '.tsx', '.d.ts'],
+      '@typescript-eslint/parser': ['.ts', '.tsx'],
     },
     'import/resolver': {
       'node': {
--- a/src/rules/no-unresolved.js
+++ b/src/rules/no-unresolved.js
@@ -27,6 +27,11 @@ module.exports = {
     const options = context.options[0] || {};
 
     function checkSourceValue(source) {
+      // ignore type-only imports
+      if (source.parent && source.parent.importKind === 'type') {
+        return;
+      }
+
       const caseSensitive = !CASE_SENSITIVE_FS && options.caseSensitive !== false;
       const caseSensitiveStrict = !CASE_SENSITIVE_FS && options.caseSensitiveStrict;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,15 +6,18 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 
 ## [Unreleased]
 
-### Changed
-- [Refactor] switch to an internal replacement for `pkg-up` and `read-pkg-up` ([#2047], thanks [@mgwalker])
-- [patch] TypeScript config: remove `.d.ts` from [`import/parsers` setting] and [`import/extensions` setting] ([#2220], thanks [@jablko])
-
 ### Added
 - [`no-unresolved`]: add `caseSensitiveStrict` option ([#1262], thanks [@sergei-startsev])
 - [`no-unused-modules`]: add eslint v8 support ([#2194], thanks [@coderaiser])
 - [`no-restricted-paths`]: add/restore glob pattern support ([#2219], thanks [@stropho])
 - [`no-unused-modules`]: support dynamic imports ([#1660], [#2212], thanks [@maxkomarychev], [@aladdin-add], [@Hypnosphi])
+
+### Fixed
+- [`no-unresolved`]: ignore type-only imports ([#2220], thanks [@jablko])
+
+### Changed
+- [Refactor] switch to an internal replacement for `pkg-up` and `read-pkg-up` ([#2047], thanks [@mgwalker])
+- [patch] TypeScript config: remove `.d.ts` from [`import/parsers` setting] and [`import/extensions` setting] ([#2220], thanks [@jablko])
 
 ## [2.24.2] - 2021-08-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ## [Unreleased]
 
 ### Changed
-- [Refactor] switch to an internal replacement for `pkg-up` and `read-pkg-up` ([#2047], [@mgwalker])
+- [Refactor] switch to an internal replacement for `pkg-up` and `read-pkg-up` ([#2047], thanks [@mgwalker])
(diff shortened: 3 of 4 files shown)


config/typescript.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     'import/extensions': allExtensions,
     'import/external-module-folders': ['node_modules', 'node_modules/@types'],
     'import/parsers': {
-      '@typescript-eslint/parser': ['.ts', '.tsx', '.d.ts'],
+      '@typescript-eslint/parser': ['.ts', '.tsx'],
```
jablko, line 13: Unrelated nit: `.d.ts` only matters in `import/extensions`. Here it's covered by `.ts`.

ljharb:
Does typescript actually confirm resolution for untyped imports?

jablko:
> Does typescript actually confirm resolution for untyped imports?

Yes (I'm pretty sure). If I `rm --recursive node_modules/@types` and `tsc --build`, I get:
```
index.ts:1:24 - error TS2307: Cannot find module 'mdast' or its corresponding type declarations.

1 import * as mdast from "mdast";
                         ~~~~~~~

index.ts:3:24 - error TS2307: Cannot find module 'unist' or its corresponding type declarations.

3 import * as unist from "unist";
                         ~~~~~~~
```

I can't think of a TypeScript setup where that wouldn't be the case ...

ljharb:
Those are types packages. I’m talking about untyped commonjs packages, that the lint rule enforces resolution on.

jablko:
> Those are types packages. I’m talking about untyped commonjs packages, that the lint rule enforces resolution on.

Sorry, good point: TypeScript doesn't complain that `node_modules/mdast` is missing (unless `node_modules/@types/mdast` is also missing). If eslint-import-resolver-typescript *is* installed, neither does `import/no-unresolved`.

That works for me: There's no mdast untyped package. But if I add `export function notImplemented(): void;` -> `node_modules/@types/mdast/index.d.ts` and call `mdast.notImplemented()` from my `index.ts`, I get `Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'mdast'` at runtime (sad but predictable).

As far as I know, eslint-import-resolver-typescript can get the kind of import ([type-only or value](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#basic-concepts)) from the TypeScript parser -- based on usage, regardless of the [`import type` syntax](https://www.typescriptlang.org/docs/handbook/modules.html#importing-types)? Or maybe it merely parses without analysis ...

Ideally eslint-import-resolver-typescript would only consider `node_modules/@types` to satisfy type-only imports, and `import/no-unresolved` would error on value imports. Which implies `import/no-unresolved: error`, as today.

&therefore; my mistake, thanks for spotting! I'll close this and maybe investigate eslint-import-resolver-typescript + import kinds.

jablko:
Reopening this after revising the title because it carries on from the previous discussion: Although TypeScript doesn't confirm that *value* imports can be resolved, it does confirm that [*type-only* imports](https://www.typescriptlang.org/docs/handbook/modules.html#importing-types) can.

&therefore; instead of disabling `import/no-unresolved` in the `plugin:import/typescript` preset, like in #1726, disable it for type-only imports like in #1057.

Also, `import/no-unresolved` currently misses imports that can't be resolved if it finds a `.d.ts` with the same name. For example, if you `import * as name from "module-name"` and `import/no-unresolved` should error because `module-name` can't be resolved, except that a `module-name.d.ts` exists, then it won't in fact report an error. Instead you'll get `Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'module-name'` at runtime.

&therefore; remove `.d.ts` from the `plugin:import/typescript` preset.

The result is that `import/no-unresolved` will catch imports that can't be resolved to a `.ts`/`.tsx`/`.js`/`.jsx` implementation, and TypeScript will catch unresolvable type-only imports (which `import/no-unresolved` will ignore).

### Details

Removing `.d.ts` should only affect you if you have an import with a `.d.ts` but no `.ts`/`.tsx`/`.js`/`.jsx` implementation -- in which case you probably want the error? If you don't want the error because you're only using types and eliding the import at runtime, your options are: make it an explicit type-only import, or restore `.d.ts` to `extensions` in your own ESLint config?

In addition to *explicit* type-only imports, I investigated ignoring *implicit* type-only imports: when you `import * as name from "module-name"`, but never use `name` in a value context (only types). What I found was:
1. If a `.ts`/`.tsx`/`.js`/`.jsx` can't be resolved, `import/no-unresolved` should probably still error, because depending on the [importsNotUsedAsValues](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues) option you'll get `Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'module-name'` at runtime:
   - `error`: error at *compile time*
   - `preserve`: error at *runtime*
   - `remove`: no error (this just happens to be the default)
2. We couldn't ignore implicit type-only imports if we wanted to because it's done with `canConvertImportDeclarationToTypeOnly()` which is an internal API?

So I think the choice is between:
1. Keep `.d.ts` in the preset:
   - miss value imports without a resolvable `.ts`/`.tsx`/`.js`/`.jsx` implementation
   - miss implicit type-only imports that error if `importsNotUsedAsValues` is `preserve`
2. Remove `.d.ts`:
   - error on value imports without a resolvable implementation (good, right?)
   - error on implicit type-only imports that otherwise only error if `importsNotUsedAsValues` is `preserve` (must make them explicit restore `.d.ts` yourself)

Note that case 1 (keep `.d.ts`) will *still* error on implicit type-only imports from `@types` packages, because only eslint-import-resolver-typescript understands those:
- This PR fixes `import/no-unresolved` on *explicit* type-only imports, in both cases (so now you can silence these `import/no-unresolved` `@types` errors by making them explicit)
- You still might want eslint-import-resolver-typescript to find extraneous `@types` dependencies (no-extraneous-dependencies)

src/rules/no-unresolved.js
```diff
@@ -24,6 +24,11 @@ module.exports = {
 
   create(context) {
     function checkSourceValue(source) {
+      // ignore type-only imports
+      if (source.parent.importKind === 'type') {
+        return;
+      }
```
ljharb, line 30: can we get some tests for this? ideally both using `import type` and also regular `import`

ljharb (changes requested):

src/rules/no-unresolved.js
jablko: You bet -- I cargo-culted these tests from some of the other type-only tests but they seem to be failing on Node *.[45]? I don't recognize the pattern ... I'll keep investigating. Do you spy anything obvious?
https://github.com/import-js/eslint-plugin-import/blob/35e14312a47812816fdd618c5dbceddc809ceaf2/tests/src/rules/no-unresolved.js#L418-L436

Also, any idea what's with the "TypeScript TypeScript TypeScript" in `1) TypeScript TypeScript TypeScript no-unresolved ignore type-only valid import type { JSONSchema7Type } from "@types/json-schema";:`? https://github.com/import-js/eslint-plugin-import/pull/2220/checks?check_run_id=3576242736#step:5:72

jablko: I think I solved it: `parser !== require.resolve('typescript-eslint-parser')` like https://github.com/import-js/eslint-plugin-import/blob/7784948fc5024476f0861b7b8bb20603f28cf809/tests/src/rules/no-duplicates.js#L420

Tests are passing now -- did I do it right?

ljharb (approved):
This seems reasonable to me. `.d.ts` is of course redundant when we have `.ts`; and `no-unresolved` shouldn't bother checking type imports since tsc/flow does that for us. Thanks!

jablko:
Thank you!

PR #2226, opened 2021-09-12
[Fix] `order`: Fix import ordering in TypeScript module declarations

### LINKED ISSUE #2217: import/order false positives for TypeScript module declarations
The `import/order` rule gives false positives for imports inside TypeScript module declarations.

Practical example:

```ts
declare module 'monaco-editor/esm/vs/base/common/cancellation' {
  export enum CancellationToken {
    None,
  }
}

declare module 'monaco-editor/esm/vs/editor/contrib/documentSymbols/documentSymbols' {
  import { ITextModel, languages } from 'monaco-editor';
  import { CancellationToken } from 'monaco-editor/esm/vs/base/common/cancellation';

  export function getDocumentSymbols(
    model: ITextModel,
    flat: boolean,
    token: CancellationToken,
  ): Promise<languages.DocumentSymbol[]>;
}

declare module 'monaco-editor/esm/vs/editor/editor.worker' {
  import { worker } from 'monaco-editor/esm/vs/editor/editor.api';

  export function initialize(
    fn: (ctx: worker.IWorkerContext, createData: unknown) => unknown,
  ): void;
}
```

This is happening, because `import/order` collects all import statements and assumes they are in the program body.

I would expect the `import/order` rule to be applied to each body separately. Currently these are the `Program` body and the `TSModuleBlock` body, but who knows what the future holds.

Alternatively the rule could be applied to imports which are direct children `Program` body only.

remcohaszing:
Without this, `import/order` checks if all imports in a file are sorted. The autofix would then move all imports to the type of the file, breaking TypeScript module declarations.

Closes #2217

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -336,7 +336,7 @@ function registerNode(context, importEntry, ranks, imported, excludedImportTypes
   }
 }
 
-function isModuleLevelRequire(node) {
+function getRequireBlock(node) {
   let n = node;
   // Handle cases like `const baz = require('foo').bar.baz`
   // and `const foo = require('foo')()`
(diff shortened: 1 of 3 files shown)


src/rules/order.js
```diff
@@ -670,22 +683,24 @@ module.exports = {
...
-          imported,
+          getBlockImports(block),
           pathGroupsExcludedImportTypes
         );
       },
       'Program:exit': function reportAndReset() {
-        if (newlinesBetweenImports !== 'ignore') {
-          makeNewlinesBetweenReport(context, imported, newlinesBetweenImports);
-        }
+        for (const imported of importMap.values()) {
```
ljharb, line 691: would prefer importMap.forEach here

ljharb (approved):
Seems great!

Issue #909, opened 2017-08-04
Request: option to let `no-duplicates` force type & normal imports to be de-duplicated


TiddoLangerak:
#225 asked to allow both type & normal imports to be allowed together, e.g.

```javascript
import {someFunction} from '../src/foo';
import type {someType} from '../src/foo';
```

This was requested because according to the author there was no way to merge these together at that point in time.

However, currently flow *does* support merging these imports together:

```
import { someFunction, type someType } from '../src/foo';
```

I think it would therefore be useful if `no-duplicates` could be configured to have the same behaviour as before #225, i.e. to require type & normal imports to be merged.

sibelius:
this should be optional

ljharb:
I agree this should be controlled by an option.

loyd:
Some progress?

remcohaszing:
Regular imports and type imports can be merged in TypeScript as well by the same logic. The difference with flow is that in TypeScript, the type imports will look like regular imports.

```ts
import {someFunction} from '../src/foo';
import type {someType} from '../src/foo';
```

becomes

```ts
import {someFunction, someType} from '../src/foo';
```

SimenB:
TypeScript 4.5 also supports inline `type`: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names

ljharb:
We'd probably have to detect TS version regardless to support that, and we may want it to be a separate option. Let's explore it in #2229.

SimenB:
I'd assume somebody picking up #2229 would be welcome, so I suggest starting there 🙂

PR #2229, opened 2021-09-17
[New] `no-duplicates`: add `combineTypeImports` option to de-duplicate flow type imports

christianvuerings:
## Description

[#334](https://github.com/import-js/eslint-plugin-import/pull/334) always duplicated flow type imports. In this diff we add the `combineTypeImports` option to be able to de-duplicate flow type imports.

TypeScript support can be added later, for now we only support combining regular & flow type imports.

Fixes: #909

## Related PRs / Issues

- https://github.com/import-js/eslint-plugin-import/pull/334: Check duplicates for normal imports and flow type imports separately
- https://github.com/import-js/eslint-plugin-import/issues/909: Request: option to let no-duplicates force type & normal imports to be de-duplicated

### DIFF
--- a/docs/rules/no-duplicates.md
+++ b/docs/rules/no-duplicates.md
@@ -14,12 +14,11 @@ is different in two key ways:
 Valid:
 ```js
 import SomeDefaultClass, * as names from './mod'
-// Flow `type` import from same module is fine
+// By default the flow `type` import from same module is fine (see combineTypeImports)
 import type SomeType from './mod'
 ```
 
-...whereas here, both `./mod` imports will be reported:
-
+Invalid:
 ```js
 import SomeDefaultClass from './mod'
 
@@ -36,7 +35,8 @@ The motivation is that this is likely a result of two developers importing diffe
 names from the same module at different times (and potentially largely different
 locations in the file.) This rule brings both (or n-many) to attention.
 
-### Query Strings
+## Options
+### considerQueryString
 
 By default, this rule ignores query strings (i.e. paths followed by a question mark), and thus imports from `./mod?a` and `./mod?b` will be considered as duplicates. However you can use the option `considerQueryString` to handle them as different (primarily because browsers will resolve those imports differently).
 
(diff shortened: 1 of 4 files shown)


docs/rules/no-duplicates.md
```diff
@@ -61,6 +61,25 @@ import SomeDefaultClass from './mod?minify'
 import * from './mod.js?minify'
 ```
 
+### combineTypeImports
+
+By default, this rule ignores flow & typescript type imports. However you can use the option `combineTypeImports` to combine type imports with regular imports.
```
ljharb, line 66: can you explain why this is a good thing, or how it would apply to TS? In TS, it's preferred to use `import type` for all type imports, and `import` only for value imports.

Does Flow not have a separate `import type`?

christianvuerings: > can you explain why this is a good thing, or how it would apply to TS? In TS, it's preferred to use `import type` for all type imports, and `import` only for value imports.

TypeScript allows to not use `import type` for type imports. `typescript-eslint` has a rule to specify the syntax: [typescript-eslint/consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md). As you are saying it's not the preferred way, so it shouldn't be the default, but it is something users might want to set.

The main reason we would like to combine type and regular imports in our flow code is to make it easy for developers to see what the dependencies on a certain file. Some of our files have both regular and type exports so when they are imported, it's confusing to not see them combined:

## Before
```
// @flow
import { type I18nContext } from 'app/packages/i18n/useI18n';
// Other imports
import useI18n, { I18nContextProvider } from 'app/packages/i18n/useI18n';
```

## After
```
// @flow
import useI18n, { I18nContextProvider, type I18nContext } from 'app/packages/i18n/useI18n';
```

> Does Flow not have a separate import type?

Flow supports both `import type Foo` and `import { type Foo }` syntax.

ljharb: I can’t imagine why you’d prefer the combined form over the separate import vs import type statements, but at least i understand what the option would do.

Since they can’t be combined in TS, this option would only ever be useful for Flow. I’m not sure supporting only flow is worth the complexity at this time due to the current state of Flow usage in the ecosystem.

christianvuerings: @ljharb understood, closing this PR for now.

ljharb: Looks like in TS 4.5+, they can be combined. I'll reopen.

christianvuerings:
Closing since I no longer plan to work on this. Anyone should feel free to open a new PR and can use this PR as inspiration.

ljharb:
That’s fine, but let’s leave it open so this PR can be updated later :-)

PR #2231, opened 2021-09-20
import/extensions: Reverse logic of `ignorePackages` with `never`

Drafter500:
`ignorePackages` flag always works in one direction, it only allows skipping extensions in packages in `always` mode (default), but it is impossible to make it reverse: allow having extensions for packages in `never` mode. Currently in combination with `never`, it does nothing. By the name of the flag, I expected it to stop complaining about package imports in any situation. 

For example, with this config:
`['error', 'never', {ignorePackages: true} ]`

It should forbid extensions except for packages:
```js
import Component from './Component'; // OK
import Component from './Component.jsx'; // Not OK
import baz from 'package/module-one'; // OK
import bizz from 'package/module-two.js'; // Should be also OK!!! 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 - [`no-unused-modules`]: add eslint v8 support ([#2194], thanks [@coderaiser])
 - [`no-restricted-paths`]: add/restore glob pattern support ([#2219], thanks [@stropho])
 - [`no-unused-modules`]: support dynamic imports ([#1660], [#2212], thanks [@maxkomarychev], [@aladdin-add], [@Hypnosphi])
+- [`import/extensions`]: revert `ignorePackages` with `never` ([#2231], thanks [@Drafter500])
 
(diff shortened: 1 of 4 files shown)


ljharb:
Why should that be ok?? Packages without “exports” don’t need extensions, and packages with it should be defining their paths without them as well. Can you provide an example where both forms are valid, and you *want* the extension on third-party code?

Drafter500:
> Why should that be ok?? Packages without “exports” don’t need extensions, and packages with it should be defining their paths without them as well. Can you provide an example where both forms are valid, and you _want_ the extension on third-party code?

With new version of `lit` some parts requires extensions when importing them:
https://github.com/lit/lit/issues/1986#issuecomment-870084329

ljharb:
@Drafter500 in that case, the solution is for the `resolve` package to add "exports" support, and then for this plugin to not try to check extensions on any package with "exports" - it'd only be if a package provides two "exports" entries that point to the same file, and one had an extension and one didn't, and you *also* preferred the one WITH an extension, that this PR would come into play - and I'm not sure why anyone would prefer using the extension when it's possible to omit it.

PR #2232, opened 2021-09-20
[Tests] fix skip usage

jablko:
Conditionally skipping tests doesn't work this way, I think?


tests/src/rules/max-dependencies.js
```diff
@@ -91,7 +91,7 @@ ruleTester.run('max-dependencies', rule, {
   ],
 });
 
-context('TypeScript', { skip: semver.satisfies(eslintPkg.version, '>5.0.0') }, () => {
```
jablko, line 94: `{ skip: ... }` isn't a thing, I think? We're passing the callback as the third argument -- if this is the [describe()/context() implementation](https://github.com/mochajs/mocha/blob/8421974e8d252e2ef11602d15bed178e4f5f4534/lib/interfaces/bdd.js#L41), then that argument is ignored?

tests/files/with-typescript-dev-dependencies/package.json
```diff
@@ -1,5 +1,5 @@
 {
   "devDependencies": {
-    "@types/json-schema": "*"
+    "a": "*"
```
jablko, line 3: Now that these tests are actually running, they need a resolved import in order to pass? Otherwise no-extraneous-dependencies ignores them:
https://github.com/import-js/eslint-plugin-import/blob/fd85369fc219c40862d2afcfb11ad053387ba967/src/rules/no-extraneous-dependencies.js#L176

tests/src/rules/max-dependencies.js
ljharb: you're right; this is the way tape works, not BDD-style testing. Thanks for the fix.

ljharb (approved):
Thanks, this seems great

tests/src/rules/max-dependencies.js
```diff
@@ -91,7 +91,7 @@ ruleTester.run('max-dependencies', rule, {
...
+(semver.satisfies(tsPkg.version, '>=3.8') ? describe.skip : describe)('TypeScript', () => {
```
jablko, line 94: Now that the tests are actually running, this predicate wasn't working either. The root reason these tests conditionally fail isn't eslint-typescript-parser or the ESLint version, it's that type-only imports were [added in TypeScript 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html), and we conditionally test older versions: https://github.com/import-js/eslint-plugin-import/blob/fd85369fc219c40862d2afcfb11ad053387ba967/tests/dep-time-travel.sh#L28-L29

Predicating tests on eslint-typescript-parser or the ESLint version coincidentally works because we only install @typescript-eslint/parser on ESLint >= 5 and only downgrade TypeScript on ESLint < 3: https://github.com/import-js/eslint-plugin-import/blob/fd85369fc219c40862d2afcfb11ad053387ba967/tests/dep-time-travel.sh#L16-L18

Predicating tests directly on TypeScript 3.8 is more obvious and less error prone, I think?

I haven't investigated exhaustively, this doesn't explain the removed comment `TODO: figure out why these tests fail in eslint 4` unless it truly meant `< 4`, but the fact that type-only imports didn't exist before TypeScript 3.8 is a probable explanation?

tests/src/rules/max-dependencies.js
jablko: Ah, that makes sense. Thanks for explaining the confusion!

I'm done now, please merge when ready.

jablko: Sorry, I take that back: there is a mistake. One sec ...

tests/src/rules/max-dependencies.js
ljharb: I think you're right, and pivoting based on the TS version makes the most sense.

ljharb:
I just rebased to tweak the commit messages; two tests are failing.

tests/src/rules/max-dependencies.js
jablko: Thanks, but I was mistaken ... In fact the root reason is that type-only imports were added to @typescript-eslint/typescript-estree in https://github.com/typescript-eslint/typescript-eslint/pull/1697, whereas typescript-eslint-parser depends on [the older typescript-estree](https://unpkg.com/browse/typescript-eslint-parser@22.0.0/package.json#L49), so doesn't handle them.

eslint-plugin-import installs @typescript-eslint/parser and typescript-eslint-parser in parallel, hence @typescript-eslint/typescript-estree and typescript-estree, so I can't think of a clever way to predicate the tests on the package versions. Since typescript-eslint-parser never handles type-only imports the right thing to do is to exclude it like before, I think?

ljharb: In that case it seems fine to check the parser, yes.

jablko:
> I just rebased to tweak the commit messages;

Got them, thanks!

> two tests are failing.

Hopefully this does the trick: We downgrade eslint-import-resolver-typescript on Node.js < 8: https://github.com/import-js/eslint-plugin-import/blob/fd85369fc219c40862d2afcfb11ad053387ba967/tests/dep-time-travel.sh#L32-L35

1.0.2 [doesn't resolve .js extensions](https://unpkg.com/browse/eslint-import-resolver-typescript@1.0.2/index.js#L6) (I presume) and this test needs a resolved import: https://github.com/import-js/eslint-plugin-import/blob/fd85369fc219c40862d2afcfb11ad053387ba967/src/rules/no-extraneous-dependencies.js#L176

tests/src/rules/max-dependencies.js
jablko: Done. Thanks!

tests/src/rules/no-extraneous-dependencies.js
```diff
@@ -380,74 +378,46 @@ ruleTester.run('no-extraneous-dependencies', rule, {
...
 });
 
-// TODO: figure out why these tests fail in eslint 4
-describe('TypeScript', { skip: semver.satisfies(eslintPkg.version, '^4') }, function () {
-  getTSParsers().forEach((parser) => {
-    const parserConfig = {
-      parser,
-      settings: {
-        'import/parsers': { [parser]: ['.ts'] },
-        'import/resolver': { 'eslint-import-resolver-typescript': true },
```
ljharb, line 390: should this maybe be kept, but add `node: true`?

jablko: Yes, good idea. Done.

PR #1237, opened 2018-11-19
[Generic Import Callback] Make callback for all import once in rules

ljqx:
This PR fixes #1230.

Not only for `import/extensions`, this PR also adds supports of all 
1) `ImportDeclaration`, 
2) `ExportNamedDeclaration`, 
3) `ExportAllDeclaration`, 
4) static require `CallExpression` 

for rules `import/extensions`, `import/max-dependencies`, `import/no-extraneous-dependencies`, `import/no-internal-modules`, `import/no-nodejs-modules`, `import/no-restricted-paths`, `import/no-self-import`, `import/no-webpack-loader-syntax` by abstracting out the common part for these callbacks.

Tests are added for `import/extensions`.

PR #2233, opened 2021-09-21
[Refactor] `no-unresolved`, `no-extraneous-dependencies`: moduleVisitor usage

jablko:
I overlooked the second moduleVisitor callback parameter (the ImportDeclaration/ImportExpression) in #2220. Doesn't make a functional difference, just tidier than recovering `node` from `node.source`: https://github.com/import-js/eslint-plugin-import/blob/4f0f560544b4d6dcd2d1b7d5880bd5238099f979/utils/moduleVisitor.js#L34

### DIFF
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -162,7 +162,6 @@ function reportIfMissing(context, deps, depsOptions, node, name) {
   // Do not report when importing types
   if (
     node.importKind === 'type' ||
-    (node.parent && node.parent.importKind === 'type') ||
     node.importKind === 'typeof'
   ) {
     return;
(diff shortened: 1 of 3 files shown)


src/rules/no-extraneous-dependencies.js
```diff
@@ -162,7 +162,6 @@ function reportIfMissing(context, deps, depsOptions, node, name) {
   // Do not report when importing types
   if (
     node.importKind === 'type' ||
-    (node.parent && node.parent.importKind === 'type') ||
```
jablko, line 165: Relatedly this was introduced [here](https://github.com/import-js/eslint-plugin-import/pull/1820/files#diff-4c4c694d76da5ddc56956bb432ce2bc90fbbea0404ee005e66ef5f4284bf5a94R114) and rendered obsolete [here](https://github.com/import-js/eslint-plugin-import/commit/93e060f63f3c6f557c7fea9cddc0018c033b058d#diff-4c4c694d76da5ddc56956bb432ce2bc90fbbea0404ee005e66ef5f4284bf5a94R210), I think.

[This comment](https://github.com/import-js/eslint-plugin-import/issues/1618#issuecomment-643082389) is describing `node.source`, which is what `node` used to be, prior to #1237. That requires the `node.parent.importKind` term. After, it's the ImportDeclaration/ImportExpression, which satisfies the `node.importKind` term instead.

ljharb (approved):
thanks, this is great

PR #2234, opened 2021-09-21
[New] no extraneous type-only dependencies

jablko:
If eslint-import-resolver-typescript is installed, instead of ignoring type-only imports entirely, it would be handy if `no-extraneous-dependencies` reported undeclared e.g. `@types` dependencies? In other words, resolve type-only imports with eslint-resolver-typescript or not at all, which is what `no-extraneous-dependencies` does right now.

eslint-import-resolve-typescript is required because other resolvers can't resolve types, so can't tell whether they resolve to e.g. `json-schema` or `@types/json-schema` and so can't tell if the dependency is declared or missing.

TypeScript compilation confirms that types are resolved, but it doesn't confirm that those dependencies are declared in the package.json, so `no-extraneous-dependencies` can be beneficial.


package.json
```diff
@@ -71,7 +71,7 @@
     "cross-env": "^4.0.0",
     "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0",
     "eslint-import-resolver-node": "file:./resolvers/node",
-    "eslint-import-resolver-typescript": "^1.0.2 || ^1.1.1",
+    "eslint-import-resolver-typescript": "^1.0.2 || ^2.5.0",
```
jablko, line 74: eslint-import-resolver-typescript handles `@types` packages since [version 2.0.0](https://unpkg.com/browse/eslint-import-resolver-typescript@2.0.0/lib/index.js#L44).

src/rules/no-extraneous-dependencies.js
```diff
@@ -117,6 +117,28 @@ function optDepErrorMessage(packageName) {
     `not optionalDependencies.`;
 }
 
+function resolveTypeOnly(modulePath, context) {
+  const sourceFile = context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename();
+  const configResolvers = context.settings['import/resolver'];
```
jablko, line 122: Copied from: https://github.com/import-js/eslint-plugin-import/blob/4f0f560544b4d6dcd2d1b7d5880bd5238099f979/utils/resolve.js#L130-L139

src/rules/no-extraneous-dependencies.js
```diff
@@ -196,7 +219,7 @@ function reportIfMissing(context, deps, depsOptions, node, name) {
 
     if (
       declarationStatus.isInDeps ||
-      (depsOptions.allowDevDeps && declarationStatus.isInDevDeps) ||
+      ((depsOptions.allowDevDeps || isTypeOnly) && declarationStatus.isInDevDeps) ||
```
jablko, line 222: It's common practice to declare type-only dependencies in devDependencies -- that was among the reasons for originally ignoring type-only imports: #1618

package.json
ljharb: i'd prefer to continue testing on v1.1.1
```suggestion
    "eslint-import-resolver-typescript": "^1.0.2 || ^1.1.1 || ^2.5.0",
```

tests/src/rules/no-extraneous-dependencies.js
```diff
@@ -378,7 +380,7 @@ ruleTester.run('no-extraneous-dependencies', rule, {
   ],
 });
 
-describe('TypeScript', () => {
+(semver.satisfies(typescriptResolverPkg.version, '>=2.1.0') ? describe : describe.skip)('TypeScript', () => {
```
ljharb, line 383: this means the files in here won't be tested on v1 (v1.0 or v1.1) of the TS resolver. can we preserve that?

utils/resolve.js
```diff
@@ -235,4 +235,6 @@ function resolve(p, context) {
   }
 }
 resolve.relative = relative;
+resolve.resolverReducer = resolverReducer;
+resolve.requireResolver = requireResolver;
```
ljharb, line 239: it'd be ideal to add some unit tests for these. i know there aren't any direct tests of the utils package, but it might be time to add some.

tests/src/rules/no-extraneous-dependencies.js
jablko: Yes but ... This only works with >=2.0.0 of the resolver because earlier versions didn't look in `@types` packages, so testing against them doesn't confirm any special type-only treatment because they return the same thing as the node resolver? Like we exclude typescript-eslint-parser because it predates type-only imports.

I can see adding a "Legacy TypeScript resolver" test and running that in case of an earlier version. I think I'd prefer not, and keep the tests focused on the desired behavior (like we exclude typescript-eslint-parser). Maybe there's a reason it's important though? I'm happy either way!

utils/resolve.js
jablko: I've added some now. What do you think?

tests/src/rules/no-extraneous-dependencies.js
ljharb: Mainly semver - things that work should keep working, and the tests ensure that's the case.

jablko: Fair enough. I've now added a test to confirm that the behavior of earlier versions is unchanged?

erlichmen:
is this PR still relevant? with the latest (eslint v8,   eslint-import-resolver-typescript v2.5. and  eslint-plugin-import: v2.25.3) I'm still getting  "ESLint: '@types/XXXX' should be listed in the project's dependencies, not devDependencies. (import/no-extraneous-dependencies).

ljharb:
@jablko it'd be great to get this rebased :-)

PR #2238, opened 2021-09-24
[New] give resolvers access to parserServices

jablko:
Is there a way for resolvers to access [parserServices](https://eslint.org/docs/developer-guide/working-with-rules#the-context-object)? If not, what do you think about enabling it, maybe like so?

### DIFF
--- a/utils/resolve.js
+++ b/utils/resolve.js
@@ -80,19 +80,20 @@ exports.fileExistsWithCaseSync = function fileExistsWithCaseSync(filepath, cache
   return result;
 };
 
-function relative(modulePath, sourceFile, settings) {
-  return fullResolve(modulePath, sourceFile, settings).path;
+function relative(modulePath, sourceFile, context) {
+  return fullResolve(modulePath, sourceFile, context).path;
 }
 
-function fullResolve(modulePath, sourceFile, settings) {
+function fullResolve(modulePath, sourceFile, context) {
   // check if this is a bonus core module
-  const coreSet = new Set(settings['import/core-modules']);
+  const coreSet = new Set(context.settings['import/core-modules']);
   if (coreSet.has(modulePath)) return { found: true, path: null };
 
   const sourceDir = path.dirname(sourceFile);
-  const cacheKey = sourceDir + hashObject(settings).digest('hex') + modulePath;
+  const cacheKey =
+    sourceDir + hashObject(context.settings).digest('hex') + modulePath;
 
-  const cacheSettings = ModuleCache.getSettings(settings);
+  const cacheSettings = ModuleCache.getSettings(context.settings);
 
   const cachedPath = fileExistsCache.get(cacheKey, cacheSettings);
   if (cachedPath !== undefined) return { found: true, path: cachedPath };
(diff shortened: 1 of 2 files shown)


utils/resolve.js
```diff
@@ -114,7 +120,12 @@ function fullResolve(modulePath, sourceFile, settings) {
     }
 
     function v2() {
-      return resolver.resolve(modulePath, sourceFile, config);
+      return resolver.resolve(
+        modulePath,
+        sourceFile,
+        config,
+        context.parserServices,
```
jablko, line 127: Pass `context.parserServices` -> `resolver.resolve()`.

ljharb:
I guess it seems fine, but we’d need tests, and also what’s the use case? Does any parser actually provide info here, and does any resolver look for it?

utils/resolve.js
```diff
@@ -80,19 +80,20 @@ exports.fileExistsWithCaseSync = function fileExistsWithCaseSync(filepath, cache
   return result;
 };
 
-function relative(modulePath, sourceFile, settings) {
-  return fullResolve(modulePath, sourceFile, settings).path;
+function relative(modulePath, sourceFile, context) {
```
ljharb, line 83: this is actually a breaking change for the utility package, so that seems quite suboptimal

jablko: Can I resolve that by accepting either the new value (context) or the old (settings), as in the latest commit?

utils/resolve.js
```diff
@@ -80,19 +80,25 @@ exports.fileExistsWithCaseSync = function fileExistsWithCaseSync(filepath, cache
...
+  return fullResolve(
+    modulePath,
+    sourceFile,
+    // Accept context or settings, for backward compatibility
+    !('settings' in context) ? { settings: context } : context,
```
ljharb, line 88: I’m pretty skeptical there’s always a “settings” in context, consistently from eslint 2 - 7.

jablko: Well, we pass `context.settings` -> it today, so we already rely on that, don't we?
https://github.com/import-js/eslint-plugin-import/blob/dd814245b7f769a84235f2ffb7204287d47001ba/utils/resolve.js#L220
https://github.com/import-js/eslint-plugin-import/blob/dd814245b7f769a84235f2ffb7204287d47001ba/src/ExportMap.js#L434

ljharb: i guess that's true; if it wasn't an object it'd throw since it's not checked for truthiness. Still, is there a more robust way to check? perhaps eslint provides a Context constructor or something we could use to compare it to?

jablko: Good question ... Here's where it's constructed today, I think: https://github.com/eslint/eslint/blob/cf34e5cf5ed5d09eb53c16cca06821c4e34b7b70/lib/linter/linter.js#L910-L954

Looking at that, I'm not sure what would be more robust, and consistent between ESLint versions?

ljharb: Another alternative is to always pass `settings`, and just *also* pass `context`?

jablko: Totally -- I propose overloading the argument because I think it's functionally equivalent, and slightly tidier than calling `relative(modulePath, sourceFile, context.settings, context)`, but no strong feelings either way ...

ljharb: it's definitely tidier - we could also allow `relative(modulePath, sourceFile, null, context)`, and adapt it later whenever we do a semver-major

PR #2239, opened 2021-09-28
[Docs] order: update the description of the `pathGroupsExcludedImportTypes` option

liby:
Motivation: #2156 
> #1665 I read this issue and knowing why we had this PR, but I think they use pathGroupsExcludedImportTypes in the wrong way. If we need to let pathGroup works, sometimes we have to add pathGroupsExcludedImportTypes as [].

@Mairu @Akiq2016 @yola-0316 PTAL.

### DIFF
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -131,7 +131,7 @@ Properties of the objects
 ### `pathGroupsExcludedImportTypes: [array]`:
 
 This defines import types that are not handled by configured pathGroups.
-This is mostly needed when you want to handle path groups that look like external imports.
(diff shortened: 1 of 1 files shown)


Mairu:
Sorry, I missed this PR when it was opened, but yes the changes are correct and the current documentation is wrong. Using patterns as excluded types works accidentally because the default types are removed.

ljharb:
it's very unfortunate that the fork was deleted, making this PR unrecoverable, so now we need #3036 as a duplicate :-/

please never delete a fork that has open PRs.

PR #2240, opened 2021-09-28
Resolve extended TypeScript configuration files

### LINKED ISSUE #1908: [import/default] typescript - esModuleInterop value not read correctly
The following PR was created to detect the esModuleInterop parameter in tsconfig.json. https://github.com/benmosher/eslint-plugin-import/commit/0b585a14d605eb059050bb07ab3c2a2abf3961a6

It seems that the tsconfig.json file is loaded like a text file. So any extended configuration provided in the  "extends" parameter of the tsconfig.json file won't be detected. 

**Expected:**
I can set the esModuleInterop value in a tsconfig.base.json that will be extended by tsconfig.json and the rule import/default will work

**Here you can find a reproduction of the bug:**
https://github.com/Ebooth/eslint-plugin-import-esModuleInterop-bug

Also it could be nice to add support for monorepo as the current code works only in you have one typescript project.

mrmckeb:
This PR resolves #1908.

In a comparison I ran locally, the output is:
```js
// New method - includes extended configuration
{
  options: {
    esModuleInterop: true,
    forceConsistentCasingInFileNames: true,
    noFallthroughCasesInSwitch: true,
    skipLibCheck: true,
    strict: true,
    baseUrl: '[redacted]/tsconfig.json',
    allowJs: true,
    jsx: 1,
    target: 99,
    module: 99,
    lib: [ 'lib.dom.d.ts', 'lib.esnext.d.ts' ],
    noEmit: true,
    moduleResolution: 2,
    resolveJsonModule: true,
    isolatedModules: true,
    incremental: true,
    paths: { '~/*': [ './*' ], '@*': [ '../shared/*' ] },
    pathsBasePath: '[redacted]/tsconfig.json',
    configFilePath: undefined
  }
  // ...
}

// Old method - only references the config to extend.
{
  extends: '[extended-config-name]',
  compilerOptions: {
    baseUrl: '.',
    allowJs: true,
    jsx: 'preserve',
    target: 'ESNext',
    module: 'ESNext',
    lib: [ 'DOM', 'ESNext' ],
    noEmit: true,
    moduleResolution: 'node',
    resolveJsonModule: true,
    isolatedModules: true,
    incremental: true,
    paths: { '~/*': [Array], '@*': [Array] }
  },
  // ...
}
```

For reference, the shared config has:
```json
{
  "compilerOptions": {
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "noFallthroughCasesInSwitch": true,
    "skipLibCheck": true,
    "strict": true
  }
}
```

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -1,4 +1,5 @@
 import fs from 'fs';
+import { dirname } from 'path';
 
 import doctrine from 'doctrine';
 
@@ -18,7 +19,7 @@ import { tsConfigLoader } from 'tsconfig-paths/lib/tsconfig-loader';
 
 import includes from 'array-includes';
 
-let parseConfigFileTextToJson;
+let ts;
 
 const log = debug('eslint-plugin-import:ExportMap');
 
@@ -525,12 +526,15 @@ ExportMap.parse = function (path, content, context) {
     });
     try {
       if (tsConfigInfo.tsConfigPath !== undefined) {
-        const jsonText = fs.readFileSync(tsConfigInfo.tsConfigPath).toString();
-        if (!parseConfigFileTextToJson) {
-          // this is because projects not using TypeScript won't have typescript installed
-          ({ parseConfigFileTextToJson } = require('typescript'));
-        }
-        return parseConfigFileTextToJson(tsConfigInfo.tsConfigPath, jsonText).config;
+        // Projects not using TypeScript won't have `typescript` installed.
+        if (!ts) { ts = require('typescript'); }
+  
+        const configFile = ts.readConfigFile(tsConfigInfo.tsConfigPath, ts.sys.readFile);
+        return ts.parseJsonConfigFileContent(
+          configFile.config,
+          ts.sys,
+          dirname(tsConfigInfo.tsConfigPath),
+        );
       }
     } catch (e) {
       // Catch any errors
@@ -545,11 +549,11 @@ ExportMap.parse = function (path, content, context) {
     }).digest('hex');
     let tsConfig = tsConfigCache.get(cacheKey);
     if (typeof tsConfig === 'undefined') {
-      tsConfig = readTsConfig();
+      tsConfig = readTsConfig(context);
       tsConfigCache.set(cacheKey, tsConfig);
     }
 
-    return tsConfig && tsConfig.compilerOptions ? tsConfig.compilerOptions.esModuleInterop : false;
+    return tsConfig && tsConfig.options ? tsConfig.options.esModuleInterop : false;
   }
 
   ast.body.forEach(function (n) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This change log adheres to standards from [Keep a CHANGELOG](http://keepachangel
 ### Fixed
 - [`no-unresolved`]: ignore type-only imports ([#2220], thanks [@jablko])
 - [`order`]: fix sorting imports inside TypeScript module declarations ([#2226], thanks [@remcohaszing])
+- [`default`], `ExportMap`: Resolve extended TypeScript configuration files ([#2240], thanks [@mrmckeb])
 
 ### Changed
 - [Refactor] switch to an internal replacement for `pkg-up` and `read-pkg-up` ([#2047], thanks [@mgwalker])
@@ -915,6 +916,7 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#2240]: https://github.com/import-js/eslint-plugin-import/pull/2240
 [#2233]: https://github.com/import-js/eslint-plugin-import/pull/2233
 [#2226]: https://github.com/import-js/eslint-plugin-import/pull/2226
 [#2220]: https://github.com/import-js/eslint-plugin-import/pull/2220
@@ -1509,6 +1511,7 @@ for info on changes for earlier releases.
 [@mgwalker]: https://github.com/mgwalker
 [@MikeyBeLike]: https://github.com/MikeyBeLike
 [@mplewis]: https://github.com/mplewis
+[@mrmckeb]: https://github.com/mrmckeb
 [@nickofthyme]: https://github.com/nickofthyme
 [@nicolashenry]: https://github.com/nicolashenry
 [@noelebrun]: https://github.com/noelebrun
--- a/tests/files/typescript-extended-config/index.d.ts
+++ b/tests/files/typescript-extended-config/index.d.ts
@@ -1,0 +1,3 @@
+export = FooBar;
+
+declare namespace FooBar {}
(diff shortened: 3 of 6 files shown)


dglsparsons (approved):

src/ExportMap.js
```diff
@@ -785,3 +763,32 @@ function makeSourceCode(text, ast) {
     return new SourceCode({ text, ast });
   }
 }
+
+/**
+ * Returns a project's TypeScript configuration.
+ */
+export function readTsConfig(context) {
```
ljharb, line 770: relying on hoisting is a bad practice; can we move this back up to where it was defined previously? that will have the side bonus of making the diff cleaner

tests/src/ExportMap.js
```diff
@@ -0,0 +1,21 @@
+import { expect } from 'chai';
+import { readTsConfig } from '../../src/ExportMap';
```
ljharb, line 2: let's please not export things solely to test them; can these test fixtures be used to text the actual *effects* of the bug, as reported in the original issue?

ljharb (changes requested):

tests/src/ExportMap.js
mrmckeb: I'll make that change now :)

src/ExportMap.js
mrmckeb: I'm also not a fan of relying on hoisting, but noted it was already in use in this file - and assumed it was a project convention, sorry. Moved back and will push an update soon.

ljharb: I haven't yet enabled the rule that forbids it; there's a lot of legacy coding patterns in the project to clean up. Thanks :-)

src/ExportMap.js
```diff
@@ -525,31 +526,34 @@ ExportMap.parse = function (path, content, context) {
...
     try {
       if (tsConfigInfo.tsConfigPath !== undefined) {
-        const jsonText = fs.readFileSync(tsConfigInfo.tsConfigPath).toString();
-        if (!parseConfigFileTextToJson) {
-          // this is because projects not using TypeScript won't have typescript installed
-          ({ parseConfigFileTextToJson } = require('typescript'));
-        }
-        return parseConfigFileTextToJson(tsConfigInfo.tsConfigPath, jsonText).config;
+        // Projects not using TypeScript won't have `typescript` installed.
+        if (!ts) (ts = require('typescript'));
```
ljharb, line 530:
```suggestion
        if (!ts) { ts = require('typescript'); }
```

src/ExportMap.js
```diff
@@ -525,31 +526,34 @@ ExportMap.parse = function (path, content, context) {
...
+        );
       }
     } catch (e) {
       // Catch any errors
     }
-
+  
     return null;
-  }
+  }
```
ljharb, line 544:
```suggestion
  }
```

src/ExportMap.js
```diff
@@ -525,31 +526,34 @@ ExportMap.parse = function (path, content, context) {
...
+          configFile.config,
+          ts.sys,
+          dirname(tsConfigInfo.tsConfigPath),
+        );
       }
     } catch (e) {
       // Catch any errors
     }
-
+
```
ljharb, line 542:
```suggestion

```

src/ExportMap.js
```diff
@@ -525,31 +526,34 @@ ExportMap.parse = function (path, content, context) {
...
-        if (!parseConfigFileTextToJson) {
-          // this is because projects not using TypeScript won't have typescript installed
-          ({ parseConfigFileTextToJson } = require('typescript'));
-        }
-        return parseConfigFileTextToJson(tsConfigInfo.tsConfigPath, jsonText).config;
+        // Projects not using TypeScript won't have `typescript` installed.
+        if (!ts) (ts = require('typescript'));
+  
+        const configFile = ts.readConfigFile(tsConfigInfo.tsConfigPath, ts.sys.readFile);
+        return ts.parseJsonConfigFileContent(
```
ljharb, line 533: can you confirm which versions of typescript offer this method?

ljharb (changes requested):

src/ExportMap.js
mrmckeb: Sorry, was saving without formatting - Prettier was aggressively attacking the file. Resolved!

src/ExportMap.js
mrmckeb: I'm looking now. I see references to this from 2015:
https://github.com/microsoft/TypeScript/issues/5276

Based on the dates here, that would indicate it was around pre-v3.
https://github.com/microsoft/TypeScript/releases?after=v3.0.0

mrmckeb: It looks like it was added around 1.7.2.
https://github.com/Microsoft/TypeScript/blob/v1.7.2/src/compiler/commandLineParser.ts#L416

It's not in 1.6.2, at least not in this file.

ljharb: We only test on v2 and v3 (and should add v4) so I'm comfortable not caring about old v1. Thanks!

mrmckeb:
Thanks for the feedback @ljharb. Let me know if any other changes are needed and I can make them tomorrow :)

ljharb (approved):
Looks great!

mrmckeb:
The errors I'm seeing on Appveyor are:
```
## Running `apt-get update` for you...
+ apt-get update
Hit:1 http://security.ubuntu.com/ubuntu focal-security InRelease
Ign:2 https://deb.nodesource.com/node_16.x focal InRelease
Err:3 https://deb.nodesource.com/node_16.x focal Release
  Certificate verification failed: The certificate is NOT trusted. The certificate chain uses expired certificate.  Could not handshake: Error in the certificate verification. [IP: 23.220.206.33 443]
Hit:4 http://archive.ubuntu.com/ubuntu focal InRelease
Hit:5 http://archive.ubuntu.com/ubuntu focal-updates InRelease
Hit:6 http://archive.ubuntu.com/ubuntu focal-backports InRelease
Reading package lists...
wsl : E: The repository 'https://deb.nodesource.com/node_16.x focal Release' does not have a Release file.
At line:1 char:1
+ wsl curl -sL 'https://deb.nodesource.com/setup_${nodejs_version}.x' ` ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (E: The reposito...a Release file.:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
Error executing command, exiting
Command executed with exception: E: The repository 'https://deb.nodesource.com/node_16.x focal Release' does not have a Release file.
```

Which seems like it could be something else? Windows 11 has kind of been released in the last day or so, could be related?

mrmckeb:
Thanks for the updates @ljharb!

ljharb:
Seems like appveyor fixed itself.

PR #2246, opened 2021-10-05
WIP: [Fix] use import/ignore in more rules

### LINKED ISSUE #2244: Docs on import/ignore either incorrect or unclear
> In practice, this means rules other than `no-unresolved` will not report on any imports with (absolute filesystem) paths matching this pattern.

In my eslint config I am attempting to ignore two aliases created by SvelteKit: `$lib` and `$app`. However, `no-relative-parent-imports` is still tripping on those in Svelte files which reference them. Am I misunderstanding the documentation?

I think a change like this could work, although it seems like having it somewhere more central might be better so that each rule did not have to do this:

```
diff --git a/src/rules/no-relative-parent-imports.js b/src/rules/no-relative-parent-imports.js
index 8e36962..b4da147 100644
--- a/src/rules/no-relative-parent-imports.js
+++ b/src/rules/no-relative-parent-imports.js
@@ -2,6 +2,7 @@ import moduleVisitor, { makeOptionsSchema } from 'eslint-module-utils/moduleVisi
 import docsUrl from '../docsUrl';
 import { basename, dirname, relative } from 'path';
 import resolve from 'eslint-module-utils/resolve';
+import isIgnored from '../../utils/ignore';
 
 import importType from '../core/importType';
 
@@ -21,6 +22,10 @@ module.exports = {
     function checkSourceValue(sourceNode) {
       const depPath = sourceNode.value;
 
+      if (isIgnored(depPath)) {
+        return;
+      }
+
       if (importType(depPath, context) === 'external') { // ignore packages
         return;
       }
```
I will be happy to submit a PR once we have good direction for a proper solution.

wallw-bits:
This is to match the description of the setting in the readme.

resolves #2244

### DIFF
--- a/src/rules/default.js
+++ b/src/rules/default.js
@@ -1,5 +1,6 @@
 import Exports from '../ExportMap';
 import docsUrl from '../docsUrl';
+import isIgnored from '../../utils/ignore';
 
 module.exports = {
   meta: {
@@ -13,6 +14,9 @@ module.exports = {
   create(context) {
 
     function checkDefault(specifierType, node) {
+      if (node.source && isIgnored(node.source.value, context)) {
+        return;
+      }
 
       const defaultSpecifier = node.specifiers.find(
         specifier => specifier.type === specifierType
--- a/src/rules/named.js
+++ b/src/rules/named.js
@@ -1,6 +1,7 @@
 import * as path from 'path';
 import Exports from '../ExportMap';
 import docsUrl from '../docsUrl';
+import isIgnored from '../../utils/ignore';
 
 module.exports = {
   meta: {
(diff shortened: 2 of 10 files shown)


wallw-bits:
Maybe I was unclear on the usage of `utils/ignore.js`, I'm using it on the source of imports, and it may have been intended for resolved absolute paths only.

ljharb:
Certainly since paths could look very different based on where they're imported from, the resolved path is what's needed to be used.

wallw-bits:
Hmm. I don't think that makes much sense from the dev use perspective though. Consider the case where we are trying to ignore a particular alias:

```javascript
import foo from '$alias/foo';
```
If `$alias` resolves to `src/my-path/`, and then we add an `import/ignore` for `/my-path/`, then we have effectively ignored all imports in `my-path`, even ones which do not use the alias.

e.g.
```javascript
// src/my-path/bar.js
import foo from './foo'; // is ignored, so if we muck up a rule here, and rules obey `import/ignore`, then it won't be caught
```

ljharb:
The spirit of the rule has nothing to do with what the path looks like - the rule's meant to ensure you don't import things from a parent directory, regardless of how you get to them.

PR #2269, opened 2021-10-21
[Fix] `no-duplicates` now removes duplicate import specifiers and sorts specifiers alphabetically #1812 #1956

### LINKED ISSUE #1812: no-duplicates should remove duplicate elements
[`no-duplicates`](https://github.com/benmosher/eslint-plugin-import/blob/v2.20.2/docs/rules/no-duplicates.md) should remove duplicate Imported values.

This case is most commonly seen when resolving merge conflicts where identical imports can be copied. It would be helpful if the fix were able to de-duplicate the imported values themselves not only the paths.

```ts
// bad
import { foo } from '../foo'
import { bar } from '../bar'
import { foo } from '../foo'

// current fix
import { foo, foo } from '../foo' // <-- creates duplicate imported values
import { bar } from '../bar'

// idealized fix
import { foo } from '../foo' // <-- removed duplicated value
import { bar } from '../bar'
```

### Edge cases

```ts
// used with `as`
import { foo as foo1 } from '../foo'
import { bar } from '../bar'
import { foo as foo2 } from '../foo'

// idealized fix to avoid naming collisions 
import { foo as foo1, foo as foo2 } from '../foo' // <-- removed duplicated value
import { bar } from '../bar'
```

### LINKED ISSUE #1956: no-duplicates does not add whitespace after commas
I had some code like this

```js
import { a } from "pkg"
import { b } from "pkg"
```

"import/no-duplicates" did work, but it left `a` and `b` glued together like so

```js
import { a,b } from "pkg"
```

Having the imports glued by commas upsets Prettier. A sensible solution is to add a whitespace after the comma(s).

```js
import { a, b } from "pkg"
```

fingerartur:
Fixes #1812. Fixes #1956.


tests/src/rules/no-duplicates.js
```diff
@@ -402,7 +401,7 @@ import {x,y} from './foo'
     // #2027 long import list generate empty lines
     test({
       code: "import { Foo } from './foo';\nimport { Bar } from './foo';\nexport const value = {}",
-      output: "import { Foo , Bar } from './foo';\nexport const value = {}",
+      output: "import { Bar, Foo } from './foo';\nexport const value = {}",
```
ljharb, line 404: this suggests it's a breaking change. Why is this order reversed? Foo appears first in the original source code; it should appear first in the autofix.

tests/src/rules/no-duplicates.js
```diff
@@ -119,49 +119,48 @@ ruleTester.run('no-duplicates', rule, {
...
       code: "import'./foo'; import {x} from './foo'",
-      output: "import {x} from'./foo'; ",
+      output: "import { x } from './foo'; ",
       errors: ['\'./foo\' imported multiple times.', '\'./foo\' imported multiple times.'],
     }),
 
     test({
       code: "import './foo'; import { /*x*/} from './foo'; import {//y\n} from './foo'; import {z} from './foo'",
-      output: "import { /*x*///y\nz} from './foo';   ",
+      output: "import { z } from './foo'; import { /*x*/} from './foo'; import {//y\n} from './foo'; ",
```
ljharb, line 157: this isn't much of an improvement; it ends up leaving behind two useless import statements just for the comments.

tests/src/rules/no-duplicates.js
```diff
@@ -119,49 +119,48 @@ ruleTester.run('no-duplicates', rule, {
...
       code: "import {x} from './foo'; import {} from './foo'",
-      output: "import {x} from './foo'; ",
+      output: "import { x } from './foo'; ",
       errors: ['\'./foo\' imported multiple times.', '\'./foo\' imported multiple times.'],
     }),
 
     test({
       code: "import {x} from './foo'; import {} from './foo'; import {/*c*/} from './foo'; import {y} from './foo'",
-      output: "import {x/*c*/,y} from './foo';   ",
+      output: "import { x, y } from './foo';  import {/*c*/} from './foo'; ",
```
ljharb, line 133: same here; the comment must appear after the x and before the y, or else it's likely not to make sense.

tests/src/rules/no-duplicates.js
```diff
@@ -119,49 +119,48 @@ ruleTester.run('no-duplicates', rule, {
 
     test({
       code: "import { x, /* x */ } from './foo'; import {//y\ny//y2\n} from './foo'",
-      output: "import { x, /* x */ //y\ny//y2\n} from './foo'; ",
```
ljharb, line 122: why is this example no longer autofixable?

src/core/utils/array.js
```diff
@@ -0,0 +1,11 @@
...
+ * Same as the classic array.flat() method, which I cannot use
+ * because that does not pass some CI pipelines. I'm guessing
+ * there is somehow differently configured transpilation.
+ * Didn't have time to get into the details.
+ */
+export const arrayFlat = (arrayOfArrays) => {
+  return arrayOfArrays.reduce((acc, item) => {
+    return acc.concat(item)
+  }, [])
+}
```
ljharb, line 11: instead of inlining this utility, please use https://npmjs.com/array.prototype.flat.

Transpilation only changes syntax; array methods are API.

tests/src/rules/no-duplicates.js
```diff
@@ -412,6 +411,20 @@ import {x,y} from './foo'
...
+    /**
+     * Sadly I cannot run this test because I am getting a fatal parsing error:
+     *
+     * `AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: A fatal parsing error occurred: Parsing error: Identifier 'Foo' has already been declared`
+     *
+     * If there is some option to disable this check, we could uncomment this test again.
+     */
+    // #1812 merged imports should not have duplicate import specifiers
+    // test({
+    //   code: "import { Foo, Foo1 } from './foo';\nimport { Foo, Bar } from './foo';",
```
ljharb, line 424: indeed, this is invalid code because the `Foo` is duplicated. It wouldn't parse in the first place, so eslint can't fix it.

ljharb (changes requested):

fingerartur:
I have a feeling this MR may not get merged due to it's lack of support for inline comments inside import statements. For anybody who prefers no-duplicates to remove duplicate import specifiers and sort them alphabetically and does not need support for comments inside import statements, you may use my fork:

```
yarn add -D fingerartur/eslint-plugin-import
```

ljharb:
@fingerartur it doesn’t have to _support_ autofixing inline comments, but it definitely can’t incorrectly autofix them. An easy path to getting this _P_ R merged <sub>(there’s no such thing as an MR except on gitlab)</sub> would be to skip the autofixer whenever a comment is present - then it only autofixes in the safe cases.

Is that a change you’re willing to make?

maldito-dot (commented):
@maldito-dot

PR #2275, opened 2021-10-26
[Fix] `extensions`: resolve type-only imports with TypeScript resolver

jablko:
So as not to apply .js `import/extensions` rules to type-only imports, when a .js is resolvable.

### DIFF
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -1,6 +1,9 @@
 import path from 'path';
 
-import resolve from 'eslint-module-utils/resolve';
+import resolve, {
+  requireResolver,
+  resolverReducer,
+} from 'eslint-module-utils/resolve';
 import { isBuiltIn, isExternalModule, isScoped } from '../core/importType';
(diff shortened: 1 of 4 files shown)


ljharb:
hmm - none of these tests fail without your fix.

eturino:
@ljharb @jablko could we extend this to optionally cover all `.ts` files?
I am migrating a codebase to ESM and I want the linter to complain if we are importing something like
```ts
import blah from "./blah";
```
instead of 
```ts
import blah from "./blah.js";
```

The problem is that the file is `blah.ts`, but we need to import it as `blah.js`. I'd like to have a config to the extensions rule so that we could specify `.jsx` for `tsx` files and `.js` for `ts` files.

ljharb:
@eturino that's not something I'm likely to accept; an extension should only be present when the config mandates it *and* when it's present on the filesystem, and omitting the extension is the best practice. Either way, this PR isn't the place to discuss it.

PR #2283, opened 2021-10-30
[Fix] add missing optional peer dependencies

jdanil:
The plugin uses `@typescript-eslint/parser` as the default parser for typescript files in the exported typescript config. `eslint-import-resolver-node` is also used as the default resolver.

This PR adds the packages as optional peer dependencies so that they can be safely accessed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,11 @@
     "typescript": "^2.8.1 || ~3.9.5",
     "typescript-eslint-parser": "^15 || ^22.0.0"
   },
+  "peerDependenciesMeta": {
+    "@typescript-eslint/parser": {
+      "optional": true
+    }
+  },
   "peerDependencies": {
     "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8"
   },
--- a/utils/package.json
+++ b/utils/package.json
@@ -25,6 +25,14 @@
     "url": "https://github.com/import-js/eslint-plugin-import/issues"
   },
   "homepage": "https://github.com/import-js/eslint-plugin-import#readme",
+  "peerDependenciesMeta": {
+    "@typescript-eslint/parser": {
+      "optional": true
+    },
+    "eslint-import-resolver-node": {
+      "optional": true
+    }
+  },
   "dependencies": {
     "debug": "^3.2.7",
     "find-up": "^2.1.0",
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ rules:
 
 You may use the following shortcut or assemble your own config using the granular settings described below.
 
-Make sure you have installed [`@typescript-eslint/parser`] which is used in the following configuration. Unfortunately NPM does not allow to list optional peer dependencies.
+Make sure you have installed [`@typescript-eslint/parser`] which is used in the following configuration.
 
 ```yaml
 extends:
--- a/utils/CHANGELOG.md
+++ b/utils/CHANGELOG.md
@@ -4,6 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This change log adheres to standards from [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## Unreleased
+- add missing optional peer dependencies ([#2283], thanks [@jdanil])
 
 ## v2.7.1 - 2021-10-13
 


utils/package.json
```diff
@@ -25,6 +25,14 @@
     "url": "https://github.com/import-js/eslint-plugin-import/issues"
   },
   "homepage": "https://github.com/import-js/eslint-plugin-import#readme",
+  "peerDependenciesMeta": {
+    "@typescript-eslint/parser": {
+      "optional": true
+    },
+    "eslint-import-resolver-node": {
```
ljharb, line 32: This isn’t needed, because it’s already available as a dependency, so the node resolution algorithm means it’s already accessible. There are some broken package managers that require this unnecessary peer dep declaration to make it accessible, but I’m not going to add a breaking change just to cater to broken package managers.

ljharb (changes requested):
This is a breaking change for npm versions that lack optional peer dep support.

jdanil:
> This is a breaking change for npm versions that lack optional peer dep support.

Hi @ljharb. I'm not sure I understand why this would be a breaking npm <7. If the package manager doesn't support optional peer dependencies, then wouldn't the behaviour be the same as it is now?

ljharb:
Peer dependencies are *always* required, unless the package manager supports optional peer deps and it's also declared as such.

So, since adding a non-optional peer dependency is a breaking change, this would be a breaking change as well.

jdanil:
This change doesn't add the dependencies to `peerDependencies` though, so there shouldn't be any impact there. They're only added to `peerDependenciesMeta` as optional. So packages that do not support `peerDependenciesMeta` will have no impact.

utils/package.json
jdanil: `eslint-import-resolver-node` is a dependency of `eslint-plugin-import`, but not of `eslint-module-utils` where it is required by default.

ljharb:
ah, i see what you mean.

That confuses me tho. peerDependenciesMeta is only intended to describe things listed in peerDependencies. It's not supposed to have any effect to put something in there that's not mentioned in peer deps.

Is this something you're observing in npm, or in an alternative client?

jdanil:
If marking a dependency as optional in `peerDependenciesMeta` it implicitly considers that to be `"dependency": "*"` if no specific version ranges have been listed in `peerDependencies`.

So it seems like that would allow us to avoid any breaking changes as there is no need to modify `peerDependencies`.

I'm using yarn myself.

ljharb:
Presumably you're using PnP? normal yarn matches npm afaik, which doesn't require anything to be marked as a peer dep to be "accessible".

jdanil:
Yep I am running PnP in CI for now and node modules linker locally. The node modules linker in both yarn and npm seem to work in this case as they are accessing these transitive dependencies. But since they are being explicitly required in these packages it would seem to be correct for them to be declared as peer dependencies.

ljharb:
I believe that's because yarn itself has hardcoded these examples for this plugin.

It's not "more correct" because there's nothing correct about requiring a declaration. The correct thing is for transitive dependencies to be nondeterministically available to the application.

teohhanhui:
Accessing transitive dependencies is inherently unsafe, regardless of package manager. Things might just break when your dependencies change their dependencies...

ljharb:
In the case of the TS parser, you'd have it as an explicit top-level dep if it's used in your config (or a shared config would have it as a peer dep). In the case of the node resolver, it's a runtime dep of this plugin, so we can be certain we'll never remove it.

jdanil:
`eslint-plugin-import` isn't the only consumer of `eslint-module-utils` though. There are other plugins like https://github.com/azz/eslint-plugin-monorepo which use the utils to resolve paths. If you're saying that these packages too should add `eslint-import-resolver-node` as a dependency if they intend to use the default resolver then that seems like exactly what peer dependencies are designed to indicate.

Even if we're just considering the node modules linker, without having these peer dependencies explicitly declared, strict node modules layouts like what pnpm uses would not necessarily have access to these packages.

In your view, what is the purpose of optional peer dependencies is if not for this?

utils/package.json
ljharb: eslint-module-utils actually *dynamically* requires `eslint-import-resolver-X`, for every `X` defined in the user's eslint settings.

That the node resolver is the default doesn't mean someone has to use it, and that you can specify a thousand different resolvers (that eslint-module-utils will dutifully try to require) doesn't mean eslint-module-utils should be specifying it.

This kind of dynamic requiring is simply a normal part of the way the entire eslint ecosystem works, and any package manager that breaks it is *itself broken*.

ljharb:
For one, anything npm doesn't do isn't something I think needs to be catered to - npm is the standard. That said, npm will be adding a pnpm-like "isolated mode" soon, at which point I'd be obliged to address it somehow.

See https://github.com/import-js/eslint-plugin-import/pull/2283#discussion_r739965061 for why eslint-module-utils is correct as-is - it, like eslint itself, dynamically requires things based on config, and thus, it's always up to consumers to ensure that the things their config references are installed.

utils/package.json
jdanil: Yes, and if a consumer is using a non-default resolver, it is up to them to declare it as a dependency. `eslint-module-utils` is hard coded to use `eslint-import-resolver-node` by default, which is why it makes sense for it to be declared as an _optional_ peer dependency.

jdanil:
npm may have been grandfathered into node, but I wouldn't say it is the "standard". Node now provides yarn and pnpm out of the box via corepack.

I still don't understand what you think the purpose of optional peer dependencies are if you think that all packages should have access to all transitive dependencies.

You obviously have some very strong opinions about what package manager users of `eslint-plugin-import` should use. I'm not going to argue. I just thought I'd try to contribute a fix to resolve the issue of undeclared dependencies, which it sounds like you will eventually have to address once it becomes an issue with npm.

ljharb:
The sole purpose of optional peer deps is for something that, if present, must conform to a version range, but need not be present.

utils/package.json
bradennapier: Correct, this is the right behavior and this causes any projects using yarn v3 to fail without custom configuration.  Perhaps the yarn documentation will make it more clearn as-to why: https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

ljharb: @bradennapier that link is an opinion, not a fact, and contradicts the way parts of the npm ecosystem (including eslint itself) have actually worked for over a decade.

wmertens:
> The sole purpose of optional peer deps is for something that, if present, must conform to a version range, but need not be present.

I would argue that the purpose of optional peer deps is to notify the package manager that you would like access to these peer deps if they are in the project. If there is official documentation stating otherwise I'd love to find out.

I'm writing a package manager of sorts for Nix, where it would be most efficient if each package would declare their (optional) (peer) dependencies correctly, so that those can be symlinked together instead of copied.

I can confirm that pnpm treats `peerDependenciesMeta` packages that are not mentioned elsewhere as peerDependencies matching any version. Npm and Yarn copy things together so they leak access that way.

wmertens:
Incidentally, utils also requires eslint to find espree, which means eslint or espree are peer dependencies. Since every project that uses this plugin also uses eslint, I'd argue that eslint should be a peer dependency.

utils/package.json
wmertens: To be fair, almost all of the ecosystem follows that opinion, and only some plugins of eslint and babel cause trouble...

ljharb:
That’s not why it means - utils looks for espree/eslint inside a try/catch, which isn’t a peer relationship. Peer deps aren’t for defining “allowed dependencies to access”, optional or otherwise.

utils/package.json
ljharb: It's a problem for anything that has static configuration, which may include typescript as well as a number of other ecosystems.

Either way, there are valid use cases for not making that convention a constraint.

LuckyMona: But if `eslint-module-utils` don't declare `eslint-import-resolver-node` as its peer dep, it will cause bugs which you can only fix it by add `eslint-import-resolver-node` to your monorepo project's root path. But that's a very bad practice.

wmertens: @LuckyMona I added the node resolver to my project's devDeps and it seems to just work, no special tricks needed apart from the optional peer dep on eslint.
What error are you seeing?

LuckyMona: @wmertens hi master,

> "I added the node resolver to my project's devDeps"

Yes, Just only this method can avoid the bug, but which I think is a very bad practice for these reasons:

1. Although this method can work, it's still a phantom dependency case, which we should try to avoid
2. In this case, it can't work: 
    Suppose you developed a npm package, and the package required `eslint-plugin-import` or `eslint-module-utils`, then if  the ancestor consumer of your package uses pnpm, you have to require the ancestor consumer  install `eslint-import-resolver-node` in its devDeps, which almost can't be done. Otherwise the project will report error.

wmertens: Hmm, IMHO the resolver is a plug-in, to be picked by the project. In the situation you describe I would expect that helper package to import the plugin somehow and take care of it, but as a direct consumer I'm responsible for the plugins.

In any case, unless these resolver plugins are the ever-only possible plugins, they should be treated like all plugins. Otherwise you're just moving the problem.

LuckyMona: Do you mean I should add node resolver to my npm package's devDeps ? But It won't work if you use pnpm, because pnpm's install mechanism forbid phantom dependency, so `eslint-module-utils` can't get node resolver from my npm package and will report error

JounQin: You'd better provide a reproduction instead. AFAIK, pnpm hoists all eslint related dependencies by default.

LuckyMona: You can try to reproduce it as follows:

One project has A in its Deps, A has B and C in its Deps, C has D in its Deps,  C also require B in its code but don't declare it in its package.json. Then C won't get B from its father A and will report error. (Yarn 3 as the project's package manager and pnpm linker as yarn's nodeLinker)

Given more spare time, I can give a demo.

ljharb: @LuckyMona fwiw this exact kind of conflict is why pnpm/PnP's approach is fundamentally flawed - it's simply not *possible* for this package to explicitly list every plugin package that a user might want to instruct it to load. As such, the only "solution" is either a) hardcoding special cases into their own resolvers, or b) allowing anything you explicitly depend on to be required by anything else you explicitly depend on.

wmertens: Or c) allow specifying plugins by full path.

ljharb: That’s already allowed, but that breaks more than a decade of expectations in the node ecosystem - paths should never be absolute, and node_modules shouldn’t ever have to be hardcoded.

wmertens: Right - that's what require.resolve is for :)

I think we need an example that breaks. I'm in Camp Pure Declarations but plugins cannot be known beforehand so cannot be declared, and AFAIK things just work right now @LuckyMona

wmertens: What is the error you're getting? I don't understand how these settings might be problematic

JounQin: A reproduction is really required here. 😅

ljharb: @LuckyMona eslint config must be able to be static - not JS, but JSON/ini. As such, require.resolve is not available in every case to the config.

package.json
```diff
@@ -91,48 +78,43 @@
...
+  },
+  "peerDependenciesMeta": {
+    "@typescript-eslint/parser": {
+      "optional": true
+    }
   },
   "peerDependencies": {
-    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8 || ^9"
+    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8"
   },
```
nikolawork, line 104: `eslint-plugin-import` also requires `eslint-import-resolver-node`. This becomes apparent when using npm's `install-strategy=linked` ([link](https://docs.npmjs.com/cli/v11/using-npm/config#install-strategy)). I have made a PR for this in #3266, but would adding `eslint-import-resolver-node` to `peerDependenciesMeta` in this PR be enough to fix the issue?

```suggestion
  "peerDependenciesMeta": {
    "@typescript-eslint/parser": {
      "optional": true
    },
    "eslint-import-resolver-node": {
      "optional": true
    }
  },
  "peerDependencies": {
    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8 || ^9 || ^10"
  },
```

nikolawork: Or do we also need to add `eslint-import-resolver-node` to `peerDependencies`?

```suggestion
  "peerDependenciesMeta": {
    "@typescript-eslint/parser": {
      "optional": true
    },
    "eslint-import-resolver-node": {
      "optional": true
    }
  },
  "peerDependencies": {
    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8 || ^9 || ^10",
    "eslint-import-resolver-node": "^0.4.0"
  },
```

PR #2299, opened 2021-11-10
[Fix] `no-default-import`: limit highlight to the word "default"

pmcelhaney:
Instead of reporting the entire export default declaration:

image

Only reporting the "default" token: 

image

### DIFF
--- a/src/rules/no-default-export.js
+++ b/src/rules/no-default-export.js
@@ -20,15 +20,17 @@ module.exports = {
 
     return {
       ExportDefaultDeclaration(node) {
-        context.report({ node, message: preferNamed });
+        const { loc } = context.getSourceCode().getFirstTokens(node)[1] || {};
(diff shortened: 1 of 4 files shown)


src/rules/no-default-export.js
```diff
@@ -20,19 +20,20 @@ module.exports = {
       `Do not alias \`${local.name}\` as \`default\`. Just export ` +
       `\`${local.name}\` itself instead.`;
 
+
     return {
       ExportDefaultDeclaration(node) {
-        context.report({ node, message: preferNamed });
+        context.report({ node, message: preferNamed, loc: context.getSourceCode().getFirstTokens(node, 2)[1].loc });
```
pmcelhaney, line 26: The first tokens are "export" and "default". We want to highlight the loc[ation] of the "default" token.

ljharb (approved):
Seems reasonable, thanks

Issue #1931, opened 2020-10-25
[no-unused-modules] The "path" argument must be of type string. Received undefined


exodus4d:
After ESLint upgrade `7.11.0`  ➜  `7.12.0` I noticed an error:

- ESLint Version: **7.12.0** / parser: `@typescript-eslint/parser`
- Node Version: **15.0.1**
- npm Version: **7.0.3**
- OS: **Win10**

```cmd
Oops! Something went wrong! :(

ESLint: 7.12.0

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at validateString (internal/validators.js:124:11)
    at Object.extname (path.js:752:5)
    at loadConfigFile (G:\webpages\pathfinder\node_modules\@eslint\eslintrc\lib\config-array-factory.js:305:18)
    at ConfigArrayFactory._loadConfigData (G:\webpages\pathfinder\node_modules\@eslint\eslintrc\lib\config-array-factory.js:613:42)
    at ConfigArrayFactory._loadExtendedBuiltInConfig (G:\webpages\pathfinder\node_modules\@eslint\eslintrc\lib\config-array-factory.js:803:25)
    at ConfigArrayFactory._loadExtends (G:\webpages\pathfinder\node_modules\@eslint\eslintrc\lib\config-array-factory.js:780:29)
    at ConfigArrayFactory._normalizeObjectConfigDataBody (G:\webpages\pathfinder\node_modules\@eslint\eslintrc\lib\config-array-factory.js:724:25)
    at _normalizeObjectConfigDataBody.next (<anonymous>)
    at ConfigArrayFactory._normalizeObjectConfigData (G:\webpages\pathfinder\node_modules\@eslint\eslintrc\lib\config-array-factory.js:669:20)
    at _normalizeObjectConfigData.next (<anonymous>)
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! pathfinder-eve@2.0.0 eslint: `eslint ".eslintrc.json" ".\js\app\conf\appConfig.mjs"`
npm ERR! Exit status 2
```
The origin issue (https://github.com/eslint/eslint/issues/13789) was opened inESLint project, because it was unclear where it came from.  
I could narrow the Error down to [import/no-unused-modules](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unused-modules.md):
```
 "import/no-unused-modules": [
[...]

PR #2305, opened 2021-11-17
[Fix] `importType`: avoid crashing on a non-string

lencioni:
While writing imports in VSCode recently, I started to get errors throwm
from eslint that appear to originate from eslint-plugin-import.

Here's the strack trace from the output panel:

```
[Error - 3:52:43 PM] TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at validateString (internal/validators.js:124:11)
    at isAbsolute (path.js:1029:5)
    at isAbsolute (/path/to/repo/node_modules/eslint-plugin-import/src/core/importType.js:17:10)
    at typeTest (/path/to/repo/node_modules/eslint-plugin-import/src/core/importType.js:92:7)
    at resolveImportType (/path/to/repo/node_modules/eslint-plugin-import/src/core/importType.js:105:10)
    at reportIfMissing (/path/to/repo/node_modules/eslint-plugin-import/src/rules/no-extraneous-dependencies.js:170:7)
    at commonjs (/path/to/repo/node_modules/eslint-plugin-import/src/rules/no-extraneous-dependencies.js:267:7)
    at checkSourceValue (/path/to/repo/node_modules/eslint-module-utils/moduleVisitor.js:29:5)
    at checkSource (/path/to/repo/node_modules/eslint-module-utils/moduleVisitor.js:34:5)
    at /path/to/repo/node_modules/eslint/lib/linter/safe-emitter.js:45:58
```

I am able to trigger this consistently when writing an import, before I
start the open quote, like this:

```
import foo from
```

I found https://github.com/import-js/eslint-plugin-import/issues/1931
which pointed at a problem in the resolvers causing this problem. We do
use some custom resolvers with this plugin, so I tried disabling that
but the problem persisted.

Thankfully, I was able to reproduce this error in the test suite.

It would be good if this plugin handled this scenario gracefully instead
of throwing an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,23 +6,26 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## [Unreleased]
 
+### Fixed
+- `importType`: avoid crashing on a non-string' ([#2305], thanks [@ljharb])
+
 ## [2.25.3] - 2021-11-09
 
 ### Fixed
-- [`extensions`]: ignore unresolveable type-only imports ([#2270], [#2271], [@jablko])
-- `importType`: fix `isExternalModule` calculation ([#2282], [@mx-bernhard])
-- [`no-import-module-exports`]: avoid false positives with a shadowed `module` or `exports` ([#2297], [@ljharb])
+- [`extensions`]: ignore unresolveable type-only imports ([#2270], [#2271], thanks [@jablko])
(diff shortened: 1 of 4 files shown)


ljharb:
I don't think it's possible to have a test here, since this is invalid syntax, so eslint can't parse it. That said, I can repurpose this into a bug fix given that stack trace, thanks!

ljharb (approved):

PR #2312, opened 2021-11-25
[Fix] Minor typo in prefer-default-export.md

kkyaruek:


ljharb:
I'm not sure this is a typo; it's correct (and the same) both before and after your change.

kkyaruek:
Yes, but I thought it might be unintended because eslint doesn't recommend anonymouse default export and `const bar` is never used.

https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-anonymous-default-export.md

I'm sorry I didn't write the explanation above. 
Thank you for your time.

ljharb (approved):

Issue #2314, opened 2021-11-30
False positive of `no-import-module-exports` on `module` property imported from package


BarryThePenguin:
Similar to #2181, I'm seeing a false positive when using `qunit` with ember.

`qunit` exports a `module` property that allows tests to be grouped under a module. It also has `module.only()`, `module.skip()`, and `module.todo()` methods that apply behaviour to all tests in that module.

```js
import { module, test } from 'qunit'

module.skip('A module to test', function () {
  test('A test', function () {
    ...
  })
})
```

The `no-import-modules-exports` is reporting these methods as errors.

ljharb:
Fixed by #2315.

PR #2315, opened 2021-11-30
[Fix] `no-import-module-exports`: avoid a false positive for import variables

BarryThePenguin:
This lets the `no-import-module-exports` rule determine if a variable was defined by an import statement. For #2314


ljharb (approved):
Very sorry for the unacceptably long review delay here. This looks awesome!

PR #2323, opened 2021-12-09
fix: detect case-sensitive filesystems correctly under yarn PnP

charlessuh:
yarn in PnP mode uses a virtual filesystem, and adding the `preferUnplugged: true` flag forces it to write this package to the real filesystem. This is needed so the case-sensitive detection logic in `resolve.js` works properly.


ljharb:
Can you link to the documentation here?

This sounds like a bug in yarn PnP, to be honest; why should i add bytes to the package.json for all users, when yarn could fix it for yarn pnp users?

charlessuh:
No, this is intentional in yarn: https://github.com/yarnpkg/berry/issues/1762

Juat a little background: The virtual filesystem in yarn is an optimization, so when installing packages you are not making a gazillion real filesystems calls like `npm` does to create the `node_modules` folder.

This field in `package.json` simply tells yarn to place this package somewhere on the real filesystem.

Also, IMO a few extra bytes in the `package.json` is pretty trivial. It's also arguably not unreasonable to decentralize this kind of data across potentially millions of packages, rather than centralize this kind of data in one place.

ljharb:
On the contrary; the minority of yarn users are who should pay the cost, not everyone else.

charlessuh:
Just out of curiosity (I’m not intending to defend yarn or fight any ideological battle here), but do you aim for this package to specifically only support `npm`, or does it aim to be as support as many users as possible given there are three package managers in the JS ecosystem (npm, pnpm, yarn)?

If this package only supports `npm`, then sure, even adding a single byte in this package to support any other package manager would be inconveniencing the only users of this package, who should be using `npm`.

ljharb:
There’s one package manager that is the standard and has the majority, and whose behavior is correct by definition; I’m happy to support the others when it doesn’t hurt the primary audience.

If yarn’s virtual filesystem does not exactly mimic the native one, then it has a bug, and it should be fixed.

charlessuh:
> There’s one package manager that is the standard and has the majority, and whose behavior is correct by definition

> If yarn’s virtual filesystem does not exactly mimic the native one, then it has a bug, and it should be fixed.

While I can appreciate that standpoint, I’m not sure how widespread this opinion that by /definition/ every single `npm` implementation detail is automatically standard and correct, and there is no room for package managers to differ.

If you take a look at the PnP page (https://yarnpkg.com/features/pnp#compatibility-table), you can see a large variety of packages like `webpack`, `eslint`, `babel`, `prettier`, `gulp` all have added explicit compatibility with yarn PnP.

> I’m happy to support the others when it doesn’t hurt the primary audience.

Firstly, I’m open to alternative suggestions here. For example, instead of a metadata field in `package.json`, could this package detect PnP via `process.versions.pnp` and set the value of `CASE_SENSITIVE_FS` to true in that case?

Can you define what you mean by hurt? For example, would adding even a single byte to the download that would be unnecessary for a `npm` user be harmful?

ljharb:
Certainly the mechanism in this PR is likely the least intrusive one available (altho yarn-specific metadata should really be inside a yarn object, instead of littering the top level with stuff)

Some of it is philosophical - just because some major packages have been effectively strong-armed into adding this metadata, doesn't mean that endorsing this dark pattern is something I want to do as a package author.

charlessuh:
Okay. I'm trying to think of other ways to try to solve this, such as passing an explicit configuration option to tell this package whether the filesystem is case-sensitive or not (instead of having it detect it automatically), or perhaps using `pkg-dir` to detect the case-sensitivity of the project's package.json (which of course is on the real filesystem):

```diff
-const CASE_SENSITIVE_FS = !fs.existsSync(path.join(__dirname.toUpperCase(), 'reSOLVE.js'));
+const CASE_SENSITIVE_FS = !fs.existsSync(path.join(pkgDir.sync(__dirname).toUpperCase(), 'pacKAGE.json'));
```

ljharb:
If package.json is always on the local filesystem, that seems like totally reasonable thing to check instead, if that bypasses yarn's bug.

utils/resolve.js
```diff
@@ -10,7 +10,8 @@ const path = require('path');
 const hashObject = require('./hash').hashObject;
 const ModuleCache = require('./ModuleCache').default;
 
-const CASE_SENSITIVE_FS = !fs.existsSync(path.join(__dirname.toUpperCase(), 'reSOLVE.js'));
+const projectDir = pkgDir.sync(__dirname);
```
ljharb, line 13: is this basically just avoiding hardcoding `./package.json`?

charlessuh: Oh, whoops -- I thought `pkg-dir` skipped the current directory and started looking for a `package.json` in the parent directory.

So the logic should be trying to read `/Users/sindresorhus/foo/package.json`, not `/Users/sindresorhus/foo/node_modules/eslint-module-utils/package.json`.

```
/
└── Users
    └── sindresorhus
        └── foo
            ├── package.json
            └── node_modules
                ├── eslint-module-utils
                └── package.json
                └── resolve.js
```

ljharb: In that case, it'd be `path.join(process.cwd(), 'package.json')` - but also, that wouldn't work for a project that doesn't have a package.json (altho that's a rare use case)

merceyz:
> if that bypasses yarn's bug.

It's possible to mount the `node_modules` folder to a different drive using a different file system when using npm as well, shouldn't the check be testing on the file getting linted instead?

ljharb:
Absolutely, that's a really good point. Want to make that change?

charlessuh:
> shouldn't the check be testing on the file getting linted instead?

Can you elaborate more on what you mean? 

There are special-cases like`package.json` or files inside a published npm package where you could assume a case, but for a user-generated file, it could be named anything -- like `MyRAnDoMFiLE.js` -- and it's not clear to me how you could tell a file access succeeding on a case-sensitive filesystem (because the file is /really/ named like that) or on a case-insensitive filesystem (because the filesystem doesn't care.)

I'm sure there may be operating system APIs that could do this, but does nodejs expose any of them?

ljharb:
I don't think node has any API to expose that, unfortunately.

One technique would be to find a filename, and a variant, that don't exist - write one, and check to see if the variant suddenly exists - but this would just be another heuristic, not a guarantee.

We could use `fs.readdir(path.dirname(currentFilename))` - and find a filename that exists, but who has a capitalization variant that does not exist - and then see if the variant exists directly?

ljharb:
@charlessuh any interest in updating the check?

charlessuh:
It looks like there is already logic in `fileExistsWithCaseSync` to query an expected filename, and see if the filename is the same case when you list the directory via `readdir`: https://github.com/import-js/eslint-plugin-import/blob/main/utils/resolve.js#L53-L56.

However, note this function returns early when `CASE_SENSITIVE_FS` is true. I've updated this PR to stop checking that variable -- so case sensitivity is now queried on a per-file basis.

ljharb (approved):
Thanks! I think this has evolved into a broadly useful fix - there's no reason any system's case sensitivity is constant for every file, since one file could be on a case sensitive volume and another on a case insensitive volume.

ljharb:
@charlessuh hm, removing the `utils` change doesn't cause any of the tests to fail. Is there any way we could add a test for that?

PR #2325, opened 2021-12-11
Exclude `no-unresolved` in TypeScript config

fregante:
I think this rule isn't useful on TS codebase


ljharb:
Please open an issue discussing things like this first.

It is useful on a TS codebase, because i don’t believe TS checks CJS deps nor untyped deps.

Issue #2306, opened 2021-11-18
webpackExports: ["namedExport"] is not supported for dynamic imports


emanueleperuffo:
Webpack 5 has some new magic comments allowing better tree-shaking.
https://webpack.js.org/api/module-methods/#magic-comments

Currently, if a comment in the form `/* webpackExports: ["namedExport"] */` is added to a dynamic import, eslint shows an error when checking the with the `import/dynamic-import-chunkname` rule.

ljharb:
Sounds like we need to add support for this new feature.

Issue #1904, opened 2020-09-18
[dynamic-import-chunkname] webpackChunkName should be optional when webpackMode: "eager"


dtinth:
## Code

```js
import(
  /* webpackMode: "eager" */ '../a-module-that-requires-some-time-to-initialize'
)
```

## Expected result

(No warning)

## Actual result

> dynamic imports require a leading comment in the form /* webpackChunkName: "[a-zA-Z0-9-/_]+",? */ `eslint(import/dynamic-import-chunkname)`

## Description

From the docs:

> `'eager'`: **Generates no extra chunk.** All modules are included in the current chunk and no additional network requests are made. A Promise is still returned but is already resolved. In contrast to a static import, the module isn't executed until the call to import() is made.

`/* webpackMode: "eager" */` is handy when you want to include a module in the same chunk, but do not want to execute it when the chunk is initially loaded.

## Workaround

```js
// eslint-disable-next-line import/dynamic-import-chunkname
import(
  /* webpackMode: "eager" */ '../a-module-that-requires-some-time-to-initialize'
)
```

anomiex:
Also `webpackIgnore: true`, which also opts out of module splitting.

Issue #1840, opened 2020-06-29
`dynamic-import-chunkname` does not allow additional comments


MhMadHamster:
eslint: `v7.3.1`
eslint-plugin-import: `v2.22.0`

Eslint rule:
```
"rules": {
	"import/dynamic-import-chunkname": "error",
}
```

Code:
```javascript
import(
	/* webpackInclude: /\.json$/ */
	/* webpackChunkName: "module" */
	"./module.js"
);
```

Produces error:
```
  1:1  error  dynamic imports require a leading comment in the form /* webpackChunkName: "[0-9a-zA-Z-_/.]+",? */  import/dynamic-import-chunkname
```

Version 2.21.2 works well

ljharb:
Looks like a duplicate of #1407; i'm not sure tho why it would have worked in v2.21.

PR #2330, opened 2021-12-18
[Fix] `dynamic-import-chunkname`: add handling webpack magic comments

MhMadHamster:
This PR follows the conversation from https://github.com/import-js/eslint-plugin-import/pull/1411 and aims to add support for all of the magic comments that currently (at the point of opening this pr) supported by webpack itself. [Webpack documentation](https://webpack.js.org/api/module-methods/#magic-comments) and [ImportParserPlugin](https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/dependencies/ImportParserPlugin.js) are used as reference for rules.

This PR **not yet ready** for merge and more of a PoC (haven't updated docs/changelog yet). The reason why i consider this more of a PoC is because by introducing changes in this PR, IMO rule becomes more than a "check for a correct chunkname in a webpack dynamic import" because it now validates other magic comments as well. Following the conversation from https://github.com/import-js/eslint-plugin-import/pull/1411 i would suggest either completely removing anything unrelated to `webpackChunkname` (i.e. rule just checks for presence of `webpackChunkname` and that it is a valid string, and doesn't care for anything else in a comment) **OR** renaming the rule to (for example) `webpack-dynamic-import` keeping the changes from this PR and adding a separate flag, that you can enable, for `webpackChunkname` to be required in your dynamic imports. I am, open for other suggestions.

I dont see a point in adding "partial" support of magic comments suggested in https://github.com/import-js/eslint-plugin-import/pull/1411 You will be able to use invalid values, but rule will try to somewhat validate your comments anyway, why even validate anything aside from `webpackChunkname` then?

Related:
https://github.com/import-js/eslint-plugin-import/issues/1407
https://github.com/import-js/eslint-plugin-import/issues/1840
https://github.com/import-js/eslint-plugin-import/issues/1904
https://github.com/import-js/eslint-plugin-import/issues/2306
https://github.com/import-js/eslint-plugin-import/pull/1411

### DIFF
--- a/src/rules/dynamic-import-chunkname.js
+++ b/src/rules/dynamic-import-chunkname.js
@@ -27,10 +27,10 @@ module.exports = {
   create(context) {
     const config = context.options[0];
     const { importFunctions = [] } = config || {};
-    const { webpackChunknameFormat = '[0-9a-zA-Z-_/.]+' } = config || {};
+    const { webpackChunknameFormat = '([0-9a-zA-Z-_/.]|\\[(request|index)\\])+' } = config || {};
 
     const paddedCommentRegex = /^ (\S[\s\S]+\S) $/;
-    const commentStyleRegex = /^( \w+: (["'][^"']*["']|\d+|false|true),?)+ $/;
+    const commentStyleRegex = /^( ((webpackChunkName: .+)|((webpackPrefetch|webpackPreload): (true|false|-?[0-9]+))|(webpackIgnore: (true|false))|((webpackInclude|webpackExclude): \/.*\/)|(webpackMode: ["'](lazy|lazy-once|eager|weak)["'])|(webpackExports: (['"]\w+['"]|\[(['"]\w+['"], *)+(['"]\w+['"]*)\]))),?)+ $/;
     const chunkSubstrFormat = ` webpackChunkName: ["']${webpackChunknameFormat}["'],? `;
     const chunkSubstrRegex = new RegExp(chunkSubstrFormat);
 
@@ -83,7 +83,7 @@ module.exports = {
           context.report({
             node,
             message:
-              `dynamic imports require a leading comment in the form /*${chunkSubstrFormat}*/`,
+              `dynamic imports require a "webpack" comment with valid syntax`,
           });
           return;
         }
--- a/tests/src/rules/dynamic-import-chunkname.js
+++ b/tests/src/rules/dynamic-import-chunkname.js
@@ -5,7 +5,7 @@ import semver from 'semver';
 const rule = require('rules/dynamic-import-chunkname');
 const ruleTester = new RuleTester();
 
-const commentFormat = '[0-9a-zA-Z-_/.]+';
+const commentFormat = '([0-9a-zA-Z-_/.]|\\[(request|index)\\])+';
 const pickyCommentFormat = '[a-zA-Z-_/.]+';
 const options = [{ importFunctions: ['dynamicImport'] }];
 const pickyCommentOptions = [{
@@ -21,8 +21,9 @@ const noLeadingCommentError = 'dynamic imports require a leading comment with th
 const nonBlockCommentError = 'dynamic imports require a /* foo */ style comment, not a // foo comment';
 const noPaddingCommentError = 'dynamic imports require a block comment padded with spaces - /* foo */';
 const invalidSyntaxCommentError = 'dynamic imports require a "webpack" comment with valid syntax';
-const commentFormatError = `dynamic imports require a leading comment in the form /* webpackChunkName: ["']${commentFormat}["'],? */`;
-const pickyCommentFormatError = `dynamic imports require a leading comment in the form /* webpackChunkName: ["']${pickyCommentFormat}["'],? */`;
+const commentFormatError = `dynamic imports require a "webpack" comment with valid syntax`;
+const chunkNameFormatError = `dynamic imports require a leading comment in the form /* webpackChunkName: ["']${commentFormat}["'],? */`;
+const pickyChunkNameFormatError = `dynamic imports require a leading comment in the form /* webpackChunkName: ["']${pickyCommentFormat}["'],? */`;
 
 ruleTester.run('dynamic-import-chunkname', rule, {
   valid: [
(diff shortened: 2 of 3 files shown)


src/rules/dynamic-import-chunkname.js
```diff
@@ -27,10 +35,19 @@ module.exports = {
...
+      ['webpackPreload', '(true|false|-?[0-9]+)'],
+      ['webpackIgnore', '(true|false)'],
+      ['webpackInclude', '\\/.*\\/'],
+      ['webpackExclude', '\\/.*\\/'],
+      ['webpackMode', `["'](lazy|lazy-once|eager|weak)["']`],
+      ['webpackExports', `(['"]\\w+['"]|\\[(['"]\\w+['"], *)+(['"]\\w+['"]*)\\])`],
+    ];
     const paddedCommentRegex = /^ (\S[\s\S]+\S) $/;
-    const commentStyleRegex = /^( \w+: (["'][^"']*["']|\d+|false|true),?)+ $/;
+    const commentStyleRegex = generateRegexForMagicComments(webpackMagicComments);
```
ljharb, line 50: why are we doing regex magic here instead of just hardcoding each regex?

tests/src/rules/dynamic-import-chunkname.js
```diff
@@ -53,10 +53,34 @@ ruleTester.run('dynamic-import-chunkname', rule, {
         'someModule'
       )`,
       options: pickyCommentOptions,
-      errors: [{
-        message: pickyCommentFormatError,
-        type: 'CallExpression',
-      }],
```
ljharb, line 59: why are these tests removed? (removing tests is a red flag that it might be a breaking change)

ljharb:
(I marked the PR as a draft, since that's how one indicates a PR isn't yet ready for review)

tests/src/rules/dynamic-import-chunkname.js
MhMadHamster: > why are these tests removed? (removing tests is a red flag that it might be a breaking change)

I removed unused `errors` property, test itself is still in place. There are multiple tests which are expected to be valid, but they contain `errors` field, which is AFAIK not used, because test passes. Probably someone was just copy/pasting tests and forgot to remove it.

src/rules/dynamic-import-chunkname.js
MhMadHamster: > why are we doing regex magic here instead of just hardcoding each regex?

I though this way it's more clear what's happening and more readable overall. I feel like concatenating everything by hand in one huge regex is harder to grasp and more error prone (cause you need to do it by hand).

```
const commentStyleRegex = new RegExp(`^( ((webpackChunkName: ["']${webpackChunknameFormat}["'])|(webpackPrefetch: (true|false|-?[0-9]+))|(webpackPreload: (true|false|-?[0-9]+))|(webpackIgnore: (true|false))|(webpackInclude: \\/.*\\/)|(webpackExclude: \\/.*\\/)|(webpackMode: ["'](lazy|lazy-once|eager|weak)["'])|(webpackExports: (['"]\\w+['"]|\\[(['"]\\w+['"], *)+(['"]\\w+['"]*)\\]))),?)+ $`);
```

tests/src/rules/dynamic-import-chunkname.js
ljharb: Ahhh thanks :-) good catch!

src/rules/dynamic-import-chunkname.js
ljharb: While that’s true, dynamic regex construction is a huge code smell; i think it’s worth the clarity hit to avoid it.

MhMadHamster:
> (I marked the PR as a draft, since that's how one indicates a PR isn't yet ready for review)

Sure, however, it would be nice if someone from code owners/maintainers can make a decision in regards to rule's future and what I described in the main comment/objective, so i can finish off this PR.

ljharb:
@MhMadHamster I think it's totally reasonable for this rule - whose purpose is to validate a magic webpack comment's chunk name - to also report on a syntactically invalid webpack magic comment (or alternatively, not to crash on them). Does that answer your question?

MhMadHamster:
> @MhMadHamster I think it's totally reasonable for this rule - whose purpose is to validate a magic webpack comment's chunk name - to also report on a syntactically invalid webpack magic comment (or alternatively, not to crash on them). Does that answer your question?

Yes, thanks. I would suggest to change [this](https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/dynamic-import-chunkname.js#L86) error message to [this](https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/dynamic-import-chunkname.js#L77) because it now will fail not only on wrong chunkName comment but on anything that is not valid magic comment.

ljharb:
That seems reasonable to me for when it's an invalid magic comment; when it's valid but doesn't have a correct chunkName, then i'd expect the current error.

src/rules/dynamic-import-chunkname.js
```diff
@@ -27,10 +27,10 @@ module.exports = {
   create(context) {
     const config = context.options[0];
     const { importFunctions = [] } = config || {};
-    const { webpackChunknameFormat = '[0-9a-zA-Z-_/.]+' } = config || {};
+    const { webpackChunknameFormat = '([0-9a-zA-Z-_/.]|\\[(request|index)\\])+' } = config || {};
 
     const paddedCommentRegex = /^ (\S[\s\S]+\S) $/;
-    const commentStyleRegex = /^( \w+: (["'][^"']*["']|\d+|false|true),?)+ $/;
+    const commentStyleRegex = /^( ((webpackChunkName: .+)|((webpackPrefetch|webpackPreload): (true|false|-?[0-9]+))|(webpackIgnore: (true|false))|((webpackInclude|webpackExclude): \/.*\/)|(webpackMode: ["'](lazy|lazy-once|eager|weak)["'])|(webpackExports: (['"]\w+['"]|\[(['"]\w+['"], *)+(['"]\w+['"]*)\]))),?)+ $/;
```
MhMadHamster, line 33: Do not validate `webpackChunkName` initially, so we can throw more specific error later, if the whole comment is valid.

MhMadHamster:
@ljharb hey, can you please take a look, when you will have time

ljharb (approved):

ljharb:
oops, let me fix my rebase

PR #2332, opened 2021-12-18
[Docs] `no-unresolved`: Fix RegExp escaping in readme 

stephtr:
In the `ignore` example the backslash in `'\.img$'` was escaping the "." as a string character, not in the RegExp sense. `'\\.img$'` would be the correct string.


ljharb (approved):
Good catch!

PR #2334, opened 2021-12-20
Recognize properly resolving `@/*`-aliased imports as internal

### LINKED ISSUE #2333: Aliasing with `@` is not recognized as 'internal'
This seems to be a repeat of an issue that has been resolved and regressed twice now, see https://github.com/import-js/eslint-plugin-import/issues/2145 and https://github.com/import-js/eslint-plugin-import/issues/2249.

In my case the source code is private, but we're using a config that looks like:

```jsonc
// ...
"import/resolver": {
  "alias": [
    ["@", "./src"]
  ]
}
// ...
```

(See also https://github.com/johvin/eslint-import-resolver-alias.) And with source code like:

```jsx
// ...
import exampleInternalViaAlias from '@/exampleInternalViaAlias';

import exampleSibling from './exampleSibling';
// ...
```

When upgrading to v2.25.3 from v2.22.1, without any other changes to source code or other dependencies, we have now started getting false negatives like:

```
  3:1  error  `@/exampleInternalViaAlias` import should occur after import of `./exampleSibling`  import/order
```

Injecting some logs into the installed version of `eslint-plugin-import` in our node_modules, I can see that the import type is getting calculated as `'unknown'` (instead of `'internal'`) for `'@/exampleInternalViaAlias'`.

ombene:
Aims to close #2333.

The issue this PR addresses seems to have cropped up twice before. I've tried to include a new test case that will hopefully help prevent regressions. Also, I've tried to make the various "type" cases more independent, out of the belief that this will be more robust and future-proof. But I'd love any feedback! This is my first contribution, so I'm a little wary that I might be changing too many things at once, without the proper context for the decisions that led the code to be in its present state.

Looking at the code, it seemed like a good idea to have the check for "is this internal" stop being dependent on either "is this scoped" or "is this a module", and also to separate the "is this internal" logic from the "is this external" logic.

In doing so, it seemed prudent to put the internal path check after the external path check, so that `node_modules` (or any other external path that happens to be in the root project directory) would correctly calculate as `'external'`. In doing that, it became apparent that the [`import/internal-regex`](https://github.com/import-js/eslint-plugin-import/blob/main/README.md#importinternal-regex) configuration should take precedence, so I made that its own function and I actually decided to put it before everything—presumably the regex should take precedence over even types that would otherwise resolve as "absolute" or "built-in", say.

One final complication is `'external'` can also include import names that look like (potentially scoped) modules, even if the names don't resolve to any particular path, but only as long as the names don't resolve to an internal path. So I decided to split that "external-looking name check" logic out into its own function, which could be called separately after the the internal check. This also meant augmenting other `isExternalPath` consumers to call this new function as well.

Some other context is that I'm curious whether / why this logic around "external-looking names" is necessary. The architecture as a whole seems a lot more straightforward if `eslint-plugin-import` relies on its resolvers to provide paths, and treats any path misses as `'unknown'`—but perhaps I'm missing some important use case, like maybe web-URL-based imports? So by putting that logic into its own function, I hoped to: (a) enrich the discussion around whether / why we need this (by giving it a concrete name / location in the code), (b) make it more straightforward to delete if indeed that seems wise, or possibly better rename it to whatever use cases it serves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Fixed
 - `importType`: avoid crashing on a non-string' ([#2305], thanks [@ljharb])
 - [`first`]: prevent crash when parsing angular templates ([#2210], thanks [@ljharb])
+- `importType`: properly resolve `@/*`-aliased imports as internal ([#2334], thanks [@ombene])
 
 ### Changed
 - [`no-default-import`]: report on the token "default" instead of the entire node ([#2299], thanks [@pmcelhaney])
@@ -950,6 +951,7 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#2334]: https://github.com/import-js/eslint-plugin-import/pull/2334
 [#2305]: https://github.com/import-js/eslint-plugin-import/pull/2305
 [#2299]: https://github.com/import-js/eslint-plugin-import/pull/2299
 [#2297]: https://github.com/import-js/eslint-plugin-import/pull/2297
@@ -1571,6 +1573,7 @@ for info on changes for earlier releases.
 [@noelebrun]: https://github.com/noelebrun
 [@ntdb]: https://github.com/ntdb
 [@nwalters512]: https://github.com/nwalters512
+[@ombene]: https://github.com/ombene
 [@ota-meshi]: https://github.com/ota-meshi
 [@panrafal]: https://github.com/panrafal
 [@paztis]: https://github.com/paztis
--- a/src/core/importType.js
+++ b/src/core/importType.js
@@ -13,6 +13,11 @@ function baseModule(name) {
   return pkg;
 }
 
+function isInternalRegexMatch(name, settings) {
+  const internalScope = (settings && settings['import/internal-regex']);
+  return internalScope && new RegExp(internalScope).test(name);
+}
+
 export function isAbsolute(name) {
   return typeof name === 'string' && nodeIsAbsolute(name);
 }
@@ -25,33 +30,18 @@ export function isBuiltIn(name, settings, path) {
   return isCoreModule(base) || extras.indexOf(base) > -1;
 }
 
-export function isExternalModule(name, settings, path, context) {
-  if (arguments.length < 4) {
-    throw new TypeError('isExternalModule: name, settings, path, and context are all required');
+export function isExternalModule(name, path, context) {
+  if (arguments.length < 3) {                                                                                                                                                                              
+    throw new TypeError('isExternalModule: name, path, and context are all required');
   }
-  return (isModule(name) || isScoped(name)) && isExternalPath(name, settings, path, getContextPackagePath(context));
+  return (isModule(name) || isScoped(name)) && typeTest(name, context, path) === 'external';
 }
 
-export function isExternalModuleMain(name, settings, path, context) {
-  return isModuleMain(name) && isExternalPath(name, settings, path, getContextPackagePath(context));
-}
-
-function isExternalPath(name, settings, path, packagePath) {
-  const internalScope = (settings && settings['import/internal-regex']);
-  if (internalScope && new RegExp(internalScope).test(name)) {
-    return false;
+export function isExternalModuleMain(name, path, context) {
+  if (arguments.length < 3) {                                                                                                                                                                              
+    throw new TypeError('isExternalModule: name, path, and context are all required');
   }
-
-  if (!path || relative(packagePath, path).startsWith('..')) {
-    return true;
-  }
-
-  const folders = (settings && settings['import/external-module-folders']) || ['node_modules'];
-  return folders.some((folder) => {
-    const folderPath = nodeResolve(packagePath, folder);
-    const relativePath = relative(folderPath, path);
-    return !relativePath.startsWith('..');
-  });
+  return isModuleMain(name) && typeTest(name, context, path) === 'external';
 }
 
 const moduleRegExp = /^\w/;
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Fixed
 - `importType`: avoid crashing on a non-string' ([#2305], thanks [@ljharb])
 - [`first`]: prevent crash when parsing angular templates ([#2210], thanks [@ljharb])
+- Recognize properly resolving `@/*`-aliased imports as internal ([#2334], thanks [@ombene])
```
ljharb, line 12:
```suggestion
- `importType`: Recognize properly resolving `@/*`-aliased imports as internal ([#2334], thanks [@ombene])
```

ljharb (commented):
This seems like a good fix, but it makes me _very_ nervous to be changing this much of the core importType logic.

Could we add some test cases to affected rules as well?

ombene:
> Could we add some test cases to affected rules as well?

Yeah sounds like a great idea! In terms of new test cases, I can imagine something sensible for `import/order` and `import/no-internal-modules`. Does that sound right?

ljharb:
That sounds great. There's a lot of rules that use importType, but those along with `no-extraneous-dependencies` might be the only ones that care about whether something is "internal" or not.

ombene:
👍 I [took a crack at it](https://github.com/import-js/eslint-plugin-import/compare/4f810cd4270e57f95b255c53578a6d103db8c4aa...69c767e81ac438006bb5eddf1e73e0bebb5cd170). What do you think?

src/core/importType.js
```diff
@@ -29,29 +34,11 @@ export function isExternalModule(name, settings, path, context) {
   if (arguments.length < 4) {
     throw new TypeError('isExternalModule: name, settings, path, and context are all required');
   }
-  return (isModule(name) || isScoped(name)) && isExternalPath(name, settings, path, getContextPackagePath(context));
+  return (isModule(name) || isScoped(name)) && (isExternalPath(path, context) || isExternalLookingName(name));
```
ljharb, line 37: it seems like `isExternalModule` is missing the `isInternalPath` logic that's in `typeTest`. Should these be in sync? or perhaps, should this function use `typeTest`'s result instead?

(same question about `isExternalModuleMain`)

ombene: Great catch! Yeah just updated to use `typeTest`, that seems like a good idea

ombene: I'm also realizing that at least in theory to be as backwards compatible as possible the `settings` argument should get used if provided. So I just updated that, too

src/core/importType.js
```diff
@@ -87,17 +74,48 @@ function isRelativeToSibling(name) {
...
+  }
+  const packagePath = getContextPackagePath(context);
+  return !relative(packagePath, path).startsWith('../');
+}
+
+function isExternalLookingName(name) {
+  return isModule(name) || isScoped(name);
+}
+
+function typeTest(name, context, path, settings = context.settings) {
```
ljharb, line 109: can you remind me, when do we want to pass a `settings` that is different than `context.settings`? iow, why are we adding this fourth argument?

ombene: Good question.

My original reasoning was that since `isExternalModule` / `isExternalModuleMain` are being exported, they should be considered part of the public API (though this may be incorrect—I don't know—it seemed like a safer assumption than the opposite conclusion). And based on how they're working in main, it seems in theory possible for a consumer to pass in separate `settings` which would take precedence over `context.setting`. I wasn't sure whether that use case was important or not, but figured that any behavior change could be a separate concern for a different PR. Internally to `eslint-plugin-import`, the only source calls to `isExternalModule` that I'm finding all provide `context.settings` as settings—so, no internal use cases for `settings` being different than `context.settings`.

Anyways, because of the potential for external consumers relying on this "feature", I figured I should keep the behavior constant. At the time, the simplest way to me seemed to be to provide settings as an argument to `typeTest` as well, but alternatively we could do something like `typeTest(name, {...context, settings}, path)` from `isExternalModule`. I thought that'd be messier, but now that I'm re-thinking it seems reasonably simple, and probably a lot better than adding unnecessary complexity to `typeTest`. I'll update to that.

 I would be in favor of a change that removed the `settings` argument entirely from `isExternalModule` and `isExternalModuleMain`. If consumers do need to override `settings` for some reason, they could migrate from `isExternalModule(name, settings, path, context)` => `isExternalModule(name, {...context, settings}, path)`. My default inclination is to do that in a separate PR, but I'd be perfectly happy to include the change into this PR as well.

ombene: OK, so I just tried to do `typeTest(name, { ...context, settings }, path)` but that fails because the eslint context object has a non standard prototype, and so for example `getFilename` is now missing from the attempted copy (producing errors). With a more faithful copy, so `typeTest(name, Object.assign(Object.create(Object.getPrototypeOf(context)), context, { settings }, path)`, I'm now seeing:

```
TypeError: Cannot assign to read only property 'settings'
```

Which is probably fixable with some more work, but the solution's complexity is already making me think that this might not be the ideal path to try and follow. Any thoughts?

ljharb: If there's no internal use cases for passing a settings that's different than `context.settings`, then it's better not to permit that.

src/core/importType.js
```diff
@@ -29,29 +34,11 @@ export function isExternalModule(name, settings, path, context) {
   if (arguments.length < 4) {
     throw new TypeError('isExternalModule: name, settings, path, and context are all required');
```
ljharb, line 35: ahhh i see what you mean. I have no idea how i was able to add this exception check, but why i also didn't require `settings === context.settings` in the same check. Let's try adding that runtime check here.

or, alternatively, let's just remove the `settings` argument entirely, and not worry about the breaking change since this is plugin internals?

src/core/importType.js
```diff
@@ -29,29 +34,11 @@ export function isExternalModule(name, settings, path, context) {
...
   }
-  return (isModule(name) || isScoped(name)) && isExternalPath(name, settings, path, getContextPackagePath(context));
+  return (isModule(name) || isScoped(name)) && typeTest(name, context, path, settings) === 'external';
 }
 
 export function isExternalModuleMain(name, settings, path, context) {
```
ljharb, line 40: same here

src/core/importType.js
ombene: [Done](https://github.com/import-js/eslint-plugin-import/pull/2334/commits/670b09633d5bf3cd2ecfe75e69199f156efd43bf)

ljharb (approved):

alumni:
Cool! This also solved the tsconfig path alias `"@modules/*": ["src/modules/*"]` being identified as external in one out of two similar projects inside a monorepo. I will try to figure out what's going on in the remaining project, but this refactor is a great improvement!

PR #2341, opened 2021-12-30
[Fix] `named`/`ExportMap`: handle named imports from CJS modules that use dynamic import

### LINKED ISSUE #2294: import/named falsely reports import as not found (regression in 2.25)
eslint-plugin-import version: 2.25.0 and 2.25.2
eslint version: 7.32 and 8.1
Reproduction repo: https://github.com/ludofischer/eslint-plugin-import-error-demo

When using the `optimize` named import from the [svgo](https://github.com/svg/svgo) 2.8 library, since 2.25 eslint-plugin-import reports
```
error  optimize not found in 'svgo'  import/named
```
while the import was correctly handled as valid in eslint-plugin-import 2.24. svgo uses CommonJS with just a `main` `package.json` field that points to this file: https://github.com/svg/svgo/blob/master/lib/svgo-node.js

ludofischer:
Fix #2294
* Modify `ExportMap` to track whether the module is unambiguously ES6
* Add tests for named import with `import` as `require()`

### Proposed solution
Mark ambiguous (i.e. not unambiguously ESM) modules that contain dynamic import()
so that they can be ignored like other ambiguous modules when analysing named imports.

Previous to https://github.com/import-js/eslint-plugin-import/commit/7c382f02a6cafe676ee754a896092040dad8daf6 this was not necessary because a `null` ExportMap was returned for all ambiguous modules.

### Alternatives
Parse the CommonJS exports instead of ignoring the CommonJS modules. Rejected because analysingg CommonJS exports seems out of scope of this plugin.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -43,6 +43,10 @@ export default class ExportMap {
      */
     this.imports = new Map();
     this.errors = [];
+    /**
+     * type {'ambiguous' | 'Module' | 'Script'}
+     */
+    this.parseGoal = 'ambiguous';
   }
 
   get hasDefault() { return this.get('default') != null; } // stronger than this.has
@@ -406,7 +410,8 @@ ExportMap.parse = function (path, content, context) {
     },
   });
 
-  if (!unambiguous.isModule(ast) && !hasDynamicImports) return null;
+  const unambiguouslyESM = unambiguous.isModule(ast);
+  if (!unambiguouslyESM && !hasDynamicImports) return null;
 
   const docstyle = (context.settings && context.settings['import/docstyle']) || ['jsdoc'];
   const docStyleParsers = {};
@@ -710,6 +715,9 @@ ExportMap.parse = function (path, content, context) {
     m.namespace.set('default', {}); // add default export
   }
 
+  if (unambiguouslyESM) {
+    m.parseGoal = 'Module';
+  }
   return m;
 };
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - `importType`: avoid crashing on a non-string' ([#2305], thanks [@ljharb])
 - [`first`]: prevent crash when parsing angular templates ([#2210], thanks [@ljharb])
 - `importType`: properly resolve `@/*`-aliased imports as internal ([#2334], thanks [@ombene])
+- [`named`]/`ExportMap`: handle named imports from CJS modules that use dynamic import ([#2341], thanks [@ludofischer])
 
 ### Changed
 - [`no-default-import`]: report on the token "default" instead of the entire node ([#2299], thanks [@pmcelhaney])
@@ -951,6 +952,7 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#2341]: https://github.com/import-js/eslint-plugin-import/pull/2341
 [#2334]: https://github.com/import-js/eslint-plugin-import/pull/2334
 [#2305]: https://github.com/import-js/eslint-plugin-import/pull/2305
 [#2299]: https://github.com/import-js/eslint-plugin-import/pull/2299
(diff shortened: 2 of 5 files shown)


src/ExportMap.js
```diff
@@ -43,6 +43,13 @@ export default class ExportMap {
      */
     this.imports = new Map();
     this.errors = [];
+    /**
+     * true when we are still unsure if
+     * the module is ESM, happens when the module
+     * contains dynamic `import()` but no other
+     * ESM import/export
+     */
+    this.maybeNotEsm = false;
```
ljharb, line 52: "containing dynamic import" tells us nothing, so it'd be find to maybe name this `this.parseGoal`?

that way it could be set to `'ambiguous'` by default, `'Module'` when it finds an export or an import, `Script` when it finds a `with` statement, etc.

src/ExportMap.js
```diff
@@ -406,7 +413,8 @@ ExportMap.parse = function (path, content, context) {
     },
   });
 
-  if (!unambiguous.isModule(ast) && !hasDynamicImports) return null;
+  const maybeNotEsm = !unambiguous.isModule(ast);
+  if (maybeNotEsm && !hasDynamicImports) return null;
```
ljharb, line 417: it could still be ESM in this case - it depends on what the eslint config has set for `sourceType`.

ljharb (commented):
Tests look good!

src/ExportMap.js
ludofischer: Sounds reasonable. Basically make the three types of results explicit.

ludofischer: But what happens in the cases where the code returns a `null` `ExportMap` currently? Should  it return an `ExportMap` with `parseGoal` set to `Script` as a sort of null object?

src/ExportMap.js
ludofischer: Do you suggest adding a check on the `sourceType` value in the config? I don't see any code checking this at the moment, but maybe I am not looking for it correctly.

src/ExportMap.js
ljharb: That seems reasonable - to always have a consistent type returned.

src/ExportMap.js
ljharb: You're right, we probably aren't checking that. I think that specifically when the goal is ambiguous, we should defer to the config (and add that checking)

ludofischer:
I've applied your suggestion of turning `maybeNotEsm` into a `parseGoal` field. Is there another refactor you had in mind where this `parseGoal` should be used? Another thing that was on my mind is that the `default` rule might also fail in a similar case.

ljharb:
I'm mainly hoping that we can get these kinds of helpers in a place where it becomes trivial to support CJS files as well as ESM files.

What are your thoughts about having an "empty" ExportsMap instead of `null` for a CJS file without dynamic imports?

ludofischer:
> What are your thoughts about having an "empty" ExportsMap instead of `null` for a CJS file without dynamic imports?

I need to take a longer look at the code. My instinct would be that it increases the risk of breaking something else as we'll miss one spot that's relying on the ExportsMap being `null`.

ljharb:
Fair enough, let's do that in a follow-up PR then.

ljharb (approved):

ludofischer:
> Fair enough, let's do that in a follow-up PR then.

I feel like I am missing some context to make a follow-up. I've looked at the implementation for the rules some more and I've also looked at the issues. I cannot find  case where returning empty ExportMap from a CJS file would make a difference:
1. it would not get rid of returned `null` since `ExportMap.get()` returns `null` when the file is ignored or has an invalid extension according to the config
2. all of `dependencies`, `imports`, `errors` would be empty, it would be the same as skipping processing completely, which already happens when `ExportMap` is `null`

On an unrelated note, have you thought about using issue templates to require a link to a reproduction? I've got the impression that a lot of open issues are unactionable because they lack a reproduction.

ljharb:
@ludofischer i'm mainly thinking that it'd make it clearer for rules to use ExportsMap - since CJS Scripts have a dep graph too, and this plugin should support them.

As for an issue template, I'd be happy to look at a PR that added one :-)

PR #2342, opened 2021-12-30
[Improvement]: `no-extraneous-dependencies` add support for npm workspaces

acdibble:
This adds support for npm workspaces (I am the only one using them?) with the plugin.

The rule is set to `false` by default because the vast minority of projects will be using npm workspaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Changed
 - [`no-default-import`]: report on the token "default" instead of the entire node ([#2299], thanks [@pmcelhaney])
 - [Docs] [`order`]: Remove duplicate mention of default ([#2280], thanks [@johnthagen])
+- [`no-extraneous-dependencies`]: add npm workspaces support ([#2342], thank [@acdibble])
 
 ## [2.25.3] - 2021-11-09
 
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -1,6 +1,6 @@
 # import/no-extraneous-dependencies: Forbid the use of extraneous packages
 
-Forbid the import of external modules that are not declared in the `package.json`'s `dependencies`, `devDependencies`, `optionalDependencies`, `peerDependencies`, or `bundledDependencies`.
+Forbid the import of external modules that are not declared in the `package.json`'s `dependencies`, `devDependencies`, `optionalDependencies`, `peerDependencies`, `bundledDependencies`, or npm `workspaces`.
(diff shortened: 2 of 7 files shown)


ljharb:
Could you elaborate a bit?

A commit adding bundledDeps is fine (altho both `bundleDependencies` and `bundledDependencies` are supported names for this property, so it'll be a bit more involved than you've got here) - but I'm confused about the workspaces use case.

In a workspace child project, everything that's required/imported should still be defined in the child project's package.json.

acdibble:
Unless I've been using npm workspaces horribly incorrectly (always possible), when a  new workspace is created, it receives an entry in the root-level package.json `workspaces` array and a symlink in `node_modules`, but the child packages do not receive any sort of entries in its package.json.

So if a child wants to require a sibling, the extraneous dependency error appears.

A manual entry can be created with a `file:../sibling` entry, but it seems like they would do that automatically if that was the intended use of npm workspaces? I get an error when trying to run `npm i ./packages/a -w packages/b`, so it seems like this isn't intended configuration.

ljharb:
You're definitely *supposed* to have explicit entries in package.json, whether npm does it automatically for you or not.

No project should require or import anything it doesn't explicitly depend on.

The error you get is because you're installing file paths. Try adding an explicit entry of the dependency's name, for `*` (it can even be a peer dep and not a regular dep)

acdibble:
The [workspaces docs](https://docs.npmjs.com/cli/v7/using-npm/workspaces#using-workspaces) seem to imply that the child packages are supposed to be consumed without any further entries anywhere.

I _think_ this would be a proper application of the plugin to workspaces intended use, but I can close this if it seems like it's promoting a bad practice.

ljharb:
Let's not close it just yet; it's still worth adding bundled deps support :-)

I think that if npm workspaces isn't making it easy to, or encouraging, having explicit sibling workspace deps in the child projects' package.jsons, that's something that should be filed with npm itself (in the cli issues, or as an RFC).

acdibble:
> Let's not close it just yet; it's still worth adding bundled deps support :-)

I think the bundled dependencies support is already there, I just added some stuff next to it :).

> I think that if npm workspaces isn't making it easy to, or encouraging, having explicit sibling workspace deps in the child projects' package.jsons, that's something that should be filed with npm itself (in the cli issues, or as an RFC).

Agreed.

ljharb:
ah, whoops, i just got confused by the diff churn.

I think that if there is something workspaces-specific to add, we should add it - but if the only thing would be "make it OK to have implicit, undocumented dependencies on sibling child workspace projects", then I think it's probably better not to do that :-)

acdibble:
Agreed, thanks for your time!

PR #2212, opened 2021-08-30
[New] no-unused-modules: support dynamic imports [2]

Hypnosphi:
Replaces #1660

PR #2343, opened 2021-12-30
[utils] [patch] Fix `@babel/eslint-parser` 8 compatibility

nicolo-ribaudo:
In Babel 8, you won't be able to directly require `@babel/eslint-parser/worker/ast-info.cjs` anymore (there is a reason if the path contains "worker": it's meant to be used in workers after properly setting them up :stuck_out_tongue:). However, `@babel/eslint-parser` exposes the `visitorKeys` on its result object, similarly to how `@typescript-eslint/parser` does: you can access it without requiring any internal file.

We caught this in our integration tests: even if for some reason the error is logged without reporting a failure, in https://app.circleci.com/pipelines/github/babel/babel/8764/workflows/3516a741-7bca-4713-8b12-9f2f20310066/jobs/50483/parallel-runs/0/steps/0-103 you can see
```
Error while parsing /home/circleci/babel/eslint/babel-eslint-tests/test/fixtures/eslint-plugin-import/a.js
Line undefined, column undefined: Cannot read properties of undefined (reading 'VISITOR_KEYS')
 PASS  eslint/babel-eslint-tests/test/integration/eslint-plugin-import.js (66.694 s)
```

There aren't any tests for `@babel/eslint-parser`, and I'm not sure about how to add them from scratch because it requires a `babel.config.json` file when used with `eslint-plugin-import` (to specify which additional syntax should be supported, such as TS). However, I verified manually that this change makes our test pass.

### DIFF
--- a/utils/parse.js
+++ b/utils/parse.js
@@ -7,34 +7,27 @@ const fs = require('fs');
 
 const log = require('debug')('eslint-plugin-import:parse');
 
-function getBabelVisitorKeys(parserPath) {
+function getBabelEslintVisitorKeys(parserPath) {
   if (parserPath.endsWith('index.js')) {
     const hypotheticalLocation = parserPath.replace('index.js', 'visitor-keys.js');
     if (fs.existsSync(hypotheticalLocation)) {
       const keys = moduleRequire(hypotheticalLocation);
       return keys.default || keys;
     }
-  } else if (parserPath.endsWith('index.cjs')) {
-    const hypotheticalLocation = parserPath.replace('index.cjs', 'worker/ast-info.cjs');
-    if (fs.existsSync(hypotheticalLocation)) {
-      const astInfo = moduleRequire(hypotheticalLocation);
-      return astInfo.getVisitorKeys();
-    }
   }
   return null;
 }
 
 function keysFromParser(parserPath, parserInstance, parsedResult) {
+  // Exposed by @typescript-eslint/parser and @babel/eslint-parser
+  if (parsedResult && parsedResult.visitorKeys) {
+    return parsedResult.visitorKeys;
+  }
   if (/.*espree.*/.test(parserPath)) {
     return parserInstance.VisitorKeys;
   }
-  if (/.*(babel-eslint|@babel\/eslint-parser).*/.test(parserPath)) {
-    return getBabelVisitorKeys(parserPath);
-  }
-  if (/.*@typescript-eslint\/parser/.test(parserPath)) {
-    if (parsedResult) {
-      return parsedResult.visitorKeys;
-    }
+  if (/.*babel-eslint.*/.test(parserPath)) {
+    return getBabelEslintVisitorKeys(parserPath);
   }
   return null;
 }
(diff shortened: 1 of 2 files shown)


utils/parse.js
```diff
@@ -7,34 +7,27 @@ const fs = require('fs');
...
 
-function getBabelVisitorKeys(parserPath) {
+function getBabelEslintVisitorKeys(parserPath) {
   if (parserPath.endsWith('index.js')) {
     const hypotheticalLocation = parserPath.replace('index.js', 'visitor-keys.js');
     if (fs.existsSync(hypotheticalLocation)) {
       const keys = moduleRequire(hypotheticalLocation);
       return keys.default || keys;
     }
-  } else if (parserPath.endsWith('index.cjs')) {
```
ljharb, line 17: was this branch always only for the new Babel parser?

nicolo-ribaudo: Yes, introduced by https://github.com/babel/babel/pull/13199

ljharb: I’m confused; that seems to be for an unreleased Babel version. This branch here, in this plugin, is presumably for a released Babel version?

ljharb: Specifically, this was added in #1660 / #2212 by @Hypnosphi - any idea why it was needed then, but not needed after this PR?

Hypnosphi: Apparently, it targeted Babel 7.14: https://github.com/Hypnosphi/eslint-plugin-import/commit/513960d9f559fc98bc15a8215e4a63814e6f63d7

nicolo-ribaudo: @ljharb Currently we have both the versions in `@babel/eslint-parser`: the "default one" which works with this PR (because the internal file can be directly required), and the "worker one" which wouldn't work (it can be used with `@babel/eslint-parser/experimental-worker`).

The worker one will be enabled by default in Babel 8 (not released yet) and it will be the only one available. So this PR is mostly to future-proof this implementation, and to use the public `@babel/eslint-parser` API rather than its internal files (where we don't make any stability guarantee).

ljharb: ok - so for Babel 7, the PR is still supporting both parsers because they also provide visitorKeys, and in Babel 8, it will still work for the same reason - but the current way supports both parsers in Babel 7, but will break in Babel 8 due to “exports”?

nicolo-ribaudo: This PR works with:
- `babel-eslint` (which is no more maintained, so it will never have breaking changes and it will continue working forever)
- With `@babel/eslint-parser` 7 and 8, because it uses the `visitorKeys` option

I didn't actually think about `"exports"`: the breakage I found with the current approach in Babel 8 is that you cannot just require `@babel/eslint-parser/lib/worker/visitor-keys.cjs`, but you have to first `await require("@babel/eslint-parser/lib/worker/babel-core.js").init")` and _then_ you can require `@babel/eslint-parser/lib/worker/visitor-keys.cjs`.
We will also introduce `"exports"` in Babel 8, but since `eslint-plugin-import` uses absolute paths I don't think that it will be affected.

ljharb (approved):

utils/parse.js
ljharb: All of our absolute paths come from `require.resolve`, so anything not in "exports" wouldn't resolve in the first place.

Issue #2350, opened 2022-01-06
eslint-plugin-import v2.24.2 not working anymore?


AyushMittal1999:
From `version 2.25.0` eslint-plugin-import has started identifying cyclic dependency in dynamic imports. But we are having a huge codebase containing some cyclic dependency in dynamic imports. So, till the time we fixed that, we decided to restrict to eslint-plugin-import  `version 2.24.2`.
Using version 2.24.2, we found that it is not working now due to lapse in dependency version of `eslint-module-utils`. Latest version of `eslint-module-utils` (>= `2.7.0`) is not compatible with eslint-plugin-import `v2.24.2`.  But in package.json of eslint-plugin-import `v2.24.2` we are using  `"eslint-module-utils": "^2.6.2"` (https://github.com/import-js/eslint-plugin-import/blob/v2.24.2/package.json#L108).  We need to use `"eslint-module-utils": "2.6.2"` to use v2.24.2.

Please let me know how can we overcome this problem. Can we go for a `v2.24.3` with the required change to support the required feature. Thanks

ljharb:
I'm confused, there doesn't seem to be a bug here.

ljharb:
In what way is `eslint-module-utils` v2.7.0+ incompatible with `eslint-plugin-import` v2.24.2?

If there's a bug here, it'd be in `utils/`, and *not* in the main eslint-plugin-import package.

AyushMittal1999:
Considering this https://github.com/AyushMittal1999/eslint-testing sample repo, eslint-plugin-import is unable to identify cyclic dependency here.
But if I used raised [PR](https://github.com/AyushMittal1999/eslint-plugin-import/tree/eslint-plugin-import-fix-version-2.24.2) branch for eslint-plugin-import i.e.     `"eslint-plugin-import": "github:AyushMittal1999/eslint-plugin-import#fix-version-2.24.2"` it works fine

ljharb:
That an older version of the plugin isn't identifying things that a newer version does isn't a problem.

Are you saying that v2.24 of the plugin with v2.6 of the utils DOES report them, but v2.24 of the plugin with v2.7 of the utils does not?

AyushMittal1999:
Yes I mean,
 v2.24 of the plugin with v2.6 of the utils DOES report cyclic dependencies, but v2.24 of the plugin with v2.7 of the utils does not.

ljharb:
[...]

PR #2351, opened 2022-01-06
fix eslint-module-utils version for compatibility with v2.24.2

AyushMittal1999:
Solves https://github.com/import-js/eslint-plugin-import/issues/2350

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "debug": "^2.6.9",
     "doctrine": "^2.1.0",
     "eslint-import-resolver-node": "^0.3.6",
-    "eslint-module-utils": "^2.6.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -105,7 +105,7 @@
     "debug": "^2.6.9",
     "doctrine": "^2.1.0",
     "eslint-import-resolver-node": "^0.3.6",
-    "eslint-module-utils": "^2.6.2",
+    "eslint-module-utils": "2.6.2",
```
ljharb, line 108: this is never a correct change; everything should be using a `^` semver range.

ljharb (changes requested):

ljharb:
This was done in df8c1a8c.

PR #2374, opened 2022-01-27
[Refactor] `no extraneous dependencies` improve performance

meowtec:


src/core/importType.js
```diff
@@ -109,20 +109,23 @@ function isExternalLookingName(name) {
   return isModule(name) || isScoped(name);
 }
 
-function typeTest(name, context, path ) {
+function typeTest(name, context, getPath) {
   const { settings } = context;
   if (isInternalRegexMatch(name, settings)) { return 'internal'; }
-  if (isAbsolute(name, settings, path)) { return 'absolute'; }
+  if (isAbsolute(name, settings)) { return 'absolute'; }
```
ljharb, line 115: i very intentionally had all of these pass all three arguments, for consistency. if we're changing that, then this only takes 1 argument.
```suggestion
  if (isAbsolute(name)) { return 'absolute'; }
```

(similar feedback on all the others)

src/rules/no-extraneous-dependencies.js
```diff
@@ -30,6 +30,17 @@ function extractDepFields(pkg) {
   };
 }
 
+function getPackageDepFields(packageJsonPath) {
+  if (!depFieldCache.has(packageJsonPath)) {
+    const depFields = extractDepFields(
+      JSON.parse(fs.readFileSync(packageJsonPath, 'utf8')),
```
ljharb, line 36: what happens if the package.json file has invalid syntax, or doesn't exist?

src/core/importType.js
ljharb, line 115: i very intentionally had all of these pass all three arguments, for consistency. if we're changing that, then this only takes 1 argument.
```suggestion
  if (isAbsolute(name)) { return 'absolute'; }
```

(similar feedback on all the others)

src/core/importType.js
```diff
@@ -109,20 +109,23 @@ function isExternalLookingName(name) {
...
   if (isInternalRegexMatch(name, settings)) { return 'internal'; }
-  if (isAbsolute(name, settings, path)) { return 'absolute'; }
+  if (isAbsolute(name)) { return 'absolute'; }
+  if (isRelativeToParent(name)) { return 'parent'; }
+  if (isIndex(name)) { return 'index'; }
+  if (isRelativeToSibling(name)) { return 'sibling'; }
+
+  // `resolve` is expensive, do it only when necessary
+  const path = getPath();
   if (isBuiltIn(name, settings, path)) { return 'builtin'; }
```
ljharb, line 122: this change moves `isBuiltIn` below absolute/relativeToParent/index/relativeToSibling - this might be a breaking change for anyone relying on the `import/core-modules` setting, which currently takes precedence.

meowtec: Could we make this change in next major version?

ljharb: There's unlikely to be a major version any time soon, if ever.

ljharb:
@meowtec i'd love to incorporate these speed improvements if we can avoid the breaking change - can you update this PR to do that?

ljharb:
I've reordered the commits, and added "clear the cache on program exit" so that long-running eslint processes won't have a memory leak. I plan to pull the cache commit into main directly, and leave this PR containing just the "lazy resolve" commit so we can keep working on it.

PR #2375, opened 2022-01-27
[Fix] `export`: false positive for typescript namespace merging

### LINKED ISSUE #1964: [export] “Multiple exports” false positive on merged class and namespace
```ts
// foo.ts
export class Foo {
  bar: string;
  constructor(options: Foo.Options) {
    this.bar = options.bar;
  }
}

export namespace Foo {
  export interface Options {
    bar: string;
  }
}
```

```json
{
  "parser": "@typescript-eslint/parser",
  "plugins": ["import"],
  "rules": {
    "import/export": "error"
  }
}
```

```
/tmp/test/foo.ts
  1:14  error  Multiple exports of name 'Foo'  import/export
  8:18  error  Multiple exports of name 'Foo'  import/export

✖ 2 problems (2 errors, 0 warnings)
```

TypeScript [namespace merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-namespaces-with-classes-functions-and-enums) should not be an error. This is similar to #1549, but that issue seems to be specific to default exports.

magarcia:
Fixes #1964

Fixes false positives like this:
```ts
export class Foo { }
export namespace Foo { }
```
```ts
export function Foo() { }
export namespace Foo { }
```
```ts
export enum Foo { }
export namespace Foo { }
```

### DIFF
--- a/src/rules/export.js
+++ b/src/rules/export.js
@@ -36,13 +36,59 @@ const tsTypePrefix = 'type:';
  */
 function isTypescriptFunctionOverloads(nodes) {
   const types = new Set(Array.from(nodes, node => node.parent.type));
-  return (
-    types.has('TSDeclareFunction') &&
-    (
-      types.size === 1 ||
-      (types.size === 2 && types.has('FunctionDeclaration'))
-    )
-  );
+  return types.has('TSDeclareFunction')
+    && (
+      types.size === 1
+      || (types.size === 2 && types.has('FunctionDeclaration'))
+    );
+}
+
+/**
+ * Detect merging Namespaces with Classes, Functions, or Enums like:
+ * ```ts
+ * export class Foo { }
+ * export namespace Foo { }
+ * ```
+ * @param {Set<Object>} nodes
+ * @returns {boolean}
+ */
+function isTypescriptNamespaceMerging(nodes) {
+  const types = new Set(Array.from(nodes, node => node.parent.type));
+  const noNamespaceNodes = Array.from(nodes).filter((node) => node.parent.type !== 'TSModuleDeclaration');
+
+  return types.has('TSModuleDeclaration')
+    && (
+      types.size === 1
+      // Merging with functions
+      || (types.size === 2 && (types.has('FunctionDeclaration') || types.has('TSDeclareFunction')))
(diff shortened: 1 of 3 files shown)


ljharb:
Looks like you may need to skip the tests in older eslint versions, or on the older TS parser.

magarcia:
> Looks like you may need to skip the tests in older eslint versions, or on the older TS parser.

Sure 😄  I'm not sure how yet. Do you have an example of how to do that?

BTW: I'm setting it as a draft. After a walk/break 😅 , I realized that this fix won't work properly for:
```ts
export class Foo { }
export class Foo { }
export namespace Foo { }
```
I'll have a second look at it tomorrow, it's already quite late in my timezone.

ljharb:
Here's an example: https://github.com/import-js/eslint-plugin-import/blob/6c957e7df178d1b81d01cf219d62ba91b4e6d9e8/tests/src/rules/no-commonjs.js#L71-L96

src/rules/export.js
```diff
@@ -45,6 +45,41 @@ function isTypescriptFunctionOverloads(nodes) {
...
+      (types.size === 3 && types.has('FunctionDeclaration') && types.has('TSDeclareFunction')) ||
+      // Merging with classes or enums
+      (types.size === 2 && (types.has('ClassDeclaration') || types.has('TSEnumDeclaration')) && noNamespaceNodes.length === 1)
+    )
+  );
+
+  if (!isMerging && types.has('TSModuleDeclaration') && (types.has('TSEnumDeclaration') || types.has('ClassDeclaration') || types.has('FunctionDeclaration') || types.has('TSDeclareFunction'))) {
+    // Remove namespace nodes to error only on other types
+    Array.from(nodes).forEach(node => {
+      if (node.parent.type === 'TSModuleDeclaration') nodes.delete(node);
```
ljharb, line 76: mutating an argument seems very undesirable

magarcia: I was a little hesitant about the best way to do it, but I found a different approach implemented on [`d7acbba` (#2375)](https://github.com/import-js/eslint-plugin-import/pull/2375/commits/d7acbbac2619acf5d104fc12f393252a168ed8eb).

ljharb (approved):

PR #2381, opened 2022-02-04
[New] `no-relative-packages`: add fixer

forivall:
I was using no-relative packages and noticed it didn't have a fixer. So, this adds a fixer.


ljharb:
Looks good, but a bunch of tests are failing. Please mark it as ready for review when tests are passing :-)

forivall:
Hmm, i guess i missed committing that fix (the commit i just pushed). CI should be passing now, i'll review if there's still issues.

forivall:
I *think* i've identified an issue with the existing rule, in that, on win32, the error will show a message that the path in your scoped package should replace the forward slashes with back slashes. IIRC, import pathnames should always be posix paths. I'll check when i have direct access to my windows box, but for now, i just convert to posix paths in the fixer, and for now, i can look at the appveyor results.

ljharb (approved):

PR #2387, opened 2022-02-21
[New] `no-cycle`: add option to allow cycle via dynamic import

### LINKED ISSUE #2265: Allow cyclic dependencies for `import/no-cycle` iff mixing static/dynamic imports
## Request

Add a new option for `import/no-cycle` to allow/warn/error about cyclic dependencies if, and only if, the cycle involves dynamic imports.

## Problem Statement

There exists a use case for web apps where some modules are stored in a form of global state and then other modules import that state to update it. For example, an (overly simplistic/half-pseudo code) purpose for this might look something like:

```javascript
// state.js
import Comp1 from './Comp1';
import Comp2 from './Comp2';
import Comp3 from './Comp3';

const routeModuleMap = {
    '/path1': Comp1,
    '/path2': Comp2,
    '/path3': Comp3,
};

const routeOrder = [
    '/path1',
    '/path2',
    '/path3',
];

let state = 0;

export function getActiveModule() {
    return routeModuleMap[location.pathname];
}

export function setActiveModule(dx = 1) {
    if (dx === 0) {
        return;
    }

    if (dx < 0) {
        state = Math.max(0, state + dx);
    }

    if (dx > 0) {
        state = Math.min(routeOrder.length - 1, state + dx);
    }

    history.pushState(null, null, routeOrder[state]);
}

export default const Context = React.createContext({ getActiveModule, setActiveModule });


// parent.jsx
import Context, { getActiveModule, setActiveModule } from './state';

export default function App() {
    const Component = getActiveModule();

    return (
        <Context.Provider value={{ getActiveModule, setActiveModule }}>
            <Component />
        </Context.Provider>
    );
}


// Comp1.jsx
import Context from './state';

export default Comp1() {
    const { setActiveModule } = useContext(Context);

    return (
        <button onClick={() => setActiveModule()}>
            Go to next page
        </button>
    );
}
```

Obviously this has a circular dependency, so when bundled, it might result in `CompN is not defined` errors.
However, that can easily be resolved by mixing static and dynamic imports, which return promises instead of the code itself. This means that all files are parsed/run _before_ any promises resolve, meaning that we never have `X is not defined` errors.

i.e.

```javascript
// old
import Comp1 from './Comp1';
import Comp2 from './Comp2';
import Comp3 from './Comp3';

const routeModuleMap = {
    '/path1': Comp1,
    '/path2': Comp2,
    '/path3': Comp3,
};


// new
const routeModuleMap = {
    '/path1': React.lazy(() => import('./Comp1')),
    '/path2': React.lazy(() => import('./Comp2')),
    '/path3': React.lazy(() => import('./Comp3')),
};
```

While this is a very simple and contrived example, there are cases where some sort of global state links to modules, and each module needs to modify that global state. In these cases, mixing static with dynamic imports solves the problem of code not being defined because your bundler of choice converts `A --> B --> A` to `A --> B --> Promise.resolve(A)`.

Of course, having circular dependencies in your code is an anti-pattern and reflects brittle code that should still be refactored. This is a great plugin/rule to do exactly that: fix old code and guide new code standards.
But for gigantic repositories, those can't all be fixed at once, especially when the code is in more complex global state. For those specific situations, it'd be really helpful to be able to toggle dynamic-import-cycles using dynamic imports from allow --> warn --> error gradually as they are addressed.
This would mean **errors are still shown** for static import cycles, protecting new code while old code is refactored.

GerkinDev:
Hi,

as discussed in #2265, I added a new option, `allowUnsafeDynamicCyclicDependency`. I've tested all cases I could.

Unfortunately, I did not managed to write tests using the `TS_OLD` parser. I don't know if this is important.

Feedbacks are welcome ! Have a nice day

Fixes #2265.

### DIFF
--- a/docs/rules/no-cycle.md
+++ b/docs/rules/no-cycle.md
@@ -75,6 +75,22 @@ import { a } from './dep-a.js' // not reported as this module is external
 
 Its value is `false` by default, but can be set to `true` for reducing total project lint time, if needed.
 
+#### `allowUnsafeDynamicCyclicDependency`
+
+This option disable reporting of errors if a cycle is detected with at least one dynamic import.
+
+```js
+// bar.js
+import { foo } from './foo';
+export const bar = foo;
+
+// foo.js
+export const foo = 'Foo';
+export function getBar() { return import('./bar'); }
+```
+
+> Cyclic dependency are **always** a dangerous anti-pattern as discussed extensively in [#2265](https://github.com/import-js/eslint-plugin-import/issues/2265). Please be extra careful about using this option.
+
 ## When Not To Use It
 
 This rule is comparatively computationally expensive. If you are pressed for lint
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`no-named-default`, `no-default-export`, `prefer-default-export`, `no-named-export`, `export`, `named`, `namespace`, `no-unused-modules`]: support arbitrary module namespace names ([#2358], thanks [@sosukesuzuki])
 - [`no-dynamic-require`]: support dynamic import with espree ([#2371], thanks [@sosukesuzuki])
 - [`no-relative-packages`]: add fixer ([#2381], thanks [@forivall])
+- [`no-cycle`]: add `allowUnsafeDynamicCyclicDependency` option ([#2387], thanks [@GerkinDev])
 
 ### Fixed
 - [`default`]: `typescript-eslint-parser`: avoid a crash on exporting as namespace (thanks [@ljharb])
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -395,6 +395,7 @@ ExportMap.parse = function (path, content, context) {
           loc: source.loc,
         },
         importedSpecifiers,
+        dynamic: true,
       }]),
     });
   }
--- a/src/rules/no-cycle.js
+++ b/src/rules/no-cycle.js
@@ -33,6 +33,11 @@ module.exports = {
         type: 'boolean',
         default: false,
       },
+      allowUnsafeDynamicCyclicDependency: {
+        description: 'Allow cyclic dependency if there is at least one dynamic import in the chain',
+        type: 'boolean',
+        default: false,
+      },
     })],
   },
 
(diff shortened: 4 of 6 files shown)


GerkinDev:
Seeing the CI job https://app.travis-ci.com/github/import-js/eslint-plugin-import/builds/246801187, it seems that eslint<4 or node<8 does not parse dynamic imports in a way that allows cyclic dependencies detection with dynamic imports, Yet, it passes for later versions. I don't really know what to do. Should I remove this test ?

ljharb:
@GerkinDev no, but it can be skipped on older eslint versions.

docs/rules/no-cycle.md
```diff
@@ -75,6 +75,22 @@ import { a } from './dep-a.js' // not reported as this module is external
 
 Its value is `false` by default, but can be set to `true` for reducing total project lint time, if needed.
 
+#### `allowUnsafeDynamicCyclicDependency`
+
+This option disable reporting of errors if a cycle is detected with at least one dynamic import.
```
ljharb, line 80: the intention, i thought, was that this option would not warn on cycles where the cycle passes through a dynamic import. if the cycle is separately discovered statically, then it should still warn on it.

ljharb (commented):
I'd like to make sure we also test the case where there's a cycle that's accessed via dynamic import - that even with this option, the static cycle is warned on.

docs/rules/no-cycle.md
GerkinDev: If I understand correctly, that's the current behavior. See the tests

GerkinDev: Test cases & fix have been added in 0fe7fc18336bba59398ff693d2e7012102885ae2.

GerkinDev: Up

ljharb:
I've rebased this, and ensured no tests were removed. If everything passes, this should be good to go.

ljharb (approved):

Anber:
It looks like the changes weren't published https://unpkg.com/eslint-plugin-import@2.26.0/lib/rules/no-cycle.js

ljharb:
Yet. They were only merged 8 days ago. You can click on the merging commit hash to see that there’s no release tags attached to it.

Anber:
However, it is mentioned in the changelog as released in 2.26.0 https://github.com/import-js/eslint-plugin-import/blob/main/CHANGELOG.md#2260---2022-04-05

ljharb:
you're right, thanks, i'll fix that.

ljharb:
Note that the v2.26.0 changelog doesn't have that mistake: https://github.com/import-js/eslint-plugin-import/blob/v2.26.0/CHANGELOG.md

Issue #2291, opened 2021-11-04
`pathGroupsExcludedImportTypes` must be set to an empty array?


thany:
Here's the relevant bit of our config:
```json
    "import/order": [ "error", {
      "groups": [
        [ "builtin", "external" ],
        [ "internal" ],
        [ "parent", "sibling", "index" ],
        [ "object", "type" ]
      ],
      "pathGroups": [
        {
          "pattern": "Forms/**",
          "group": "internal",
          "position": "before"
        }
      ],
      "pathGroupsExcludedImportTypes": [],
      "newlines-between": "always"
    } ],
```

This works, I think. But I fail to understand why. The `pathGroups`, I assume, tells `import/order` the patterns that belong to `internal` imports.

But then why would I *ever* need to specify the `position` as well? Surely this is handled by `groups` already? 🤨

Also why does this *only* work when `pathGroupsExcludedImportTypes` is an empty array? I cannot understand *for the life of me* what that thing actually does. But it has to do with `pathGroups`, in some way or another, so I just started guessing at its value. The documentation is incredibly vague on this. Something is excluded - what is exlcuded, and from what? And what effect does it have on import order, and why?

Lastly, I shouldn't have to use `pathGroups` in the first place. This is a workaround for the fact that even with a parser that understands `jsconfig.json` or `tsconfigjson`, after [2 years](https://github.com/import-js/eslint-plugin-import/issues/1490) it still doesn't see that these aliased paths actually belong to `internal`. I haven't looked at the code, but it feels like `import/order` way of deterining what sort of import is external vs internal, is hardcoded, instead of taken from any kind of configuration to speak of.

I'd like to keep my configs DRY, but that seems impossible with this plugin. Can we get this fixed please?

Recap:

* `position` shouldn't be neccesary - handled by `groups` order.
* `pathGroupsExcludedImportTypes` is vague:
    * Needs better explanation
    * Might need a more sane default value
    * Should not be neccesary when just using `pathGroups` to "assign" certain paths to a group
* Please respect project config in `jsconfig.json` or `tsconfig.json`. Using one of these is the defacto standard to configure path aliases. And ESLint is aware of these when using a parser that requires them.

ljharb:
It would help if you made a PR with failing test cases.

I don't see why we'd want to respect `jsconfig.json`, which isn't really a thing, but we do respect tsconfig.json already afaik.

[...]

PR #2392, opened 2022-03-02
[Docs] `order`: use correct default value

hyperupcall:
Hai~,

In the source, the [default values](https://github.com/import-js/eslint-plugin-import/blob/b0e6f7f48945e2533e96d513248bf3e54e0c1aac/src/rules/order.js#L589) includes `"object"` for option `pathGroupsExcludedImportTypes`, but this is not reflected in the docs

This discrepancy been mentioned in [this comment](https://github.com/import-js/eslint-plugin-import/issues/2291#issuecomment-1050199872)


ljharb (approved):
Thanks!

PR #2393, opened 2022-03-02
Fix title

TheJaredWilcurt:
matches other docs pages now, easier for copy/pasting the rule


ljharb (changes requested):
Could you also update `imports-first` in the same way?

ljharb (approved):
Thanks!

PR #2395, opened 2022-03-03
[New] `order`: add `distinctGroup` option

### LINKED ISSUE #2292: Custom ordering
How can I get a certain import to customly get ordered *within* its group?

I tried:

```json
    "import/order": [ "error", {
      "groups": [
        [ "builtin", "external" ],
        [ "internal" ],
        [ "parent", "sibling", "index" ],
        [ "object", "type" ]
      ],
      "pathGroups": [
        {
          "pattern": "react",
          "group": "external",
          "position": "before"
        }
      ],
      "newlines-between": "always"
    } ],
```

Hoping that `react` will always be at the top of its group `external`, but it's not working. It's magicing up a new group, because it now requires a newline between it and any subsequent imports that are also `external`.

I cannot reference that new magic group:

```json
"groups": [
  [ "react", "builtin", "external" ],
  [ "internal" ],
  [ "parent", "sibling", "index" ],
  [ "object", "type" ]
],
```

Pure guesswork, because the documentation doesn't say how to deal with this. This of course, produces an "unknown group" error.

And no, it's going into the `builtin` group, because of the `"position": "before"` property, because then it wouldn't require a newline between it and subsequent imports.

As stated in my previewsly created issue, `pathGroupsExcludedImportTypes` is super vague, and even though there's an example of it that makes it include `react`, it's not working as expected for me. I'm not sure if I need it at all.

Ultimately what I want to achieve is:

* Order imports alphabetically within their groups, but;
* Customly order certain imports to be always at the top, like `react`.

hyperupcall:
This PR aims to close #2292. It implements what is proposed

I'll add that this functionality is really useful for emulating the behavior of TSLint's native `ordered-imports` option. This is needed for a PR I've made for [tslint-to-eslint-config](https://github.com/typescript-eslint/tslint-to-eslint-config/pull/1402).

In the linked issue:

> It seems reasonable to allow a custom pathGroup to be able to opt out of having a newline inserted around it, but I'm not sure what that schema would look like

I did a first pass implementation that seemed good - lemmie know if I need to change anything

Before this gets merged, I would like to add several more tests to ensure everything is working OK - I opened this PR hopefully for some initial feedback

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -493,7 +493,7 @@ function removeNewLineAfterImport(context, currentImport, previousImport) {
   return undefined;
 }
 
-function makeNewlinesBetweenReport(context, imported, newlinesBetweenImports) {
+function makeNewlinesBetweenReport(context, imported, newlinesBetweenImports, distinctGroup) {
   const getNumberOfEmptyLinesBetween = (currentImport, previousImport) => {
     const linesBetweenImports = context.getSourceCode().lines.slice(
       previousImport.node.loc.end.line,
@@ -502,27 +502,34 @@ function makeNewlinesBetweenReport(context, imported, newlinesBetweenImports) {
 
     return linesBetweenImports.filter((line) => !line.trim().length).length;
   };
+  const getIsStartOfDistinctGroup = (currentImport, previousImport) => {
+    return currentImport.rank - 1 >= previousImport.rank;
+  };
   let previousImport = imported[0];
 
   imported.slice(1).forEach(function (currentImport) {
     const emptyLinesBetween = getNumberOfEmptyLinesBetween(currentImport, previousImport);
+    const isStartOfDistinctGroup = getIsStartOfDistinctGroup(currentImport, previousImport);
 
     if (newlinesBetweenImports === 'always'
         || newlinesBetweenImports === 'always-and-inside-groups') {
       if (currentImport.rank !== previousImport.rank && emptyLinesBetween === 0) {
-        context.report({
-          node: previousImport.node,
-          message: 'There should be at least one empty line between import groups',
-          fix: fixNewLineAfterImport(context, previousImport),
-        });
-      } else if (currentImport.rank === previousImport.rank
-        && emptyLinesBetween > 0
+        if (distinctGroup || (!distinctGroup && isStartOfDistinctGroup)) {
+          context.report({
+            node: previousImport.node,
+            message: 'There should be at least one empty line between import groups',
+            fix: fixNewLineAfterImport(context, previousImport),
+          });
+        }
+      } else if (emptyLinesBetween > 0
         && newlinesBetweenImports !== 'always-and-inside-groups') {
-        context.report({
-          node: previousImport.node,
-          message: 'There should be no empty line within import group',
-          fix: removeNewLineAfterImport(context, currentImport, previousImport),
-        });
+        if ((distinctGroup && currentImport.rank === previousImport.rank) || (!distinctGroup && !isStartOfDistinctGroup)) {
+          context.report({
+            node: previousImport.node,
+            message: 'There should be no empty line within import group',
+            fix: removeNewLineAfterImport(context, currentImport, previousImport),
+          });
+        }
       }
     } else if (emptyLinesBetween > 0) {
       context.report({
@@ -544,6 +551,9 @@ function getAlphabetizeConfig(options) {
   return { order, caseInsensitive };
 }
 
+// TODO, semver-major: Change the default of "distinctGroup" from true to false
+const defaultDistinctGroup = true;
+
 module.exports = {
   meta: {
     type: 'suggestion',
@@ -561,6 +571,10 @@ module.exports = {
           },
           pathGroupsExcludedImportTypes: {
             type: 'array',
+          },
+          distinctGroup: {
+            type: 'boolean',
+            default: defaultDistinctGroup,
           },
           pathGroups: {
             type: 'array',
@@ -582,6 +596,7 @@ module.exports = {
                   enum: ['after', 'before'],
                 },
               },
+              additionalProperties: false,
               required: ['pattern', 'group'],
             },
           },
@@ -622,6 +637,7 @@ module.exports = {
     const newlinesBetweenImports = options['newlines-between'] || 'ignore';
     const pathGroupsExcludedImportTypes = new Set(options['pathGroupsExcludedImportTypes'] || ['builtin', 'external', 'object']);
     const alphabetize = getAlphabetizeConfig(options);
+    const distinctGroup = options.distinctGroup == null ? defaultDistinctGroup : !!options.distinctGroup;
     let ranks;
 
     try {
@@ -724,7 +740,7 @@ module.exports = {
       'Program:exit': function reportAndReset() {
         importMap.forEach((imported) => {
           if (newlinesBetweenImports !== 'ignore') {
-            makeNewlinesBetweenReport(context, imported, newlinesBetweenImports);
+            makeNewlinesBetweenReport(context, imported, newlinesBetweenImports, distinctGroup);
           }
 
           if (alphabetize.order !== 'ignore') {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`no-cycle`]: add `allowUnsafeDynamicCyclicDependency` option ([#2387], thanks [@GerkinDev])
 - [`no-restricted-paths`]: support arrays for `from` and `target` options ([#2466], thanks [@AdriAt360])
 - [`no-anonymous-default-export`]: add `allowNew` option ([#2505], thanks [@DamienCassou])
+- [`order`]: Add `distinctGroup` option ([#2395], thanks [@hyperupcall])
 
 ### Fixed
 - [`order`]: move nested imports closer to main import entry ([#2396], thanks [@pri1311])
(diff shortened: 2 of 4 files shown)


src/rules/order.js
```diff
@@ -547,7 +551,12 @@ module.exports = {
                   type: 'string',
                   enum: ['after', 'before'],
                 },
+                insertNewline: {
+                  type: 'boolean',
+                  default: true,
```
ljharb, line 556: I’m not sure if json schema defaults actually apply in every eslint version. Separately, it’s better to have boolean options treat absence the same as false. Can we reword this so it can default to false?

ljharb (commented):
Looks pretty good!

src/rules/order.js
```diff
@@ -299,7 +299,11 @@ function mutateRanksToAlphabetize(imported, alphabetizeOptions) {
 
 function computePathRank(ranks, pathGroups, path, maxPosition) {
   for (let i = 0, l = pathGroups.length; i < l; i++) {
-    const { pattern, patternOptions, group, position = 1 } = pathGroups[i];
+    const { pattern, patternOptions, group, position = 1, insertNewline } = pathGroups[i];
```
ljharb, line 302: i think maybe the property name is changed now?

LSH0924:
hi there.
This is exactly the function I am looking for. is this draft still alive?

hyperupcall:
@LSH0924 When I last made an update, I think there were issues with for example alphabetization not working correctly somehow. I don't remember the specifics, but I've been meaning to test it again and see if there is an actual issue or I was just tripping. I'm going to give it another go today or tomorrow, but if there is an issue and I'm not able to fix it this week I'm probably going to close this issue and leave it up to somebody else to continue from here.

ljharb:
@hyperupcall thanks! in either case, please do not close the PR; others can update it later instead.

hyperupcall:
Okay so the feature has been implemented, but on further thought, I think it might be better to fix the original issue in a different way. The original feature request (and mine too) was/is

> How can I get a certain import to customly get ordered within its group?

Within the implementation, the imports _are_ ordered within its group because processing imports. The catch is (as I understand it) patterns that match a particular `pathGroup[].pattern` _create a new_ group. As I understand it, these new groups always have a decimal `rank` value. But, this doesn't intuitively make sense because the `pathGroup.position` property doesn't imply a new group is created, only that it is ordered within its own `pathGroup.group`

The `noInsertNewline` property partially solves this because it allows for manually disabling newlines and making it appear that everything works. But I have found that it is quite hard to get the matching behavior in various situations (possibly impossible), and it requires extra effort

I propose an additional property called something like `pathGroupsNaturalNewlines` - it will automatically create newlines the way that makes sense. Moreover, modifications to `pathGroups` appear to not create an extra group, Below is a demonstration of the current implementation 

Here is what the file would currently look like 

<details>
<summary>file.ts</summary>

import A from "Bar";
import B from "baz";
import b from "Foo";

import J from ".";

import F from "..";
import D from "../";

import Z from "../Bar";
import O from "../baz";
import Z from "../Foo";

import L from "./Bar";
import P from "./baz";
import L from "./Foo";


</details>

This is what the file would look like if `pathGroupsNaturalNewlines` behaves as is described and is `true`:


<details>
<summary>file.ts</summary>

import A from "Bar";
import B from "baz";
import b from "Foo";

import J from ".";
import F from "..";
import D from "../";
import Z from "../Bar";
import O from "../baz";
import Z from "../Foo";

import L from "./Bar";
import P from "./baz";
import L from "./Foo";


</details>




<details>
<summary>.eslintrc.json</summary>

```json
"pathGroupsExcludedImportTypes": [],
"pathGroupsNaturalNewlines": true,
"pathGroups": [
    {
        "pattern": "./",
        "patternOptions": {
            "nocomment": true,
            "dot": true
        },
        "group": "sibling",
        "position": "before"
    },
    {
        "pattern": ".",
        "patternOptions": {
            "nocomment": true,
            "dot": true
        },
        "group": "parent",
        "position": "before"
    },
    {
        "pattern": "..",
        "patternOptions": {
            "nocomment": true,
            "dot": true
        },
        "group": "parent",
        "position": "before"
    },
    {
        "pattern": "../",
        "patternOptions": {
            "nocomment": true,
            "dot": true
        },
        "group": "parent",
        "position": "before"
    }
]
```

</details>


If `pathGroupsNaturalNewlines` solves everyone's use cases, then it might only be necessary to implement that and not go with `noInsertNewline` after all. @LSH0924 @thany would either of you still have a need for `noInsertNewline` if `pathGroupsNaturalNewlines` exists?

hyperupcall:
Since this PR is several months old (and now ready), I'll recap and provide all the necessary information:

This PR aims to close https://github.com/import-js/eslint-plugin-import/issues/2292, but approaches the solution differently than proposed.

Originally, it was proposed to implement a property for any entry in `pathGroup`, so that ordering would stay _within_ its own group. That is, when `position` was specified, this new property would ensure that a "new group" wouldn't get created and would ensure an extra newline would not be generated. The original PR implemented it this way, but it was too fragile and was difficult to use in practice.

Now, with the current code, the property was added at the top level, so it affects all `pathGroups` entries. This makes it so that every `pathGroup` with a custom `position` property will stay within its own group. As a result, the configuration required to fix this issue is more intuitive.

This PR has been undrafted and includes the relevant tests and documentation

hyperupcall:
Oops, I see the unresolved connections, and I can't appear to resolve them since I rebased over the latest commits and force pushed the result.

Both seem to be outdated, as I now default to `false` in the schema definition and the other comment was about an older property name.

src/rules/order.js
```diff
@@ -502,27 +502,34 @@ function makeNewlinesBetweenReport(context, imported, newlinesBetweenImports) {
...
+  };
   let previousImport = imported[0];
 
   imported.slice(1).forEach(function (currentImport) {
     const emptyLinesBetween = getNumberOfEmptyLinesBetween(currentImport, previousImport);
+    const isStartOfNaturalGroup = getIsStartOfNaturalGroup(currentImport, previousImport);
 
     if (newlinesBetweenImports === 'always'
-        || newlinesBetweenImports === 'always-and-inside-groups') {
+      || newlinesBetweenImports === 'always-and-inside-groups') {
```
ljharb, line 515:
```suggestion
        || newlinesBetweenImports === 'always-and-inside-groups') {
```

src/rules/order.js
```diff
@@ -502,27 +502,34 @@ function makeNewlinesBetweenReport(context, imported, newlinesBetweenImports) {
...
     return linesBetweenImports.filter((line) => !line.trim().length).length;
   };
+  const getIsStartOfNaturalGroup = (currentImport, previousImport) => {
+    return currentImport.rank - 1 >= previousImport.rank;
+  };
   let previousImport = imported[0];
 
   imported.slice(1).forEach(function (currentImport) {
     const emptyLinesBetween = getNumberOfEmptyLinesBetween(currentImport, previousImport);
+    const isStartOfNaturalGroup = getIsStartOfNaturalGroup(currentImport, previousImport);
```
ljharb, line 512: i'm a bit confused by the "natural" terminology. what's natural about this, or unnatural about the alternative?

hyperupcall: I used "natural" because to me, it wasn't intuitive that setting a `position` in a `pathGroup` would appear to automatically create a new group, and add the newline. I assumed that only each `pathGroup` entry would potentially create a new group, and that by setting a `.position`, I would only be relatively positioning the matched imports _within_ that particular group. I felt that `pathGroupsNaturalNewlines: true` changed the behavior so that it felt more natural and expected.

Maybe the `natural` thing is too unclear - possibly the option can be changed to `patternPositionWithinGroup` and the function to something like `getShouldDivideGroup()`?

ljharb: hm - so you're saying that by default, if you *don't* set `position`, a pathGroup doesn't add an additional newline, but if you *do* set it, it does, unless this option is enabled?

hyperupcall: Not quite - I'm saying that by default, if you _don't_ set `position`, a `pathGroup` _does_ add an additional newline. But, if you _do_ set it, then _two_ newlines are added, unless this option is enabled

ljharb: two newlines, like next to each other? if so, that's a bug.

maybe it's just hard to tell in the diff (－‸ლ) could you maybe add two identical test cases, one with this option and one without, so i can see the difference? (or if that's already there, point me to it with my apologies)

ljharb (commented):
Thanks, that clears it up!

Let's name this option (and the resulting variables) "distinctGroup", and default it to "true" - and include a note in the readme (and a "TODO, semver-major" comment in the code) that the next semver-major will change the default of "distinctGroup" from true to false.

src/rules/order.js
hyperupcall: For sure, I've added a test case with identical imports for when this new option is `false`. Here are the permalinks

With [`pathGroupNaturalNewlines: false`](https://github.com/hyperupcall/eslint-plugin-import/blob/053e5716a90639a3cf9f46c528e832d4f82c4004/tests/src/rules/order.js#L928). This is the default case. Here, since `position` is set in `pathGroups[0]`, it creates a second newline after the `import A` line. Likewise, with `pathGroups[1]`, it creates a second newline after the `import C` line.

With [`pathGroupNaturalNewlines: true`](https://github.com/hyperupcall/eslint-plugin-import/blob/053e5716a90639a3cf9f46c528e832d4f82c4004/tests/src/rules/order.js#L958), for the same imports, no second newlines are created. This is where the "natural" comes in - both `'a'` and `'b'` are external imports, just like `c` - and intuitively, they are grouped together, and the `position` property positions them _within_ the `'external'` group, as expected

If this is a bug fix, rather than a feature request, maybe I should remove the option all together and just make it the regular behavior?

ljharb (approved):

PR #2396, opened 2022-03-03
[Fix] `order`: move nested import entries closer to the main import entry

### LINKED ISSUE #2385: [order] move nested import entries closer to the main import entry
I want to keep `next` and `next/*` imports together
```ts
import Core from "next";
import Image from "next/image";
import Link from "next/link"
import Auth from "next-auth";
```

But the plugin forces me to break them with `next-*` imports
```diff
  import Core from "next";
+ import Auth from "next-auth";
  import Image from "next/image";
  import Link from "next/link"
- import Auth from "next-auth";
```

Same issue for the relative imports, the plugin wants me to change the order this way 
```diff
  import e1 from "../businesses";
+ import e3 from "../businesses-users";
  import e2 from "../businesses/components";
- import e3 from "../businesses-users";
```

pri1311:
Closes #2385 

Hey @ljharb! Have added a fix for the linked issue, also added 4 basic tests.

Might need to add more tests, but would like to get the solution reviewed first.

Ran all tests, nothing breaks.

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -1,6 +1,8 @@
 'use strict';
 
 import minimatch from 'minimatch';
+import includes from 'array-includes';
+
 import importType from '../core/importType';
 import isStaticRequire from '../core/staticRequire';
 import docsUrl from '../docsUrl';
@@ -244,16 +246,37 @@ function getSorter(ascending) {
   const multiplier = ascending ? 1 : -1;
 
   return function importsSorter(importA, importB) {
-    let result;
+    let result = 0;
 
-    if (importA < importB) {
-      result = -1;
-    } else if (importA > importB) {
-      result = 1;
+    if (!includes(importA, '/') && !includes(importB, '/')) {
+      if (importA < importB) {
+        result = -1;
+      } else if (importA > importB) {
+        result = 1;
+      } else {
+        result = 0;
+      }
     } else {
-      result = 0;
-    }
+      const A = importA.split('/');
+      const B = importB.split('/');
+      const a = A.length;
+      const b = B.length;
 
+      for (let i = 0; i < Math.min(a, b); i++) {
+        if (A[i] < B[i]) {
+          result = -1;
+          break;
+        } else if (A[i] > B[i]) {
+          result = 1;
+          break;
+        }
+      }
+
+      if (!result && a !== b) {
+        result = a < b ? -1 : 1;
+      }
+    }
+    
     return result * multiplier;
   };
 }
--- a/tests/src/rules/order.js
+++ b/tests/src/rules/order.js
@@ -675,6 +675,47 @@ ruleTester.run('order', rule, {
         alphabetize: { order: 'desc' },
       }],
     }),
+    // Option alphabetize: {order: 'asc'} and move nested import entries closer to the main import entry
+    test({
+      code: `
+        import a from "foo";
+        import c from "foo/bar";
+        import d from "foo/barfoo";
+        import b from "foo-bar";
+      `,
+      options: [{ alphabetize: { order: 'asc' } }],
+    }),
+    // Option alphabetize: {order: 'asc'} and move nested import entries closer to the main import entry
+    test({
+      code: `
+        import a from "foo";
+        import c from "foo/foobar/bar";
+        import d from "foo/foobar/barfoo";
+        import b from "foo-bar";
+      `,
+      options: [{ alphabetize: { order: 'asc' } }],
+    }),
+    // Option alphabetize: {order: 'desc'} and move nested import entries closer to the main import entry
+    test({
+      code: `
+        import b from "foo-bar";
+        import d from "foo/barfoo";
+        import c from "foo/bar";
+        import a from "foo";
+      `,
+      options: [{ alphabetize: { order: 'desc' } }],
+    }),
+    // Option alphabetize: {order: 'desc'} and move nested import entries closer to the main import entry with file names having non-alphanumeric characters.
+    test({
+      code: `
+        import b from "foo-bar";
+        import c from "foo,bar";
+        import d from "foo/barfoo";
+        import a from "foo";`,
+      options: [{
+        alphabetize: { order: 'desc' },
+      }],
+    }),
     // Option alphabetize with newlines-between: {order: 'asc', newlines-between: 'always'}
     test({
       code: `
@@ -2228,6 +2269,27 @@ ruleTester.run('order', rule, {
       }],
       errors: [{
         message: '`bar` import should occur before import of `Bar`',
+      }],
+    }),
+    // Option alphabetize: {order: 'asc'} and move nested import entries closer to the main import entry
+    test({
+      code: `
+        import a from "foo";
+        import b from "foo-bar";
+        import c from "foo/bar";
+        import d from "foo/barfoo";
+      `,
+      options: [{
+        alphabetize: { order: 'asc' },
+      }],
+      output: `
+        import a from "foo";
+        import c from "foo/bar";
+        import d from "foo/barfoo";
+        import b from "foo-bar";
+      `,
+      errors: [{
+        message: '`foo-bar` import should occur after import of `foo/barfoo`', 
       }],
     }),
     // Option alphabetize {order: 'asc': caseInsensitive: true}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Added
 - [`newline-after-import`]: add `considerComments` option ([#2399], thanks [@pri1311])
 
+### Fixed
+- [`order`]: move nested imports closer to main import entry ([#2396], thanks [@pri1311])
+
 ### Changed
 - [Tests] `named`: Run all TypeScript test ([#2427], thanks [@ProdigySim])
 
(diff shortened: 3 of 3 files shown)


src/rules/order.js
```diff
@@ -244,14 +244,35 @@ function getSorter(ascending) {
...
+      
+      for (let i = 0; i < Math.min(a, b); i++) {
+        if (A[i] < B[i]) {
+          result = -1;
+          break;
+        } else if (A[i] > B[i]) {
+          result = 1;
+          break;
+        }
+      }
```
ljharb, line 271: can we not check `A < B` directly, rather than needing to iterate and check each character?

tests/src/rules/order.js
```diff
@@ -675,6 +675,39 @@ ruleTester.run('order', rule, {
...
+    // Option alphabetize: {order: 'desc'} and move nested import entries closer to the main import entry
+    test({
+      code: `
+        import b from "foo-bar";
+        import d from "foo/barfoo";
+        import c from "foo/bar";
+        import a from "foo";`,
+      options: [{
+        alphabetize: { order: 'desc' },
+      }],
```
ljharb, line 709:
```suggestion
        import a from "foo";
      `,
      options: [{ alphabetize: { order: 'desc' } }],
```

tests/src/rules/order.js
```diff
@@ -675,6 +675,39 @@ ruleTester.run('order', rule, {
...
+    // Option alphabetize: {order: 'asc'} and move nested import entries closer to the main import entry
+    test({
+      code: `
+        import a from "foo";
+        import c from "foo/foobar/bar";
+        import d from "foo/foobar/barfoo";
+        import b from "foo-bar";`,
+      options: [{
+        alphabetize: { order: 'asc' },
+      }],
```
ljharb, line 698:
```suggestion
        import b from "foo-bar";
      `,
      options: [{ alphabetize: { order: 'asc' } }],
```

tests/src/rules/order.js
```diff
@@ -675,6 +675,39 @@ ruleTester.run('order', rule, {
...
+    // Option alphabetize: {order: 'asc'} and move nested import entries closer to the main import entry
+    test({
+      code: `
+        import a from "foo";
+        import c from "foo/bar";
+        import d from "foo/barfoo";
+        import b from "foo-bar";`,
+      options: [{
+        alphabetize: { order: 'asc' },
+      }],
```
ljharb, line 687:
```suggestion
        import b from "foo-bar";
      `,
      options: [{ alphabetize: { order: 'asc' } }],
```

src/rules/order.js
```diff
@@ -244,14 +244,35 @@ function getSorter(ascending) {
...
+
+    if (!importA.includes('/') && !importB.includes('/')) {
+      if (importA < importB) {
+        result = -1;
+      } else if (importA > importB) {
+        result = 1;
+      } else {
+        result = 0;
+      }
     } else {
```
ljharb, line 257: is there a reason we can't unconditionally split on slash, and iterate the array?

src/rules/order.js
pri1311: Aaah yes, makes sense. I was worried the length of the array might be a factor for comparison, so I didn't do that. Anyway, changed it now.

pri1311:
hey, have made the required changes, could you please review them?

src/rules/order.js
```diff
@@ -244,14 +244,19 @@ function getSorter(ascending) {
...
-
-    if (importA < importB) {
+    let result = 0;
+
+    const A = importA.split('/');
+    const B = importB.split('/');
+    const a = A.length;
+    const b = B.length;
+    
+    if (A < B) {
```
ljharb, line 254: oh hm, i didn't think about comparing the arrays directly - i think you could certainly compare `[0]` and `[1]` this way, but i wouldn't expect this to work.

I was assuming something like:
```js
if (A.length !== B.length) {
  return B.length - A.length;
}
for (let i = 0; i < A.length) {
  if (A[i] !== B[i]) {
    return a < b ? -1 : 1;
  }
}
return 0;
```

can we add some more test cases? My hope is that the added cases will result in one of these outcomes:
1. your current approach fails them, my approach passes them
1. your current approach passes them, my approach fails them
1. both approaches pass them or fail them

pri1311: > ```js
> if (A.length !== B.length) {
>   return B.length - A.length;
> }
> ```

This will create a problem for paths like `one/two/three/four` and `one/two`



> ```js
> for (let i = 0; i < A.length) {
>   if (A[i] !== B[i]) {
>     return a < b ? -1 : 1;
>   }
> }
> ```

I believe this will work similarly to what I have added. But like you said, I'll work on a few more test cases just to be sure about both approaches.

pri1311: > i didn't think about comparing the arrays directly

Turns out it works the way we would expect. I'd still try and test both approaches just to be sure.
image

pri1311: `import Core from "next";
import Linka from "next/authenticate";
import Linkb from "next/authenticate/link/link2";
import Image from "next/image";
import Linkc from "next/nextauth/link";
import Auth from "next-auth";
`
I believe this is a valid test case. My code parses it correctly. 
Is there any other specific test case you have in mind?

ljharb: I suppose since the `/` is the separator, and since `/` can't appear in any of the individual parts, this works - it's basically the same as `importA < importB`, though, so can we use that to avoid the overhead of stringifying the array?

src/rules/order.js
```diff
@@ -244,14 +244,19 @@ function getSorter(ascending) {
...
       result = -1;
-    } else if (importA > importB) {
+    } else if (A > B) {
       result = 1;
-    } else {
-      result = 0;
+    } else if (a !== b) {
+      result = a < b ? -1 : 1;
```
ljharb, line 259: codecov says this line isn't covered by tests; can we find a test case that covers it?

src/rules/order.js
pri1311: Directly comparing strings doesn't seem to work. Splitting will be necessary I believe.

image

src/rules/order.js
pri1311: Looks like we can get rid of that line of code as `<` operator takes care of length as well.

src/rules/order.js
ljharb: hm, i suspect this is because the array joins with a comma. What about adding some test cases with commas in the parts? like `next/auth` vs `next,auth`

pri1311: > What about adding some test cases with commas in the parts?

I am confused though, commas and a few other characters are not allowed for use for filenames right?

pri1311: > i suspect this is because the array joins with a comma.

Yep definitely creates an issue. Will have go back to the previous method of iterating through the arrays instead of direct comparison.

image

pri1311: Will make the changes and add related changes. Iterating through the array should work fine in my opinion.

pri1311:
Turns out I had reverse back to the first iteration.

`if (!importA.includes('/') && !importB.includes('/')) {` adding this particular piece of line is helping speedup the process.

Rest is basic element-wise comparison.

Have also added a test with commas. Works well in my opinion.

src/rules/order.js
```diff
@@ -244,16 +244,37 @@ function getSorter(ascending) {
...
+    let result = 0;
+
+    if (!importA.includes('/') && !importB.includes('/')) {
+      if (importA < importB) {
+        result = -1;
+      } else if (importA > importB) {
+        result = 1;
+      } else {
+        result = 0;
+      }
```
ljharb, line 256:
```suggestion
      return importB - importA;
```

src/rules/order.js
```diff
@@ -244,16 +244,37 @@ function getSorter(ascending) {
...
     } else {
-      result = 0;
-    }
+      const A = importA.split('/');
+      const B = importB.split('/');
+      const a = A.length;
+      const b = B.length;
+
+      for (let i = 0; i < Math.min(a,b); i++) {
```
ljharb, line 263:
```suggestion
      for (let i = 0; i < Math.min(a, b); i++) {
```

src/rules/order.js
```diff
@@ -244,16 +244,37 @@ function getSorter(ascending) {
...
+        if (A[i] < B[i]) {
+          result = -1;
+          break;
+        } else if (A[i] > B[i]) {
+          result = 1;
+          break;
+        }
```
ljharb, line 270: var result = A[i] - B[i];
if (result) {
  break;
}

pri1311: But we are returning the result, and then its value can be something other than -1/0/1. Will that be fine?

ljharb: Yep! As long as it's a number - zero, negative, or positive - it's a valid comparator.

src/rules/order.js
pri1311: importB and importA are strings, which would return NaN

src/rules/order.js
pri1311: My bad, even A[i] and B[i] are strings.

ljharb: Per discussion, we'll need to keep greater/less comparison here.

ljharb (approved):

c-harding:
This fix causes an unexpected consequence: we are sorting alphabetically with `['sibling','parent']` together in one group. Previously `../myfile` would be sorted before `./myotherfile`, but now they have all been reversed. It would be good if this change did not affect relative paths.

ljharb:
@c-harding see the discussion here: https://github.com/import-js/eslint-plugin-import/issues/2682#issuecomment-1387485013

PR #2398, opened 2022-03-06
[order] Sort type imports

### LINKED ISSUE #2339: `import/order`: alphabetize `import type` above (or below) `import`
> For the version where types are mixed within groups, it would be nice if ^ was actually enforced. We have a bunch of
> 
> ```
> import type A from 'a';
> import A from 'a';
> import B from 'b';
> import type B from 'b';
> ```
> 
> and it would be great to have a more strict ordering if the library is going to start being type-import aware. For reference, [@typescript-eslint/consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md) extracts type imports and places them directly above their corresponding regular imports (though it has no on-going opinion after the initial extraction).
> 
> _Originally posted by @merrywhether in https://github.com/import-js/eslint-plugin-import/issues/2070#issuecomment-841732870_

This would be nice. The main reason we sort imports is to minimize merge conflicts, and I've noticed when sorting with this plugin that it has no opinion on the order between these two lines:

```ts
import { foo } from "./foo";
import type { Foo } from "./foo";
```

Both `simple-import-sort` and, as mentioned, `@typescript-eslint/consistent-type-imports` place `import type` (or `export type`) above `import` (or `export`) when the paths are identical; that seems like it could be a good precedent to follow. 

This might be as easy as concatenating `importKind` within the alphabetize comparator; I tried this with a local monkey patch and it seemed to do the trick.

aaronadamsCA:
Here's a preliminary PR that resolves #2339. It's my first contribution, so please bear with me! 🙂

- Type imports are now sorted before regular imports, because this appears to be the overwhelming ecosystem standard.
- To reduce ambiguity, error messages can now refer to "type import" instead of just "import".
- I included `typeof` support just in case (though I don't know much about Flow).
- Tests were updated to reflect revised behaviour, and two new tests were added.

I also did some opportunistic code cleanup:

- Some test messages seemed to reflect old non-deterministic rule behaviour with "before" vs. "after" errors. Given this is now predictable, I made the tests easier to understand and maintain.
- Coding style toward the end of the test file was all over the place, so I cleaned it up.
- I made a minor performance improvement in `fixOutOfOrder`.

Let me know if there's anything I should undo here, or separate into separate PRs.

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -187,6 +187,16 @@ function canReorderItems(firstNode, secondNode) {
   return true;
 }
 
+function makeImportDescription(node) {
+  if (node.node.importKind === 'type') {
+    return 'type import';
+  }
+  if (node.node.importKind === 'typeof') {
+    return 'typeof import';
+  }
+  return 'import';
+}
+
 function fixOutOfOrder(context, firstNode, secondNode, order) {
   const sourceCode = context.getSourceCode();
 
@@ -204,7 +214,9 @@ function fixOutOfOrder(context, firstNode, secondNode, order) {
     newCode = newCode + '\n';
   }
 
-  const message = `\`${secondNode.displayName}\` import should occur ${order} import of \`${firstNode.displayName}\``;
+  const firstImport = `${makeImportDescription(firstNode)} of \`${firstNode.displayName}\``;
+  const secondImport = `\`${secondNode.displayName}\` ${makeImportDescription(secondNode)}`;
+  const message = `${secondImport} should occur ${order} ${firstImport}`;
 
   if (order === 'before') {
     context.report({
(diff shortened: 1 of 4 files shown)


stropho:
Hmm wondering what is this waiting for :) Anyway, if I was the maintainer, I would appreciate keeping the test coverage on the same level or higher.

A nice non-blocking improvement could be providing and option to sort type imports before/after the regular import. On the other hand that could be a different PR..

src/rules/order.js
```diff
@@ -174,12 +174,20 @@ function canReorderItems(firstNode, secondNode) {
   return true;
 }
 
+function makeImportDescription(node) {
+  if (node.node.importKind === 'type') {
+    return 'type import';
+  } else if (node.node.importKind === 'typeof') {
+    return 'typeof import';
```
ljharb, line 181: Let’s add tests that cover this line

src/rules/order.js
```diff
@@ -240,21 +254,30 @@ function makeOutOfOrderReport(context, imported) {
   reportOutOfOrder(context, imported, outOfOrder, 'before');
 }
 
-function getSorter(ascending) {
-  const multiplier = ascending ? 1 : -1;
+function getSorter(alphabetizeOptions) {
+  const result = alphabetizeOptions.order === 'asc' ? 1 : -1;
+
+  function normalizeCase(value) {
+    return alphabetizeOptions.caseInsensitive ? String(value).toLowerCase() : value;
```
ljharb, line 261: This function can be hoisted to module level

src/rules/order.js
```diff
@@ -240,21 +254,30 @@ function makeOutOfOrderReport(context, imported) {
...
     }
 
-    return result * multiplier;
+    if (importA.node.importKind < importB.node.importKind) {
+      return -result;
+    } else if (importA.node.importKind > importB.node.importKind) {
+      return result;
+    }
+
+    return 0;
```
ljharb, line 280: We need a test to cover this line too

tests/src/rules/order.js
```diff
@@ -2338,11 +2338,11 @@ context('TypeScript', function () {
           // Option alphabetize: {order: 'asc'}
           test({
             code: `
-              import c from 'Bar';
               import type { C } from 'Bar';
+              import c from 'Bar';
```
ljharb, line 2342: Changing existing tests implies it’s a breaking change. Can this test be reverted and a new one added?

tests/src/rules/order.js
```diff
@@ -2548,11 +2548,7 @@ context('TypeScript', function () {
               },
             ],
             errors: [
-              {
-                message: semver.satisfies(eslintPkg.version, '< 3')
-                  ? '`bar` import should occur after import of `Bar`'
-                  : /(`bar` import should occur after import of `Bar`)|(`Bar` import should occur before import of `bar`)/,
-              },
+              { message: '`bar` import should occur after import of `Bar`' },
```
ljharb, line 2551: does this test now report the same error, even in eslint 2?

ljharb (changes requested):
Thanks, this is great!

We need some more test coverage, and it needs to be rebased (there’s a conflict).

ljharb:
Regarding separate PRs, i don’t think that’s necessary, but it would be nice when rebasing to group the “cleanup” stuff into its own commits, separate from the “fix and updated tests” commits.

stropho:
Hey @aaronadamsCA , how are you? What's the status? As this feature is very useful to me right now, I'm willing to help :) If you give me access to your fork, I could move this PR a bit further ;)

ljharb:
@stropho its only been 4 days over a holiday weekend, give it some time :-)

ljharb (approved):

aaronadamsCA:
Sorry, my contributions fell behind for a while due to offline things, so thank you for completing this!

stropho:
And now, finally publishing it after a while would be great :)

PR #2399, opened 2022-03-08
[New] `newline-after-import`: add `considerComments` option

### LINKED ISSUE #2151: `newline-after-import` deosn't take care of comments
The `newline-after-import` rule doesn't hit if the import-statements are immediately followed by comments:

This piece of code is expected to produce an error, but it doesn't:
```ts
import { join } from "path";
/**
 *
 */
console.log();
```

pri1311:
Closes #2151 

Initially, the rule only checked whether some piece of code succeeded the imports. This PR alters the rule to check for any comments on the line following the imports too and enforces the `newline` rule for the same.

However, my confusion here is:
Out of the two code snippets below, which one is valid?
First Block:
```
import path from 'path';
//  some comment
import foo from 'foo';
```

Second Block:
```
import path from 'path';

//  some comment
import foo from 'foo';
```

Right now, the code I have added parses the second block to be correct.

Side note: Ran existing tests as well, nothing breaks.

### DIFF
--- a/src/rules/newline-after-import.js
+++ b/src/rules/newline-after-import.js
@@ -67,6 +67,7 @@ module.exports = {
             'type': 'integer',
             'minimum': 1,
           },
+          'considerComments': { 'type': 'boolean' },
         },
         'additionalProperties': false,
       },
@@ -75,6 +76,7 @@ module.exports = {
   create(context) {
     let level = 0;
     const requireCalls = [];
+    const options = Object.assign({ count: 1, considerComments: false }, context.options[0]);
 
     function checkForNewLine(node, nextNode, type) {
       if (isExportDefaultClass(nextNode) || isExportNameClass(nextNode)) {
@@ -87,7 +89,6 @@ module.exports = {
         nextNode = nextNode.decorators[0];
       }
 
-      const options = context.options[0] || { count: 1 };
       const lineDifference = getLineDifference(node, nextNode);
       const EXPECTED_LINE_DIFFERENCE = options.count + 1;
 
@@ -103,8 +104,32 @@ module.exports = {
             line: node.loc.end.line,
             column,
           },
-          message: `Expected ${options.count} empty line${options.count > 1 ? 's' : ''} \
-after ${type} statement not followed by another ${type}.`,
+          message: `Expected ${options.count} empty line${options.count > 1 ? 's' : ''} after ${type} statement not followed by another ${type}.`,
+          fix: fixer => fixer.insertTextAfter(
+            node,
+            '\n'.repeat(EXPECTED_LINE_DIFFERENCE - lineDifference),
+          ),
+        });
+      }
+    }
+
+    function commentAfterImport(node, nextComment) {
+      const lineDifference = getLineDifference(node, nextComment);
+      const EXPECTED_LINE_DIFFERENCE = options.count + 1;
+
+      if (lineDifference < EXPECTED_LINE_DIFFERENCE) {
+        let column = node.loc.start.column;
+
+        if (node.loc.start.line !== node.loc.end.line) {
+          column = 0;
+        }
+
+        context.report({
+          loc: {
+            line: node.loc.end.line,
+            column,
+          },
+          message: `Expected ${options.count} empty line${options.count > 1 ? 's' : ''} after import statement not followed by another import.`,
           fix: fixer => fixer.insertTextAfter(
             node,
             '\n'.repeat(EXPECTED_LINE_DIFFERENCE - lineDifference),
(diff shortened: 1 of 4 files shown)


src/rules/newline-after-import.js
```diff
@@ -113,6 +113,33 @@ after ${type} statement not followed by another ${type}.`,
...
+          column = 0;
+        }
+
+        context.report({
+          loc: {
+            line: node.loc.end.line,
+            column,
+          },
+          message: `Expected ${options.count} empty line${options.count > 1 ? 's' : ''} \
+after import statement not followed by another import.`,
```
ljharb, line 134:
```suggestion
          message: `Expected ${options.count} empty line${options.count === 1 ? '' : 's'} after import statement not followed by another import.`,
```

ljharb (commented):
I think we need to add this behavior under a new option - and add some test cases for comments that pass as-is.

pri1311:
Hey @ljharb! I have made the required changes, please review them once.

Also, how exactly do I check what line of code is not covered by the tests as suggested by codevco?

ljharb:
The codecov links should lead you to a report you can browse and see which lines aren't covered.

src/rules/newline-after-import.js
```diff
@@ -87,7 +88,7 @@ module.exports = {
         nextNode = nextNode.decorators[0];
       }
 
-      const options = context.options[0] || { count: 1 };
+      const options = context.options[0] && 'count' in context.options[0]? context.options[0] : { count: 1 };
```
ljharb, line 91:
```suggestion
      const options = context.options[0] && 'count' in context.options[0] ? context.options[0] : { count: 1 };
```
what happens when `count` is not provided but `considerComments` is?

src/rules/newline-after-import.js
```diff
@@ -113,6 +114,33 @@ after ${type} statement not followed by another ${type}.`,
       }
     }
 
+    function commentAfterImport(node, nextComment) {
+      const options = context.options[0] && 'count' in context.options[0] ? context.options[0] : { count: 1 };
```
ljharb, line 118: i'd hope we can do this defaulting in only one place

pri1311: Like a global variable? or passing as function parameters?

ljharb: not global, but a rule-level variable - or passing in, yep :-)

pri1311:
> The codecov links should lead you to a report you can browse and see which lines aren't covered.

Aah okay it works now, earlier it was showing unauthorized.

src/rules/newline-after-import.js
```diff
@@ -76,6 +76,20 @@ module.exports = {
...
+      if (context.options[0]) {
+        if ('count' in context.options[0] && 'considerComments' in context.options[0] ) {
+          return context.options[0]
+        } else if ( 'count' in context.options[0] ) {
+          return Object.assign({}, context.options[0], { considerComments: false });
+        } else {
+          return Object.assign({}, context.options[0], { count: 1 });
+        }
+      } 
+      return options = { count: 1, considerComments: false };
```
ljharb: let’s avoid conflating assignment with an expression; it probably shouldn’t return anything if it’s meant to have side effects.

pri1311: Yep, corrected it in the latest commit.

src/rules/newline-after-import.js
```diff
@@ -75,6 +76,20 @@ module.exports = {
...
+        if ('count' in context.options[0] && 'considerComments' in context.options[0] ) {
+          return context.options[0];
+        } else if ( 'count' in context.options[0] ) {
+          return Object.assign({}, context.options[0], { considerComments: false });
+        } else {
+          return Object.assign({}, context.options[0], { count: 1 });
+        }
+      } 
+      return { count: 1, considerComments: false };
+    }
```
ljharb, line 92:
```suggestion
    const options = Object.assign({ count: 1, considerComments: false }, context.options[0]);
```
will this work?

src/rules/newline-after-import.js
```diff
@@ -113,6 +127,32 @@ after ${type} statement not followed by another ${type}.`,
...
 
+    function commentAfterImport(node, nextComment) {
+      const lineDifference = getLineDifference(node, nextComment);
+      const EXPECTED_LINE_DIFFERENCE = options.count + 1;
+
+      if (lineDifference < EXPECTED_LINE_DIFFERENCE) {
+        let column = node.loc.start.column;
+
+        if (node.loc.start.line !== node.loc.end.line) {
+          column = 0;
```
ljharb, line 138: can we add a test case that hits this branch?

pri1311:
Hey @ljharb! Could you check now? I have added the necessary tests and made the suggested changes.

ljharb (approved):

nelson6e65:
When will be this option released? 👀

Issue #2128, opened 2021-06-14
`no-unused-modules` should not halt execution if ignoreExports value is not found


Zamiell:
I have an ESLint template config that contains the following:

```js
    "import/no-unused-modules": [
      "error",
      {
        missingExports: true,
        unusedExports: true,
        ignoreExports: ["src/**/*.d.ts"],
      },
    ],
 ```

This is because I want to use the `no-unused-modules` rule by default in my TypeScript projects, but I also want exclude any TypeScript definitions by default. (They will result in false positives.)

This mostly works great, but a big problem with this is that when I import the template config into a brand new TypeScript project, ESLint will refuse to execute, because no ".d.ts" files exist yet in the repository.

I think a safer default behavior here would be for ESLint to throw a warning and continue executing. Honestly, even showing a warning at all is pretty spurious in the case of a wildcard input, but I guess that could be argued either way.

ljharb:
Unfortunately there's no consistent convention or expectation with star-glob paths that match nothing. Sometimes, what you *want* is what you're asking for - "0 or more" - but sometimes what you want is "1 or more", and you do indeed want an error when nothing matches.

In this case, it's an "ignore" setting, so I do think that it makes sense to be silent when nothing is found.

PR #2403, opened 2022-03-10
[Fix] Drop NoFilesFoundError for ignoreExports

Sleavely:
Silently swallows glob patterns that match 0 files when the pattern is used for ignoring the file.

Fixes https://github.com/import-js/eslint-plugin-import/issues/2128

**PR checklist**

- [ ] `write tests` - I'm not familiar with the test setup in this project and got a bit overwhelmed from the various helpers and abstractions. Would love if someone could help me out here.
- [X] `implement feature/fix bug`
- [ ] `update docs` - N/A?
- [ ] `make a note in change log` - I realized after the first commit I had missed this; in which section and what format should I add it? "Fixes"?

### DIFF
--- a/src/rules/no-unused-modules.js
+++ b/src/rules/no-unused-modules.js
@@ -176,7 +176,15 @@ const resolveFiles = (src, ignoreExports, context) => {
   const srcFileList = listFilesToProcess(src, extensions);
 
   // prepare list of ignored files
-  const ignoredFilesList =  listFilesToProcess(ignoreExports, extensions);
+  let ignoredFilesList = [];
+  try {
+    listFilesToProcess(ignoreExports, extensions);
+  } catch (e) {
+    // pattern for ignored files is allowed to have zero matches
+    if (e.constructor.name !== 'NoFilesFoundError') {
+      throw e;
+    }
+  }
   ignoredFilesList.forEach(({ filename }) => ignoredFiles.add(filename));
 
   // prepare list of source files, don't consider files from node_modules


ljharb (changes requested):
Seems great! Can we get a regression test for this behavior?

Sleavely:
@ljharb I tried writing a regression test that I expected to fail when I commented out my change in the rule but I cant seem to figure out how the RuleRunner works. 

This is what I tried appending to `tests/src/rules/no-unused-modules.js`. Any tips?

```js
describe('do not throw an error when ignored file does not exist', () => {
  ruleTester.run('no-unused-modules', rule, {
    valid: [
      test({
        options: [{
          src: [testFilePath('./no-unused-modules/**/*.js')],
          // In a real scenario this pattern would throw an error
          ignoreExports: [testFilePath('./no-unused-modules/*does-not-exist*.js')],
          missingExports: true,
        }],
        code: 'export const test = true',
        filename: testFilePath('./no-unused-modules/file-ignored-a.js'),
      }),
    ],
    invalid: [],
  });
});
```

ljharb:
`testFilePath` takes an actual file path, not a glob, and is for telling eslint what filename to pretend the code has, not for use in options.

My expectation is that there'd be actual test fixtures on disk, and that the RuleTester snippet would just be one of those files.

Sleavely:
I've gathered that `testFilePath` is just a helper to resolve the path to the fixtures directory. The test itself is a copy of one of the existing cases in `tests/src/rules/no-unused-modules.js` but with a modified `ignoreExports` to trigger a failure. The issue I've run into is that no matter what I pass to `ignoreExports` the test seems to pass.

I do not have bandwidth to pursue this further but hope that someone familiar with the test framework will be able to meet expectations. Feel free to close this pull request at your discretion.

PR #2408, opened 2022-03-22
Add no-shallow-imports rule

joshuason:
Prevent shallow (barrel) imports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`no-named-default`, `no-default-export`, `prefer-default-export`, `no-named-export`, `export`, `named`, `namespace`, `no-unused-modules`]: support arbitrary module namespace names ([#2358], thanks [@sosukesuzuki])
 - [`no-dynamic-require`]: support dynamic import with espree ([#2371], thanks [@sosukesuzuki])
 - [`no-relative-packages`]: add fixer ([#2381], thanks [@forivall])
+- Add [`no-shallow-imports`] rule
 
 ### Fixed
 - [`default`]: `typescript-eslint-parser`: avoid a crash on exporting as namespace (thanks [@ljharb])
--- a/docs/rules/no-shallow-imports.md
+++ b/docs/rules/no-shallow-imports.md
@@ -1,0 +1,44 @@
+# import/no-shallow-imports
+
+Prevent the use of shallow (barrel) imports.
+
+## Examples
+
+Given the following:
+
+```sh
+- /
+    - dir1/
+        - example1.js
+        - index.js
+    - dir2/
+        - example2.js
+        - index.js
+    - tests/
+        - index.test.js
+    - index.js
+```
+
+Examples of **incorrect** code for this rule:
+
+```sh
+//  dir1/example1.js
+
+import example2 from '..';                  // imports from /index.js
+import example2 from '../dir2';             // imports from /dir2/index.js
+import example2 from '../dir2/index';       // same as above but explicit
+```
+
+Examples of **correct** code for this rule:
+
+```sh
+//  dir1/example1.js
+
+import example2 from '../dir2/example2';    // deep import :D
+```
+
+```sh
+//  index.test.js
+
+import * as index from '../index';          // allowed
+```
--- a/src/rules/no-shallow-imports.js
+++ b/src/rules/no-shallow-imports.js
@@ -1,0 +1,104 @@
+import fs from 'fs';
+import minimatch from 'minimatch';
+import docsUrl from '../docsUrl';
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Prevent the use of shallow imports',
+      recommended: true,
+      url: docsUrl('no-shallow-imports'),
+    },
+    schema: [
+      {
+        oneOf: [
+          {
+            type: 'object',
+            properties: {
+              allow: {
+                type: 'array',
+                items: {
+                  type: 'string',
+                },
+              },
+            },
+            additionalProperties: false,
+          },
+        ],
+      },
+    ],
+  },
+  create: context => {
+    return {
+      ImportDeclaration: node => maybeReportShallowImport(node, context),
+    };
+  },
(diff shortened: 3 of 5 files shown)


joshuason:
@ljharb Hey Jordan, just wondering if there's any feedback on this pull request. Cheers :)

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`no-named-default`, `no-default-export`, `prefer-default-export`, `no-named-export`, `export`, `named`, `namespace`, `no-unused-modules`]: support arbitrary module namespace names ([#2358], thanks [@sosukesuzuki])
 - [`no-dynamic-require`]: support dynamic import with espree ([#2371], thanks [@sosukesuzuki])
 - [`no-relative-packages`]: add fixer ([#2381], thanks [@forivall])
+- Add [`no-shallow-imports`] rule
```
ljharb, line 24: this line needs to be moved up to the "unreleased" section

docs/rules/no-shallow-imports.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Examples of **incorrect** code for this rule:
+
+```js
+//  dir1/example1.js
+
+import example2 from '..';                  // imports from /index.js
+import example2 from '../dir2';             // imports from /dir2/index.js
+import example2 from '../dir2/index';       // same as above but explicit
+```
```
ljharb, line 50: Why are these incorrect? What if `dir2/index` has a single default export?

docs/rules/no-shallow-imports.md
```diff
@@ -0,0 +1,64 @@
...
+```js
+//  dir1/example1.js
+
+import example2 from '../dir2/example2';    // deep import :D
+```
+
+```js
+//  index.test.js
+
+import * as index from './index';          // allowed
```
ljharb, line 63: `import *` is a substandard pattern that should never be encouraged.

ljharb (changes requested):
I don't think this should work by looking at the consuming side - i think it should work by looking at the *exporting* side.

joshuason:
> I don't think this should work by looking at the consuming side - i think it should work by looking at the _exporting_ side.

Hmmm, sorry, I don't understand what you mean by this. How would the exporting side preemptively know what's consuming it?

CHANGELOG.md
joshuason: Good point! :)

ljharb:
It wouldn't. The problem is caused by a file re-exporting something.

If you don't re-export anything, it's impossible to have a "barrel" approach unless you put everything in the same file.

joshuason:
Why is file re-exporting a bad thing? Should we be discouraging nested barrel exports?

ljharb:
It's a bad thing because it creates an otherwise nonexistent need for tree-shaking - which due to the nature of JS, can only ever do a half-assed job at deleting unused code.

Yes, nobody should *ever* be using barrel exports, nested or otherwise, and everything should always be deep-imported from the most granular place it can be.

src/rules/no-shallow-imports.js
```diff
@@ -0,0 +1,104 @@
...
+  // If the file does not match the regexps and is a shallow import, then report it
+  if (!doesFileMatchAllowRegexps && isShallowImport(node, context)) {
+    return context.report({
+      node,
+      message: 'Please deep import!',
+    });
+  }
+};
+
+const isCalledIndex = filename => filename.slice(0, 5) === 'index';
```
brunocrosier, line 56:
```suggestion
const isCalledIndex = filename => filename.split('.')[0] === 'index';
```

Think this might be slightly better because otherwise:

```js
isCalledIndex('index-lorem-ipsum.tsx') 
```
will be `true` where as it should be `false`.

ljharb:
```suggestion
const isCalledIndex = filename => path.basename(filename, path.extname(filename)) === 'index';
```

docs/rules/no-shallow-imports.md
brunocrosier: Should `import *` be disallowed by this rule?

src/rules/no-shallow-imports.js
```diff
@@ -0,0 +1,104 @@
...
+
+const isShallowImport = (node, context) => {
+  const filenamePath = context.getFilename();
+  const sourcePath = node.source.value;
+  const segments = sourcePath.split('/');
```
brunocrosier, line 61:
```suggestion
```

src/rules/no-shallow-imports.js
```diff
@@ -0,0 +1,104 @@
...
+  const lastSegment = segments[segments.length - 1];
```
brunocrosier, line 62:
```suggestion
  const lastSegment = path.basename(filenamePath, path.extname(filenamePath));
```

docs/rules/no-shallow-imports.md
ljharb: Not necessarily, but the docs shouldn't suggest it.

brunocrosier: I agree it shouldn't suggest it, but I think it is important for the docs to show that the pattern is still allowed with this rule - as users may assume that it is disallowed or vice versa. Maybe the docs should show it as correct but discouraged / not recommended?

```suggestion
import * as index from './index';          // allowed but discouraged (to disallow this, use the `import/no-namespace` rule)
```

src/rules/no-shallow-imports.js
brunocrosier: @ljharb `path.basename(filename, path.extname(filename)) === 'index'` would evaluate to `false` if the filename is `index.test.js` for example

`filename.split('.')[0] === 'index'` might be safer - what do you think?

docs/rules/no-shallow-imports.md
ljharb: Sure, that's a reasonable compromise.

src/rules/no-shallow-imports.js
ljharb: Sure, that sounds good

PR #2411, opened 2022-03-22
[Docs] `order`: TS code examples should use TS code blocks

MM25Zamanian:


ljharb (changes requested):
Why? All of the examples are JS, none of them are TS.

MM25Zamanian:
In a examples use `import type...` and this for typescript

ljharb:
touché :-) thanks then!

ljharb (approved):

PR #2412, opened 2022-03-23
docs: remove global-install in readme

aladdin-add:
as said in the docs, In eslint v6+, users have to install the plugins locally.
refs: https://eslint.org/docs/user-guide/migrating-to-6.0.0#-plugins-and-shareable-configs-are-no-longer-affected-by-eslints-location


ljharb (approved):
Thanks!

Issue #2414, opened 2022-03-25
`no-unused-modules` reports missing exports for `export *`


robin-hartmann:
The rule `no-unused-modules` incorrectly reports `No exports found` for files which only contain `export * from 'foo'` or `export * as bar from 'buz'` statements, if `missingExports` is set to `true`.

To be precise, all of the following files get reported as `No exports found`:

```ts
export * from 'foo';
```

```ts
export * from 'foo';
export * from 'bar';
```

```ts
export * as foo from 'foo';
```

```ts
export * as foo from 'foo';
export * as bar from 'bar';
```

```ts
export * from 'foo';
export * as bar from 'bar';
```

robin-hartmann:
I've already spent multiple hours trying to fix it myself, but I never managed to get all tests to complete successfully, I only managed to break different ones... but I opened a PR (#2415) to add more tests (which are failing at the moment, of course).

robin-hartmann:
I also just added my current version of the "fixed" tests with only 1 failing test to the PR to gather some input or as a jumping-off point for someone else.

PR #2415, opened 2022-03-25
[Tests]: `no-unused-modules`: add tests for star export

robin-hartmann:
See issue #2414.


qpwo:
Very interested in this. If you could also follow imports/exports through tsconfig path aliases, then this would be a universal dead-code-detection system.

PR #2419, opened 2022-03-30
[Refactor] `no-cycle`: Add per-run caching of traversed paths

Nokel81:
- This leads to about a 5x speed up

Signed-off-by: Sebastian Malton <sebastian@malton.name>

### DIFF
--- a/src/rules/no-cycle.js
+++ b/src/rules/no-cycle.js
@@ -9,7 +9,8 @@ import { isExternalModule } from '../core/importType';
 import moduleVisitor, { makeOptionsSchema } from 'eslint-module-utils/moduleVisitor';
 import docsUrl from '../docsUrl';
 
-// todo: cache cycles / deep relationships for faster repeat evaluation
+const traversed = new Set();
+
 module.exports = {
   meta: {
     type: 'suggestion',
@@ -87,7 +88,6 @@ module.exports = {
       }
 
       const untraversed = [{ mget: () => imported, route:[] }];
-      const traversed = new Set();
       function detectCycle({ mget, route }) {
         const m = mget();
         if (m == null) return;
@@ -101,7 +101,7 @@ module.exports = {
             // Ignore only type imports
             !isOnlyImportingTypes,
           );
-          
+
           /*
           If cyclic dependency is allowed via dynamic import, skip checking if any module is imported dynamically
           */
@@ -138,7 +138,11 @@ module.exports = {
       }
     }
 
-    return moduleVisitor(checkSourceValue, context.options[0]);
+    return Object.assign(moduleVisitor(checkSourceValue, context.options[0]), {
+      'Program:exit': () => {
+        traversed.clear();
+      },
+    });
   },
 };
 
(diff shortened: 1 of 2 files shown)


Nokel81:
I don't think that the CI failure is related to this PR...

fforres:
@Nokel81 @toerndev merged elsewhere? in this project?

toerndev:
@fforres Oh no I'm sorry, I must be sleep deprived. I thought I saw the exact same lines by the same author on the main branch, and somehow forgot that I had just cherry-picked it myself to test it...

But this fix seems like low-hanging fruit, assuming that it's acceptable to do static caching inside the rule.

pastelsky:
@ljharb Do you think you can give this one a look? `no-cycle` is easily the long pole on multiple large projects, and having a more persistent cache could help with that.

src/rules/no-cycle.js
```diff
@@ -40,8 +42,8 @@ module.exports = {
       },
     })],
   },
-
-  create(context) {
+  _traversed: traversed,
```
ljharb, line 45: You can see this on this line, as well as…

tests/src/rules/no-cycle.js
```diff
@@ -17,6 +18,10 @@ const testVersion = (specifier, t) => _testVersion(specifier, () => Object.assig
 
 const testDialects = ['es6'];
 
+beforeEach(() => {
+  rule._traversed.clear();
```
ljharb, line 22: this one, where without this hack, the *tests themselves* will fail, because the cache is lasting too long. This is the same failure that consumers will see.

One possible alternative is to add a `Program:exit` visitor, that clears the Set (after removing this line) - but someone would have to test that change and see if it still provided the same speedup.

ljharb (changes requested):
It looks like what this PR does is to move the cache from being done per *run* to being done per *application lifetime* - this will cause lots of problems with integrations in editors, or in anything where eslint is long-running.

Nokel81:
> It looks like what this PR does is to move the cache from being done per linting run to being done per application lifetime - this will cause lots of problems with integrations in editors, or in anything where eslint is long-running.

Yes that is the case. I guess in cases where eslint is longer running this would be incorrect. The reason why I came across this solution to the speed problem is that when doing a manual invocation of `eslint` (either via a terminal or CI) then there is no need to re-compute the entire graph traversal for every single file.

ljharb:
I totally agree - could you try adding a Program:exit visitor and clearing the Set there? Then you wouldn't need to expose it with `_traverse` nor clear it in the tests.

Nokel81:
Will try

Nokel81:
Passed the tests, need to check if it is actually faster

ljharb (approved):
This seems much better - let's give it a shot.

pastelsky:
@ljharb could you cut a release for `eslint-plugin-import`, it looks like it's been a while, and the latest master runs quite a bit faster than the latest version (2.26.0) on a large project for me. Though I haven't been able to test yet if the improvement is due to this PR in particular or other unpublished changes from the past.

PR #2421, opened 2022-04-01
[Deps] update tsconfig-paths

Alkarex:
Security fix.
See https://github.com/dividab/tsconfig-paths/pull/197
Weaknesses CWE-1321
CVE ID CVE-2021-44906

> Minimist <=1.2.5 is vulnerable to Prototype Pollution via file index.js, function setKey() (lines 69-95).

(This currently triggers cascade alerts in many packages using eslint-plugin-import)

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,6 @@
     "minimatch": "^3.1.2",
     "object.values": "^1.1.5",
     "resolve": "^1.22.0",
-    "tsconfig-paths": "^3.13.0"
+    "tsconfig-paths": "^3.14.1"
   }
 }


ljharb:
This is fine, but also 100% unnecessary, because it's using a semver range. Reinstall eslint-plugin-import and you'll get an updated copy of all its transitive deps.

rjhilgefort:
> This is fine, but also 100% unnecessary, because it's using a semver range. Reinstall eslint-plugin-import and you'll get an updated copy of all its transitive deps.

I removed `eslint-plugin-import`, yarn installed (minimist `1.2.0` went away), installed `eslint-plugin-import` and the bad version of `minimist` was back as a dep of `json5@1.0.1`. Reinstalling doesn't seem to work.

ljharb:
I can't speak to yarn, but npm would properly update it since json5 uses ^ (https://unpkg.com/browse/json5@1.0.1/package.json).

ljharb:
This was updated in 0e80ee37c, so, closing this PR.

PR #2424, opened 2022-04-05
[Docs] `no-useless-path-segments`: fix paths

s-h-a-d-o-w:
Really not much to say here. Somebody simply overlooked that `../` refers to the parent directory. 🙂
(Had me worried about whether this rule works correctly. Which it does.)

I also added the root `index.js` because of the `import "."` further below.


ljharb (approved):
Thanks!

PR #2428, opened 2022-04-07
[Test] import/named should support export = assignment of object in TS

ProdigySim:
This pattern is valid typescript, but `import/named` can't find resolve the imports. Relevant issue: #1984 

[Here's a demo showing that it should work in TS](https://codesandbox.io/s/ts-export-assignment-demo-l2mym?file=/src/index.ts)

This test currently does not run, as no TS tests in `named.js` are currently running: waiting for #2427 

Once the test is wired up, it fails with:

```
  1) TypeScript named [TypeScript] valid import { FooBar } from './typescript-export-assign-object':

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Should have no errors but had 1: [
  {
    ruleId: 'named [TypeScript]',
    severity: 1,
    message: "FooBar not found in './typescript-export-assign-object'",
    line: 1,
    column: 10,
    nodeType: 'Identifier',
    endLine: 1,
    endColumn: 16
  }
]
      + expected - actual

      -1
      +0
```

Issue #1984, opened 2021-02-01
Exports assignments not being picked up by no-unused-modules


paleite:
First of all, thanks for a great plugin! This has aided our team and saved us HOURS of debugging and it makes our imports and exports really consistent 🙂  
I'm using export assignments (`export = `, see https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require ) in some of my TypeScript modules and `no-unused-modules` claims the modules aren't exporting anything.

Background, I'm writing my ESLint config in TypeScript, and want the compiled TypeScript code to say `module.exports = eslintConfig`

My ESLint config for the rule looks like this:
`"import/no-unused-modules": ["error", { missingExports: true }]`

And the module looks like this:

**`index.ts`**:

```
const eslintConfig = {
  rules: {
    "import/no-unused-modules": ["error", { missingExports: true }],
  },
};

export = eslintConfig;
```

`no-unused-modules` with `missingExports` set to `true` will claim `index.ts` does not export a module.

ljharb:
I think #1974 may have fixed this, but is not yet released.

paleite:
Interesting, but I'm not sure it's the same.

I looked through the PR, but I couldn't find anything regarding exports in there, only regarding imports.

Either way, I'll make sure to monitor the status of this. Thanks again 🙂

ljharb:
It would be great if you wanted to create a PR with test cases - that would answer it immediately :-)

ProdigySim:
I think this may be fixed, I tried to create a test case directly for it and it passes:
https://github.com/ProdigySim/eslint-plugin-import/commit/308e8ca29623d9868919b66fb7e944d9fd6c516b

I am still having issues with exports of this style in my project, but it may be due to misconfiguration or outdated versions. I'll try to update here if I fix it.

ljharb:
Thanks!

@paleite can you confirm? I'll reopen if it's still a problem for you.

ProdigySim:
@ljharb actually, I think none of the TypeScript tests in `named.js` are wired up to actually run... it looks like the tests are collected in `invalid`/`valid` arrays but they are not run at all. So I was accidentally not running my test.

I pushed up another commit to run the tests and now my test case fails + another TS test is actually failing already:

https://github.com/import-js/eslint-plugin-import/commit/1e4d413f7ce74345e424c74196f54c35d7dcde53

PR #2432, opened 2022-04-12
[Tests] `named`: Add direct test for `export =` assignment in TS

ProdigySim:
This pattern is valid typescript, but `import/named` can't find resolve the imports. Relevant issue: #1984 

[Here's a demo showing that it should work in TS](https://codesandbox.io/s/ts-export-assignment-demo-l2mym?file=/src/index.ts)

Test fails with

```
  1) TypeScript named [TypeScript] valid import { FooBar } from './typescript-export-assign-object':

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Should have no errors but had 1: [
  {
    ruleId: 'named [TypeScript]',
    severity: 1,
    message: "FooBar not found in './typescript-export-assign-object'",
    line: 1,
    column: 10,
    nodeType: 'Identifier',
    endLine: 1,
    endColumn: 16
  }
]
      + expected - actual

      -1
      +0
```

Reopened from #2428

### DIFF
--- a/tests/files/typescript-export-assign-object/index.ts
+++ b/tests/files/typescript-export-assign-object/index.ts
@@ -1,0 +1,5 @@
+const someObj = {
+  FooBar: 12,
+};
+
+export = someObj;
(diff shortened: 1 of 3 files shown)


ljharb:
I've rebased this, and looking over it, it's actually correct - `export =` syntax creates only a default export, at least when esModuleInterop is on (without which, TS's module system is broken anyways). You can't named-import from a module that uses that.

ljharb (approved):

PR #2435, opened 2022-04-12
Change the way existence of the path is checked. 

lerayne:
Previously it was reading the source dir to get the list of its contents and then checking if the parsedPath.base is in the list. Not only this is not optimal, but also can cause an access issue if the script executor doesn't have access to project's parent folder, for example in CI/CD env.

### DIFF
--- a/utils/resolve.js
+++ b/utils/resolve.js
@@ -68,8 +68,7 @@ exports.fileExistsWithCaseSync = function fileExistsWithCaseSync(filepath, cache
   if (dir === '' || parsedPath.root === filepath) {
     result = true;
   } else {
-    const filenames = fs.readdirSync(dir);
-    if (filenames.indexOf(parsedPath.base) === -1) {
+    if (!fs.existsSync(path.join(dir, parsedPath.base))) {
       result = false;
     } else {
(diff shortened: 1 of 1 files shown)


utils/resolve.js
```diff
@@ -68,8 +68,7 @@ exports.fileExistsWithCaseSync = function fileExistsWithCaseSync(filepath, cache
   if (dir === '' || parsedPath.root === filepath) {
     result = true;
   } else {
-    const filenames = fs.readdirSync(dir);
-    if (filenames.indexOf(parsedPath.base) === -1) {
+    if (!fs.existsSync(path.join(dir, parsedPath.base))) {
       result = false;
     } else {
       result = fileExistsWithCaseSync(dir, cacheSettings, strict);
```
ljharb, line 74: it seems like on a case-insensitive system, this will still trigger the permissions issue you had?

ljharb (commented):
This appears to have been added all the way back in fc4f99ecd.

utils/resolve.js
lerayne: I will check

ljharb: any update?

PR #2436, opened 2022-04-14
[Fix] `no-unresolved`, `extensions`: ignore type-only exports

Lukas-Kullmann:
In #2220 and #2270, a fix was released that ignored type-only imports for the rules `no-unresolved` and `extensions` respectively.

However type-only **ex**ports are not ignored.

The rules could always be tricked by this code anyway:

```typescript
import type { A } from 'non-existing';

export { A };
```

the code above is equivalent to

```typescript
export type { A } from 'non-existing`;
```

but the second one would throw an error while the first one does not.

This PR fixes this inconsistencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`no-cycle`]: add ExportNamedDeclaration statements to dependencies ([#2511], thanks [@BenoitZugmeyer])
 - [`dynamic-import-chunkname`]: prevent false report on a valid webpack magic comment ([#2330], thanks [@mhmadhamster])
 - [`export`]: do not error on TS export overloads ([#1590], thanks [@ljharb])
+- [`no-unresolved`], [`extensions`]: ignore type only exports ([#2436], thanks [@Lukas-Kullmann])
 
 ### Changed
 - [Tests] [`named`]: Run all TypeScript test ([#2427], thanks [@ProdigySim])
@@ -1014,6 +1015,7 @@ for info on changes for earlier releases.
 [#2466]: https://github.com/import-js/eslint-plugin-import/pull/2466
 [#2440]: https://github.com/import-js/eslint-plugin-import/pull/2440
 [#2438]: https://github.com/import-js/eslint-plugin-import/pull/2438
+[#2436]: https://github.com/import-js/eslint-plugin-import/pull/2436
 [#2427]: https://github.com/import-js/eslint-plugin-import/pull/2427
 [#2424]: https://github.com/import-js/eslint-plugin-import/pull/2424
 [#2419]: https://github.com/import-js/eslint-plugin-import/pull/2419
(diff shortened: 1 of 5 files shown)


src/rules/extensions.js
```diff
@@ -164,8 +164,8 @@ module.exports = {
       ) || isScoped(importPath);
 
       if (!extension || !importPath.endsWith(`.${extension}`)) {
-        // ignore type-only imports
-        if (node.importKind === 'type') return;
+        // ignore type-only imports and exports
+        if (node.importKind === 'type' || node.importKind === 'type') return;
```
ljharb, line 168: These are the same condition.

Could you add a test that fails without this suggestion:
```suggestion
        if (node.importKind === 'type' || node.exportKind === 'type') return;
```

ljharb (changes requested):
Looks great overall!

src/rules/extensions.js
Lukas-Kullmann: There were already tests, but I screwed up the conditional here. Thanks for pointing that out!

Is it ok now?

ljharb: I didn't check; did the tests fail with the typo? if so, then it's good now :-) (altho there's still a bunch of failing tests)

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Changed
 - [Tests] `named`: Run all TypeScript test ([#2427], thanks [@ProdigySim])
 
+### Fixed
+- [`no-unresolved`, `extensions`]: ignore type only exports ([#2436], thanks [@Lukas-Kullmann])
+
```
ljharb, line 17:
```suggestion
### Fixed
- [`no-unresolved`, `extensions`]: ignore type only exports ([#2436], thanks [@Lukas-Kullmann])

### Changed
- [Tests] `named`: Run all TypeScript test ([#2427], thanks [@ProdigySim])

```

tests/src/rules/extensions.js
```diff
@@ -613,6 +613,14 @@ describe('TypeScript', () => {
             ],
             parser,
           }),
+          test({
+            code: 'export type T from "./typescript-declare";',
```
wsuwt, line 617: Hi guys, I think the code here should be named export instead in order to be valid. Test failed because parser couldn't understand what exactly it is trying to export.

```suggestion
            code: 'export type { T } from "./typescript-declare";',
```

tests/src/rules/extensions.js
```diff
@@ -624,6 +632,15 @@ describe('TypeScript', () => {
             ],
             parser,
           }),
+          test({
+            code: 'export T from "./typescript-declare";',
```
wsuwt, line 636: Also, an invalid statement here will throw `Parsing error: Declaration or statement expected`. I am suggesting a named export here as well.

```suggestion
            code: 'export { T } from "./typescript-declare";',
```

tests/src/rules/extensions.js
wsuwt: @ljharb @Lukas-Kullmann Could you guys take a look at this ? I think these changes should make this PR ready.

tests/src/rules/extensions.js
ljharb: indeed, https://github.com/tc39/proposal-export-default-from is still a stage 1 proposal

ljharb (approved):
Seems good

PR #2438, opened 2022-04-21
[Tests] `no-cycle`: add passing test cases

georeith:
This adds a (currently failing) test case for the issue: https://github.com/import-js/eslint-plugin-import/issues/2439

### DIFF
--- a/tests/src/rules/no-cycle.js
+++ b/tests/src/rules/no-cycle.js
@@ -267,5 +267,23 @@ ruleTester.run('no-cycle', rule, {
       parser: parsers.BABEL_OLD,
       errors: [error(`Dependency cycle via ./flow-types-depth-two:4=>./es6/depth-one:1`)],
     }),
+    test({
+      code: 'import { foo } from "./intermediate-ignore"',
+      errors: [
+        {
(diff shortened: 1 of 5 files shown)


tests/src/rules/no-cycle.js
```diff
@@ -30,6 +30,10 @@ ruleTester.run('no-cycle', rule, {
     test({ code: 'var foo = require("@scope/foo")' }),
     test({ code: 'var bar = require("./bar/index")' }),
     test({ code: 'var bar = require("./bar")' }),
+    // correctly passes
+    test({ code: 'import { foo } from "./ignore.js"' }),
+    // incorrectly fails
+    test({ code: 'import { foo } from "./intermediate-ignore.js"' }),
```
ljharb, line 36: actually it turns out BOTH of these should be failing - since the tests have `depth-zero` configured as the "current filename" for the `code`, BOTH of them have a cycle, so BOTH of them should be failing!

ljharb (approved):
I don't think this fixes the linked issue, but the tests are now correctly passing (because the cases are both detected as cycles)

PR #2444, opened 2022-04-28
[meta] replace `git.io` link in comments with the original URL

### LINKED ISSUE #2443: Git.io deprecation notice
Hello, I've found that some of your files in this repo is still using https://git.io/, and this service will be discontinued in the future. (ref: [Git.io deprecation](https://github.blog/changelog/2022-04-25-git-io-deprecation/))
Theses files are: utils/moduleVisitor.js

> This is an automated message from NPMMirror(https://github.com/npmmirror), a bot developed by Nova Kwok(https://github.com/n0vad3v).

liby:
Fix #2443


ljharb (approved):
Thanks! I'll rebase and remove the word "chore", since i find that to be an incredibly toxic term to use for something that is actually meaningful and important.

Issue #1810, opened 2020-06-08
`no-unresolved` is not aware of `exports` definition in `package.json`


ignatiusreza:
From [this announcement](https://medium.com/@nodejs/announcing-core-node-js-support-for-ecmascript-modules-c5d6dc29b663)

> Package entry points and the package.json “exports” field
> 
> There are now two fields that can define entry points for a package: “main” and“exports”. The “main” field is supported in all versions of Node.js, but its capabilities are limited: it only defines the main entry point of the package. A new package.json field “exports” can also define the main entry point, along with subpaths. It also provides encapsulation, where only the paths explicitly defined in “exports” are available for importing. “exports” applies to both CommonJS and ES module packages, whether used via require or import.

Might need to be fix in [`resolve`](https://github.com/browserify/resolve/pull/224) package..

ljharb:
Indeed, that’s where it needs to be fixed.

However, for back compact, you should have a file at the location the exports key would otherwise resolve to anyways - so even when it’s fixed in resolve, you’d still want it fixed now.

plus-ultra-git:
When to fix?

inlined:
Will this be resolved any time soon? We're going to have to yank `no-unresolved` from our Google code samples because it can no longer handle our SDKs.

ljharb:
@inlined why would your Google SDKs be using `exports`, without `main`, in a way that's not backwards-compatible??

inlined:
We _are_ using `main`. We also use `exports` for submodule paths so we can omit the `lib` folder generated by TypeScript.

For example, `./lib/v2/providers/https.js` is exported at `./v2/https`. See github.com/firebase/firebase-functions for more info.

ljharb:
Right - but that's not backwards-compatible to pre-exports node. However, if you created an actual `./v2/https.js` file that re-exported that provider, then not only would your package be backwards-compatible, but you wouldn't be blocked on `resolve` getting `exports` support.

inlined:
We control our execution environment and do not need to be backwards compatible to versions that don't support exports.

ljharb:
@inlined i'm not "of the opinion that exports should not use renaming features", i'm of the opinion that you have a really simple workaround while you wait.

None of my "opinions" are delaying `exports` support - the work is difficult and nobody else is doing it, and I have limited time. It will be done eventually.

ljharb:
[...]

PR #2447, opened 2022-05-04
[Deps] upgrade tsconfig-paths to ^4.0.0

F3n67u:
This deps upgrade will introduce a breaking change, because `tsconfig-paths` upgrade its `json5` version to v2 which does not support node 4. see https://github.com/dividab/tsconfig-paths/pull/198

If we want to upgrade `tsconfig-paths` to the latest, I think we could do it on the next major release and drop the node v4 support. But I am sure if it is accepted. @ljharb what do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,6 @@ matrix:
   - os: osx
     env: ESLINT_VERSION=3
     node_js: 6
-  - os: osx
-    env: ESLINT_VERSION=2
-    node_js: 4
 
   fast_finish: true
 
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.26.0",
   "description": "Import with sanity.",
   "engines": {
-    "node": ">=4"
+    "node": ">=6"
   },
   "main": "lib/index.js",
   "directories": {
@@ -111,6 +111,6 @@
     "minimatch": "^3.1.2",
     "object.values": "^1.1.5",
     "resolve": "^1.22.0",
-    "tsconfig-paths": "^3.14.1"
+    "tsconfig-paths": "^4.0.0"
   }
 }


MichaelDeBoey:
@ljharb It would be nice to be able to use `tsconfig-paths`' new features (especially supporting `jsconfig.json`) though, so how would you propose we should go forward in getting that done?

ljharb:
@MichaelDeBoey what tooling besides TS supports `jsconfig.json`?

MichaelDeBoey:
@ljharb We're supporting it in the Remix repo to have a `jsconfig.json` file.
Updating the codebase to use `tsconfig-paths` v4 to do this instead of our custom solution breaks our tests (see https://github.com/remix-run/remix/pull/3074) because `eslint-plugin-import` & `eslint-import-resolver-typescript` aren't updated (yet).

https://github.com/alexgorbatchev/eslint-import-resolver-typescript/pull/104 was already merged & dropped supporting Node v4, I think it's fair for `eslint-plugin-import` to do the same, as Node v4 is really old and really long EOL.

If we merge this PR, this would be a major release.
People who still want to use Node v4 (they should have updated a long time ago imo) can still use the current major version if they really want to stick with Node v4.

nied:
v4 of `tsconfig-paths` also updates json5 to 2.2.1 which drops a version of `minimist` that contains a security issue: https://github.com/dividab/tsconfig-paths/pull/198. We're currently updating all our dependencies to remove the vulnerable version of `minimist` and this package is on the list. Let me know if we can help in any way to get this merged!

ljharb:
@nied a) that "security issue" is not one, because it's a self-attack; b) tsconfig-paths v3 depends via semver range on a version of minimist that is NOT vulnerable, so that motivation is nonexistent.

ljharb:
@kentcdodds you could ask the same question to the tsconfig-paths maintainers who opted to do a breaking change instead of adding the feature you need in a v3 minor ¯\\\_(ツ)\_/¯

MichaelDeBoey:
@ljharb It could indeed be in a major release, but they opted not to do it for the reasons I said above:
- Node v4 is EOL a long time, so people should have updated a long time ago
- People who still want to use older Node versions could stay at the older version of `tsconfig-paths`

ljharb:
> People who still want to use older Node versions could stay at the older version of tsconfig-paths

no, we (this plugin) can't, because we want the new feature and also want to use node 4.

It is wildly inappropriate and harmful for project maintainers to try to "push" people to upgrade. Folks who haven't upgraded can't; nobody needs the push.

kentcdodds:
Can you help me understand who specifically would be harmed by dropping Node v4 support in this package? They would have to both:
1. Be using Node v4
2. Need to upgrade their version of eslint-plugin-import

Can you give examples of specific people who fall into this category? I can give you examples of specific people who are harmed by not pursuing this. My personal expectation is that anyone on Node v4 is already locked-in on their dependencies and not upgrading anything anyway.

ljharb:
There seems to be multiple issues, needs, and philosophical concerns here, so let me type a mini-essay to summarize:

## Breaking changes in packages are harmful for the ecosystem
In general, when a package (something that can be used transitively, which notably does NOT really include eslint plugins) has a breaking change, that causes a number of very harmful and costly downstream effects. It means that consumers have to choose whether to make their own breaking change, or be stuck on an older major; it means that if there's ever a CVE on that package, consumers transitively depending on the old major are far more likely to be stuck on it.

Additionally, there's a number of popular packages/maintainers that do *unnecessary* major bumps - iow, the major bump is correct, but the actual breaking thing was unnecessary. For example, dropping a node version or using new syntax when there's not a strong reason to do so. In many cases, the harm and cost of these unnecessary major bumps is worse than if the package had never existed - in other words, it can cancel out the lifetime benefit of the package itself.

As such, I have developed an intense avoidance for breaking changes in all of my packages, because I don't want to inflict hundreds of millions of dollars of person-hour cost on the entire industry unnecessarily.

## dropping platforms does not force change, and change should not be forced

Dropping support for an EOL version of node, or a browser, doesn't help anybody. Nobody stays on an old version of node or a browser by choice, they do it because they *have no other choice*. The reasons might be inexperience, or lack of resources, or all sorts of things - but the reasons are irrelevant.

Punishing these users doesn't help them upgrade, it just disrupts their roadmap by forcing them to deprioritize things that might be what's keeping their business alive, and instead to upgrade. Whether that's a better outcome or not is *not your place to decide*. If you need to drop a platform because of a compelling reason, great! That's what semver-major is for. Just please don't pretend there's some kind of positive effect - it's a large negative effect, that is hopefully outweighed by whatever motivation you have for the major bump.

## People on old platforms don't update dependencies

This expectation is simply false; they update everything they can, it's just that deps that do the hostile things described above prevent them from doing so.

## "they can just use an old version"

No, they can't, for a ton of reasons: vulnerabilities, often in transitive dependencies; bugfixes; new features. Forcing them to update node just to get a new feature in "not node" makes no sense, unless that new feature requires functionality only present in newer node.

ljharb:
Now, let's speak to this specific PR.

If a feature is compelling enough, especially if it actually requires a newer version of node (this one surely does not, except that tsconfig-paths dropped support), then it might be *worth* the breaking change. However, jsconfig.json does not feel compelling to me. TS users use tsconfig.json; .editorconfig is a universal standard; but jsconfig.json is an upstart format that is not widely supported. Tools supporting a new format _can_ be helpful when it's low-cost, but as described above, a major change is costly. Separately, tools supporting minimally used formats can be harmful: Webpack supported "concord modules" or something for years, adding tons of bloat and complexity for every one of its users, for a module format that virtually zero people ever cared about.

Can you help me understand what about jsconfig.json is a compelling feature that a non-TS user would want? (note the dep is _ts_ config paths, so it should only be relevant for TS users)

nied:
> it means that if there's ever a CVE on that package, consumers transitively depending on the old major are far more likely to be stuck on it

You literally have an open CVE in https://github.com/advisories/GHSA-xvch-5gv4-984h as long as you're not merging this.

By staying on Node 4, people are vulnerable to a lot of security issues that have been fixed in later Node versions (look at the list at https://www.cvedetails.com/vulnerability-list/vendor_id-12113/Nodejs.html) so keeping Node 4 support to be able to fix CVEs is kind of a moot point imho - they're already at risk. You could argue that another issue will not make the situation better, and that's fair, but this way you are actively hurting people who actually are trying to stay on top of security issues, dependabot alerts etc. I get that you want to keep backwards compatibility but at this point it's getting a bit absurd.

If you care so much about the consumers of your package, surely you don't want open CVEs in your repository, right?

I see two ways of moving forward here:
1. Merge this PR
2. If you desperately want to keep Node 4 support, decide that you want to drop the `tsconfig-paths` dependency altogether.

ljharb:
@nied that CVE isn’t one; it’s a self-attack when you craft command input to a CLI - one we’re not using at all. That CVE can remain open forever and it doesn’t matter.

Thus, i never need to drop tsconfig-paths 3, nor upgrade it, because this CVE is invalid (like most CVEs in the JS ecosystem).

kentcdodds:
Looks like we're not going to convince you. Bummer.

/me unsubscribes 🤷‍♂️

ljharb:
@kentcdodds certainly not if you're unwilling to answer https://github.com/import-js/eslint-plugin-import/pull/2447#issuecomment-1119864544 ¯\\\_(ツ)\_/¯

cmawhorter:
> that CVE isn’t one; it’s a self-attack when you craft command input to a CLI - one we’re not using at all [...]

i hear you that it's not a high priority, but i'd disagree that it's a non-issue.  vuln code is vuln code and even if it's not immediately/publicly an problem doesn't mean it can't be used to escalate privileges/whatever somewhere, somehow, on some misconfigured machine out there probably holding all our medical records and built by some high school intern.

> Dropping support for an EOL version of node, or a browser, doesn't help anybody

this. node's EOL has caused me pain in the ecosystem via strict adherence by maintainers. but... there has to be a line _somewhere_ for when support for old node versions is dropped.  i'd ask -- should users of newer node versions get a slap in the name of older support?

ljharb:
It’s not vulnerable code, it’s a CLI arg parser that isn’t invoked with user input by this plugin. There exists no possible exploit here. So yes, it does mean it can never be used for that - at least with the same likelihood as all other code that doesn’t yet have a known vulnerability. This reaction to the mere existence of a CVE is just the FUD caused by the broken CVE system.

Why does there have to be a line somewhere?

cmawhorter:
> It’s not vulnerable code, it’s a CLI arg parser that isn’t invoked with user input by this plugin

if this package was the only dependency used i'd agree that this is def not critical, but consider an app that has two dependencies:
- one is this package, with the vulnerable minimist version (as you've said, minimist is not used)
- but, another uses minimist@^1.x, but _does_ use it

since the problem minimist version satisfies both dependencies, npm can dedupe it into the vulnerable one, and your app is hit.

so just because _this package_ isn't an issue doesn't mean that it's problems won't bleed elsewhere.  

> Why does there have to be a line somewhere?

does eslint-plugin-import support node v0.10? there is always a line _somewhere_, and it's figuring out where that line should be.  you've stated your reasoning and i'm certainly not going to fork this, so i'll just thank you for your work and leave it at that.

ljharb:
In that case you’d just depend directly on a safe range of minimist and it’d work - or, that other package would update it.

It technically should support node 0.10 because eslint 2 does, and we support eslint 2, but we’ve never prioritized the work to do it.

cmawhorter:
> or, that other package would update it

but what if that _other package_ is part of a supply chain attack that is
designed to work in conjunction with this package?

ljharb:
You can continue contriving scenarios, but it’s never an unrelated package’s problem to fix. npm’s deduping doesn’t ever make this plugin at all responsible for another package’s vulnerability.

ljharb:
At this point, there’s only three viable paths forward:
1. tsconfig-paths backports a fix in v3, which they really should be doing regardless
1. this PR rots and we stay on v3 forever, or until something is _actually_ worth a breaking change
1. someone can explain why jsconfig.json is a compelling feature, as opposed to a harmful xkcd 927

01taylop:
So will this change never be merged? I am getting frustrated because we use this dependency in an internal linting tool which is used by hundreds of projects within our company. That's a lot of Dependabot alerts to dismiss and as much as I enjoy this plugin and am grateful for it, am feeling quite let down.

ljharb:
@01taylop nobody's tried to answer this question yet: https://github.com/import-js/eslint-plugin-import/pull/2447#issuecomment-1119864544

ljharb:
Thanks, that's helpful. Both of those are antipatterns imo, but they are likely to be common needs, and unfortunately at a certain level of usage even bad practices sometimes need to be supported. I'll think on this.

JounQin:
I searched `tsconfig-paths` in this repository, and can only find it used in

https://github.com/import-js/eslint-plugin-import/blob/b2f6ac8eedac22a241f9d295bcdd4eef4e1c85cf/src/ExportMap.js#L526-L550

It is using `tsconfig-paths` for searching `tsconfig` files and then `require('typescript')` which is very expensive, I would suggest adding a new option `esModuleInterop: boolean` for this plugin instead of relying on `tsconfig-paths` or `typescript` at all. And also `@typescript-eslint/eslint-parser` does not use `tsconfig-paths`, so maybe @bradzacher can help to find a better way to get if `esModuleInterop` is configured without `tsconfig-paths`.

ljharb:
I think it would be a loss to lose the single source of truth of `esModuleInterop`, but I'm quite happy to use an alternative method of reading tsconfig (and handling arbitrary extends) if one is available.

bradzacher:
in ts-eslint we don't search for tsconfigs because we make people specify them explicitly via `parserOptions.project` - we've talked about searching for them, but it does require some consideration because it can sometimes be difficult to figure out exactly what tsconfig to pick for a given file (it's not always just "traverse up the tree and find the first `tsconfig.json`").

in terms of loading the tsconfig - we use TS for this directly:
https://github.com/typescript-eslint/typescript-eslint/blob/3d552895af436ef0a983a34a9b708030c9b5ebc7/packages/typescript-estree/src/create-program/createDefaultProgram.ts#L33-L38
https://github.com/typescript-eslint/typescript-eslint/blob/3d552895af436ef0a983a34a9b708030c9b5ebc7/packages/typescript-estree/src/create-program/useProvidedPrograms.ts#L66-L81

I don't know if there's another option for loading a `tsconfig` and resolving its `extends`.

JounQin:
> I don't know if there's another option for loading a tsconfig and resolving its extends.

@bradzacher 

I'm using https://github.com/privatenumber/get-tsconfig in `eslint-import-resolver-typescript@v3`, it resolves `extends` very well without depending on `typescript`!

ljharb:
@JounQin what would happen for users of an older TS parser, which we still support?

If we can come up with a way to expand the resolver API in a nonbreaking way, that'd be fine.

As for switching to `get-tsconfig` from `tsconfig-paths`, if it replicates the functionality then I'm perfectly fine with it.

JounQin:
> what would happen for users of an older TS parser, which we still support?

Which version do you mean?

> If we can come up with a way to expand the resolver API in a nonbreaking way, that'd be fine.

I think adding arguments into `resolve` function won’t be breaking.

> As for switching to get-tsconfig from tsconfig-paths, if it replicates the functionality then I'm perfectly fine with it.

`get-tsconfig` should be able to replace `tsconfig-paths` + `typescript` at the same time!

ljharb:
We support v2-v5 of `@typescript-eslint/parser`, and v15-v22 of `typescript-eslint-parser`.

JounQin:
> We support v2-v5 of `@typescript-eslint/parser`, and v15-v22 of `typescript-eslint-parser`.

When no `project` nor `tsconfigRootDir` defined in `parserOptions`, `get-tsconfig` will just fallback to `process.cwd()` as previous. Nothing would be broken.

JounQin:
For anyone may be interested, I've just forked this plugin as `eslint-plugin-i` at https://github.com/JounQin/eslint-plugin-i#eslint-plugin-i, they should just be compatible with any version of each other except `eslint-plugin-i` uses [get-tsconfig](https://github.com/privatenumber/get-tsconfig) to replace [tsconfig-paths](https://github.com/dividab/tsconfig-paths) and heavy [typescript](https://github.com/microsoft/TypeScript) under the hood

So you can just use `npm install -D eslint-plugin-import@npm:eslint-plugin-i` to replace original `eslint-plugin-import`, that's all.

And It will rebase and try to release in order to sync with the upstream every day, see [.github/workflows/rebase-upstream.yml](https://github.com/JounQin/eslint-plugin-i/blob/fork-release/.github/workflows/rebase-upstream.yml) for details.

n-studio:
json5 has a vulnerability for `< 2.2.2`

https://github.com/advisories/GHSA-9c47-m6qq-7p4h

ljharb:
@n-studio like most CVEs in the JS ecosystem, that one's a false positive for the code path in question. Prototype Pollution, in particular, is only a vulnerability if you're handling unsanitized data in production, and this is an eslint plugin - there's nothing to attack except yourself.

air2:
Nobody is using node 4. Please please just drop node 4 support. In theory it is a breaking change, but practicly it is not. The same with the CVE in theory it is a security risk, in practice it is not. BUT The code for the json5 v1 lib is not maintained, the code for v2 is maintained. Therefor please consider upgrading.

ljharb:
@air2 since our use of json5 won't change, its maintenance status is irrelevant. Until it actually stops *working*, there's nothing broken that needs fixing.

There are definitely people who still use node 0.10, let alone node 4, it's just not very common. Whether they're using eslint, or this plugin, isn't something that can be known, but "time has passed" is an absurd reason to impose unnecessary breakage.

pkuczynski:
`npm audit` started to fail now due to this issue. My CI builds rely on successful `npm audit` and there is no way to ignore a single dependency. What shall I, as a fan of `eslint-plugin-import`, do in such a case?

air2:
Well I saw a commit on the json5 v1 branch to fix this issue, I am hoping this mess is resolved soon, although it is indeed not a real security issue in this use case, it clutters the npm audit so real issues are not easily noticed anymore.

rattkin:
> Now, let's speak to this specific PR.
> 
> If a feature is compelling enough, especially if it actually requires a newer version of node (this one surely does not, except that tsconfig-paths dropped support), then it might be _worth_ the breaking change. However, jsconfig.json does not feel compelling to me. TS users use tsconfig.json; .editorconfig is a universal standard; but jsconfig.json is an upstart format that is not widely supported. Tools supporting a new format _can_ be helpful when it's low-cost, but as described above, a major change is costly. Separately, tools supporting minimally used formats can be harmful: Webpack supported "concord modules" or something for years, adding tons of bloat and complexity for every one of its users, for a module format that virtually zero people ever cared about.
> 
> Can you help me understand what about jsconfig.json is a compelling feature that a non-TS user would want? (note the dep is _ts_ config paths, so it should only be relevant for TS users)

Everybody who means well and takes their work seriously uses TS already. Especially ESLint users.
Existing bugs (or dangerous features) could mean further, yet undiscovered bugs.
Removing Node 4 that was created in the last century, is 14 versions old, and deprecated before some of us weren't even programming (me included) does not warrant a major version.

mashpie:
FWIW: eslint dropped supporting node <12 more than one year ago https://github.com/eslint/eslint/blob/main/package.json#L169

node v4 support was dropped in eslint april 2018 https://github.com/eslint/eslint/commit/b77846d5a4af266d7b5a752fa0d4a403f7177f24 perfectly aligned with release schedule of nodejs itself https://github.com/nodejs/release#release-schedule.

---

side-note on "attack yourself": any _local_ or _dev only_ vulnerability could even be worse than remote exploitable. If affected, your privileges might help compromising your local machine, CI/CD or any deployment target. Some thoughts for example https://security.stackexchange.com/questions/256114/are-devdependencies-in-node-js-exploitable & https://snyk.io/blog/npm-security-preventing-supply-chain-attacks/

rattkin:
> @air2 since our use of json5 won't change, its maintenance status is irrelevant. Until it actually stops _working_, there's nothing broken that needs fixing.
> 
> There are definitely people who still use node 0.10, let alone node 4, it's just not very common. Whether they're using eslint, or this plugin, isn't something that can be known, but "time has passed" is an absurd reason to impose unnecessary breakage.

Yet you are causing (prevent fixing) unnecessary breakage to many users. Certainly much more than hypothetic Node 4 ESLint users.

ljharb:
If json5 publishes a v1 backport that fixes it, then you won't have to see this false positive (most JS CVEs are false positives).

@rattkin i take my work seriously, and I mean well, and I stridently avoid using TS, so I'm going to hide your hostile and slanderous comment.

There IS NO BREAKAGE. There is only an incorrect vulnerability warning.

ljharb:
@pkuczynski unfortunately this is why relying on npm audit to pass in CI simply isn't a viable approach - you should use an alternative that allows ignoring by CVE and code path.

mashpie:
ref: https://github.com/import-js/eslint-plugin-import/issues/2625#issuecomment-1367501447

Okay, "they" backported - but https://github.com/json5/json5/pull/298#issuecomment-1367095053

looks like you'll loose v4 support anyways

ljharb:
No, in that case i'd just have to pin json5 to a vulnerable version.

mashpie:
Alright! Let’s all pin to deps to vulnerable versions - Hope your bugbounty is in place. Sry for sarcasm, but please don’t teach to ignore CVE‘s

ljharb:
@mashpie more accurately i'd have to (in this hypothetical scenario) pin to a working version, which happens to be included in the CVE.

I'm not teaching people to ignore CVEs. I'm teaching people to investigate CVEs and ignore the ones that are false positives, rather than naively and blindly believing that "audit warning === vulnerability".

PraveenVerma17:
This is really bad just to support node v4 you guys are blocking genuine PR, not sure which all idiots still use node v4. 
Finally i have removed this library from my eslint plugin. 
Good luck for lib anyway your users are going down as per npm chart.

pkuczynski:
> @pkuczynski unfortunately this is why relying on npm audit to pass in CI simply isn't a viable approach - you should use an alternative that allows ignoring by CVE and code path.

Any suggestions?

pmweeks98:
If node 4 isn't supported by latest eslint why is an eslint plugin supporting node 4? It requires minimum node 12 atm

ljharb:
@pmweeks98 this plugin supports eslint 2-8, so it supports everything eslint 2 or later supports.

ljharb:
@pkuczynski there's a ton of vendors that offer this - snyk, whitesource, etc - and it's also relatively trivial, i'd think, to write a custom ci task around `npm audit --json` to filter out the code paths you know aren't vulnerable. There's also `overrides`, ofc.

pmweeks98:
Why cant a new version be released that drops support for older versions.

This version will still be available for anyone who's using older eslint versions

ljharb:
@pmweeks98 because that would cause harm to those that depend on it - non-latest major lines aren't updated, and don't get security updates, which is the very concern being expressed here.

A new version of json5 v1 will be released soon, and then eslint-plugin-import won't have to do _anything_, and the audit warning won't complain anymore, so I don't see the point in releasing a breaking change here just for that.

PythonCoderAS:
I don't understand how it's necessary for an eslint plugin to support an 8-year-old node runtime.

ljharb:
@PythonCoderAS i hope you realize that your lack of understanding doesn’t change the necessity.

DonovanDMC:
It's all fine and good that you don't want to fix vulnerabilities levied against you (save your breath, I've already read that it doesn't truly effect you) but end users don't know that. They see 3 high security vulnerabilities in packages they use (typically dependency of a dependency) and freak out. Beyond either breaking our dependency trees by force installing an older version, or downgrading every other plugin that has moved on to eslint 8 (the most recent version that isn't effected is `2.24.1`, which apparently doesn't support eslint 8). You're telling us we all need to take our entire linting suite and *downgrade* them just because you're too full of yourself to accept dropping a node version that's 7 years old, or let our users sit there and wonder why we aren't fixing security vulnerabilities? I hope someone forks this and does a better job than you have.

n-studio:
> because that would cause harm to those that depend on it - non-latest major lines aren't updated, and don't get security updates, which is the very concern being expressed here.

@ljharb If I'm not mistaken, node 4 is an LTS and its maintenance ended back in 2018, almost 5 years ago. I don't think the argument of security stands anymore. If a developer keeps using node 4, I think they have other pressing security issues than getting security fixes for an eslint plugin. Just my 2 cents.

PraveenVerma17:
Owner does not want to fix this high severity issue, According to him its not an issue and people must live with it. 
I would suggest to remove this plugin from your eslint. Till the time other fork or he fixes it, removing plugin will not break your code.

kmturley:
It would only take a few minutes to release a patched version which would fix the issue automatically for 99% of users.
Instead you are asking every single dependent app (9 million weekly downloads) to manually update and test their code...

I've already spent and hour trying to find a solution for this, isn't your package meant to save developers time? rather than waste it?

ljharb:
@kmturley you're wrong, i can't possibly release a patch - it'd be a major version, a breaking change. It would cost me thousands of additional hours in support time, which virtually ZERO of you are paying me for.

ljharb:
It's clear there's no productive discussion to be had here.

The only way this audit warning will go away for the foreseeable future is if a json v1 patch is released, or if a tsconfig-paths v3 patch is released. Although the tsconfig maintainers have chosen not to do that, sadly, the json maintainers are planning to in the coming days (during a holiday week, mind you, which is the same week in which you're all yelling at me about not doing the specific unpaid work you want).

Posts that are abusive/hostile, and violate our CoC, will not be tolerated. Additionally, claims of "<thing that i don't personally care about but others do> doesn't matter" is simply invalid reasoning and is useless to volunteer.

ljharb:
v1.0.2 of json5 has been published, so this invalid vulnerability warning should no longer appear - just update your lockfiles.

pkuczynski:
> v1.0.2 of json5 has been published, so this invalid vulnerability warning should no longer appear - just update your lockfiles.

It still does, as it is defined as `json5  <2.2.2`. I guess we need to wait for the update on the CEV/npm audit side?

ljharb:
@pkuczynski yep, that's here: https://github.com/github/advisory-database/pull/1541

jordanbtucker:
@ljharb Sorry for the disruption I caused by not backporting the fix before the vulnerability was published. I'll try to avoid that in the future considering how popular json5@v1 still is. Hopefully the security advisory will be updated soon.

ljharb:
@jordanbtucker you have nothing to blame at all; it's a holiday week and you've gone above and beyond to get the fix out, and I appreciate it! That some users are entitled, hostile, or reductionist doesn't diminish your effort.

PraveenVerma17:
Either don’t do charity, if you do charity you are entitled to be blamed. Period.

ljharb:
@PraveenVerma17 i think you really don't understand the concept of charity - and open source isn't charity. The software license says "as is" which means "you are owed nothing, ever". If you're a jerk, you're entitled to be blocked. Period.

coldMater:
Finally. 
https://github.com/dividab/tsconfig-paths/commit/9721a98718c9f6a35bb4029292204aaa90474bab

jordanbtucker:
The security advisory is finally updated and json5@1.0.2 is recognized as patching CVE-2022-46175.

zackdotcomputer:
(Moving over from yet-another-update-this-dependency issue that was closed in favor of this discussion)

Want to throw my support in favor of accepting this PR as well. Fully agreed with the 7 months of discussion above that maintaining support for Node 4 (whose last release which will turn 5 years old in just a few months) as the main reason to reject a new major release of a dependency does not strike me as a sound decision.

Blocking the advancement of dependencies to new major versions, even if they require dropping support for one arcane release of Node, is just shifting the burden of applying ongoing patches to back-catalogued versions down the chain to this package's dependencies. Eventually, one of them will not do that work either and an update will have to be accepted.

Rather than waiting for that, applying these updates gradually over time, via major version updates and dropping Node versions one-by-one, allows for the whole ecosystem to move gradually with you. This is, in my opinion, vastly preferable to waiting for an unpatched issue to one day force a more abrupt and larger update, or force developers who must stay compliant with security standards like SOC to figure out how to abandon your package.

mashpie:
@zackdotcomputer exactly!

air2:
Gradually dropping support for old (node) versions and keeping dependencies up to date is IMHO the best way to support the community and saving all people using this package a lot of time.

ljharb:
My long experience tells me that the reverse is true - it's *dropping* support for versions that causes the most pain, as evidenced by this very PR - if tsconfig-paths hadn't made a v4, and had maintained backwards compatibility (or if json5 hadn't made a v2) then all of this would have just been automatic.

Breaking changes are the greatest possible cost to the community, and I don't want to contribute to that.

air2:
Well in my 18+ years of software development experience i understand your position. To me it's always a line of balance. The world around us is progressing, so therefor we have to flow with it. Of course being backwards compatible is nice, but in the end it always is just a lot of work. And there is this line of balance. How much work you want to put into flowing with the progress and being backward compatible. Json5 etc will flow further. By not flowing with those packages, in the end it will be much more work to catch up in the event of some problem that causes you to have to move forward. But again it is your package. So you decide. Its not a democratic environment. 

Owh and off topic just as not to you personal. I never intended to be rude to point people to an other package. I just wanted to help. I am sorry if i caused bad feelings. Because the other package just uses all your work, only works around the thing that was causing a problem for many people. But again i never wanted to behave badly.

jordevo (approved):

oegea (approved):

PythonCoderAS:
I don't understand why there's such a desire for remaining compatible with such an old node version because to my knowledge the `import` statement is not even supported until Node 12

zackdotcomputer:
@PythonCoderAS it was introduced in 8 but behind the experimental-modules flag until 12. But that's somewhat irrelevant here since your linting environment might use a different platform than your deployed environment - e.g. you could be writing Typescript that uses `import` which is eventually compiled to use only `require` before you go to actually run it. You'd still want to lint the `import`s even if you're not using them at runtime.

The main question is whether it's imperative for this project to continue to support people who are doing their development or linting on archaic versions of Node, irrespective of where they eventually deploy the code.

albertodiazdorado:
Can we please put this PR on hold and upgrade `json5` to version `1.0.2` at the very least? Is there any reason why we shouldn't do that short-term?

Karibash:
From typescript 5.0, it is now possible to specify an array in the "extends" field of the tsconfig file.
However, as tsconfig-paths does not yet support array specification in the "extends" field, an error occurs when specifying an array in "extends".
This issue is expected to be fixed in the following PR.
https://github.com/dividab/tsconfig-paths/pull/245

If eslint-plugin-import is to support the latest TypeScript, tsconfig-paths needs to be updated.

ljharb:
@Karibash alternatively, tsconfig-paths could backport that support to v3, which has 18 million weekly downloads to v4's 3-4 million, and reach more people.

marvinhagemeister:
Given that the download count of `eslint-plugin-import` [is around 19-20m](https://www.npmjs.com/package/eslint-plugin-import) could it be that this plugin is the sole contributor to the `tsconfig-paths` v3 download count?

ljharb:
@marvinhagemeister that's entirely possible. However, we can't upgrade to v4 because of the node versions it drops, so the impact is the same.

ljharb:
@PraveenVerma17 don't tell me what i must do - i, in fact, NEVER need to drop support for node 4. node's support of any node version is irrelevant.

PraveenVerma17:
I just request everyone to stop using this library, owner is silly and arrogant. 
Enjoy your shitty lib and eat this lib you fk retarded

n-studio:
> I just request everyone to stop using this library, owner is silly and arrogant. Enjoy your shitty lib and eat this lib you fk retarded

This kind of language is unacceptable in open source. Please stop using github.

WoodyWoodsta:
No-one should be using Node versions which have reached EOL in production. Anybody interested in delivering secure and maintainable software should know this to be true. Node's release schedule provides everyone with **many years** to update, and failing to do so **is the developer's responsibility**.

---

@ljharb I think there are two options for a community who are interested in delivering the quality owed to users:
1. This project releases a major version dropping unsupported versions of Node, or
2. The community forks and maintains a secure and updated version of `eslint-plugin-import` which respects and follows the release schedule of the platform upon which it is built.

`tsconfig-paths` has no reason, nor obligation, to backport TS 5 to an older version of their project.

pmweeks98:
This conversation has been going on a year in circles. @ljharb does not want to upgrade and while I do not agree with the decision it is their package, their choice.

If think @WoodyWoodsta suggestion for a fork is possibly the only path forward here if people want an upgraded tsconfig-paths. Continuing this discussion is pointless

ljharb:
@WoodyWoodsta Something reaching EOL is irrelevant in many production use cases, and just because a security bug exists doesn't mean it applies. I certainly _hope_ everyone's upgrading node as much as they can - but I refuse to allow my packages to be the reason why they're forced to endure that risk and cost.

Indeed, tsconfig-paths has no obligation - but there's a HUGE reason, which is that the primary source of their downloads is stuck on v3.

If people want to make and maintain a fork, go for it! Considering in the last half-decade or more, a very very small number of people have made PRs, let alone expressed interest in maintaining it or provided funding, I suspect you'll find that "shoulds" and outrage aren't sufficient to sustainably maintain a project.

n-studio:
> Something reaching EOL is irrelevant in many production use cases

Really? It doesn't sound right.

> but I refuse to allow my packages to be the reason why they're forced to endure that risk and cost.

I would see the glass half-full rather than half-empty, your package would encourage people to secure their application.

The risk and cost of keeping legacy is way higher than the risk and cost of updating.

WoodyWoodsta:
@ljharb No outrage from me - just a lot of head scratching. I could throw benefit of the doubt into supporting a version that fell just short of the current maintenance range, but v4 is way outside of that bracket.

The point that makes the least sense to be is that `eslint`, **the tool to which yours is a plugin**, drops support for:
- Node 10, 13 + 15, in eslint v8
- Node 8 in eslint v7
- Node 6 in eslint v6
- Node 4 in eslint v5 (2018)

So what exactly are you trying to achieve by supporting a version of the platform that was long dropped by your plugin's parent?

--

Over the next week I'll consider forking the project publicly. I **cannot promise** anything, but otherwise, my decision for my own organisation's use of the plugin is between creating a private fork (again, forced anti-community behaviour) or just not using the plugin altogether. We'd gladly not lint imports in our projects over keeping back the versions of the rest of our tool-chain.

WoodyWoodsta:
Ah. Turns out https://www.npmjs.com/package/eslint-plugin-i is a fork which already exists as a result of this PR. Mind-blowing that one can watch fragmentation of the community unfold yet still claim that maintaining an archaic version of Node is "not hurting the community".

eranhirsch:
The whole purpose of using a lint is to tell you you might be doing something wrong, I'd say using node 4 in mid 2023 to run your linting is probably one of the advices, based on it being the advice of the node team, and the eslint team.

Users of node4 can always just use an older version of this plugin... they are probably using an older version of eslint anyway, and an older version of typescript too (that's how I found this issue, because of the new array format for extends).

Feels like without proper numbers of how many of these users exist the whole discussion is kinda moot. There might be a handful of users who this decision actually serves, vs likely at least several magnitudes more users who this doesn't serve.

These decisions are never pure, they should be compromises between competing trade-offs.

Sparticuz:
Nothing but problems after upgrading to Typescript 5 and using an extends array. Maybe @eslint-community can step in and fork the package to allow for modern programming practices.

ljharb:
@eranhirsch 1% of users might still be thousands of human beings.

@Sparticuz maintaining backwards compatibility _is_ a modern programming practice, and given that basically nobody's stepped up to help me maintain the plugin, it's pretty unlikely anybody will ever meaningfully fork it. Good luck tho!

eranhirsch:
That wasn't the point. Even if 1% is thousands, there are 100 fold other users you aren't serving with this choice.

A good feature of ts 5 is no longer co operable with this plugin. Meaning users of both need to pick 1 instead of enjoying using both.

This decision you are making is material, its not theoretical. I had to change my tsconfig in a way that i didn't want because of this, and so will most likely hundreds of thousands of users as more breaking changes are added to ts

Sparticuz:
> nobody's stepped up to help me maintain the plugin

Oh, I assumed that this PR was fixing the issue?

ljharb:
@Sparticuz this PR updates a dependency, forcing a breaking change, which incurs a *lot* of maintenance cost on me - a very different thing than someone stepping up to help *maintain* the plugin.

Backports are certainly viable, but that adds *even more* burden on me.

ljharb:
@eranhirsch that's totally true, and it's very unfortunate. My hope is that tsconfig-paths will backport their support to v3, or, be willing to accept PR(s) from me to restore node 4 support to v4 - the latter is where i'd much rather invest my time.

Sparticuz:
> @Sparticuz this PR updates a dependency, forcing a breaking change, which incurs a _lot_ of maintenance cost on me - a very different thing than someone stepping up to help _maintain_ the plugin.
> 
> Backports are certainly viable, but that adds _even more_ burden on me.

I guess we'll never know who will or won't step up if the change is never merged.

eranhirsch:
> @eranhirsch that's totally true, and it's very unfortunate. My hope is that tsconfig-paths will backport their support to v3, or, be willing to accept PR(s) from me to restore node 4 support to v4 - the latter is where i'd much rather invest my time.

Is there a pr we or issues in that repo we can follow?

ljharb:
Nope, I haven't had time to file an issue yet asking.

mashpie:
https://owasp.org/Top10/A06_2021-Vulnerable_and_Outdated_Components/

eranhirsch:
> @eranhirsch "backport future changes" is a burden, which is why i prefer if tsconfig-paths restores node 4 compat rather than backporting things to v3. The thing that always puts the least burden on everyone is maintaining maximal backwards compatibility for the maximum amount of time - anything else causes harm.

But what if they don't want to do that? Isn't this causing harm too? Did eslint cause harm when they dropped support for node4? Did node? I'm not being sarcastic, I'm just don't understand how you measure harm?

mashpie:
https://owasp.org/Top10/A06_2021-Vulnerable_and_Outdated_Components/

pmweeks98:
> Someone still using an older version of eslint can still use us.

Is this not true if you introduce a breaking change- they would just have to use a specific version instead of latest?

mashpie:
https://owasp.org/Top10/A06_2021-Vulnerable_and_Outdated_Components/ - outdated… Not off topic

Sparticuz:
Why doesn't this package support Node 0.10 or 0.12?

ljharb:
@mashpie if you keep reposting comments i've hidden, you'll be blocked from the org. It is off-topic, and I'm quite familiar with OWASP, thanks.

ljharb:
@Sparticuz presumably because at the time v2 was created (long before I took over maintenance), testing was only done on node 4.

kmturley:
Node metrics for downloads per day:
```
"v18.15.0": 243310,
"v16.13.0": 177585,
"v14.21.3": 107756,
"v19.8.1": 81883,
"v12.22.12": 37941,
"v10.24.1": 20658,
"v8.12.0": 9027,
"v6.11.1": 3285,
"v4.1.1": 1194,
```
https://nodejs.org/metrics/

AWS ended support for v10 and lower in their SDK in 2021
https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-node-js-runtimes-10-x-in-the-aws-sdk-for-javascript-v2/

Google Cloud only supports v10+
https://cloud.google.com/appengine/docs/standard/nodejs/runtime

Azure functions deprecated v10 in 2022
https://azure.microsoft.com/en-us/updates/azure-functions-support-for-node-10-is-ending-on-30-september-2022/

Are there cloud services which support Node v4? because I can't find any

ljharb:
@kmturley i'm not sure what good there is in citing statistics; a version of node doesn't need to be downloaded to be used - it could be cached internally, or on a docker image, or whatever.

01taylop:
> @pmweeks98 anything that stops this plugin from functioning on eslint 2 or higher is a breaking change. If we did do a breaking change, and say, dropped eslint < 7, then those users would be forced to upgrade eslint to use new versions of this plugin.

I've enjoyed following this thread and have taken a lot from it. However, there is one thing I find nonsensical. This change would drop support for Node 4, which if I'm not mistaken ESLint did in their 5.0.0 release (ESLint is now up to version 8.x.y. which drops support for Node 10).

I think it is fair to say that users of Node 4 are not particularly fussed about supporting the latest versions, and that almost all of their dependencies would be major versions behind the latest due to their Node version. A new feature of an ESLint plugin which is not available to them would likely go unnoticed, let alone cause an issue, and certainly not be the reason they upgrade their Node version.

"The only thing that never changes, is that everything changes", and although Louis L'Amour wasn't talking about software engineering, I think the quote is quite apt to this thread. Dropping Node 4 support and supporting TypeScript 5 can open up more opportunities for ESLint plugins, support a huge number of engineers, and be more resilient to future changes.

Sparticuz:
Eslint 9 will also break this plug in because of dropping the old config system, so it's only a matter of time before this package has to have breaking changes or truly become obsolete.

ljharb:
@Sparticuz #2556, which can be added in a semver-minor, will address that concern.

mashpie:
> @mashpie if you keep reposting comments i've hidden, you'll be blocked from the org. It is off-topic, and I'm quite familiar with OWASP, thanks.

I‘ve been posting from mobile. Doublepost before ready. Last one was with comment. Beeing forced to use an outdated package by another package introduces risks. 

I don‘t say yours should drop support for old versions, but forcing old versions is risky.

ljharb:
It’s only risky if there’s security issues that apply - this is an eslint plugin, to which virtually none apply.

knpwrs:
Adding this to my `package.json` resolves the issue for me:

```json
  "overrides":{
    "tsconfig-paths": "^4.0.0"
  }
```

ljharb:
Hmm - I do wonder if we could change the tsconfig-paths dependency to `^3 || ^4`, and how old of an npm would be smart about not installing v4 when `engines.node` doesn't match (since absent those smarts, it'd install v4 automatically and be a breaking change).

If someone wants to try that out and report back, that would be a path forward here that potentially avoids a breaking change in this library.

air2:
I wonder if you will find anybody who uses node 4. I really doubt this project has any node 4 users

stianjensen:
> Hmm - I do wonder if we could change the tsconfig-paths dependency to `^3 || ^4`, and how old of an npm would be smart about not installing v4 when `engines.node` doesn't match (since absent those smarts, it'd install v4 automatically and be a breaking change).
> 
> 
> 
> If someone wants to try that out and report back, that would be a path forward here that potentially avoids a breaking change in this library.

Unfortunately, it seems like that is not the default behavior:
https://docs.npmjs.com/cli/v9/using-npm/config#engine-strict

ljharb:
@stianjensen i'm not talking about engine-strict; npm 7+, at least, will prefer an in-range dep with compatible engines over an in-range dep without them. I'm not sure how far back that behavior goes, though.

stianjensen:
This thread seems to indicate it wasn't always that way 🤔 

https://github.com/npm/npm/issues/3662

repl-krishna-g:
@F3n67u Can this PR merged? This will help us to fix json5 vulnerability in our project.

ljharb:
No, it won't - json5 v1 has a fix for the vulnerability, and you can update that without this PR landing.

so1ve:
Bro, WHY NOT MERGE THIS PR?
Support your baby node v4?
Node v4 has been abandoned by its parents, Open js foundation, but you want to pick it up as a treasure?????
Why are you developing websites not thinking about supporting mobile phones from twenty years ago? There are still many people using them now.
PLEASE, MERGE THIS PR!

so1ve:
Bro, WHY NOT MERGE THIS PR?
Support your baby node v4?
Node v4 has been abandoned by its parents, Open js foundation, but you want to pick it up as a treasure?????
Why are you developing websites not thinking about supporting mobile phones from twenty years ago? There are still many people using them now.
PLEASE, MERGE THIS PR!

so1ve:
Yes, agreed. eslint-plugin-i supports typescript 5 and has a better performance, so use it as a drop-in replacement is a good solution. And, apologize for my inpatient @ljharb.

so1ve:
Just want to provide a solution. Not a SPAM. I really love this plugin and I don't want to fight with somebody.

@ljharb The current situation is, you don't want to drop node4 support, right? How about just simply inline that package and use babel to polyfill it?

ljharb:
Transpiling code you didn’t author is dangerous, and i don’t have the capacity to take over maintaining (an inlined fork of) tsconfig-paths.

benmccann:
A very big reason this project doesn't support Node 12+ features (https://github.com/import-js/eslint-plugin-import/issues/1810) is because of its strict adherence to backwards compatibility. Breaking changes are undoubtedly harmful. But I think we also need to weigh it against the harm of making features so complicated and expensive to implement by needing to support every version since Node 0.4 that they can't progress (https://github.com/browserify/resolve/pull/224#issuecomment-1610128580).

ljharb:
@benmccann I continuously weigh it, and my stance remains.

benmccann:
Perhaps we could at least update the `engines` field to be accurate? It currently says `"node": ">=4"`, but it should really say something more like `"node": ">=4 <12"` given that this plugin is not capable of handling Node 12's `export` syntax, which has been widely adopted across the ecosystem

ljharb:
It works just fine in modern node. `engines.node` is in no way a statement about which of node's features are supported.

ljharb:
@benmccann either way, arguments about `exports` are off topic for this particular PR. If tsconfig-paths v5 restores the engines support we need, or if tsconfig-paths v4 receives the feature backport we need, then we have a path forward - otherwise, we don't.

benmccann:
Sure, I'll make future comments about that feature on the relevant issue (https://github.com/import-js/eslint-plugin-import/issues/1810). I just wanted to keep the conversation about which versions are supported in one place. I'll disagree with "works just fine" though given that it's giving incorrect output.

nocive:
For those landing here and going through this nightmare of a thread, you can use `eslint-plugin-i` as a replacement for this plugin if you're bothered by the reported CVEs.

Sparticuz:
It's not just the cve. This package doesn't even work if you use an extends array in your tsconfig.

ljharb:
@Sparticuz yep, unfortunately that feature wasn't backported to v3 either.

ptrxyz:
Simply using `eslint-plugin-i` seems to do it for me ...

coreyward:
Just want to chime in to say that this should have been a 5 second task to update some dependencies in a library of my own that uses eslint-plugin-import, but instead I have a thread of hypothetical opinions talking about avoiding costing theoretical developers time. It's yours and you do what you want, but please get real about the impact and why you are doing it because it has nothing to do with creating a better experience for developers, as the dozens and dozens of them here can attest.

tkrotoff (approved):

Tarektouati (approved):

NutchapolSal:
If you insist that you have to support Node v4, have you considered splitting releases into v2 and v3? v2 can keep support for old Node versions while v3 get updated dependencies and use new features. Bug fixes that apply to both can be merged into both lines. I think it is better than the current situation.

domdomegg:
If people are concerned about the CVEs (which you shouldn't be - see here), you can update json5 now there is a fix on v1.

If people want support for multiple extends is tsconfig, there is a backport that would allow that here: https://github.com/dividab/tsconfig-paths/pull/260

coreyward:
> If people are concerned about the CVEs (which you shouldn't be - see here), you can update json5 now there is a fix on v1.

@domdomegg I think you missed a link ("see here").

Most corporate security policies require CVEs to be reviewed and have resolution tracked as part of the processes designed to meet security standards (SOC2, ISO 27001, NIST, etc). Determining that affected code is unused requires a fair amount of investigation, especially for common sub-dependencies like `import-js`—changes to dependencies can change the code exercised in a sub-dependency. Asserting that the risk is tolerable is sometimes okay, but when an alternative package by a similar name is available and well maintained the position is rather unpalatable. The only real move for those of us working on commercial projects is to adopt the alternative package.

ljharb:
@coreyward I think you missed that the CVE is *fixed* in json5 v1 - in other words, even for corporations that have incorrectly rigid security policies, *the CVE no longer applies if you update your deps, in-range*.

ljharb:
Since the desired functionality is now backported into tsconfig-paths v3, there's no longer any reason to upgrade to v4, and this PR is no longer needed.

PR #2464, opened 2022-06-01
[Tests] `order`: groups in groups (failing test only)

bobaaaaa:
I have noticed, that the `import/order` rule got a regression for groups in groups config (TypeScript). I am pretty sure that this was working at some point. If I remove the inner group, the test passes. Or did I miss something?


ljharb:
Thanks for the failing test! I'll see if i can come up with a fix.

Issue #2142, opened 2021-07-02
[import/no-restricted-paths] configuration items target and form should support array format


benboba:
Sometimes I want to set certain paths in batches to prohibit relying on the same path, or a certain path to prohibit relying on multiple paths.

Now I need to set one by one, and hope to support the incoming array for batch setting.

e.g.
`'import/no-restricted-paths': [2, { "zones": [ {
    "target": "modules/a",
    "from": ["modules/b", "modules/c"],
}, {
    "target": ["modules/b", "modules/c"],
    "from": "modules/d",
} ] }]`

ljharb:
Seems reasonable.

benboba:
Maybe it can support a RegExp or RegExp string, which can be more flexible.

ljharb:
Definitely not; that’s a floodgate to CVEs.

PR #2466, opened 2022-06-01
feat(2142): `no-restricted-paths` - add support for arrays in `from` and `target` fields

AdriAt360:
Issue is: https://github.com/import-js/eslint-plugin-import/issues/2142

### Context

Sometimes I want to set certain paths in batch to prohibit relying on the same path, or a certain path to prohibit relying on multiple paths.

### Changes

In this PR
- I update the schema of this rule
  - before: `from` and `target` are only a string
  - after: these fields can be either a string or an array of string
- I change the behavior + add tests
  - all the elements in `from` and `target` are taken into account the same way it was before
  - I also ensure the homogeneity of the `from` elements to stay aligned with the current homogeneity checks between `from` and `except`
- I update the related doc and add new examples

### Note

I advise to read this PR commit by commit as there are quite a few changes that are easier to review step by step, e.g. extracting functions before using them inside loops


src/rules/no-restricted-paths.js
```diff
@@ -29,8 +29,8 @@ module.exports = {
             items: {
               type: 'object',
               properties: {
-                target: { type: 'string' },
-                from: { type: 'string' },
+                target: { type: ['string', 'array'] },
+                from: { type: ['string', 'array'] },
```
ljharb, line 33: i think this needs to be `oneOf`, so it can be either a string, or an array of unique strings

src/rules/no-restricted-paths.js
```diff
@@ -56,78 +56,125 @@ module.exports = {
     const basePath = options.basePath || process.cwd();
     const currentFilename = context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename();
     const matchingZones = restrictedPaths.filter((zone) => {
-      const targetPath = path.resolve(basePath, zone.target);
+      const allTargets = zone.target instanceof Array ? zone.target : [zone.target];
```
ljharb, line 59: `instanceof` should never be used.
```suggestion
      const allTargets = [].concat(zone.target);
```

src/rules/no-restricted-paths.js
```diff
@@ -56,78 +56,125 @@ module.exports = {
...
+    }
 
     function isValidExceptionPath(absoluteFromPath, absoluteExceptionPath) {
       const relativeExceptionPath = path.relative(absoluteFromPath, absoluteExceptionPath);
 
       return importType(relativeExceptionPath, context) !== 'parent';
     }
 
+    function areBothGlobPatternAndAbsolutePath(areGlobPatterns) {
+      return areGlobPatterns.some((isGlob) => isGlob) && areGlobPatterns.some((isGlob) => !isGlob);
```
ljharb, line 79: these conditions can't ever both be true.

src/rules/no-restricted-paths.js
```diff
@@ -141,34 +188,34 @@ module.exports = {
...
+
+          if (!hasValidExceptions) {
+            reportInvalidException(node);
+            return;
+          }
+
+          const pathIsExcepted = isPathException(absoluteImportPath);
+          if (pathIsExcepted) {
+            return;
+          }
```
ljharb, line 212: let's use `.filter` here for all of these, so the part in the forEach is small

tests/src/rules/no-restricted-paths.js
```diff
@@ -255,7 +322,143 @@ ruleTester.run('no-restricted-paths', rule, {
         } ],
       } ],
       errors: [ {
-        message: 'Restricted path exceptions must be glob patterns when`from` is a glob pattern',
```
ljharb, line 258: oof, this missing space passed before??

tests/src/rules/no-restricted-paths.js
```diff
@@ -255,7 +322,143 @@ ruleTester.run('no-restricted-paths', rule, {
...
+    // array with single element
+    test({
+      code: 'import b from "../server/b.js"',
+      filename: testFilePath('./restricted-paths/client/a.js'),
+      options: [ {
+        zones: [ {
+          target: ['./tests/files/restricted-paths/client'],
+          from: './tests/files/restricted-paths/server',
+        } ],
+      } ],
```
ljharb, line 341:
```suggestion
      options: [
        {
          zones: [
            {
              target: ['./tests/files/restricted-paths/client'],
              from: './tests/files/restricted-paths/server',
            }
          ],
        }
      ],
```
and similar formatting in all the test cases

src/rules/no-restricted-paths.js
```diff
@@ -56,78 +56,125 @@ module.exports = {
...
+      const isGlobPattern = areGlobPatterns.every((isGlob) => isGlob);
+
+      return allZoneFrom.map(singleZoneFrom => {
+        const absoluteFrom = path.resolve(basePath, singleZoneFrom);
+
+        if (isGlobPattern) {
+          return computeGlobPatternPathValidator(absoluteFrom, zoneExcept);
+        } else {
+          return computeAbsolutePathValidator(absoluteFrom, zoneExcept);
+        }
```
ljharb, line 176:
```suggestion
        }
        return computeAbsolutePathValidator(absoluteFrom, zoneExcept);
```

src/rules/no-restricted-paths.js
```diff
@@ -56,78 +56,125 @@ module.exports = {
...
       return {
         isPathRestricted,
         hasValidExceptions,
         isPathException,
         reportInvalidException,
       };
+    }
+
+    const makePathValidators = (zoneFrom, zoneExcept = []) => {
+      const allZoneFrom = zoneFrom instanceof Array ? zoneFrom : [zoneFrom];
```
ljharb, line 160:
```suggestion
      const allZoneFrom = [].concat(zoneFrom);
```

ljharb (changes requested):

src/rules/no-restricted-paths.js
AdriAt360: @ljharb they can, for example with `from: ["./*", "./"]`
- `areGlobPatterns` will be an array with values `[true, false]` (the first one is a glob pattern, the second one isn't)
- that's exactly the situation I want to avoid, to keep the homogeneity between `from` and `except` as it was before

do you agree with this with this context?

docs/rules/no-restricted-paths.md
```diff
@@ -124,3 +127,66 @@ The following import is not considered a problem in `my-project/client/sub-modul
...
+   └── b.js
+```
+
+and the current configuration is set to:
+
+```
+{ "zones": [ {
+    "target": ["./tests/files/restricted-paths/two/*", "./tests/files/restricted-paths/three/*",
+    "from": ["./tests/files/restricted-paths/one", "./tests/files/restricted-paths/three"]
+} ] }
```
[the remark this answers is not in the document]
AdriAt360: @ljharb ok for me 👍🏽 https://github.com/import-js/eslint-plugin-import/pull/2466/commits/b59175dc42e696b005c3e237c3571249ecd46d27

src/rules/no-restricted-paths.js
AdriAt360: @ljharb agree! fixed in https://github.com/import-js/eslint-plugin-import/pull/2466/commits/bd987a55ff959f3239cee4f7506ef5f30ac96448

src/rules/no-restricted-paths.js
AdriAt360: @ljharb good point 👍🏽 
fixed in https://github.com/import-js/eslint-plugin-import/pull/2466/commits/95d16861d1ec2c0f0902b57330356857ee152540

src/rules/no-restricted-paths.js
AdriAt360: @ljharb good point 👍🏽 
what do you think about this new version with `.filter` and new functions? https://github.com/import-js/eslint-plugin-import/pull/2466/commits/2d8aad50b618dd7ac63e11cdd3ce871d79cbb582

tests/src/rules/no-restricted-paths.js
AdriAt360: @ljharb looks good 👍🏽 https://github.com/import-js/eslint-plugin-import/pull/2466/commits/76883fd77dc5bf42e2cee1c3af30243412f5fbb6

src/rules/no-restricted-paths.js
AdriAt360: @ljharb seems like a formatting preference, ok to align on yours 👍🏽 https://github.com/import-js/eslint-plugin-import/pull/2466/commits/6b4ea9f5f09db7e440883a87e4627c349b6f7f4e

src/rules/no-restricted-paths.js
AdriAt360: @ljharb good point 👍🏽 
fixed in https://github.com/import-js/eslint-plugin-import/pull/2466/commits/95d16861d1ec2c0f0902b57330356857ee152540

src/rules/no-restricted-paths.js
ljharb: ah, i see what you mean, thanks.

src/rules/no-restricted-paths.js
```diff
@@ -29,8 +29,26 @@ module.exports = {
...
               properties: {
-                target: { type: 'string' },
-                from: { type: 'string' },
+                target: {
+                  oneOf: [
+                    { type: 'string' },
+                    {
+                      type: 'array',
+                      items: { type: 'string' },
+                      uniqueItems: true,
```
ljharb, line 38:
```suggestion
                      uniqueItems: true,
                      minLength: 1,
```

src/rules/no-restricted-paths.js
```diff
@@ -29,8 +29,26 @@ module.exports = {
...
+                    },
+                  ],
+                },
+                from: {
+                  oneOf: [
+                    { type: 'string' },
+                    {
+                      type: 'array',
+                      items: { type: 'string' },
+                      uniqueItems: true,
```
ljharb, line 49:
```suggestion
                      uniqueItems: true,
                      minLength: 1,
```

src/rules/no-restricted-paths.js
```diff
@@ -56,78 +74,138 @@ module.exports = {
     const basePath = options.basePath || process.cwd();
     const currentFilename = context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename();
     const matchingZones = restrictedPaths.filter((zone) => {
-      const targetPath = path.resolve(basePath, zone.target);
+      const allTargets = [].concat(zone.target);
+      const allTargetPaths = allTargets.map(target => path.resolve(basePath, target));
+      return allTargetPaths.some(targetPath => isMatchingTargetPath(currentFilename, targetPath));
```
ljharb:
```suggestion
      return [].concat(zone.target)
        .map(target => path.resolve(basePath, target))
        .some(targetPath => isMatchingTargetPath(currentFilename, targetPath));
```

ljharb (approved):
I'll make these changes in a rebase. Thanks!

AdriAt360:
@ljharb my pleasure!
thanks for your review and suggestions 🙏🏽 
how do you want to proceed for the next steps: as I have no write access, I guess you will merge this pull request yourself and close the related issue when done?

ljharb:
Yes, I'll merge it myself, and the `fixes #2142` i edited into your OP will cause github to auto-close it upon merging.

AdriAt360:
@ljharb indeed, I missed your edit, all perfect then =)

ljharb:
@AdriAt360 hm, looks like this PR is causing the native windows tests to fail: https://ci.appveyor.com/project/ljharb/eslint-plugin-import/build/job/xqkwnmajs2svu218

AdriAt360:
@ljharb as there are multiple tests with this name and I can't reproduce locally, I just added some comments to identify the failing test https://github.com/import-js/eslint-plugin-import/pull/2466/commits/d3fa030e36021f60152e7a2a31c7db8c984989df
I hope it will help investigating and keep you posted

ljharb:
looks like number 7: https://ci.appveyor.com/project/ljharb/eslint-plugin-import/builds/43753304/job/dshd09a4o6tucs0a

AdriAt360:
@ljharb still no way to repro on my side, so I added another test to help me investigate

AdriAt360:
@ljharb ok, seems that this specific glob pattern (`./some/path/*`) is not accepted by windows
I updated and added some tests to check some other common patterns

AdriAt360:
@ljharb I guess I found the issue, but as it impacts all the tests, you might want to give an opinion
as stated in https://stackoverflow.com/a/39111164/2814072, `path.resolve` always resolves to an absolute path where `path.join` might or not
problem is:
- in `test/utils`, the `testFilePath` function used `path.join`, but in the source code of the `no-restricted-paths` rule, only `path.resolve` is used
- in case of glob patterns, `minimatch` will match the file only if both the glob and the file path are absolute, as the glob pattern is
image
- I guess on Windows, `path.join` resolved to a relative path, explaining the failing tests

If this hypothesis is correct, the pipeline should be all green now, so the next step would be to clean the tests I added to investigate here, unless you prefer keeping them

AdriAt360:
@ljharb ok, wrong hypothesis
another hint could be `process.cwd()`

<details><summary>More details here</summary>

- from https://www.geeksforgeeks.org/node-js-process-cwd-method/, I can see that the current working directory resolves with back-slashes
image
- from https://www.npmjs.com/package/minimatch, I can see that the glob pattern should contain only forward-slashes
image
- but in our situation
  - `target` is defined correctly (only `/`)
  - if no `basePath` is provided, `process.cwd()` is used as `basePath` to build `targetPath`
  - meaning on Windows native, the target path pattern could resolve with `\` in some conditions (I guess, don't know why it could not be everytime)
I opened [this other PR](https://github.com/import-js/eslint-plugin-import/pull/2472) to validate this hypothesis (expected output: the new test I added there is red too), I'll close it 

</details>

As it seems be working with the `from` and `except` parts, I change to use `Minimatch.match` as it's done there, it should fix this for real: https://github.com/import-js/eslint-plugin-import/pull/2466/commits/4392f46c22d062dedacc68142e6b92530087ba03

AdriAt360:
hi there @ljharb 👋🏽 
as my previous attempt wasn't successful either, here is another try to fix the existing issue on the Windows Native pipeline
- I decided to use the `resolve` function the same way it's [done for the imports](https://github.com/AdriAt360/eslint-plugin-import/blob/de331693613c5a13b6f624367b4b90ba3b454072/src/rules/no-restricted-paths.js#L218) https://github.com/import-js/eslint-plugin-import/pull/2466/commits/de331693613c5a13b6f624367b4b90ba3b454072
- I also fixed the tests accordingly https://github.com/import-js/eslint-plugin-import/pull/2466/commits/2f9ba90677fd40baa22f617466b961be23ae7ecb

I hope this will make the pipeline green 🤞🏽

AdriAt360:
@ljharb how do you want to proceed next here?
- as it was already broken (confirmed by https://github.com/import-js/eslint-plugin-import/pull/2472), I think it would not be more harmful to merge this current version (without my last 2 commits obviously)
- in the meantime, I'm not fond of ignoring a broken pipeline
- but I must admit I run out of new ideas for this windows native pipeline problem

ljharb:
hmm. let's reopen #2472, and update it so that it skips the broken tests in windows. then we can land that, rebase this, and land it.

AdriAt360:
@ljharb I just commented the failing test and removed my previous investigation commits -> this PR seems ready to be merged
the AppVeyor build failed but it seems like a timeout on a random test outside this PR's scope: could you rerun this check?
By the way, feel free to rebase if you feel the 2 commits https://github.com/import-js/eslint-plugin-import/pull/2466/commits/786b9b397ae11f314ebe026cea0bd58f0a5c8e18 and https://github.com/import-js/eslint-plugin-import/pull/2466/commits/1196129605726be1c217e30931e9b2a7234ad062 should be merged into only one

ljharb (approved):

src/rules/no-restricted-paths.js
AdriAt360: resolving here as it seems to be ok

AdriAt360:
@ljharb it seems that this PR is ready to be merged, but feel free to tell if there is still something waiting on what I could be helpful :)

PR #2473, opened 2022-06-09
[New] `consistent-type-specifier-style`: add rule

### LINKED ISSUE #2469: New rule: consistent inline type specifier usage
In both flow and TS you can mark an import as a type (or typeof for flow) via a specifier.
Both languages support two forms:

top-level, which marks all names in the import as type-only:
```ts
import type Foo from 'Foo';
import type {Bar} from 'Bar';
// ts only 
import type * as Bam from 'Bam';
// flow only
import typeof Baz from 'Baz';
```

inline, which marks just one name in the import as type-only:
```ts
import {type Foo} from 'Foo';
// flow only
import {typeof Bar} from 'Bar';
```

Users care about consistency when it comes to the use of inline type imports - they either want them, or they don't.
I would like to propose a rule which enforces either for, or against the use of inline type specifiers.

The rule would have an option which would enforce one way or another, i.e.
```ts
type Options = [
  'perfer-inline' | 'prefer-top-level',
];
```

### ✅ Valid always

```ts
import type Foo from 'Foo'; // default specifiers are always good
import type * as NS from 'Bar'; // namespace specifiers are always good
```

### ✅ Valid with `prefer-top-level`, but ❌ Invalid with `prefer-inline`

```ts
import type {Foo} from 'Foo';
import type Foo, {Bar} from 'Foo';
// flow only
import typeof {Foo} from 'Foo';
```

### ❌ Invalid with `prefer-top-level`, but ✅ Valid with `prefer-inline`

```ts
import {type Foo} from 'Foo';
import Foo, {type Bar} from 'Foo';
// flow only
import {typeof Foo} from 'Foo';
```

## Fixer

For `prefer-top-level`:
```ts
import {Foo, type Bar, typeof Baz} from 'Foo';
// becomes
import {Foo} from 'Foo';
import type {Bar} from 'Bar';
import typeof {Baz} from 'Baz';
```

For `prefer-inline`:
```ts
import type {Foo, Bar} from 'Foo';
// becomes
import {type Foo, type Bar} from 'Foo';

import type Foo, {Bar} from 'Foo';
// becomes
import type Foo from 'Foo';
import {type Bar} from 'Foo';
```

Note that the fixer need not necessarily concern itself with de-duplicating the imports, as `import/no-duplicates` can be turned on by the user to handle that concern.

bradzacher:
Fixes #2469

Adds a new rule as specced in #2469
To test this I had to bump the TS version to 4.5, as that was the first version that supported inline type specifiers.

### DIFF
--- a/docs/rules/consistent-type-specifier-style.md
+++ b/docs/rules/consistent-type-specifier-style.md
@@ -1,0 +1,87 @@
+# import/consistent-type-specifier-style
+
+In both Flow and TypeScript you can mark an import as a type-only import by adding a "kind" marker to the import. Both languages support two positions for marker.
+
+**At the top-level** which marks all names in the import as type-only and applies to named, default, and namespace (for TypeScript) specifiers:
+
+```ts
+import type Foo from 'Foo';
+import type {Bar} from 'Bar';
+// ts only
+import type * as Bam from 'Bam';
+// flow only
+import typeof Baz from 'Baz';
+```
+
+**Inline** with to the named import, which marks just the specific name in the import as type-only. An inline specifier is only valid for named specifiers, and not for default or namespace specifiers:
+
+```ts
+import {type Foo} from 'Foo';
+// flow only
+import {typeof Bar} from 'Bar';
+```
+
+## Rule Details
+
+This rule either enforces or bans the use of inline type-only markers for named imports.
+
+This rule includes a fixer that will automatically convert your specifiers to the correct form - however the fixer will not respect your preferences around de-duplicating imports. If this is important to you, consider using the [`import/no-duplicates`] rule.
+
+[`import/no-duplicates`]: ./no-duplicates.md
+
+## Options
+
+The rule accepts a single string option which may be one of:
+
+- `'prefer-inline'` - enforces that named type-only specifiers are only ever written with an inline marker; and never as part of a top-level, type-only import.
+- `'prefer-top-level'` - enforces that named type-only specifiers only ever written as part of a top-level, type-only import; and never with an inline marker.
+
+By default the rule will use the `prefer-inline` option.
+
+## Examples
+
+### `prefer-top-level`
+
+❌ Invalid with `["error", "prefer-top-level"]`
+
+```ts
+import {type Foo} from 'Foo';
+import Foo, {type Bar} from 'Foo';
+// flow only
+import {typeof Foo} from 'Foo';
+```
+
+✅ Valid with `["error", "prefer-top-level"]`
+
+```ts
+import type {Foo} from 'Foo';
+import type Foo, {Bar} from 'Foo';
+// flow only
+import typeof {Foo} from 'Foo';
+```
+
+### `prefer-inline`
+
+❌ Invalid with `["error", "prefer-inline"]`
+
+```ts
+import type {Foo} from 'Foo';
+import type Foo, {Bar} from 'Foo';
+// flow only
+import typeof {Foo} from 'Foo';
+```
+
+✅ Valid with `["error", "prefer-inline"]`
+
+```ts
+import {type Foo} from 'Foo';
+import Foo, {type Bar} from 'Foo';
+// flow only
+import {typeof Foo} from 'Foo';
+```
+
+## When Not To Use It
+
+If you aren't using Flow or TypeScript 4.5+, then this rule does not apply and need not be used.
+
+If you don't care about, and don't want to standardize how named specifiers are imported then you should not use this rule.
--- a/src/rules/consistent-type-specifier-style.js
+++ b/src/rules/consistent-type-specifier-style.js
@@ -1,0 +1,216 @@
+import docsUrl from '../docsUrl';
+
+function isComma(token) {
+  return token.type === 'Punctuator' && token.value === ',';
+}
+
+function removeSpecifiers(fixes, fixer, sourceCode, specifiers) {
+  for (const specifier of specifiers) {
+    // remove the trailing comma
+    const comma = sourceCode.getTokenAfter(specifier, isComma);
+    if (comma) {
+      fixes.push(fixer.remove(comma));
+    }
+    fixes.push(fixer.remove(specifier));
+  }
+}
+
+function getImportText(
+  node,
+  sourceCode,
+  specifiers,
+  kind,
+) {
+  const sourceString = sourceCode.getText(node.source);
+  if (specifiers.length === 0) {
+    return '';
+  }
+
+  const names = specifiers.map(s => {
+    if (s.imported.name === s.local.name) {
+      return s.imported.name;
+    }
+    return `${s.imported.name} as ${s.local.name}`;
+  });
+  // insert a fresh top-level import
+  return `import ${kind} {${names.join(', ')}} from ${sourceString};`;
+}
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Enforce or ban the use of inline type-only markers for named imports',
+      url: docsUrl('consistent-type-specifier-style'),
+    },
+    fixable: 'code',
+    schema: [
+      {
+        type: 'string',
+        enum: ['prefer-inline', 'prefer-top-level'],
+        default: 'prefer-inline',
+      },
+    ],
+  },
+
+  create(context) {
+    const sourceCode = context.getSourceCode();
+
+    if (context.options[0] === 'prefer-inline') {
+      return {
+        ImportDeclaration(node) {
+          if (node.importKind === 'value' || node.importKind == null) {
+            // top-level value / unknown is valid
+            return;
+          }
+
+          if (
+            // no specifiers (import type {} from '') have no specifiers to mark as inline
+            node.specifiers.length === 0 ||
+            (node.specifiers.length === 1 &&
+              // default imports are both "inline" and "top-level"
+              (node.specifiers[0].type === 'ImportDefaultSpecifier' ||
+                // namespace imports are both "inline" and "top-level"
+                node.specifiers[0].type === 'ImportNamespaceSpecifier'))
+          ) {
+            return;
+          }
+
+          context.report({
+            node,
+            message: 'Prefer using inline {{kind}} specifiers instead of a top-level {{kind}}-only import.',
+            data: {
+              kind: node.importKind,
+            },
+            fix(fixer) {
+              const kindToken = sourceCode.getFirstToken(node, { skip: 1 });
+
+              return [].concat(
+                kindToken ? fixer.remove(kindToken) : [],
+                node.specifiers.map((specifier) => fixer.insertTextBefore(specifier, `${node.importKind} `)),
+              );
+            },
+          });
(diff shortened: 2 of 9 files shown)


bradzacher:
image

big oof - testing against all of these old versions of things is really tough to get right

docs/rules/consistent-type-specifier-style.md
```diff
@@ -0,0 +1,87 @@
...
+import {typeof Bar} from 'Bar';
+```
+
+## Rule Details
+
+This rule either enforces or bans the use of inline type-only markers for named imports.
+
+This rule includes a fixer that will automatically convert your specifiers to the correct form - however the fixer will not respect your preferences around de-duplicating imports. If this is important to you, consider using the [`import/no-duplicates`] rule.
+
+[`import/no-duplicates`]: ./no-duplicates.md
```
snewcomer, line 30: support for `inlineTypeImport` https://github.com/import-js/eslint-plugin-import/pull/2475

tests/src/rules/consistent-type-specifier-style.js
```diff
@@ -0,0 +1,291 @@
...
+      }],
+    },
+    {
+      code: "import { type Foo, Bar } from 'Foo';",
+      output: "import {  Bar } from 'Foo';\nimport type {Foo} from 'Foo';",
+      options: ['prefer-top-level'],
+      errors: [{
+        message: 'Prefer using a top-level type-only import instead of inline type specifiers.',
+        type: 'ImportSpecifier',
+      }],
```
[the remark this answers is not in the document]
bradzacher: ohh that's a really good test case I forgot about.
this one is even better though:
```ts
import AValue, { type Foo } from 'Foo';
```
Because it should fix to
```ts
import AValue from 'Foo';
import type Foo from 'Foo';
```
i.e. we have to remove the empty `{}` as well

I wonder... @ljharb - do you have a preference here?
Should there be a new rule which enforces against empty named specifier blocks, or should we just bring that complexity into each lint rule?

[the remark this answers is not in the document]
ljharb: @bradzacher Isn't that invalid syntax? `import {} from` is already invalid.

If it's valid to have `import type {} from` then i think we just need to handle that in individual rules.

(Sorry for the delayed response; i'd like to land this soon so I'll keep a closer eye on it)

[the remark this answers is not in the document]
bradzacher: > Isn't that invalid syntax? import {} from is already invalid

According to [the spec](https://262.ecma-international.org/6.0/#sec-imports) `{}` is allowed:

```
NamedImports :
    { }
    { ImportsList }
    { ImportsList , }
```

Based on testing - all eslint parsers allow it.
Also the flow/hermes/typescript/babel parsers allow `import type {}`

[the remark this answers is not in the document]
ljharb: oh oof, gross. ok then i think a) we'd need a separate rule to ban empty named import/export blocks, and b) this rule's autofixer should just always be removing empty blocks when removing duplicates.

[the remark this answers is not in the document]
bradzacher: there is *some* additional complexity that will get added to the fixer for this, but we can definitely do it!

The other option is to let this fixer purely focus on moving specifiers around, and then let the new `no-empty-named-blocks` focus on cleaning up the empty curlies - similar to how the fixer right now does not attempt to consolidate imports, and just moves things around.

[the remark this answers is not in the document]
ljharb: I suppose that's an option too - but I think in the common case people will not want empty braces, and it's nice when autofixers optimize for the common case.

That said, certainly if someone _wants_ empty braces for some reason, that's only possible if this autofixer leaves them behind :-/

src/rules/consistent-type-specifier-style.js
```diff
@@ -0,0 +1,174 @@
...
+              const kindToken = sourceCode.getFirstToken(node, { skip: 1 });
+
+              if (kindToken) {
+                yield fixer.remove(kindToken);
+              }
+
+              for (const specifier of node.specifiers) {
+                yield fixer.insertTextBefore(specifier, `${node.importKind} `);
+              }
+            },
```
ljharb, line 55:
```suggestion
            fix(fixer) {
              const kindToken = sourceCode.getFirstToken(node, { skip: 1 });

              return [].concat(
                kindToken ? fixer.remove(kindToken) : [],
                node.specifiers.map((specifier) => fixer.insertTextBefore(specifier, `${node.importKind} `),
              );
            }
```

src/rules/consistent-type-specifier-style.js
```diff
@@ -0,0 +1,174 @@
...
+              },
+              *fix(fixer) {
+                // remove just the specifiers and insert the new imports after
+                yield* removeSpecifiers(typeSpecifiers);
+                yield* removeSpecifiers(typeofSpecifiers);
+                yield fixer.insertTextAfter(node, `\n${newImports}`);
+
+                function* removeSpecifiers(
+                  specifiers,
+                ) {
```
ljharb, line 135:
```suggestion
                function* removeSpecifiers(specifiers) {
```

ljharb (commented):
I'd prefer to avoid use of generators and loops, but that's also something I can clean up once the tests are passing.

How does a Flow user who isn't using the inline `typeof` syntax already, autofix to `import typeof`?

I think that if we can get this to leave behind the empty braces, we'll be OK pending the other comments. I'd love to land this soon.

bradzacher:
Now that the TS 4.8 release is done, I should be able to get back to this this week (hopefully).

ljharb:
I'm holding off on the next release for this :-) looking forward to you getting back to it.

bradzacher:
assuming test pass - this should be good to go now.
I updated the fixer to (a) cleanup unnecessary commas and (b) cleanup empty named blocks

src/rules/consistent-type-specifier-style.js
```diff
@@ -0,0 +1,227 @@
...
+          if (
+            // no specifiers (import type {} from '') have no specifiers to mark as inline
+            node.specifiers.length === 0 ||
+            (node.specifiers.length === 1 &&
+              // default imports are both "inline" and "top-level"
+              (node.specifiers[0].type === 'ImportDefaultSpecifier' ||
+                // namespace imports are both "inline" and "top-level"
+                node.specifiers[0].type === 'ImportNamespaceSpecifier'))
+          ) {
+            return;
```
ljharb, line 39: can this line be covered by a test case?

src/rules/consistent-type-specifier-style.js
```diff
@@ -0,0 +1,227 @@
...
+        const typeSpecifiers = [];
+        const typeofSpecifiers = [];
+        const valueSpecifiers = [];
+        let defaultSpecifier = null;
+        for (const specifier of node.specifiers) {
+          if (specifier.type === 'ImportDefaultSpecifier') {
+            defaultSpecifier = specifier;
+            continue;
+          } else if (specifier.type !== 'ImportSpecifier') {
+            continue;
```
ljharb, line 94: also this one?

src/rules/consistent-type-specifier-style.js
```diff
@@ -0,0 +1,227 @@
...
+          return `import ${kind} {${names.join(', ')}} from ${sourceString};`;
+        }
+      },
+    };
+  },
+};
+
+function isComma(token) {
+  return token.type === 'Punctuator' && token.value === ',';
+}
```
ljharb, line 226: let's move this to the top of the module, so we aren't relying on hoisting

ljharb (commented):
Looks great - have these items been addressed?
> How does a Flow user who isn't using the inline typeof syntax already, autofix to import typeof?

> leave behind the empty braces

bradzacher:
> leave behind the empty braces

The fixer should no longer leave behind empty braces.

> How does a Flow user who isn't using the inline typeof syntax already, autofix to import typeof?

If I'm understanding your question correctly:

- If they don't want to use the inline specifiers at all, then they can use the `prefer-top-level` setting.
- If they want to use inline `type` with top-level `typeof`, the rule does not currently have handling for this (could be a future enhancement if people want it!).

src/rules/consistent-type-specifier-style.js
bradzacher: We could actually delete this check entirely.
It's covered above by the checks on line 80-84.
It's a syntax error to have a namespace specifier with any other kind of specifier.

This is just me doing my due diligence because I'm used to covering all of the bases due to TS types.

Would you like me to delete it?

src/rules/consistent-type-specifier-style.js
bradzacher: This should already be covered by:
- https://github.com/bradzacher/eslint-plugin-import/blob/2df9be6b7edd150c9ac94b2d479ea51f5fbe66d0/tests/src/rules/consistent-type-specifier-style.js#L16-L19
- https://github.com/bradzacher/eslint-plugin-import/blob/2df9be6b7edd150c9ac94b2d479ea51f5fbe66d0/tests/src/rules/consistent-type-specifier-style.js#L40-L43
- https://github.com/bradzacher/eslint-plugin-import/blob/2df9be6b7edd150c9ac94b2d479ea51f5fbe66d0/tests/src/rules/consistent-type-specifier-style.js#L207

and a few others

ljharb: if it were, then i wouldn't expect to see the "no coverage" annotation

src/rules/consistent-type-specifier-style.js
ljharb: if it's not used, then it's probably best to delete it.

ljharb:
Rebased; looks like this is good to go, pending any "uncovered line" annotations (https://github.com/import-js/eslint-plugin-import/pull/2473#discussion_r963255391 and https://github.com/import-js/eslint-plugin-import/pull/2473#discussion_r963255298)

src/rules/consistent-type-specifier-style.js
bradzacher: Weird - I'm not sure what's happening with the coverage report.
Locally I can see it's getting hit in the tests:
image

ljharb (approved):

latin-1:
https://github.com/microsoft/TypeScript/issues/47118

```ts
import type { X } from "...";
```

```ts
import { type X } from "...";
```

These two code snippets behave differently. Perhaps we should add a special treatment for this.

ljharb:
It's not clear what the difference is from that issue. Can you elaborate?

latin-1:
https://github.com/typescript-eslint/typescript-eslint/issues/6338
I posted more details here. I'm not sure which plugin this should be handled in.
It's a mess. We may need two or more rounds of code fixes to make it work.

ljharb:
eslint on the CLI runs multiple passes, so it's only in an editor that you're likely to run into this problem.

Issue #2470, opened 2022-06-04
addition to rule: import/no-duplicates


snewcomer:
The proposal here is to improve this rule and handle inline type modifiers

```
import { type Foo } from 'T';
import { type Bar } from 'T';
//^^^^^^^^^^^^^^^^^^^^ Error: duplicate import from module 'T' (import/no-duplicates)
```

Fixer
```
import { type Foo,  type Bar } from 'T';
```

ref https://github.com/import-js/eslint-plugin-import/issues/2469 and https://github.com/typescript-eslint/typescript-eslint/pull/5050#discussion_r889606802

ljharb:
I absolutely agree this should be autofixable by no-duplicates.

nerdjfpb:
I'm new to the codebase, but can I take this issue?

snewcomer:
@nerdjfpb Happy if you did!  If you don't happen to get to it by this weekend, I'll plan on jumping in.

nerdjfpb:
Some personal issue came up and I don't think I can jump on this in next 1/2 week. I think you can start working on this @snewcomer

PR #2475, opened 2022-06-10
[New] `no-duplicates`: support inline type import with `prefer-inline` option

snewcomer:
close https://github.com/import-js/eslint-plugin-import/issues/2470

ref https://github.com/typescript-eslint/typescript-eslint/pull/5050

Typescript `import { type A } from 'util/cool-linter'` was introduced in 4.5 https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#type-on-import-names

This PR seeks to add TypeScript support to the existing rule `no-duplicates`.

```
❌ Invalid `["error", "prefer-inline"]`

import { AValue, type AType } from './mama-mia'
import type { BType } from './mama-mia'

✅ Valid with `["error", "prefer-inline"]`

import { AValue, type AType, type BType } from './mama-mia'
```

### DIFF
--- a/docs/rules/no-duplicates.md
+++ b/docs/rules/no-duplicates.md
@@ -67,6 +67,33 @@ import SomeDefaultClass from './mod?minify'
 import * from './mod.js?minify'
 ```
 
+### Inline Type imports
+
+TypeScript 4.5 introduced a new [feature](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#type-on-import-names) that allows mixing of named value and type imports. In order to support fixing to an inline type import when duplicate imports are detected, `prefer-inline` can be set to true.
+
+Config:
+
+```json
+"import/no-duplicates": ["error", {"prefer-inline": true}]
+```
+
+<!--tabs-->
+
+❌ Invalid `["error", "prefer-inline"]`
+
+```js
+import { AValue, type AType } from './mama-mia'
+import type { BType } from './mama-mia'
+```
+
+✅ Valid with `["error", "prefer-inline"]`
+
+```js
+import { AValue, type AType, type BType } from './mama-mia'
+```
+
+<!--tabs-->
+
 ## When Not To Use It
 
 If the core ESLint version is good enough (i.e. you're _not_ using Flow and you _are_ using [`import/extensions`](./extensions.md)), keep it and don't use this.
--- a/src/rules/no-duplicates.js
+++ b/src/rules/no-duplicates.js
@@ -1,5 +1,7 @@
 import resolve from 'eslint-module-utils/resolve';
 import docsUrl from '../docsUrl';
+import semver from 'semver';
+import typescriptPkg from 'typescript/package.json';
 
 function checkImports(imported, context) {
   for (const [module, nodes] of imported.entries()) {
@@ -7,7 +9,7 @@ function checkImports(imported, context) {
       const message = `'${module}' imported multiple times.`;
       const [first, ...rest] = nodes;
       const sourceCode = context.getSourceCode();
-      const fix = getFix(first, rest, sourceCode);
+      const fix = getFix(first, rest, sourceCode, context);
 
       context.report({
         node: first.source,
@@ -25,7 +27,7 @@ function checkImports(imported, context) {
   }
 }
 
-function getFix(first, rest, sourceCode) {
+function getFix(first, rest, sourceCode, context) {
   // Sorry ESLint <= 3 users, no autofix for you. Autofixing duplicate imports
   // requires multiple `fixer.whatever()` calls in the `fix`: We both need to
   // update the first one, and remove the rest. Support for multiple
@@ -108,10 +110,19 @@ function getFix(first, rest, sourceCode) {
 
     const [specifiersText] = specifiers.reduce(
       ([result, needsComma], specifier) => {
+        const isTypeSpecifier = specifier.importNode.importKind === 'type';
+
+        const preferInline = context.options[0] && context.options[0]['prefer-inline'];
+        // a user might set prefer-inline but not have a supporting TypeScript version.  Flow does not support inline types so this should fail in that case as well.
+        if (preferInline && !semver.satisfies(typescriptPkg.version, '>= 4.5')) {
+          throw new Error('Your version of TypeScript does not support inline type imports.');
+        }
+
+        const insertText = `${preferInline && isTypeSpecifier ? 'type ' : ''}${specifier.text}`;
         return [
           needsComma && !specifier.isEmpty
-            ? `${result},${specifier.text}`
-            : `${result}${specifier.text}`,
+            ? `${result},${insertText}`
+            : `${result}${insertText}`,
           specifier.isEmpty ? needsComma : true,
         ];
       },
@@ -257,6 +268,9 @@ module.exports = {
           considerQueryString: {
             type: 'boolean',
           },
+          'prefer-inline': {
+            type: 'boolean',
+          },
         },
         additionalProperties: false,
       },
@@ -290,6 +304,9 @@ module.exports = {
       const map = moduleMaps.get(n.parent);
       if (n.importKind === 'type') {
         return n.specifiers.length > 0 && n.specifiers[0].type === 'ImportDefaultSpecifier' ? map.defaultTypesImported : map.namedTypesImported;
+      }
+      if (n.specifiers.some((spec) => spec.importKind === 'type')) {
+        return map.namedTypesImported;
       }
 
       return hasNamespace(n) ? map.nsImported : map.imported;
--- a/tests/src/rules/no-duplicates.js
+++ b/tests/src/rules/no-duplicates.js
@@ -1,5 +1,5 @@
 import * as path from 'path';
-import { test as testUtil, getNonDefaultParsers, parsers } from '../utils';
+import { test as testUtil, getNonDefaultParsers, parsers, tsVersionSatisfies, typescriptEslintParserSatisfies } from '../utils';
 import jsxConfig from '../../../config/react';
 
 import { RuleTester } from 'eslint';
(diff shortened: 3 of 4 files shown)


docs/rules/no-duplicates.md
```diff
@@ -61,6 +61,27 @@ import SomeDefaultClass from './mod?minify'
...
 ```
 
+### Inline Type imports
+
+TypeScript 4.5 introduced a new [feature](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#type-on-import-names) that allows mixing of named value and type imports. In order to support fixing to an inline type import when duplicate imports are detected, `inlineTypeImport` can be set to true.
+
+Config:
+
+```json
+"import/no-duplicates": ["error", {"inlineTypeImport": true}]
```
snewcomer, line 71: > consistent with this [PR](https://github.com/import-js/eslint-plugin-import/pull/2473/files), we could use `prefer-inline`

package.json
```diff
@@ -92,7 +92,7 @@
     "safe-publish-latest": "^2.0.0",
     "semver": "^6.3.0",
     "sinon": "^2.4.1",
-    "typescript": "^2.8.1 || ~3.9.5",
+    "typescript": "^2.8.1 || ^3.9.5 || ~4.7.3",
```
snewcomer, line 95: > Will wait for this PR to get tests passing

https://github.com/import-js/eslint-plugin-import/pull/2473/files#diff-f4543b3f6041f0b4eb2b25d9f4e6030faf96ccad7d8fb3a5209c80eea0b59c41

jasongerbes:
Would it be possible to extend this rule to also test exports?

snewcomer:
@jasongerbes Great q! I'm not seeing inline export type throw any warnings so I'm guessing it is possible!  I didn't see any mention in the release notes.  Either it is an option or a different rule.  After this PR is merged, I imagine we could consider it then.  I don't have an answer right now.

```
export { type A };
```

snewcomer:
cc @ljharb @bradzacher This is ready for your review.

src/rules/no-duplicates.js
```diff
@@ -108,10 +108,17 @@ function getFix(first, rest, sourceCode) {
 
     const [specifiersText] = specifiers.reduce(
       ([result, needsComma], specifier) => {
+        let insertText = specifier.text;
+        const isTypeSpecifier = specifier.importNode.importKind === 'type';
+        const inlineTypeImport = context.options[0] &&
+          context.options[0]['inlineTypeImport'];
+        if (inlineTypeImport && isTypeSpecifier) {
+          insertText = `type ${insertText}`;
+        }
```
ljharb, line 117:
```suggestion
        const isTypeSpecifier = specifier.importNode.importKind === 'type';
        const inlineTypeImport = context.options[0] && context.options[0]['inlineTypeImport'];
        const insertText = `${inlineTypeImport && isTypeSpecifier ? 'type ' : ''}${specifier.text}`;
```

tests/src/rules/no-duplicates.js
```diff
@@ -522,5 +522,99 @@ context('TypeScript', function () {
           }),
         ],
       });
+
+      if (!tsVersionSatisfies('>= 4.5') || !typescriptEslintParserSatisfies('>= 5.7.0')) {
+        return;
+      }
+
+      ruleTester.run('no-duplicates inline type', rule, {
```
ljharb, line 530: it'd be nice to use `[].concat()` and a ternary (adding the test cases or `[]`) rather than a totally separate `ruleTester.run`, but it's not a blocker for the PR.

ljharb (commented):
Looks good!

snewcomer:
Hi @ljharb!  Just seeing if I could get another friendly review when/if you have time!

src/rules/no-duplicates.js
```diff
@@ -108,10 +108,13 @@ function getFix(first, rest, sourceCode) {
 
     const [specifiersText] = specifiers.reduce(
       ([result, needsComma], specifier) => {
+        const isTypeSpecifier = specifier.importNode.importKind === 'type';
+        const inlineTypeImport = context.options[0] && context.options[0]['inlineTypeImport'];
```
ljharb, line 112:
```suggestion
        const inlineTypeImport = context.options[0] && context.options[0].inlineTypeImport;
```

src/rules/no-duplicates.js
```diff
@@ -290,6 +296,8 @@ module.exports = {
       const map = moduleMaps.get(n.parent);
       if (n.importKind === 'type') {
         return n.specifiers.length > 0 && n.specifiers[0].type === 'ImportDefaultSpecifier' ? map.defaultTypesImported : map.namedTypesImported;
+      } else if (n.specifiers.some((spec) => spec.importKind === 'type')) {
```
ljharb, line 299:
```suggestion
      }
      if (n.specifiers.some((spec) => spec.importKind === 'type')) {
```

ljharb (commented):
Looks good! I'm comfortable with this option name, but `preferInline` would work as well.

What about someone who supports inline type imports, but wants to force non-inline ones? Perhaps this should be an option with a value of "separate" versus "inline"?

snewcomer:
@ljharb Thanks for the reviews and comments!  So this PR is part of a larger effort.

1. fixes a Type import to Inline-Type via `prefer-inline` or `prefer-top-level` - https://github.com/import-js/eslint-plugin-import/pull/2473. 
2. Fixes a Value import to Inline Type - https://github.com/typescript-eslint/typescript-eslint/pull/5050.
3. (this PR) adds support for resolving duplicate Type imports to Inline Types to a single import source.

I think 1) is what you are referencing, correct?

snewcomer:
I updated to `prefer-inline` to be consistent with https://github.com/import-js/eslint-plugin-import/pull/2473.  Does that seem ok?

ljharb:
Yep, 1 is what I'm referencing

tests/src/rules/no-duplicates.js
```diff
@@ -428,99 +428,225 @@ context('TypeScript', function () {
...
+              message: "'./foo' imported multiple times.",
+            },
+          ],
+        }),
+      ].concat(!tsVersionSatisfies('>= 4.5') || !typescriptEslintParserSatisfies('>= 5.7.0') ? [] : [
+        test({
+          code: "import {type x} from './foo'; import type {y} from './foo'",
+          ...parserConfig,
+          options: [{ 'prefer-inline': false }],
+          output: `import {type x,y} from './foo'; `,
```
snewcomer, line 543: I want to point out this test.  no-duplicates will collapse these imports like this even before this PR.  Is there a different way we would want to handle this case?  When combined with this [PR](https://github.com/typescript-eslint/typescript-eslint/pull/5050), it would convert `y` to `type y`.  @ljharb

ljharb: Is this a difference in Flow vs TS?

snewcomer: I believe so, but I don't think I fully grok the question as well 😄 so sorry if not a direct answer.  From my very limited knowledge, Flow doesn't have inline type imports.  So this test is a bit obtuse and really just shows that for someone who wrote their imports this way, they still have a path (with other rules) to correct the imports to proper inline types.

ljharb: I thought Flow used to be the only thing that has inline type imports, and TS only recently added them?

snewcomer: Oh!  I'm not familiar with Flow but did some looking.  I'm only seeing top level type imports but I didn't spend a lot of time looking.  And yep, TS added them in 4.5 so this is the last PR in a [bucket](https://github.com/import-js/eslint-plugin-import/pull/2475#issuecomment-1304888131) of 3 PRs to allow codebases to migrate from top level (separate from value imports) type imports to inline type imports.

ljharb: If there’s any autofix that would conflict based on using TS or Flow, then for those we need to use suggestions instead.

snewcomer: One thing to note is that this will only happen if TS > 4.5.  A Flow project would presumably fail upstream if someone tried `import { type x }`.  So I think this test case is not applicable to Flow

snewcomer: @bradzacher Do you have any additional pieces of info to add here?

ljharb: @snewcomer meaning, the rule detect TS 4.5+ and only does this autofix in that case?

snewcomer: More-so you won't even be able to build the project if you use invalid syntax.  In this case below, this is a project TS < 4.5.  I would assume similar things would happen in Flow if using inline types (assuming they don't have support for them at all)

Screen Shot 2022-11-25 at 7 17 23 AM

ljharb: Right - and an eslint autofix should never cause a project to become unbuildable or change semantics. That's what "safe" means in the context of autofixing.

snewcomer: So to clarify before I investigate an addition here - I need to prevent the case where somebody erroneously sets ‘prefer-inline: true’ but their TS version (or Flow) doesn’t support inlining types?

snewcomer: > erroneously sets ‘prefer-inline: true’

To clarify further, meaning they opt-into something that their project does not support.

snewcomer: I'm sorry to bother.  Always waffling whether to bother busy maintainers like yourself.  Do you happen to have a thought here?

> So to clarify before I investigate a new commit to this PR - I need to prevent the case where somebody erroneously sets ‘prefer-inline: true’ but their TS version (or Flow) doesn’t support inlining types?

snewcomer: Thank you @ljharb!  How does [this](https://github.com/import-js/eslint-plugin-import/pull/2475/commits/f495fd931c8b48d300f5ce9d3c7274f2fffb35b6) look?

ljharb: Looks great!

ljharb (approved):
Looks great!

ljharb:
I'm not worried about the node 6 travis failure (it's due to npm 6 breaking it, and travis not yet updating nvm to work around it) nor about the uncovered line.

src/rules/no-duplicates.js
```diff
@@ -291,6 +305,9 @@ module.exports = {
       if (n.importKind === 'type') {
         return n.specifiers.length > 0 && n.specifiers[0].type === 'ImportDefaultSpecifier' ? map.defaultTypesImported : map.namedTypesImported;
       }
+      if (n.specifiers.some((spec) => spec.importKind === 'type')) {
+        return map.namedTypesImported;
+      }
```
snewcomer, line 310: This seems to be the source of error for a few folks

PR #2481, opened 2022-06-22
[New] `no-extranious-dependencies`: Support package.json files in parent folders

luxaritas:
Resolves https://github.com/import-js/eslint-plugin-import/issues/1913

I am not particularly happy with the implementation (in particular, the error passing and "requireExact" handling to match the current behavior feel misplaced and/or mishandled), but it's at the point that I feel it would be useful to get feedback (both on how the options are handled and the implementation).


docs/rules/no-extraneous-dependencies.md
```diff
@@ -17,6 +17,22 @@ This rule supports the following options:
...
+`devDependencies`. If set to `false`, it will override `considerInParents` for `devDependencies`.
+
+`considerOptionalInParents`: If set to `true`, then the rule will take `package.json` files in parent directories into account for
+`optionalDependencies`. If set to `false`, it will override `considerInParents` for `optionalDependencies`.
+
+`considerPeerInParents`: If set to `true`, then the rule will take `package.json` files in parent directories into account for
+`peerDependencies`. If set to `false`, it will override `considerInParents` for `peerDependencies`.
+
+`considerBundledInParents`: If set to `true`, then the rule will take `package.json` files in parent directories into account for
+`bundledDependencies`. If set to `false`, it will override `considerInParents` for `bundledDependencies`.
```
ljharb, line 33: oof, this is a lot of new schema options to add. maybe `considerInParents` takes an array which must have 1-5 of "prod", "dev", "peer", "optional", "bundled" in it?

src/rules/no-extraneous-dependencies.js
```diff
@@ -30,7 +29,64 @@ function extractDepFields(pkg) {
...
+  }
+
+  if (parentOptions.considerDev || (parentOptions.consider && parentOptions.considerDev !== false)) {
+    packages.forEach(pkg => Object.assign(packageContent.devDependencies, pkg.devDependencies));
+  } else {
+    Object.assign(packageContent.devDependencies, packages[0].devDependencies);
+  }
+
+  if (parentOptions.considerOptional || (parentOptions.consider && parentOptions.considerOptional !== false)) {
+    packages.forEach(pkg => Object.assign(packageContent.optionalDependencies, pkg.optionalDependencies));
```
ljharb, line 69:
```suggestion
    packages.forEach(pkg => { Object.assign(packageContent.optionalDependencies, pkg.optionalDependencies); });
```
since we're not using the arrow functions' return value (same elsewhere)

utils/readPkgUp.js
```diff
@@ -35,10 +36,40 @@ function stripBOM(str) {
...
+  if (opts.multi) {
+    const allPackages = [];
+    let searchPath = opts.cwd;
+    const root = path.parse(searchPath).root;
+    while (searchPath) {
+      const pkg = readPkgUp(Object.assign({}, opts, {
+        multi: false,
+        cwd: searchPath,
+        requireExact: opts.requireExact && allPackages.length === 0,
+      }));
```
ljharb, line 48: it would be ideal not to complicate this largely inlined dependency with a new option.

luxaritas:
FYI, I have seen these comments (thanks! good points). Hopefully I'll have time within the next week or two to address them.

luxaritas:
Sorry for the delay! Rewrote this with your feedback and it is significantly simpler. :) As a bonus, this should also improve caching, since it's no longer necessary to touch the file system to look for the closest package.json. Let me know if there's anything further I can improve!

src/rules/no-extraneous-dependencies.js
```diff
@@ -268,6 +291,7 @@ module.exports = {
           'packageDir': { 'type': ['string', 'array'] },
           'includeInternal': { 'type': ['boolean'] },
           'includeTypes': { 'type': ['boolean'] },
+          'considerInParents': { 'type': ['array'] },
```
ljharb, line 294:
```suggestion
          considerInParents: {
             type: 'array',
             uniqueItems: true,
             additionalItems: false,
             items: {
               type: 'string',
               enum: [
                'prod',
                'dev',
                'peer',
                'bundled',
                'optional',
              ],
```

docs/rules/no-extraneous-dependencies.md
```diff
@@ -37,7 +37,16 @@ There are 2 boolean options to opt into checking extra imports that are normally
 "import/no-extraneous-dependencies": ["error", {"includeInternal": true, "includeTypes": true}]
 ```
 
-Also there is one more option called `packageDir`, this option is to specify the path to the folder containing package.json.
+To take `package.json` files in parent directories into account, use the `considerInParents` option.
+This takes an array of strings, which can include `"prod"`, `"dev"`, `"peer"`, `"optional"`, `"bundled"`.
```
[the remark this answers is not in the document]
ljharb: can we specify what the default value is here? (iow, it's an array of all 5, right?)

ljharb (commented):
Looks great!

src/rules/no-extraneous-dependencies.js
luxaritas: - You removed the quotes around considerInParents were removed, but all the others in the schema have quotes - is there a reason for the difference or should it be one or the other consistently?
- It might be a good idea to update the schema for the other items to specify that they're arrays of strings? Though that would presumably be best in another PR?

docs/rules/no-extraneous-dependencies.md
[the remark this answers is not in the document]
luxaritas: The default I used was an empty array, to match the current behavior. I could see a good argument for changing it (it's technically more correct as far as node's package resolution), but I imagine that would be a breaking change. I'll leave that decision up to you!

src/rules/no-extraneous-dependencies.js
ljharb: doesn't matter here, but i'll be enabling an eslint rule to properly omit the quotes where possible :-)

While it would be great to do that, that'd be a breaking change, and it's not worth doing one just for that.

docs/rules/no-extraneous-dependencies.md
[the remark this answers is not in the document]
ljharb: I think that’s fine! Let’s explicitly mention in the docs that an empty array is the default.

luxaritas:
Resolved all the feedback - let me know if you want me to squash.

WangHoHan:
What is the status of this PR?

luxaritas:
Waiting for rereview as far as I'm concerned

src/rules/no-extraneous-dependencies.js
```diff
@@ -277,7 +315,7 @@ module.exports = {
   create(context) {
     const options = context.options[0] || {};
     const filename = context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename();
-    const deps = getDependencies(context, options.packageDir) || extractDepFields({});
+    const deps = getDependencies(context, options.packageDir, options.considerInParents || []) || extractDepFields({});
```
ljharb, line 318:
```suggestion
    const deps = getDependencies(context, options.packageDir, new Set(options.considerInParents || [])) || extractDepFields({});
```

src/rules/no-extraneous-dependencies.js
```diff
@@ -40,16 +40,41 @@ function extractDepFields(pkg) {
...
+    if ((!parsedPackage || considerInParents.length > 0) && packageDir !== path.parse(packageDir).root) {
+      const parentDepFields = getPackageDepFields(path.dirname(packageDir), considerInParents, false);
+
+      Object.keys(depFields).forEach(depsKey => {
+        if (
+          (depsKey === 'dependencies' && includes(considerInParents, 'prod') )||
+          (depsKey === 'devDependencies' && includes(considerInParents, 'dev')) ||
+          (depsKey === 'peerDependencies' && includes(considerInParents, 'peer')) ||
+          (depsKey === 'optionalDependencies' && includes(considerInParents, 'optional')) ||
+          (depsKey === 'bundledDependencies' && includes(considerInParents, 'bundled'))
```
ljharb, line 64:
```suggestion
          (depsKey === 'dependencies' && considerInParents.has('prod')) ||
          (depsKey === 'devDependencies' && considerInParents.has('dev')) ||
          (depsKey === 'peerDependencies' && considerInParents.has('peer')) ||
          (depsKey === 'optionalDependencies' && considerInParents.has('optional')) ||
          (depsKey === 'bundledDependencies' && considerInParents.has('bundled'))
```

src/rules/no-extraneous-dependencies.js
```diff
@@ -40,16 +40,41 @@ function extractDepFields(pkg) {
   };
 }
 
-function getPackageDepFields(packageJsonPath, throwAtRead) {
-  if (!depFieldCache.has(packageJsonPath)) {
-    const depFields = extractDepFields(readJSON(packageJsonPath, throwAtRead));
-    depFieldCache.set(packageJsonPath, depFields);
+function getPackageDepFields(packageDir, considerInParents, requireInDir) {
+  const cacheKey = JSON.stringify({ packageDir, considerInParents });
```
ljharb, line 44:
```suggestion
  const cacheKey = JSON.stringify({ packageDir, considerInParents: [...considerInParents] });
```

src/rules/no-extraneous-dependencies.js
```diff
@@ -40,16 +40,41 @@ function extractDepFields(pkg) {
...
+  if (!depFieldCache.has(cacheKey)) {
+    // try in the current directory, erroring if the user explicitly specified this directory
+    // and reading fails
+    const parsedPackage = readJSON(path.join(packageDir, 'package.json'), requireInDir);
+    const depFields = extractDepFields(parsedPackage || {});
+
+    // If readJSON returned nothing, we want to keep searching since the current directory didn't
+    // have a package.json. Also keep searching if we're merging in some set of parents dependencies.
+    // However, if we're already at the root, stop.
+    if ((!parsedPackage || considerInParents.length > 0) && packageDir !== path.parse(packageDir).root) {
```
ljharb, line 55:
```suggestion
    if ((!parsedPackage || considerInParents.size > 0) && packageDir !== path.parse(packageDir).root) {
```

ljharb (changes requested):
This looks great! Sorry for the long review delay.

My comments are about converting the internal `considerInParents` to a Set for perf. It'd also be nice to add more tests such that all of the dependency types are covered.

luxaritas:
Interesting idea to use set. I'd be interested to see that benchmarked given the small sizes and additional spread when creating the cache key, though I wouldn't immediately know how to do that myself. :) I can certainly add a few more tests.

ljharb:
yeah lol for such a small array it probably doesn't matter, but might as well use the proper data structure :-p

luxaritas:
@ljharb Can you help me understand why two of the tests I added are failing? It looks like this rule is trying to `resolve` any found import, even if it isn't present, and returns successfully if it isn't found? That.. can't be right, can it?

ljharb:
Hmm, no, I don't know why it's failing :-/ i just rebased it so results are fresh.

florian-bitgrip:
Are there any updates on this? I would really love this feature, since it prevents specifying lots of workspaces in the eslintrc file to work around nesting

ljharb:
@florian-bitgrip if you'd like to help, please comment with a link to a rebased branch with test failures fixed (NOT a new PR) and I'll pull the changes into here.

Thompson1985 (approved):

PR #2483, opened 2022-06-22
Add an option for `no-unresolved` rule that allows one to resolve .ts files from .js imports

ukstv:
One problem that prevents this package from being useable on TS codebase, is that it treats a file extension literally. TypeScript-based ES modules require you to import TS files using .js extension:
```
import * as foo from './foo.js'
```
The statement above really imports content of `./foo.ts` file.

This PR adds a new boolean option `ts`, set to `false` by default. If it is set to `true`, a statement above in a TS codebase is considered valid. If the plugin can not resolve `./foo.js`, it tries to also resolve `./foo.ts`, and then goes on as before.

### DIFF
--- a/src/rules/no-unresolved.js
+++ b/src/rules/no-unresolved.js
@@ -17,6 +17,7 @@ module.exports = {
 
     schema: [
       makeOptionsSchema({
+        ts: { type: 'boolean', default: false },
         caseSensitive: { type: 'boolean', default: true },
         caseSensitiveStrict: { type: 'boolean', default: false },
       }),
@@ -34,8 +35,15 @@ module.exports = {
 
       const caseSensitive = !CASE_SENSITIVE_FS && options.caseSensitive !== false;
       const caseSensitiveStrict = !CASE_SENSITIVE_FS && options.caseSensitiveStrict;
+      const useTS = options.ts;
 
-      const resolvedPath = resolve(source.value, context);
+      let resolvedPath = resolve(source.value, context);
+      if (useTS) {
+        resolvedPath = resolvedPath ||
+          resolve(source.value.replace(/\.js$/, '.ts'), context) ||
+          resolve(source.value.replace(/\.jsx$/, '.tsx'), context);
+      }
+
 
       if (resolvedPath === undefined) {
         context.report(
--- a/tests/files/app.tsx
+++ b/tests/files/app.tsx
@@ -1,0 +1,1 @@
+export default function App() {};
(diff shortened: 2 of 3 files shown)


ntucker:
probably want to have it try .tsx as well

ukstv:
Okay, now it tries to resolve `.jsx?` to `.ts` or `.tsx`.

Maxim-Mazurok (changes requested):
I think that some tests for this need to be added to https://github.com/import-js/eslint-plugin-import/blob/main/tests/src/rules/no-unresolved.js

ukstv:
@Maxim-Mazurok Added the tests. What is left for the PR to be merged?

Maxim-Mazurok:
Thanks! I think the only thing left is to get some maintainer to approve it. I didn't really check the code, but now that we have tests I think that maintainers approval is more likely to happen.

ukstv:
Good. Let me properly rebase that, and it is ready to merge then.

ukstv:
Rebased now. Hope it is good to merge. Feel free to add comments/critique, so that I can amend the PR.

ljharb:
Since this deals with TS-specific resolution, is there a reason this couldn't/shouldn't be an option on the TS eslint resolver? (cc @bradzacher)

ukstv:
Should it be a default there, or hidden behind a flag, like here?

bradzacher:
@ukstv consider raising an issue there to discuss with the maintainers! You may find there is already one?

JounQin:
I don't understand, https://github.com/import-js/eslint-import-resolver-typescript already supports `exports` field in `package.json`, this is an unnecessary change in the core module.

ukstv:
> I don't understand, https://github.com/import-js/eslint-import-resolver-typescript already supports `exports` field in `package.json`, this is an unnecessary change in the core module.

`exports` clause in `package.json` has nothing to do with this change. What we are trying to achieve is having a check if a local `./foo.js` exists because of corresponding `./foo.ts` exists.

JounQin:
> What we are trying to achieve is having a check if a local ./foo.js exists because of corresponding ./foo.ts exists.

It is also supported by https://github.com/import-js/eslint-import-resolver-typescript

JounQin:
As above, please use https://github.com/import-js/eslint-import-resolver-typescript instead.

ntucker:
I cannot find the equivalent of "import/order" in [eslint-import-resolver-typescript](https://github.com/import-js/eslint-import-resolver-typescript).

ljharb:
How is that unsafe?

JounQin:
> [import-js/eslint-plugin-import#importextensions](https://github.com/import-js/eslint-plugin-import#importextensions) the docs probably need to be updated to recommend installing this instead of adding extensions which is unsafe?

I'm going to PR to add a section about TypeScript usage in README, if you want to do that, feel free to tell me and take it.

PR #2491, opened 2022-07-07
Default-rule implemented with parsers.all

greatsage-raphael:
Closes #2482.

### DIFF
--- a/tests/src/rules/default.js
+++ b/tests/src/rules/default.js
@@ -228,9 +228,6 @@ ruleTester.run('default', rule, {
           ],
         }),
    ]),
-
-   
-
 });
 
 // #311: import of mismatched case


tests/src/rules/default.js
```diff
@@ -1,305 +1,229 @@
 import path from 'path';
-import { test, testVersion, SYNTAX_CASES, getTSParsers, parsers } from '../utils';
+import { test, testVersion, SYNTAX_CASES, parsers as utilsParsers, getTSParsers } from '../utils';
+import  parsers  from '../parsers';
 import { RuleTester } from 'eslint';
 import semver from 'semver';
 import { version as tsEslintVersion } from 'typescript-eslint-parser/package.json';
+import fromEntries from "object.fromentries"
+
+
```
ljharb, line 9:
```suggestion
```

tests/src/rules/default.js
```diff
@@ -1,305 +1,229 @@
...
+ 
 
 import { CASE_SENSITIVE_FS } from 'eslint-module-utils/resolve';
 
-const ruleTester = new RuleTester();
+
 const rule = require('rules/default');
 
+
+
```
ljharb, line 17:
```suggestion
```

tests/src/rules/default.js
```diff
@@ -1,305 +1,229 @@
...
+const parserOptions = {
+  ecmaVersion: 2018,
+  sourceType: 'module',
+  ecmaFeatures: {
+    jsx: true,
+  },
+
+  settings: {
+    "import/parsers": fromEntries(getTSParsers().map((parser)=>[parser, [".ts"]] ))
+
```
ljharb, line 27:
```suggestion
  settings: {
    'import/parsers': fromEntries(getTSParsers().map((parser) => [parser, ['.ts']]))
```
you may want to enable eslint autofixing in vscode :-)

tests/src/rules/default.js
```diff
@@ -1,305 +1,229 @@
...
+    "import/parsers": fromEntries(getTSParsers().map((parser)=>[parser, [".ts"]] ))
+     
+  }
+};
+
+const ruleTester = new RuleTester({ parserOptions });
+
 ruleTester.run('default', rule, {
-  valid: [].concat(
+  valid: parsers.all([
```
ljharb, line 34:
```suggestion
  valid: parsers.all([].concat(
```

tests/src/rules/default.js
```diff
@@ -1,305 +1,229 @@
...
+    // es2022: Arbitrary module namespace identifier names
+    testVersion('>= 8.7', () => ({
+      code: 'export { "default" as bar } from "./bar"',
+      parserOptions: {
+        ecmaVersion: 2022,
+      },
+    })),
+
+    ...SYNTAX_CASES,
+   ]),
```
ljharb, line 158:
```suggestion
     SYNTAX_CASES,
   )),
```
the spread is no longer needed if we use concat

tests/src/rules/default.js
```diff
@@ -1,305 +1,229 @@
...
+
+  invalid: parsers.all([
```
ljharb, line 160:
```suggestion
  invalid: parsers.all([].concat(
```

tests/src/rules/default.js
```diff
@@ -1,305 +1,229 @@
...
+      code: 'import twofer from "./broken-trampoline"',
+      errors: ['No default export found in imported module "./broken-trampoline".'],
+    }),
+
+    // #328: * exports do not include default
+    test({
+      code: 'import barDefault from "./re-export"',
+      errors: ['No default export found in imported module "./re-export".'],
+    }),
+
```
ljharb, line 195:
```suggestion
```

tests/src/rules/default.js
```diff
@@ -1,305 +1,229 @@
...
+
+    //typescript context
+    //context('TypeScript', function ()
```
ljharb, line 198:
```suggestion
```

tests/src/rules/default.js
```diff
@@ -1,305 +1,229 @@
...
+      errors: ['No default export found in imported module "./re-export".'],
+    }),
+
+
+    //typescript context
+    //context('TypeScript', function () 
+    test({
           code: `import foobar from "./typescript"`,
-          parser,
+
```
ljharb, line 201:
```suggestion
          features: ['ts'],
```
and same for all the TS cases

tests/src/rules/default.js
```diff
@@ -1,305 +1,229 @@
...
         test({
           code: `import Foo from "./typescript-export-as-default-namespace"`,
-          parser,
           settings: {
-            'import/parsers': { [parser]: ['.ts'] },
             'import/resolver': { 'eslint-import-resolver-typescript': true },
           },
-          parserOptions: {
-            tsconfigRootDir: path.resolve(__dirname, '../../files/typescript-no-compiler-options/'),
-          },
```
ljharb, line 302: i think we still need this one

tests/src/rules/default.js
```diff
@@ -310,7 +234,25 @@ context('TypeScript', function () {
             },
           ],
         }),
-      ],
-    });
-  });
+   ]),
+
+   
+
```
ljharb, line 240:
```suggestion
```

tests/src/rules/default.js
```diff
@@ -310,7 +234,25 @@ context('TypeScript', function () {
...
+      }),
+     ]),
+    invalid:parsers.all([ 
+      test({
+        code: 'import bar from "./Named-Exports"',
+        errors: ['No default export found in imported module "./Named-Exports".'],
+      }),
+    ]),
+  });
+}
```
ljharb, line 258: these can be inlined as well, with something like:
```jsx
CASE_SENSITIVE_FS ? [] : test({
  code: 'import foo from "./jsx/MyUncoolComponent.jsx"',
}),
```
as long as the lists are using concat

ljharb (changes requested):

tests/src/rules/default.js
greatsage-raphael: Thank you. I'll do that proptly

tests/src/rules/default.js
greatsage-raphael: okay.

tests/src/rules/default.js
greatsage-raphael: Just that one or should it be all the parserOptions should be re-added

ljharb: We should be keeping any parserOptions that are specific to the test case and can't be generically inferred from features

greatsage-raphael: okay

PR #2504, opened 2022-07-20
feat: replace tsconfig-paths + typescript with get-tsconfig

JounQin:


vendors/get-tsconfig.js
```diff
@@ -0,0 +1,4 @@
+// source https://github.com/privatenumber/get-tsconfig
```
JounQin, line 1: We can not use the npm package because it's using `optional catch binding`.

.eslintignore
```diff
@@ -7,6 +7,7 @@ tests/files/with-syntax-error
 tests/files/just-json-files/invalid.json
 tests/files/typescript-d-ts/
 resolvers/webpack/test/files
+vendors
```
ljharb, line 10: Vendoring is an obsolete and harmful practice, so if get-tsconfig doesn't support the node versions we support (by transpiling down to them, testing on them, and explicitly declaring engines for them), and is unwilling to do so, then this is a nonstarter.

Are they willing to do so?

ljharb (changes requested):

.eslintignore
JounQin: Maybe I can fork a down-leveled version?

ljharb: That doesn't seem much better than just sticking with the older tsconfig-paths that we have.

JounQin: get-tsconfig replaces tsconfig-paths + TypeScript at the same time, and it improves the performance a lot.

JounQin: And also `jsconfig.json` is supported without breaking.

JounQin: And how is vendors worse then pin old version of `tsconfig-paths`? A pinned version is just same as vendors, and we can also try to find a better way to downlevel.

package.json
```diff
@@ -18,7 +18,8 @@
     "memo-parser/{*.js,LICENSE,*.md}"
   ],
   "scripts": {
-    "prebuild": "rimraf lib",
+    "downlevel-get-tsconfig": "sed 's/catch{/catch(_){/g; s/Object.getOwnPropertyDescriptors/Object.getOwnPropertyDescriptors||require(\"object.getownpropertydescriptors\")/g; s/Object.getOwnPropertySymbols/Object.getOwnPropertySymbols||require(\"get-own-property-symbols\")/g' node_modules/get-tsconfig/dist/index.js > src/get-tsconfig.js",
```
JounQin, line 21: @ljharb How about this one? We  use the npm module and add compatibility for own supported versions.

.eslintignore
ljharb: Including minified code in a package that is not generated by the repo itself causes tons of false positives on security scanning tools, and is a wonderful way to create a future problem when someone sneaks malicious code into it.

We are simply not going to vendor; it's not an acceptable choice.

Pinning an old version is 100% fine; it's a separate dependency, it's immutable. It is in no way the same as vendoring.

ljharb: It's open source, you're free to do that.

JounQin: Also it maybe time to consider a breaking change instead for this project. Then a lot of legacy dependency `issues` can be resolved.

package.json
JounQin: This is very different with original `vendors` approach.

ljharb: That still requires us to check in minified code output that we didn't author, and then patch it.

.eslintignore
ljharb: I don't see sufficient value in doing so.

JounQin: requiring `typescript` inside is not effective method for performance.

package.json
JounQin: We have test cases, I don't think that would be a big issue.

JounQin:
For anyone may be interested, I've just forked this plugin as `eslint-plugin-i` at https://github.com/JounQin/eslint-plugin-i#eslint-plugin-i, they should just be compatible with any version of each other except `eslint-plugin-i` uses [get-tsconfig](https://github.com/privatenumber/get-tsconfig) to replace [tsconfig-paths](https://github.com/dividab/tsconfig-paths) and heavy [typescript](https://github.com/microsoft/TypeScript) under the hood, so `project` option and `jsconfig.json` are supported out of box.

So you can just use `npm install -D eslint-plugin-import@npm:eslint-plugin-i` to replace original `eslint-plugin-import`, that's all.

And It will rebase and try to release in order to sync with the upstream every day, see [.github/workflows/rebase-upstream.yml](https://github.com/JounQin/eslint-plugin-i/blob/fork-release/.github/workflows/rebase-upstream.yml) for details.

MichaelDeBoey:
@ljharb Is there anything we still need in order to get this one merged? 🤔

If so, is there anything I could to to help out?

ljharb:
@MichaelDeBoey its not clear it can be, since get-tsconfig doesn’t match our support matrix either, and I’m highly disinclined to ever vendor a dep in this post-90s world.

Issue #2461, opened 2022-05-30
`import/no-cycle` does not detect cycles when using named exports with a source


BenoitZugmeyer:
## Overview

I am facing an issue where the `no-cycle` rule fails to detect cycles when using named exports with a source (ex: `export { bar } from "./bar.js"`). The issue does not occur when using "export all" (ex: `export * from "./bar.js"`).

## How to reproduce

```
$ cat .eslintrc.js
module.exports = {
  plugins: ["import"],
  parserOptions: {
    ecmaVersion: "latest",
    sourceType: "module",
  },
  rules: {
    "import/no-cycle": "error",
  },
};

$ cat index.js
export { bar } from "./reexport.js";
export const foo = 1;

$ cat reexport.js
export { bar } from "./bar.js";

$ cat bar.js
import { foo } from "./index.js";
export const bar = foo + 1;

$ npx eslint .
(no output)
```

### Actual result

The dependency cycle is not detected

### Expected result

The dependency cycle is detected. For example, when changing `reexport.js` to an "export all", the dependency loop is correctly detected:

```
$ cat reexport.js
export * from "./bar.js"

$ npx eslint .
/.../index.js
  1:1  error  Dependency cycle via ./bar.js:1  import/no-cycle

✖ 1 problem (1 error, 0 warnings)
```

## Fix suggestion

It seems we don't capture dependencies of `ExportNamedDeclaration` nodes. This patch seems to fix the issue:
```diff
diff --git src/ExportMap.js src/ExportMap.js
index e18797a..fdaecec 100644
--- src/ExportMap.js
+++ src/ExportMap.js
@@ -611,6 +611,9 @@ ExportMap.parse = function (path, content, context) {
     }
 
     if (n.type === 'ExportNamedDeclaration') {
+      const getter = captureDependency(n, n.exportKind === 'type');
+      if (getter) m.dependencies.add(getter);
+
       // capture declaration
       if (n.declaration != null) {
         switch (n.declaration.type) {
```

JounQin:
Can you raise a PR and add the test case?

BenoitZugmeyer:
See https://github.com/import-js/eslint-plugin-import/pull/2511

PR #2511, opened 2022-07-27
[Fix] `no-cycle`: add ExportNamedDeclaration statements to dependencies

BenoitZugmeyer:
This PR is a fix proposal for #2461. It adds `ExportNamedDeclaration`  to dependencies in the same way as `ImportDelcaration` does. I'm not familiar with the logic here, so please review carefully.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -501,6 +501,26 @@ ExportMap.parse = function (path, content, context) {
     m.reexports.set(s.exported.name, { local, getImport: () => resolveImport(nsource) });
   }
 
+  function captureDependencyWithSpecifiers(n) {
+    // import type { Foo } (TS and Flow)
+    const declarationIsType = n.importKind === 'type';
+    // import './foo' or import {} from './foo' (both 0 specifiers) is a side effect and
+    // shouldn't be considered to be just importing types
+    let specifiersOnlyImportingTypes = n.specifiers.length > 0;
+    const importedSpecifiers = new Set();
+    n.specifiers.forEach(specifier => {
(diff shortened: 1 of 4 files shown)


src/ExportMap.js
```diff
@@ -497,6 +497,28 @@ ExportMap.parse = function (path, content, context) {
...
+    const declarationIsType = n.importKind === 'type';
+    // import './foo' or import {} from './foo' (both 0 specifiers) is a side effect and
+    // shouldn't be considered to be just importing types
+    let specifiersOnlyImportingTypes = n.specifiers.length;
+    const importedSpecifiers = new Set();
+    n.specifiers.forEach(specifier => {
+      if (supportedImportTypes.has(specifier.type)) {
+        importedSpecifiers.add(specifier.type);
+      }
+      if (specifier.type === 'ImportSpecifier') {
```
ljharb, line 511: should the more specific check come first, and `else if` to the more general check?

src/ExportMap.js
```diff
@@ -497,6 +497,28 @@ ExportMap.parse = function (path, content, context) {
...
+        importedSpecifiers.add(specifier.imported.name || specifier.imported.value);
+      }
+
+      // import { type Foo } (Flow)
+      specifiersOnlyImportingTypes =
+        specifiersOnlyImportingTypes && specifier.importKind === 'type';
```
ljharb, line 517: i smell prettier
```suggestion
      specifiersOnlyImportingTypes = specifiersOnlyImportingTypes && specifier.importKind === 'type';
```

ljharb (approved):
LGTM modulo comments

Issue #2529, opened 2022-08-23
Does not cache ambiguous modules


stenin-nikita:
### Description

The linter runs for a very long time if the "import/no-cycle" rule is enabled on a large project with typescript. I think the problem is that eslint does not cache all modules. For example, these files will always be parsed again:

- https://github.com/aws/aws-sdk-js/blob/master/clients/all.d.ts
- https://unpkg.com/browse/date-fns@2.29.2/typings.d.ts

Minimal repository for reproduce: https://github.com/stenin-nikita/eslint-plugin-import-bug

### Possible solution

This line https://github.com/import-js/eslint-plugin-import/blob/68cea3e6b6fe5fd61e5cf2e2c7c0be9e8dc597cb/src/ExportMap.js#L349

replace with:

```js
if (exportMap == null) {
  exportCache.set(cacheKey, null);
  return null;
}
```

Locally, this change reduced the check time from 20 to 4 minutes for me

stenin-nikita:
May be a similar problem here #2076

ljharb:
That definitely seems like a worthy change. Care to make a PR?

stenin-nikita:
ОК. I did a PR and wrote a test

stenin-nikita:
@ljharb When will a version with this fix be released?

ljharb:
Fixed by #2531.

Releases aren't done on a particular schedule, but hopefully soon.

PR #2531, opened 2022-08-24
[Fix] `ExportMap`: add caching after parsing for an ambiguous module

stenin-nikita:
This PR is a fix for #2529


ljharb (approved):
Seems legit

PR #2541, opened 2022-09-03
[New] `no-extraneous-dependencies`: added includeInternal option to validate imports of internal dependencies

### LINKED ISSUE #1678: Enhancement: Allow no-extraneous-dependencies to check internal Imports
In our monorepo I have setup that anything prefixed with out npm-scope is marked as internal, ie. `^@scope-.*/`. This is working really well for us.

We are now at a stage where we are going to need to publish some of our packages to be consumed in other projects/repos. Due to this I now wish to use `no-extraneous-dependencies` to ensure that everyone is rembering to include this internal package references.

I did consider doing this by removing the internal regex, but this then effects the import sortinng (as we keep our internal deps grouped seperatly to our external deps - thanks to the `order` rule)

What I would like to see is to add an option to `no-extraneous-dependencies` to allow it to check internal imports based on a regex (either the internal regex or a newly defined on the rule config) as well as external.

I would be happy to work on a PR to add this option.

bdwain:
…alidate imports of internal dependencies

Fixes #1678

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`no-restricted-paths`]: support arrays for `from` and `target` options ([#2466], thanks [@AdriAt360])
 - [`no-anonymous-default-export`]: add `allowNew` option ([#2505], thanks [@DamienCassou])
 - [`order`]: Add `distinctGroup` option ([#2395], thanks [@hyperupcall])
+- [`no-extraneous-dependencies`]: Add `includeInternal` option ([#2541], thanks [@bdwain])
 
 ### Fixed
 - [`order`]: move nested imports closer to main import entry ([#2396], thanks [@pri1311])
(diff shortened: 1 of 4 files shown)


ljharb (approved):
Solid, thanks!

ljharb:
Looks like a ton of tests are failing.

bdwain:
it looks like you pushed a change? I had 

```
  if (typeOfImport !== 'external' && (typeOfImport !== 'internal' || !depsOptions.verifyInternalDeps)) {
    return;
  }
```
but now it's 

```
  if (
    typeOfImport !== 'external'
    && (depsOptions.verifyInternalDeps && typeOfImport !== 'internal')
  ) {
```

but that looks wrong. it should be

```
  if (
    typeOfImport !== 'external'
    && (!depsOptions.verifyInternalDeps || typeOfImport !== 'internal')
  ) {
```

ljharb:
oh duh, yes, that's my bad. fixing :-)

bdwain:
thanks!

bdwain:
@ljharb any plans to do a release soon? Would love to start using this change.

PR #2543, opened 2022-09-04
[New] `no-extraneous-dependencies`: added includeTypes option to validate type imports

### LINKED ISSUE #2542: Verifying type imports
This one is similar to #1678 I think, but I'm interested in making this an error for no-extraneous-dependencies

```ts
import type { MyType } from 'my-missing-dep`
```

the situation where this comes up for me is in a monorepo with yarn workspaces, specifically for internal dependencies, but I guess I don't see why it needs to be limited to those. I have scenarios where I build some packages but not others, and that is determined by the package.json dependency tree. I tested in my project and I can get a missing package.json dependency to still build for a type import, which is where a lint rule would come in handy

Here's an example. If package A imports from package B, but forgets to mark it as a dependency in package.json, normally specifying that I want to build B won't build A either, so this wouldn't need a lint rule because it'd break. But if A happens to get built for another reason (it was already built and never deleted, or package C is also being built, which specifies a dep on A), it could mask the build/runtime error that I would get for missing the dependency. That's where the lint rule would come into play by making it clear at build time when I am missing the dependency. 

This applies to types as well regular imports, so I think having an option to enable the check for those would be useful. I'm not 100% sure why it would make sense to ever disable TBH, so I kind of feel like it should be on by default, or not even an option, but that can be a separate question since I know it's a breaking change.

Thoughts?

bdwain:
Fixes #2542

@ljharb I'm not totally if this will work with Flow or only Typescript, or what the current Flow support is like. I couldn't get a Flow unit test to fail with this new option, but I don't see what in the code would be different between the two type systems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`no-anonymous-default-export`]: add `allowNew` option ([#2505], thanks [@DamienCassou])
 - [`order`]: Add `distinctGroup` option ([#2395], thanks [@hyperupcall])
 - [`no-extraneous-dependencies`]: Add `includeInternal` option ([#2541], thanks [@bdwain])
+- [`no-extraneous-dependencies`]: Add `includeTypes` option ([#2543], thanks [@bdwain])
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


ljharb (approved):
LGTM

bdwain:
thanks @ljharb

Issue #2400, opened 2022-03-08
no-unused-modules needs an ignoreImports option


saiichihashimoto:
While `ignoreExports` allows us to ignore unused exports for specific files, we need to be able to ignore used imports.

### Why?
Let's assume there's three files:

```
- entry.js
- utility.js
- utility.spec.js
```

Both `entry.js` and `utility.spec.js` import from `utility.js`. Here, we'd want to include `entry.js` and `utility.spec.js` in `ignoreExports` since one is an entry point and the other is a test. We're ensured if both files stop importing from `utility.js` then the eslint rule will error, which is good!

The issue is when just `entry.js` stops importing from `utility.js`, which is what we want this rule to identify. Right now, the tests will keep `utility.js` from _ever_ being identified as unused, even though practically it's unused by the application. I'd have to manually discover that the application isn't using `utility.js`. This will never happen; I'd assume that the change in `entry.js` would identify this for me, but I'd have to manually check if `utility.spec.js` is the only one importing it, which negates the value of this rule. If I could declare in options to `ignoreImports` from `**/*.spec.*`, then `utility.js` will be identified as unused, which would be good!.

Currently, there's no way to accomplish this. Including `**/*.spec.*` in `ignoreExports`, setting `src` to `['**/!(*.spec.js)']`, or having an override where `excludedFiles` includes `**/*.spec.*` all do the same thing, which isn't what I want. We specifically need to identify files where we want to ignore their imports. The moment we write a unit test or a storybook story, we're essentially disabling the useful of this rule.

ljharb:
Typically, you'd apply the rule to your codebase, and use `src` to limit to just non-test files, and use `ignoreExports` to indicate entry points.

It sounds like the primary issue you're running into is that you have no NON entry points?

ljharb:
ahhh - you're saying that `utility.js`, being NOT an entry point, and being only imported from non-`src` files, should be reported as unused? I do agree with that expectation.

saiichihashimoto:
Exactly! My solution was to introduce an `ignoreImports` option to identify them (since using negates in `src` is very awkward) but whatever solves the problem.

DamienCassou:
I have exactly the same problem: If a file is only imported by tests, I consider it unused and would like eslint to tell me I can safely delete it.

ljharb:
[...]

PR #2557, opened 2022-09-27
feat(IgnoreImports): allow ignored imports in no-extraneous-dependency

kopax-polyconseil:
Fix : https://github.com/import-js/eslint-plugin-import/issues/2400

Idk how to use globs here, so it's no glob, and support `./` for rootDir

### DIFF
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -200,6 +200,7 @@ function reportIfMissing(context, deps, depsOptions, node, name) {
 
   if (
     declarationStatus.isInDeps ||
+    depsOptions.allowIgnoreImports ||
     (depsOptions.allowDevDeps && declarationStatus.isInDevDeps) ||
     (depsOptions.allowPeerDeps && declarationStatus.isInPeerDeps) ||
     (depsOptions.allowOptDeps && declarationStatus.isInOptDeps) ||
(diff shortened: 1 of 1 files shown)


src/rules/no-extraneous-dependencies.js
```diff
@@ -250,6 +252,16 @@ function testConfig(config, filename) {
   ));
 }
 
+function testIgnoreImports(config, filename) {
+  if (!config) {
+    return false
+  }
+  return config.some(c => (
+    filename.includes(value.substr(0, 2) === './' ? value.replace('.', process.cwd()) : value)
```
ljharb, line 260: where does `value` come from here?

also, let's use slice over substr.

src/rules/no-extraneous-dependencies.js
```diff
@@ -280,6 +293,7 @@ module.exports = {
     const deps = getDependencies(context, options.packageDir) || extractDepFields({});
 
     const depsOptions = {
+      allowIgnoreImports: testIgnoreImports(options.ignoreImports, filename) !== false,
```
ljharb, line 296:
```suggestion
      ignoreImports: testIgnoreImports(options.ignoreImports, filename) !== false,
```

ljharb (changes requested):
Can you please add tests covering this?

Issue #2450, opened 2022-05-09
import/no-restricted-paths: Allow `import type` imports


golergka:
It would be very useful for `no-restricted-paths` rule to have an option to allow `import types` for a zone, as these imports are erased at compile time anyway.

ksandin:
This would be awesome for [tRPC](https://trpc.io/) projects!

ljharb:
Related: #2608

danvk:
Related (same?): #1899

soryy708:
This one duplicates 1899, so I'd like to consolidate the conversation in one place.
Both issues had good activity and references to them, so it's unclear which one to close and which to keep open.
I decided to close this one because the other one is older and has a Pull Request referring to it.

PR #2459, opened 2022-05-19
[Tests] `no-restricted-paths`: `import type` tests

golergka:
I wanted to work on https://github.com/import-js/eslint-plugin-import/issues/2450, but decided to add tests for existing behaviour first, to ensure that subsequent work doesn't break it.

I've copied over all existing tests that had `import` and changed it to `import type`.

However, I have one failure currently that I can't figure out how to fix:

```
  1) Typescript no-restricted-paths invalid import type a from "../client/a"
import type c from "./c.ts":

      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Should have 2 errors but had 1: [
  {
    ruleId: 'no-restricted-paths',
    severity: 1,
    message: 'Unexpected path "../client/a" imported in restricted zone.',
    line: 1,
    column: 20,
    nodeType: 'Literal',
    endLine: 1,
    endColumn: 33
  }
]
      + expected - actual

      -1
      +2
      
      at testInvalidTemplate (node_modules/eslint/lib/rule-tester/rule-tester.js:773:24)
      at Context.<anonymous> (node_modules/eslint/lib/rule-tester/rule-tester.js:997:29)
      at processImmediate (node:internal/timers:462:21)
```

PR #2574, opened 2022-10-18
Fix: one failing test case from for `import type` cases in `no-restricted-paths`

azyzz228:
#2450 

In proposed solution for #2450, there was one test case which was failing. I pulled the changes of that developer, updated repo with latest changes, then fixed one failing test case.


ljharb:
When there's an existing PR, like #2459, please don't create a new one - instead, comment a link to your branch and I can pull in the changes.

azyzz228:
will do! Thanks

ljharb:
Now that it's open though, please leave this open so I can keep them in sync.

PR #2582, opened 2022-10-29
Automate docs with eslint-doc-generator

bmish:
I built this CLI tool [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) for automating the generation of the README rules list table and rule doc title/notices for ESLint plugins. It follows common documentation conventions from this and other top ESLint plugins and will help us standardize documentation across ESLint plugins (and generally improve the usability of custom rules through better documentation and streamline the process of adding new rules). It has 100% test coverage. It's already used in other popular plugins like [eslint-plugin-react](https://github.com/jsx-eslint/eslint-plugin-react/pull/3469).

Notable changes:
* Moved rule categories from README to common `meta.docs.category` property and used `eslint-doc-generator --split-by meta.docs.category` to continue displaying multiple lists in the README
* Moved rule descriptions from README to standard `meta.docs.description` property which is consumable by tooling
* Enabled [eslint-plugin/require-meta-docs-description](https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/require-meta-docs-description.md) for ensuring we have consistent descriptions defined in all rules

Rule doc headers now have consistent titles and helpful notices about autofixers and what configs the rules are enabled/disabled/warn in. The README rules list now has an attractive table containing much more information about the rules.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -71,6 +71,7 @@
         "eslint-plugin/no-deprecated-report-api": "off",
         "eslint-plugin/prefer-replace-text": "error",
         "eslint-plugin/report-message-format": "error",
+        "eslint-plugin/require-meta-docs-description": ["error", { "pattern": "^(Enforce|Ensure|Prefer|Forbid).+\\.$" }],
         "eslint-plugin/require-meta-schema": "error",
         "eslint-plugin/require-meta-type": "error",
 
--- a/README.md
+++ b/README.md
@@ -13,111 +13,85 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
 
 ## Rules
 
-### Static analysis
+<!-- begin auto-generated rules list -->
 
-* Ensure imports point to a file/module that can be resolved. ([`no-unresolved`])
-* Ensure named imports correspond to a named export in the remote file. ([`named`])
-* Ensure a default export is present, given a default import. ([`default`])
-* Ensure imported namespaces contain dereferenced properties as they are dereferenced. ([`namespace`])
-* Restrict which files can be imported in a given folder ([`no-restricted-paths`])
-* Forbid import of modules using absolute paths ([`no-absolute-path`])
-* Forbid `require()` calls with expressions ([`no-dynamic-require`])
-* Prevent importing the submodules of other modules ([`no-internal-modules`])
-* Forbid webpack loader syntax in imports ([`no-webpack-loader-syntax`])
-* Forbid a module from importing itself ([`no-self-import`])
-* Forbid a module from importing a module with a dependency path back to itself ([`no-cycle`])
-* Prevent unnecessary path segments in import and require statements ([`no-useless-path-segments`])
-* Forbid importing modules from parent directories ([`no-relative-parent-imports`])
-* Prevent importing packages through relative paths ([`no-relative-packages`])
-
-[`no-unresolved`]: ./docs/rules/no-unresolved.md
-[`named`]: ./docs/rules/named.md
-[`default`]: ./docs/rules/default.md
-[`namespace`]: ./docs/rules/namespace.md
-[`no-restricted-paths`]: ./docs/rules/no-restricted-paths.md
-[`no-absolute-path`]: ./docs/rules/no-absolute-path.md
-[`no-dynamic-require`]: ./docs/rules/no-dynamic-require.md
-[`no-internal-modules`]: ./docs/rules/no-internal-modules.md
-[`no-webpack-loader-syntax`]: ./docs/rules/no-webpack-loader-syntax.md
-[`no-self-import`]: ./docs/rules/no-self-import.md
-[`no-cycle`]: ./docs/rules/no-cycle.md
-[`no-useless-path-segments`]: ./docs/rules/no-useless-path-segments.md
-[`no-relative-parent-imports`]: ./docs/rules/no-relative-parent-imports.md
-[`no-relative-packages`]: ./docs/rules/no-relative-packages.md
+💼 Configurations enabled in.\
+⚠️ Configurations set to warn in.\
+🚫 Configurations disabled in.\
+❗ Set in the `errors` configuration.\
+☑️ Set in the `recommended` configuration.\
+⌨️ Set in the `typescript` configuration.\
+🚸 Set in the `warnings` configuration.\
+🔧 Automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/user-guide/command-line-interface#--fix).\
+💡 Manually fixable by [editor suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).\
+❌ Deprecated.
 
 ### Helpful warnings
 
-
-* Report any invalid exports, i.e. re-export of the same name ([`export`])
-* Report use of exported name as identifier of default export ([`no-named-as-default`])
-* Report use of exported name as property of default export ([`no-named-as-default-member`])
-* Report imported names marked with `@deprecated` documentation tag ([`no-deprecated`])
-* Forbid the use of extraneous packages ([`no-extraneous-dependencies`])
-* Forbid the use of mutable exports with `var` or `let`. ([`no-mutable-exports`])
-* Report modules without exports, or exports without matching import in another module ([`no-unused-modules`])
-* Prevent empty named import blocks ([`no-empty-named-blocks`])
-
-[`export`]: ./docs/rules/export.md
-[`no-named-as-default`]: ./docs/rules/no-named-as-default.md
-[`no-named-as-default-member`]: ./docs/rules/no-named-as-default-member.md
-[`no-deprecated`]: ./docs/rules/no-deprecated.md
-[`no-extraneous-dependencies`]: ./docs/rules/no-extraneous-dependencies.md
-[`no-mutable-exports`]: ./docs/rules/no-mutable-exports.md
-[`no-unused-modules`]: ./docs/rules/no-unused-modules.md
-[`no-empty-named-blocks`]: ./docs/rules/no-empty-named-blocks.md
+| Name                                                                   | Description                                                                           | 💼   | ⚠️    | 🚫 | 🔧 | 💡 | ❌  |
+| :--------------------------------------------------------------------- | :------------------------------------------------------------------------------------ | :--- | :---- | :- | :- | :- | :- |
+| [export](docs/rules/export.md)                                         | Forbid any invalid exports, i.e. re-export of the same name.                          | ❗ ☑️ |       |    |    |    |    |
+| [no-deprecated](docs/rules/no-deprecated.md)                           | Forbid imported names marked with `@deprecated` documentation tag.                    |      |       |    |    |    |    |
+| [no-empty-named-blocks](docs/rules/no-empty-named-blocks.md)           | Forbid empty named import blocks.                                                     |      |       |    | 🔧 | 💡 |    |
+| [no-extraneous-dependencies](docs/rules/no-extraneous-dependencies.md) | Forbid the use of extraneous packages.                                                |      |       |    |    |    |    |
+| [no-mutable-exports](docs/rules/no-mutable-exports.md)                 | Forbid the use of mutable exports with `var` or `let`.                                |      |       |    |    |    |    |
+| [no-named-as-default](docs/rules/no-named-as-default.md)               | Forbid use of exported name as identifier of default export.                          |      | ☑️ 🚸 |    |    |    |    |
+| [no-named-as-default-member](docs/rules/no-named-as-default-member.md) | Forbid use of exported name as property of default export.                            |      | ☑️ 🚸 |    |    |    |    |
+| [no-unused-modules](docs/rules/no-unused-modules.md)                   | Forbid modules without exports, or exports without matching import in another module. |      |       |    |    |    |    |
 
 ### Module systems
 
-* Report potentially ambiguous parse goal (`script` vs. `module`) ([`unambiguous`])
-* Report CommonJS `require` calls and `module.exports` or `exports.*`. ([`no-commonjs`])
-* Report AMD `require` and `define` calls. ([`no-amd`])
-* No Node.js builtin modules. ([`no-nodejs-modules`])
-* Forbid imports with CommonJS exports ([`no-import-module-exports`])
+| Name                                                               | Description                                                          | 💼 | ⚠️ | 🚫 | 🔧 | 💡 | ❌  |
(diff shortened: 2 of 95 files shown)


tests/src/rules/no-commonjs.js
```diff
@@ -96,19 +96,19 @@ ruleTester.run('no-commonjs', require('rules/no-commonjs'), {
     ]),
 
     // exports
-    { code: 'exports.face = "palm"', output: 'exports.face = "palm"', errors: [ { message: EXPORT_MESSAGE }] },
-    { code: 'module.exports.face = "palm"', output: 'module.exports.face = "palm"', errors: [ { message: EXPORT_MESSAGE }] },
-    { code: 'module.exports = face', output: 'module.exports = face', errors: [ { message: EXPORT_MESSAGE }] },
-    { code: 'exports = module.exports = {}', output: 'exports = module.exports = {}', errors: [ { message: EXPORT_MESSAGE }] },
-    { code: 'var x = module.exports = {}', output: 'var x = module.exports = {}', errors: [ { message: EXPORT_MESSAGE }] },
+    { code: 'exports.face = "palm"', output: null, errors: [ { message: EXPORT_MESSAGE }] },
```
bmish, line 99: Autofixed by [eslint-plugin/prefer-output-null](https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/prefer-output-null.md).

.eslintrc
```diff
@@ -69,8 +69,10 @@
         "eslint-plugin/meta-property-ordering": "error",
         "eslint-plugin/no-deprecated-context-methods": "error",
         "eslint-plugin/no-deprecated-report-api": "off",
+        "eslint-plugin/prefer-output-null": "off",
```
bmish, line 72: Can't enable [eslint-plugin/prefer-output-null](https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/prefer-output-null.md) yet since one of the older versions of ESLint that we test against doesn't. support `output: null`.

docs/rules/default.md
```diff
@@ -1,5 +1,9 @@
 # import/default
 
+💼 This rule is enabled in the following configs: ❗ `errors`, ✅ `recommended`.
```
ljharb, line 3: let's use the black and white check here for "recommended" also

docs/rules/extensions.md
```diff
@@ -1,4 +1,6 @@
-# import/extensions - Ensure consistent use of file extension within the import path
```
ljharb, line 1: i'd prefer to see this tagline preserved somewhere

docs/rules/no-absolute-path.md
```diff
@@ -1,4 +1,6 @@
-# import/no-absolute-path: Forbid import of modules using absolute paths
```
ljharb, line 1: also this one

docs/rules/no-deprecated.md
```diff
@@ -1,4 +1,8 @@
-# `import/no-deprecated`
+# import/no-deprecated
+
+💼 This rule will _warn_ in the `stage-0` config.
```
ljharb, line 3: let's exclude this undocumented config if we can :-)

docs/rules/no-dynamic-require.md
```diff
@@ -1,4 +1,6 @@
-# import/no-dynamic-require: Forbid `require()` calls with expressions
```
ljharb, line 1: preserve

docs/rules/no-extraneous-dependencies.md
```diff
@@ -1,4 +1,6 @@
-# import/no-extraneous-dependencies: Forbid the use of extraneous packages
```
ljharb, line 1: preserve

docs/rules/no-restricted-paths.md
```diff
@@ -1,4 +1,6 @@
-# import/no-restricted-paths: Restrict which files can be imported in a given folder
```
ljharb, line 1: preserve

docs/rules/no-unassigned-import.md
```diff
@@ -1,4 +1,6 @@
-# import/no-unassigned-import: Forbid unassigned imports
```
ljharb, line 1: preserve

src/rules/consistent-type-specifier-style.js
```diff
@@ -40,7 +40,8 @@ module.exports = {
   meta: {
     type: 'suggestion',
     docs: {
-      description: 'Enforce or ban the use of inline type-only markers for named imports',
+      category: 'Style guide',
```
ljharb, line 43: what's the consequence of adding these categories?

bmish: This property is not use by ESLint but is commonly used by ESLint plugins for tooling/documentation purposes. In this case, I have used it so the tool can generate the separated lists thanks to `--split-by meta.docs.category`.

ljharb: ok, so you're just matching what was previously in the readme, in the rule metadata?

bmish: Yep. Rule metadata should be the source of truth.

docs/rules/default.md
bmish: Updated using `--config-emoji recommended,☑️`.

docs/rules/no-deprecated.md
bmish: Updated using `--ignore-config stage-0`.

docs/rules/no-absolute-path.md
bmish: I added this line to the rule doc body.

docs/rules/no-dynamic-require.md
bmish: There's an equivalent sentence a few lines down.

docs/rules/no-extraneous-dependencies.md
bmish: There's already a sentence "Forbid the import of external modules that are not declared..." that should capture this.

docs/rules/no-restricted-paths.md
bmish: > In order to prevent such scenarios this rule allows you to define restricted zones where you can forbid files from imported if they match a specific path.

This line from the body might capture this already.

docs/rules/no-unassigned-import.md
bmish: This line from the body might capture this already:

> This rule aims to remove modules with side-effects by reporting when a module is imported but not assigned.

docs/rules/extensions.md
bmish: A few of these rule docs had the rule description in their title. Note that these descriptions are all preserved in `meta.docs.description` which is used in the README rules table. If you wanted, we could include the rule descriptions in all the titles using the option [`--rule-doc-title-format`](https://github.com/bmish/eslint-doc-generator#--rule-doc-title-format) option. But for the most part, these descriptions are already captured in the rule doc bodies.

ljharb: I think it's useful for the exact phrase to appear somewhere in the doc - it doesn't have to be in the title ofc, but somewhere. Can that be automated/enforced somehow?

docs/rules/no-extraneous-dependencies.md
ljharb: Per https://github.com/import-js/eslint-plugin-import/pull/2582#discussion_r1008920390, i think the exact description phrase text needs to appear in the doc somewhere.

docs/rules/no-dynamic-require.md
ljharb: Per https://github.com/import-js/eslint-plugin-import/pull/2582#discussion_r1008920390, i think the exact description phrase text needs to appear in the doc somewhere.

docs/rules/no-restricted-paths.md
ljharb: Per https://github.com/import-js/eslint-plugin-import/pull/2582#discussion_r1008920390, i think the exact description phrase text needs to appear in the doc somewhere.

docs/rules/no-unassigned-import.md
ljharb: Per https://github.com/import-js/eslint-plugin-import/pull/2582#discussion_r1008920390, i think the exact description phrase text needs to appear in the doc somewhere.

docs/rules/extensions.md
bmish: Today, there are many plugins that include the rule description directly in the rule doc title, which is why we have [`--rule-doc-title-format desc-parens-prefix-name`](https://github.com/bmish/eslint-doc-generator#--rule-doc-title-format). That's probably the best/easiest way to achieve this within existing conventions.

That said, if that's still not satisfactory, then I'm thinking of some possible options that could achieve what you want.

| Potential Option | Description | Upside | Downsides |
| --- | --- | --- | --- | 
| `--rule-doc-header-include-description` | Whether to include the rule description below any notices at the bottom of the rule doc header (default `false`). | The rule description will be auto-generated in the rule doc header. | This isn't always the best place to put the rule description. Often, rule docs provide general context first, and then provide the description later (potentially in a "Rule details" section). |
| `--rule-doc-section-include-description` | Specifies the name of the section of the rule doc that should begin with the rule description. Pass option with no section name to include the description at the bottom of the rule doc header. | Ensures the rule description is mentioned in the appropriate section, often a "Rule details" section. This is a more flexible version of `--rule-doc-header-include-description`. | Most complex implementation.
| `--rule-doc-mention-description` | Whether to require the rule description to be mentioned somewhere in the rule doc (default `false`). | Allows greatest flexibility in placement. | Not auto-generated. The descriptions could end up placed in inconsistent positions. |

My recommendation in order of preference: 

1. Leave as-is. Rule docs already capture the spirit of the description even if the wording doesn't exactly match.
2. Include rule descriptions in the rule doc titles.
3. Add `--rule-doc-section-include-description` option.
4. Add `--rule-doc-mention-description` option.

Let me know what you think. If we want to implement a new option, then it would probably be best to [file an issue](https://github.com/bmish/eslint-doc-generator/issues/new) for this and avoid blocking the overall PR, since this only affected a few rule docs anyway.

ljharb: Hmm. I agree it needn't block the PR, regardless.

I think a simple tagline is important top-level context, so I'd prefer `--rule-doc-section-include-description`. Is that already implemented? If so, let's use it! If not, we can address it in the future.

README.md
```diff
@@ -13,111 +13,84 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
...
-[`no-useless-path-segments`]: ./docs/rules/no-useless-path-segments.md
-[`no-relative-parent-imports`]: ./docs/rules/no-relative-parent-imports.md
-[`no-relative-packages`]: ./docs/rules/no-relative-packages.md
+<!-- begin auto-generated rules list -->
+
+💼 Configurations enabled in.\
+☑️ Enabled in the `recommended` configuration.\
+☑️<sup>⚠️</sup> Warns in the `recommended` configuration.\
+❗ Enabled in the `errors` configuration.\
+⚠️ Warns in the `warnings` configuration.\
```
ljharb, line 22: it's a little confusing to have the same emoji for the "warnings" config as for a modifier on another config to indicate that it warns.

bmish: How about using a new emoji for the `warnings` config via `--config-emoji warnings,🚸`. I pushed the update.


### Before:

💼 Configurations enabled in.\
☑️ Enabled in the `recommended` configuration.\
☑️<sup>⚠️</sup> Warns in the `recommended` configuration.\
❗ Enabled in the `errors` configuration.\
⚠️ Warns in the `warnings` configuration.\
⌨️<sup>🚫</sup> Disabled in the `typescript` configuration.

### After:

💼 Configurations enabled in.\
☑️ Enabled in the `recommended` configuration.\
☑️<sup>⚠️</sup> Warns in the `recommended` configuration.\
❗ Enabled in the `errors` configuration.\
🚸<sup>⚠️</sup> Warns in the `warnings` configuration.\
⌨️<sup>🚫</sup> Disabled in the `typescript` configuration.

README.md
```diff
@@ -13,111 +13,84 @@ This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, a
...
+| [no-deprecated](docs/rules/no-deprecated.md)                           | Forbid imported names marked with `@deprecated` documentation tag.                    |                                     |    |    |    |
+| [no-empty-named-blocks](docs/rules/no-empty-named-blocks.md)           | Forbid empty named import blocks.                                                     |                                     | 🔧 | 💡 |    |
+| [no-extraneous-dependencies](docs/rules/no-extraneous-dependencies.md) | Forbid the use of extraneous packages.                                                |                                     |    |    |    |
+| [no-mutable-exports](docs/rules/no-mutable-exports.md)                 | Forbid the use of mutable exports with `var` or `let`.                                |                                     |    |    |    |
+| [no-named-as-default](docs/rules/no-named-as-default.md)               | Forbid use of exported name as identifier of default export.                          | ☑️<sup>⚠️</sup> <br>🚸<sup>⚠️</sup> |    |    |    |
```
ljharb, line 37: it seems like the superscripted emoji is still rendering on a new line, and isn't staying attached to the base emoji?

bmish: Ugh. I fixed some other issues with this but looks like I'm still struggling to get the emoji superscript to stay attached to its base character. Tracking this with: https://github.com/bmish/eslint-doc-generator/issues/204

README.md
ljharb: Looks great!

README.md
bmish: I tried everything to fix the emoji superscripts but they turned out to be too fragile. This led me to remove the superscripts entirely and switch to what I believe is a significantly better information architecture with a separate rules list column per severity and many other improvements:
* https://github.com/bmish/eslint-doc-generator/pull/205

ljharb (approved):
Looks great!

A possible enhancement would be if there's a way to provide tooltips (like an img `alt`) on all the emojis.

bmish:
@ljharb based on my research, it's not possible to add alt text / tooltips on emojis in GitHub markdown. It could be nice but hopefully it won't be needed anyway.

bmish:
Seems like there's an unrelated failure on CI with `continuous-integration/appveyor/pr — AppVeyor build failed`? I'd recommend merging and addressing that separately.

```
  1) dynamic imports no-unused-modules valid 
        import App from './App';
      :
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

ljharb:
@bmish i'll rerun it til it passes :-)

Issue #2586, opened 2022-11-04
Typescript Docs Question


Nfinished:
The following bit is unclear to me.

```
# TypeScript

You may use the following shortcut or assemble your own config using the granular settings described below.

Make sure you have installed [`@typescript-eslint/parser`] and [`eslint-import-resolver-typescript`] which are used in the following configuration.

```yaml
extends:
  - eslint:recommended
  - plugin:import/recommended
# the following lines do the trick
  - plugin:import/typescript
settings:
  import/resolver:
    # You will also need to install and configure the TypeScript resolver
    # See also https://github.com/import-js/eslint-import-resolver-typescript#configuration
    typescript: true
    node: true


[`@typescript-eslint/parser`]: https://github.com/typescript-eslint/typescript-eslint/tree/HEAD/packages/parser
[`eslint-import-resolver-typescript`]: https://github.com/import-js/eslint-import-resolver-typescript
```

Does "You may use the following shortcut" and "the following lines do the trick" indicate that `plugin:import/typescript` should preclude any other configuration? Or do I still need to add the configuration outlined in https://github.com/import-js/eslint-import-resolver-typescript#configuration

Thank you!

ljharb:
it says lines, plural, so you'd need all of them.

Nfinished:
Including configuring the resolver or are there defaults?

```
# You will also need to install and configure the TypeScript resolver
# See also https://github.com/import-js/eslint-import-resolver-typescript#configuration
    ```

Nfinished:
I understand that I need it but is the ootb configuration enough or do I need to follow the configuration steps listed here? https://github.com/import-js/eslint-import-resolver-typescript#configuration

ljharb:
You need to follow all the steps listed there - that's why they're listed.

Nfinished:
Can you see how I found `You may use the following shortcut or assemble your own config using the granular settings described below.` unclear given the rest of the section? What _is_ the following shortcut? As I'm reading it I only see a guide to assembling my own config.

ljharb:
Indeed I can; happy to review a PR that makes it more clear.

To clarify, "all those lines" are the shortcut - the granular settings are the other sections of the document.

Nfinished:
Ah understood, thank you. It also threw me that `eslint-import-resolver-typescript`'s config example uses the bare `import` plugin rather that `import/recommended` so it's unclear if any of that comes with `/recommended`

soryy708:
[...]

PR #2588, opened 2022-11-04
Increase clarity around typescript configuration

Nfinished:
From a discussion with @ljharb in https://github.com/import-js/eslint-plugin-import/issues/2586


ljharb (approved):
Thanks!

PR #2589, opened 2022-11-05
[Fix] add missing param to function

Fdawgs:
Functions are passed more arguments than they have params declared.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -547,7 +547,7 @@ ExportMap.parse = function (path, content, context) {
 
   const source = makeSourceCode(content, ast);
 
-  function readTsConfig() {
+  function readTsConfig(context) {
     const tsConfigInfo = tsConfigLoader({
       cwd:
         (context.parserOptions && context.parserOptions.tsconfigRootDir) ||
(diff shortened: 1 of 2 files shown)


src/core/importType.js
```diff
@@ -112,11 +112,11 @@ function isExternalLookingName(name) {
...
   if (isInternalRegexMatch(name, settings)) { return 'internal'; }
-  if (isAbsolute(name, settings, path)) { return 'absolute'; }
+  if (isAbsolute(name)) { return 'absolute'; }
   if (isBuiltIn(name, settings, path)) { return 'builtin'; }
-  if (isRelativeToParent(name, settings, path)) { return 'parent'; }
-  if (isIndex(name, settings, path)) { return 'index'; }
-  if (isRelativeToSibling(name, settings, path)) { return 'sibling'; }
+  if (isRelativeToParent(name)) { return 'parent'; }
+  if (isIndex(name)) { return 'index'; }
+  if (isRelativeToSibling(name)) { return 'sibling'; }
```
ljharb, line 119: these are kept this way intentionally, for consistency; please revert.

src/ExportMap.js
```diff
@@ -579,7 +579,7 @@ ExportMap.parse = function (path, content, context) {
     }).digest('hex');
     let tsConfig = tsConfigCache.get(cacheKey);
     if (typeof tsConfig === 'undefined') {
-      tsConfig = readTsConfig(context);
+      tsConfig = readTsConfig();
```
ljharb, line 582: in this case, the bug is that `readTsConfig` *isn't* defining a `context` argument - so this line should be reverted, and line 550 fixed.

ljharb (changes requested):

src/ExportMap.js
Fdawgs: Thanks, updated accordingly in https://github.com/import-js/eslint-plugin-import/pull/2589/commits/c7dc38bb5663db9e2c2e6693102255c0790f96a9

ljharb (approved):

PR #2590, opened 2022-11-08
[actions] update GitHub Actions

Fdawgs:
This PR bumps GitHub Actions to their latest major versions, which now use Node 16 internally instead of Node 12 (EOL as of end of April 2022, no longer receives security updates):

- Changelog for `actions/checkout` can be [found here](https://github.com/actions/checkout/blob/main/CHANGELOG.md)
- Changelog for `codecov/codecov-action` actions can be [found here](https://github.com/codecov/codecov-action/blob/main/CHANGELOG.md)

### DIFF
--- a/.github/workflows/node-4+.yml
+++ b/.github/workflows/node-4+.yml
@@ -99,7 +99,7 @@ jobs:
             eslint: 5
 
     steps:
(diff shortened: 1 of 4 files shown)


ljharb (approved):
Thanks!

PR #2593, opened 2022-11-11
[New] `extensions`: add auto fixer!

kumavis:
doesn't seem to work in all cases - fails silently in some cases as you can see in the tests where there are expected errors but `ouput` === `code`.

first time writing a fixer so I hope I'm doing it right.

adding the missing file extension does not support query params. would be easier with node's `path`, I found it hard to work with the URL since it would not differentiate between `x` and `./x`, but maybe i was holding it wrong

### DIFF
--- a/tests/src/rules/extensions.js
+++ b/tests/src/rules/extensions.js
@@ -8,7 +8,10 @@ ruleTester.run('extensions', rule, {
   valid: [
     test({ code: 'import a from "@/a"' }),
     test({ code: 'import a from "a"' }),
-    test({ code: 'import dot from "./file.with.dot"' }),
+    test({
+      code: 'import dot from "./file.with.dot"',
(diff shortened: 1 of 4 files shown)


kumavis:
hi @ljharb/
am i doing this right?

tests/src/rules/extensions.js
```diff
@@ -151,6 +159,7 @@ ruleTester.run('extensions', rule, {
   invalid: [
     test({
       code: 'import a from "a/index.js"',
+      output: 'import a from "a/index"',
```
ljharb, line 162: this isn't a safe autofix unless we can validate that `a/index` and `a/index.js` resolve to the same thing.

if we can't add that validation using custom resolvers, then this would need to be a suggestion instead of an autofix.

Issue #1996, opened 2021-02-25
webpack resolver crashes with { alias: { "some-module": false } }


jtbandes:
eslint version: 7.20.0
eslint-plugin-import version: 2.22.1

According to the [Webpack docs](https://webpack.js.org/configuration/resolve/) on resolve settings, a module can be ignored by providing a value of `false`. This is the replacement for the now-deprecated [`null-loader`](https://github.com/webpack-contrib/null-loader).

> Setting `resolve.alias` to `false` will tell webpack to ignore a module.
> ```js
> module.exports = {
>   //...
>   resolve: {
>     alias: {
>       'ignored-module': false,
>       './ignored-module': false,
>     },
>   },
> };
> ```

But when this configuration is used, and `import/namespace` (or another rule from this plugin) tries to resolve the namespace, eslint crashes:
```
Oops! Something went wrong! :(

ESLint: 7.20.0

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of Buffer or URL. Received type boolean (false)
    at Object.statSync (node:fs:1125:10)
    at Function.ExportMap.for (.../node_modules/eslint-plugin-import/src/ExportMap.js:305:20)
    at Function.get (.../node_modules/eslint-plugin-import/src/ExportMap.js:293:10)
    at forEach (.../node_modules/eslint-plugin-import/src/rules/namespace.js:49:35)
    at Array.forEach (<anonymous>)
    at Program (.../node_modules/eslint-plugin-import/src/rules/namespace.js:81:14)
    at .../node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (.../node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (.../node_modules/eslint/lib/linter/node-event-generator.js:256:26)
```

An example of when this would be used is when a module is imported only in dev mode, but is stubbed out for production builds. At runtime it will never be used, but the import still exists in the code so the linter tries to resolve it.

It might make sense to treat this the same way as `externals`:

https://github.com/benmosher/eslint-plugin-import/blob/f2db74a347a030c15a387826500cdc98203eac1e/resolvers/webpack/index.js#L142-L145

jvanderen1:
I am getting the same error when I have the following in my webpack config:

```js
{
  ...,
  resolve: {
    ...,
    fallback: {
      https: false,
    }
[...]

Issue #2594, opened 2022-11-12
eslint 8 throwing error in CI


snewcomer:
> Error: Error while loading rule 'import/no-unused-modules': Failed to load plugin 'import' declared in '.eslintrc': Cannot find module '/home/runner/work/eslint-plugin-import/eslint-plugin-import/node_modules/eslint-plugin-import/lib/index.js'. Please verify that the package.json has a valid "main" entry

> Occurred while linting /home/runner/work/eslint-plugin-import/eslint-plugin-import/tests/files/just-json-files/invalid.json
[71](https://github.com/import-js/eslint-plugin-import/actions/runs/3449533983/jobs/5757534657#step:5:72)

See https://github.com/import-js/eslint-plugin-import/actions/runs/3449533983/jobs/5757534657#step:5:66

Happened within the last 4 days.  Last versioned release of eslint was 8 days ago.  Node also release 8 days ago.

ljharb:
npm 9 was promoted to "latest" in that timeframe.

I won't be able to properly investigate until next week.

PR #2598, opened 2022-11-15
[Fix] `no-unused-modules`: `checkPkgFieldObject` filters boolean fields

mpint:
Fixes the error reported in [this issue](https://github.com/import-js/eslint-plugin-import/issues/1996) where `boolean` fields aren't handled nicely by `path.join`.

According to the `package.json` [browser field spec](https://github.com/defunctzombie/package-browser-field-spec#ignore-a-module), a `false` field denotes an ignored module in the browser.
```
  // package.json

  "browser": {
    "browserObject": "./index.js",
    "an-ignored-module": false
  }
```


ljharb (approved):
Seems good to me!

mpint:
> Seems good to me!

Thanks. Will merge when https://github.com/import-js/eslint-plugin-import/issues/2594 is resolved 👍

Issue #2600, opened 2022-11-17
Allow `prefer-default-export` to report for files without default exports


imhoffd:
I noticed that the [prefer-default-export](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/prefer-default-export.md) rule only triggers when there is a single named export in a file. I would like for it to trigger for any file that doesn't have a default export: whether it has multiple named exports, a single named export, or no exports at all (but each of these could be configuration options for this rule).

The following patterns would be considered warnings:

```js
// bad1.js

// There is only a single module export and it's a named export.
export const foo = 'foo';
```

```js
// bad2.js

// There is more than one named export in the module.
export const foo = 'foo';
export const bar = 'bar';
```

The following patterns would not be considered warnings:

```js
// good1.js

// There is a single export and it's a default export.
const foo = 'foo';
export default foo;
```

```js
// good2.js

// There is a default export.
export const foo = 'foo';
const bar = 'bar';
export default bar;
```

ljharb:
The default is built assuming that a file with multiple named exports is a bag of things - reducers, constants, etc. It certainly seems reasonable to add an option that would warn on any file missing a default export, though.

imhoffd:
Thanks for that confirmation!

To elaborate on why this might be useful, in our codebase React components are default exports and their prop interfaces are named exports. It's easy enough to miss adding the `default` keyword to the component function, and this rule currently doesn't help with that.

In our particular case, our files are named the same as their default export (via [canonical/filename-match-exported](https://github.com/gajus/eslint-plugin-canonical#eslint-plugin-canonical-rules-filename-match-exported)). The idea is that files are self-contained modules revolving around a single, default declaration (the main thing) which may have zero to many named declarations (supplementary things that don't really warrant a file of their own, but are useful outside the module and as such are exported).

Issue #2118, opened 2021-06-04
[BUG] TypeError: Cannot read property 'replace' of undefined Occurred while linting


Alecell:
I'm getting an annoying error that occurs every time I type the word `import`.

```
[Error - 11:30:59 AM] ESLint stack trace:
[Error - 11:30:59 AM] TypeError: Cannot read property 'replace' of undefined
Occurred while linting /home/alecell/projects/AFYA/medcel.site.v6/src/pages/components/g2SwtuVEbBfUM36JWk3Rsn/g2SwtuVEbBfUM36JWk3Rsn.tsx:2
    at checkFileExtension (/home/alecell/projects/AFYA/medcel.site.v6/node_modules/eslint-plugin-import/lib/rules/extensions.js:147:52)
    at checkSourceValue (/home/alecell/projects/AFYA/medcel.site.v6/node_modules/eslint-module-utils/moduleVisitor.js:29:5)
    at checkSource (/home/alecell/projects/AFYA/medcel.site.v6/node_modules/eslint-module-utils/moduleVisitor.js:34:5)
    at /home/alecell/projects/AFYA/medcel.site.v6/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (/home/alecell/projects/AFYA/medcel.site.v6/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/home/alecell/projects/AFYA/medcel.site.v6/node_modules/eslint/lib/linter/node-event-generator.js:293:26)
    at NodeEventGenerator.applySelectors (/home/alecell/projects/AFYA/medcel.site.v6/node_modules/eslint/lib/linter/node-event-generator.js:322:22)
    at NodeEventGenerator.enterNode (/home/alecell/projects/AFYA/medcel.site.v6/node_modules/eslint/lib/linter/node-event-generator.js:336:14)
    at CodePathAnalyzer.enterNode (/home/alecell/projects/AFYA/medcel.site.v6/node_modules/eslint/lib/linter/code-path-analysis/code-path-analyzer.js:711:23)
```

THis error pops up on the right bottom corner of the VS code:
![image](https://user-images.githubusercontent.com/13935760/120826618-07cee100-c531-11eb-868e-5cfeb5b4b358.png)

I've disabled my eslint plugins one by one until I realize that this error only occurs only when I have this plugin o the plugin list.

Versions:
[...]

PR #2602, opened 2022-11-24
[New] `prefer-default-export`: add "target" option

azyzz228:
Solution for #2600.

I have added new option for the `prefer-default-export` that will check any exporting file to have a default export.

###### schema:

```
"import/prefer-default-export": [
 	( "off" | "warn" | "error" ),
	{ "target": "single" | "any" } // default is "single"
]
```
###### Config options

* **single** (current implementation of the rule, I did not change anything here): When there is only a single export from a module, prefer using default export over named export.

All examples from below are from the [rule descriptions](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/prefer-default-export.md).

Considered warnings:

```
// bad.js

// There is only a single module export and it's a named export.
export const foo = 'foo';

```

The following patterns are not warnings:

```
// good1.js

// There is a default export.
export const foo = 'foo';
const bar = 'bar';
export default bar;
```

```
// good2.js

// There is more than one named export in the module.
export const foo = 'foo';
export const bar = 'bar';
```

```
// good3.js

// There is more than one named export in the module
const foo = 'foo';
const bar = 'bar';
export { foo, bar }
```

```
// good4.js

// There is a default export.
const foo = 'foo';
export { foo as default }
```

```
// export-star.js

// Any batch export will disable this rule. The remote module is not inspected.
export * from './other-module'
```

newly added config option:

* **any**: any exporting file must contain default export.

No warnings:

```
//has default export
export default function bar() {};
```

```
// has default export
let foo;
export { foo as default }
```

```
//contains multiple exports AND default export﻿
export const a = 5;
export function bar(){};
let foo;
export { foo as default }`
```

```
// does not contain any exports => file is not checked by the rule
import * as foo from './foo';﻿
```

```
// export-star.js

// Any batch export will disable this rule. The remote module is not inspected.
export * from './other-module'
```

Following are considered as warnings:

```
//has 2 named exports, but no default export
export const foo = 'foo';
export const bar = 'bar';
```

```
// does not have default export
let foo, bar;
export { foo, bar }
```

```
// does not have default export
export { a, b } from "foo.js"﻿
```

```
// does not have default export
let item;
export const foo = item;
export { item };`
```

### Questions

I am not sure how to migrate the test cases below to new option "any". I do not understand what's going on here and what is exported.

```
export const { foo, bar } = item;
```

```
export const { foo, bar: baz } = item;
```

```
export const { foo: { bar, baz } } = item;
```

```
export const [a, b] = item;
```

I will update docs after your review.

### DIFF
--- a/docs/rules/prefer-default-export.md
+++ b/docs/rules/prefer-default-export.md
@@ -2,9 +2,43 @@
 
 <!-- end auto-generated rule header -->
 
-When there is only a single export from a module, prefer using default export over named export.
+In exporting files, this rule checks if there is default export or not.
 
 ## Rule Details
+
+##### rule schema:
+
+```javascript
+"import/prefer-default-export": [
+    ( "off" | "warn" | "error" ),
+	{ "target": "single" | "any" } // default is "single"
+]
+```
+
+### Config Options
+
+There are two options available: `single` and `any`. By default, if you do not specify the option, rule will assume it is `single`.
+
+#### single
+
+**Definition**: When there is only a single export from a module, prefer using default export over named export.
+
+How to setup config file for this rule:
+
+```javascript
+// you can manually specify it
+"rules": {
+    "import/prefer-default-export": [
+        ( "off" | "warn" | "error" ),
+        { "target": "single" }
+    ]
+}
+
+// config setup below will also work
+"rules": {
+    "import/prefer-default-export": "off" | "warn" | "error"
+}
+```
 
 The following patterns are considered warnings:
 
@@ -57,4 +91,96 @@ export { foo as default }
 
 // Any batch export will disable this rule. The remote module is not inspected.
 export * from './other-module'
+```
+
+#### any
+
+**Definition**: any exporting file must contain a default export.
+
+How to setup config file for this rule:
+
+```javascript
+// you have to manually specify it
+"rules": {
+    "import/prefer-default-export": [
+        ( "off" | "warn" | "error" ),
+        { "target": "any" }
+    ]
+}
+```
+
+
+The following patterns are *not* considered warnings:
+
+```javascript
+// good1.js
+
+//has default export
+export default function bar() {};
+```
+
+```javascript
+// good2.js
+
+// has default export
+let foo;
+export { foo as default }
+```
+
+```javascript
+// good3.js
+
+//contains multiple exports AND default export
+export const a = 5;
+export function bar(){};
+let foo;
+export { foo as default }
+```
+
+```javascript
+// good4.js
+
+// does not contain any exports => file is not checked by the rule
+import * as foo from './foo';﻿
+```
+
+```javascript
+// export-star.js
+
+// Any batch export will disable this rule. The remote module is not inspected.
+export * from './other-module'
+```
+
+The following patterns are considered warnings:
+
+```javascript
+// bad1.js
+
+//has 2 named exports, but no default export
+export const foo = 'foo';
+export const bar = 'bar';
+```
+
+```javascript
+// bad2.js
+
+// does not have default export
+let foo, bar;
+export { foo, bar }
+```
+
+```javascript
+// bad3.js
+
+// does not have default export
+export { a, b } from "foo.js"﻿
+```
+
+```javascript
+// bad4.js
+
+// does not have default export
+let item;
+export const foo = item;
+export { item };
 ```
--- a/src/rules/prefer-default-export.js
+++ b/src/rules/prefer-default-export.js
@@ -2,15 +2,28 @@
 
 import docsUrl from '../docsUrl';
 
+const SINGLE_EXPORT_ERROR_MESSAGE = 'Prefer default export on a file with single export.';
+const ANY_EXPORT_ERROR_MESSAGE = 'Prefer default export to be present on every file that has export.';
+
 module.exports = {
   meta: {
     type: 'suggestion',
     docs: {
       category: 'Style guide',
-      description: 'Prefer a default export if module exports a single name.',
+      description: 'Prefer a default export if module exports a single name or multiple names.',
       url: docsUrl('prefer-default-export'),
     },
-    schema: [],
+    schema: [{
+      type: 'object',
+      properties:{
+        target: {
+          type: 'string',
+          enum: ['single', 'any'],
+          default: 'single',
+        },
+      },
+      additionalProperties: false,
+    }],
   },
 
   create(context) {
@@ -19,7 +32,8 @@ module.exports = {
     let hasStarExport = false;
     let hasTypeExport = false;
     let namedExportNode = null;
-
+    // get options. by default we look into files with single export
+    const { target = 'single' } =  context.options[0] || {};
     function captureDeclaration(identifierOrPattern) {
       if (identifierOrPattern && identifierOrPattern.type === 'ObjectPattern') {
         // recursively capture
@@ -88,8 +102,13 @@ module.exports = {
       },
 
       'Program:exit': function () {
-        if (specifierExportCount === 1 && !hasDefaultExport && !hasStarExport && !hasTypeExport) {
-          context.report(namedExportNode, 'Prefer default export.');
+        if (hasDefaultExport || hasStarExport || hasTypeExport) {
+          return;
+        }
+        if (target === 'single' && specifierExportCount === 1) {
+          context.report(namedExportNode, SINGLE_EXPORT_ERROR_MESSAGE);
+        } else if (target === 'any' && specifierExportCount > 0) {
+          context.report(namedExportNode, ANY_EXPORT_ERROR_MESSAGE);
         }
       },
     };
--- a/tests/src/rules/prefer-default-export.js
+++ b/tests/src/rules/prefer-default-export.js
@@ -7,6 +7,10 @@ import { version as tsEslintVersion } from 'typescript-eslint-parser/package.jso
 const ruleTester = new RuleTester();
 const rule = require('../../../src/rules/prefer-default-export');
 
+const SINGLE_EXPORT_ERROR_MESSAGE = 'Prefer default export on a file with single export.';
+const ANY_EXPORT_ERROR_MESSAGE = 'Prefer default export to be present on every file that has export.';
+
+// test cases for default option { target: 'single' }
 ruleTester.run('prefer-default-export', rule, {
   valid: [].concat(
     test({
@@ -108,7 +112,7 @@ ruleTester.run('prefer-default-export', rule, {
         export function bar() {};`,
       errors: [{
         type: 'ExportNamedDeclaration',
-        message: 'Prefer default export.',
+        message: SINGLE_EXPORT_ERROR_MESSAGE,
       }],
     }),
     test({
@@ -116,7 +120,7 @@ ruleTester.run('prefer-default-export', rule, {
         export const foo = 'foo';`,
       errors: [{
         type: 'ExportNamedDeclaration',
-        message: 'Prefer default export.',
+        message: SINGLE_EXPORT_ERROR_MESSAGE,
       }],
     }),
     test({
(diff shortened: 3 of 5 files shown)


src/rules/prefer-default-export.js
```diff
@@ -19,7 +33,8 @@ module.exports = {
     let hasStarExport = false;
     let hasTypeExport = false;
     let namedExportNode = null;
-
+    // get options. by default we look into files with single export
+    const { target } =  context.options[0] || { target: 'single' };
```
ljharb, line 37:
```suggestion
    const { target = 'single' } =  context.options[0] || {};
```

src/rules/prefer-default-export.js
```diff
@@ -88,8 +103,16 @@ module.exports = {
...
       'Program:exit': function () {
-        if (specifierExportCount === 1 && !hasDefaultExport && !hasStarExport && !hasTypeExport) {
-          context.report(namedExportNode, 'Prefer default export.');
+        if (target === 'single') {
+          if (specifierExportCount === 1 && !hasDefaultExport && !hasStarExport && !hasTypeExport) {
+            context.report(namedExportNode, SINGLE_EXPORT_ERROR_MESSAGE);
+          }
+          return;
+        }
+        if (target === 'any') {
```
ljharb, line 112:
```suggestion
        } else if (target === 'any') {
```

src/rules/prefer-default-export.js
```diff
@@ -88,8 +103,16 @@ module.exports = {
...
+          if ( specifierExportCount>0 && !hasDefaultExport && !hasStarExport && !hasTypeExport) {
```
ljharb, line 113:
```suggestion
          if (specifierExportCount > 0 && !hasDefaultExport && !hasStarExport && !hasTypeExport) {
```

tests/src/rules/prefer-default-export.js
```diff
@@ -7,153 +7,382 @@ import { version as tsEslintVersion } from 'typescript-eslint-parser/package.jso
...
+  test({
+    code: `
         export const foo = 'foo';
         export const bar = 'bar';`,
-    }),
-    test({
-      code: `
-        export default function bar() {};`,
-    }),
-    test({
```
ljharb, line 21: removing test cases is a red flag for a breaking change - can we keep all the existing ones with implicit defaults, and add *new* mostly-duplicate ones with the explicit target option?

ljharb (changes requested):

tests/src/rules/prefer-default-export.js
azyzz228: That was my intention, but I accidentally moved this specific test case without having its copy. I refactored the test cases to have existing ones (with implicit) untouched, and also added new duplicate ones with explicit target option set to 'any'.

src/rules/prefer-default-export.js
```diff
@@ -2,15 +2,29 @@
 
 import docsUrl from '../docsUrl';
 
+
```
ljharb, line 5:
```suggestion
```

src/rules/prefer-default-export.js
```diff
@@ -2,15 +2,29 @@
...
-    schema: [],
+    schema: [{
+      type: 'object',
+      properties:{
+        target: { 
+          type: 'string',
+          enum: ['single', 'any'],
+        },
+      },
+      additionalProperties: false,
```
ljharb, line 25: i wonder if this is a breaking change - previously, could someone have specified any options they wanted?

src/rules/prefer-default-export.js
```diff
@@ -2,15 +2,29 @@
...
+      description: 'Prefer a default export if module exports a single name or multiple names.',
       url: docsUrl('prefer-default-export'),
     },
-    schema: [],
+    schema: [{
+      type: 'object',
+      properties:{
+        target: { 
+          type: 'string',
+          enum: ['single', 'any'],
```
ljharb, line 22:
```suggestion
          enum: ['single', 'any'],
          default: 'single',
```

src/rules/prefer-default-export.js
```diff
@@ -2,15 +2,29 @@
...
+        },
+      },
+      additionalProperties: false,
+    },
+    ],
```
ljharb, line 27:
```suggestion
    }],
```

src/rules/prefer-default-export.js
```diff
@@ -88,8 +103,15 @@ module.exports = {
       },
 
       'Program:exit': function () {
-        if (specifierExportCount === 1 && !hasDefaultExport && !hasStarExport && !hasTypeExport) {
-          context.report(namedExportNode, 'Prefer default export.');
+        if (target === 'single') {
+          if (specifierExportCount === 1 && !hasDefaultExport && !hasStarExport && !hasTypeExport) {
+            context.report(namedExportNode, SINGLE_EXPORT_ERROR_MESSAGE);
+          }
+          return;
```
ljharb, line 110:
```suggestion
```

src/rules/prefer-default-export.js
```diff
@@ -88,8 +103,15 @@ module.exports = {
...
+        } else if (target === 'any') {
+          if (specifierExportCount > 0 && !hasDefaultExport && !hasStarExport && !hasTypeExport) {
```
ljharb, line 112: seems like we can early return if `hasDefaultExport || hasStarExport || hasTypeExport`, for both code paths?

tests/src/rules/prefer-default-export.js
```diff
@@ -7,153 +7,329 @@ import { version as tsEslintVersion } from 'typescript-eslint-parser/package.jso
...
+    options: [{
+      target: 'any',
+    }],
+    errors: [{
+      message: ANY_EXPORT_ERROR_MESSAGE,
+    }],
+  }),
+);
+
+ruleTester.run('prefer-default-export', rule, { valid, invalid });
```
ljharb, line 331: can we flip this back to the inline valid/invalids, so the diff isn't so noisy?

src/rules/prefer-default-export.js
azyzz228: previously, rule did not have any options so if someone specified any option, that would flag an error. I am wondering if the problem is in using `enum`. However, it works properly on my local machine and flags an error if `target !== 'single && target !== 'any'`

src/rules/prefer-default-export.js
azyzz228: yeah, that makes sense.

src/rules/prefer-default-export.js
azyzz228: I looked at error code for failing test cases. They are all same. The issue was raised for similar error before https://github.com/import-js/eslint-plugin-import/issues/2118.

tests/src/rules/prefer-default-export.js
azyzz228: will do

src/rules/prefer-default-export.js
azyzz228: seems like adding `default: 'single'` to the schema solved the issue. All the cases in older versions of `eslint` are passing now

tests/src/rules/prefer-default-export.js
azyzz228: I brought back all inline test cases as it was before, so every test case which was present before is present now, too.

docs/rules/prefer-default-export.md
```diff
@@ -2,10 +2,44 @@
...
+// you can manually specify it
+"rules":{
+    "import/prefer-default-export": [
+        ( "off" | "warn" | "error" ),
+        { "target": "single" }
+    ]
+}
+
+// config setup below will also work
+"rules":{
```
ljharb, line 38:
```suggestion
"rules": {
    "import/prefer-default-export": [
        ( "off" | "warn" | "error" ),
        { "target": "single" }
    ]
}

// config setup below will also work
"rules": {
```

docs/rules/prefer-default-export.md
```diff
@@ -58,3 +92,95 @@ export { foo as default }
 // Any batch export will disable this rule. The remote module is not inspected.
 export * from './other-module'
 ```
+
+#### any
+
+**Definition**: any exporting file must contain default export.
```
ljharb, line 98:
```suggestion
**Definition**: any exporting file must contain a default export.
```

docs/rules/prefer-default-export.md
```diff
@@ -58,3 +92,95 @@ export { foo as default }
...
+
+How to setup config file for this rule:
+
+```javascript
+// you can manually specify it
+"rules":{
```
ljharb, line 104:
```suggestion
"rules": {
```

docs/rules/prefer-default-export.md
```diff
@@ -58,3 +92,95 @@ export { foo as default }
...
+    "import/prefer-default-export": [
+        ( "off" | "warn" | "error" ),
+        { "target": "any" }
+    ]
+}
+```
+
+
+The following patterns are no warnings:
```
ljharb, line 113:
```suggestion
The following patterns are *not* considered warnings:
```

docs/rules/prefer-default-export.md
```diff
@@ -58,3 +92,95 @@ export { foo as default }
...
+```
+
+```javascript
+// export-star.js
+
+// Any batch export will disable this rule. The remote module is not inspected.
+export * from './other-module'
+```
+
+Followings are considered as warnings:
```
ljharb, line 154:
```suggestion
The following patterns are considered warnings:
```

src/rules/prefer-default-export.js
ljharb: I think it as the defaulting on line 36 instead, but glad to hear it's working.

What I'm concerned with, though, is someone passing `[2, { literallyAnything: true }]` to the rule, and previously/currently, it still works - but after this change, eslint would throw an error for this case.

src/rules/prefer-default-export.js
```diff
@@ -88,8 +102,17 @@ module.exports = {
...
-          context.report(namedExportNode, 'Prefer default export.');
+        if (hasDefaultExport || hasStarExport || hasTypeExport) {
+          return;
+        }
+        if (target === 'single') {
+          if (specifierExportCount === 1) {
+            context.report(namedExportNode, SINGLE_EXPORT_ERROR_MESSAGE);
+          }
+        } else if (target === 'any') {
+          if (specifierExportCount > 0) {
```
ljharb, line 113:
```suggestion
        if (target === 'single' && specifierExportCount === 1) {
           context.report(namedExportNode, SINGLE_EXPORT_ERROR_MESSAGE);
        } else if (target === 'any' && specifierExportCount > 0) {
```

src/rules/prefer-default-export.js
azyzz228: I tried to pass random value and it got added to the default `{ target: 'single' }` and error was flagged.

Test case:
```javascript
    test({
      code: `
              export const foo = 'foo';
              export const bar = 'bar'; 
              export default 42;`,
      options: [{
        breakTheRule: 'any',
      }],
    }),
```

error:

```
  1) prefer-default-export valid 
              export const foo = 'foo';
              export const bar = 'bar'; 
              export default 42;:
     Error: rule-tester:
        Configuration for rule "prefer-default-export" is invalid:
        Value {"breakTheRule":"any","target":"single"} should NOT have additional properties.
```

The same error is flagged when I tried to pass random option to other rule: `import/named`:

test case snippet:

```javascript
    test({
      code: 'const { baz } = require("./bar")',
      errors: [error('baz', './bar')],
      options: [{ breakNamedRule: true }],
    }),
```

error message:

```
  1) named valid const { baz } = require("./bar"):
     Error: rule-tester:
        Configuration for rule "named" is invalid:
        Value {"breakNamedRule":true} should NOT have additional properties.
```

ljharb (approved):

PR #2605, opened 2022-11-29
Update eslint-doc-generator to v1.0.0

bmish:
First stable [v1.0.0](https://github.com/bmish/eslint-doc-generator/releases/tag/v1.0.0) release of [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator).

Originally added in:
* https://github.com/import-js/eslint-plugin-import/pull/2582

The only change is that one CLI option was renamed.


ljharb (approved):
Thanks!

PR #2608, opened 2022-11-30
[Fix] `no-cycle`: Accept `import typeof`, like `import type`

### LINKED ISSUE #2607: no-cycle should accept Flow "typeof" imports, just like "type"
This is an extension of #1098 / #1343. In addition to `import type`, Flow has a syntax `import typeof`. For example ([try-flow link](https://flow.org/try/#0JYWwDg9gTgLgBDAnmAphAZnA3nArgZxQFF10UBjeAXznSghDgHIoUBDSpgbgCgeA3NlFoAuPIRJlKXOAHpZcABZt8CRSjj42IDUlRwVcAErtKAOgLFSFGDyA)):
```js
import typeof { useEffect } from 'react';

var f: useEffect; // has the same type as React.useEffect
```

Like with `import type`, these imports are transpiled away and have no effect at runtime, so they should be given the same treatment as `import type`: that is, they should be ignored.

Currently they're not.

I have a fix which I'll send in a bit.

gnprice:
Fixes #2607.

In addition to `import type`, Flow has a syntax `import typeof`. For example ([try-flow link](https://flow.org/try/#0JYWwDg9gTgLgBDAnmAphAZnA3nArgZxQFF10UBjeAXznSghDgHIoUBDSpgbgCgeA3NlFoAuPIRJlKXOAHpZcABZt8CRSjj42IDUlRwVcAErtKAOgLFSFGDyA)):

```js
import typeof { useEffect } from 'react';

var f: useEffect; // has the same type as React.useEffect
```

Like with `import type`, these imports are transpiled away and have no effect at runtime, so they should be given the same treatment as `import type`: that is, they should be ignored.

Currently they're not. This PR fixes that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`no-unresolved`], [`extensions`]: ignore type only exports ([#2436], thanks [@Lukas-Kullmann])
 - `ExportMap`: add missing param to function ([#2589], thanks [@Fdawgs])
 - [`no-unused-modules`]: `checkPkgFieldObject` filters boolean fields from checks ([#2598], thanks [@mpint])
+- [`no-cycle`]: accept Flow `typeof` imports, just like `type` ([#2608], thanks [@gnprice])
 
 ### Changed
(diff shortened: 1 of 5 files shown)


ljharb (approved):
Thanks, this is great!

gnprice:
Thanks for the swift review and merge!

Issue #2348, opened 2022-01-03
Huge `import/no-cycle` performance downgrade in 2.25.3


zaicevas:
I upgraded `eslint-plugin-import` 2.22.1 -> 2.25.3 and the performance downgrade is huge. `eslint` is `7.24.0`. 

Roughly a hundred files:

```
Rule                                       |  Time (ms) | Relative
:------------------------------------------|-----------:|--------:
import/no-cycle                            | 304818.252 |    97.0%
prettier/prettier                          |   2347.669 |     0.7%
import/named                               |    754.553 |     0.2%
```

`2.24.2` is fine:

```
Rule                                       | Time (ms) | Relative
:------------------------------------------|----------:|--------:
import/no-cycle                            |  4206.707 |    23.0%
prettier/prettier                          |  1914.555 |    10.5%
react/prefer-stateless-function            |   849.317 |     4.6%
```

ljharb:
Thanks, that's quite large.

The only changes that touched ExportMap in that range are 9a744f73 and #2212.

Do you have a TypeScript codebase? Also, can you confirm that v2.25.4 (which contains #2341) has the same speed drop?

zaicevas:
> Do you have a TypeScript codebase? 

Yes

> can you confirm that v2.25.4 (which contains #2341) has the same speed drop?

Yes:

```
Rule                                          |  Time (ms) | Relative
:---------------------------------------------|-----------:|--------:
import/no-cycle                               | 300578.485 |    95.2%
prettier/prettier                             |   2384.299 |     0.8%
react/prefer-stateless-function               |    866.305 |     0.3%
```

ljharb:
That strongly suggests it’s #2240. cc @mrmckeb

mrmckeb:
Are you able to share a small reproduction of the issue @zaicevas? Or if not, can you share your `tsconfig.json`?

zaicevas:
> Are you able to share a small reproduction of the issue @zaicevas?

Sorry, won't be able to do that due to lack of time.

`tsconfig.json`:
```
{
  "compilerOptions": {
    "target": "ES2016",
    "module": "commonjs",
    "jsx": "react-jsx",
    "noEmit": true,
    "strict": true,
    "noImplicitAny": false,
    "noImplicitThis": false,
    "baseUrl": "./app/",
    "esModuleInterop": true,
    "allowUnreachableCode": false,
    "isolatedModules": true
  },
  "include": [
[...]

Issue #2076, opened 2021-05-15
`no-cycle` hangs since v2.23.0


alumni:
With both v2.23.0 and v2.23.1, running `eslint` on the entire project hangs. Also, it will make vscode eslint plugin hang as well when saving a file (with fix on save enabled).

`.eslintrc`:
```jsonc
{
  "parser": "@typescript-eslint/parser",
  "parserOptions": {
    "project": "tsconfig.json"
  },
  "rules": {
    "import/no-cycle": ["error", { "ignoreExternal": true, "maxDepth": 7 }]
  },
  "settings": {
    "import/parsers": {
      "@typescript-eslint/parser": [".ts", ".tsx"]
    },
    "import/resolver": {
      "typescript": {
        "project": "tsconfig.json"
      }
    }
  }
}
```
`tsconfig.json`:
```jsonc
{
    "compilerOptions": {
        "paths": {
            "@common": ["src/app/common"],
        }
    }
}
```

ljharb:
I suspect this might be due to #2077, but I'll leave this open pending confirmation that it's still an issue with v0.23.2 (not yet released).

Either way, does the issue still occur with `no-cycle` disabled?

alumni:
> Either way, does the issue still occur with no-cycle disabled?

It's fine with `no-cycle` disabled. We had it anyway disabled in the pre-commit hook to speed things up.

In v2.32.2 `eslint --rule 'import/no-cycle: 0'` will finish in a reasonable amount of time (without cache <1min), but simply running `eslint` will not finish even after 10min.

ljharb:
Does it hang running on the CLI, or only in vscode?

alumni:
As mentioned above, both the `eslint` cli and the vscode extension hang. What I noticed:
* `eslint` cli hangs when run for the entire project
* `eslint` cli doesn't hang when `no-cycle` is disabled
* vscode doesn't hang when a single file is opened (might be a bit slower than before though, not sure if it felt faster before because of caching)
* vscode hangs when a file is saved and fix on save is enabled

saharrachamimfg:
This is a critical issue. Can we please fix or revert?

SRachamim:
This is a critical issue. Can we please fix or revert?

ljharb:
Does this still occur with v2.23.4?

I’d love to fix it, but without a repro repo, it’s very difficult to debug.

alumni:
[...]

PR #2612, opened 2022-12-12
Perf: Optimize ignoreModule calls in no-cycle

DHedgecock:
Hello 👋 - Thanks for taking the time to review this PR

### Summary

This PR adds a few performance optimizations to the `no-cycle` rule to make calls to `ignoreModule` faster.

### Details

_Lots of details because digging into this was really interesting_

I was recently profiling our ESLint suite and the `no-cycle` rule was responsible for 79% of the total lint time. I found #2348 and attempted to speed up the rule by opting _IN_ to the `ignoreExternal` option for the rule, but surprisingly that *increased* our total lint time from ~5min to ~9min which seemed odd 🤔 

This comment had a very helpful hint that lead to the issue: https://github.com/import-js/eslint-plugin-import/issues/2076#issuecomment-931802708

When the `ignoreExternal` option is enabled the `ignoreModule` will call `resolve` for each module to determine if it's an external module. I added some counters and found that our repo, which has ~2,000 modules, ended up calling this check ~215,000 times (that number wasn't surprising since each file does a breadth first search of the file's entire dependency tree) so attempting to ignore external modules can actually add a lot of work calling `resolve`

### Optimization

This update optimizes this codepath by adding:

1. A new early return check for relative path imports
2. A simple object cache of evaluated modules to short circuit subsequent evaluations

#### Results

For our medium sized codebase this update:

1. Reduced total lint time from ~5min to ~2min 15s when we opted in to `ignoreExternal`
2. Reduced relative time spent in no-cycle from ~80% to ~42% 

_Note for these comparisons I disabled all the other rules that use `ExportMap` to try and ensure rule order and building the export map wasn't variable._

### DIFF
--- a/src/rules/no-cycle.js
+++ b/src/rules/no-cycle.js
@@ -10,6 +10,16 @@ import moduleVisitor, { makeOptionsSchema } from 'eslint-module-utils/moduleVisi
 import docsUrl from '../docsUrl';
 
 const traversed = new Set();
+/** 
+ * Cache of resolved paths for faster subsequent lookups
+ * @type {{ [importPath: string]: string }} 
+ */
+const resolvedPaths = {};
+/** 
+ * Cache of external module evaluations for faster subsequent lookups
+ * @type {{ [resolvedPath: string]: boolean }} 
+ */
+const externalModules = {};
 
 module.exports = {
   meta: {
@@ -52,11 +62,36 @@ module.exports = {
 
     const options = context.options[0] || {};
     const maxDepth = typeof options.maxDepth === 'number' ? options.maxDepth : Infinity;
-    const ignoreModule = (name) => options.ignoreExternal && isExternalModule(
-      name,
-      resolve(name, context),
-      context,
-    );
+    
+    const ignoreModule = (name) => {
+      // Note: This fn is optimized to eliminate call time because it may be
+      // called hundreds of thousands of times during a full repository lint run
+      // and can add minutes to the total time in large codebases
+
+      // Low cost evaluation; do not ignore if not opted in or module starts
+      // with a relative path
+      // NOTE: skipping further resolve codepath also ensures that only unique
+      // import paths are set in resolvedPaths, eg "./index.js", or "../file.js"
+      // should not be stored because there could be duplicates that resolve to
+      // different paths
+      if (!options.ignoreExternal || name[0] === '.') return false;
+
+      // High cost evaluation; cache evaluations of resolve() and
+      // isExternalModule() calls for faster subsequent lookups
+      let path = resolvedPaths[name];
+      if (path === undefined) {
+        path = resolve(name, context);
+        resolvedPaths[name]= path;
+      }
+
+      let isExternal = externalModules[path];
+      if (isExternal === undefined) {
+        isExternal = isExternalModule(name, path, context);
+        externalModules[path] = isExternal;
+      }
+      
+      return isExternal;
+    };
 
     function checkSourceValue(sourceNode, importer) {
       if (ignoreModule(sourceNode.value)) {
--- a/src/rules/no-cycle.js
+++ b/src/rules/no-cycle.js
@@ -81,7 +81,7 @@ module.exports = {
       let path = resolvedPaths[name];
       if (path === undefined) {
         path = resolve(name, context);
-        resolvedPaths[name]= path;
+        resolvedPaths[name] = path;
       }
 
       let isExternal = externalModules[path];


src/rules/no-cycle.js
```diff
@@ -10,6 +10,16 @@ import moduleVisitor, { makeOptionsSchema } from 'eslint-module-utils/moduleVisi
...
+/** 
+ * Cache of resolved paths for faster subsequent lookups
+ * @type {{ [importPath: string]: string }} 
+ */
+const resolvedPaths = {};
+/** 
+ * Cache of external module evaluations for faster subsequent lookups
+ * @type {{ [resolvedPath: string]: boolean }} 
+ */
+const externalModules = {};
```
ljharb, line 23: these caches need to be cleared in the same place `traversed` is

src/rules/no-cycle.js
```diff
@@ -52,11 +62,36 @@ module.exports = {
...
+      // called hundreds of thousands of times during a full repository lint run
+      // and can add minutes to the total time in large codebases
+
+      // Low cost evaluation; do not ignore if not opted in or module starts
+      // with a relative path
+      // NOTE: skipping further resolve codepath also ensures that only unique
+      // import paths are set in resolvedPaths, eg "./index.js", or "../file.js"
+      // should not be stored because there could be duplicates that resolve to
+      // different paths
+      if (!options.ignoreExternal || name[0] === '.') return false;
```
ljharb, line 77: couldn't a resolver choose to configure a relative path as external?

src/rules/no-cycle.js
```diff
@@ -52,11 +62,36 @@ module.exports = {
...
+
+      // High cost evaluation; cache evaluations of resolve() and
+      // isExternalModule() calls for faster subsequent lookups
+      let path = resolvedPaths[name];
+      if (path === undefined) {
+        path = resolve(name, context);
+        resolvedPaths[name]= path;
```
ljharb, line 84:
```suggestion
        resolvedPaths[name] = path;
```

src/rules/no-cycle.js
DHedgecock: The `Program:exit` hook is called at the exit of each file, so clearing the caches there cut the number of cache hits in our repo by about 50% (although it only added ~45s to our repo test case)

The high level attempt at this update was to quick filter down to "maybe external" modules like `'react'`, `'@reduxjs/toolkit'` or `'@/some/alias/path'` - where those resolved paths can be safely cached across files 🤔 

Is that possible to accomplish with any of the existing data structures?

src/rules/no-cycle.js
DHedgecock: 🤔 I suppose so, there's probably some monorepo setup out there like that...

I think removing this assumption means that we can't confidently cached the resolved module path we looked up because then there could be duplicate import paths that resolve to different files (eg import paths of `'./index.js'`).

Trying to approach this from a different angle: The current time to determine if a module is "external" is the performance bottleneck and this PR is attempting to short circuit that whenever possible. 

Do you have suggestions on an approach?

DHedgecock: What if the no-cycle rule option `ignoreExternal` could be a boolean or a regex, and if it's a regex it's used for this short circuit?

DHedgecock: I'm not sure what the best configuration fields are, but doing a quick test with a regex short circuit has about the same performance improvement as the cache implementation.

Alternative approach tested:
```js
    const ignoreModule = (name) => {
      if (!options.ignoreExternal ) return false;

      if (options.ignoreExternalRegex) {
        return options.ignoreExternalRegex.test(name);
      }

      // High cost evaluation; cache evaluations of resolve() and
      // isExternalModule() calls for faster subsequent lookups
      const path = resolve(name, context);

      let isExternal = externalModules[path];
      if (isExternal === undefined) {
        isExternal = isExternalModule(name, path, context);
        externalModules[path] = isExternal;
      }
      
      return isExternal;
    };
```

src/rules/no-cycle.js
ljharb: oh, hmm - you're right, it's less about clearing it after the file and more about clearing it after the "run" - since in editors, eslint can be long-running.

I think that in every file, there could be a different resolver, so the cache would have to be on the resolver *and* the resolver config *and*  the fully normalized file path, i think?

src/rules/no-cycle.js
ljharb: Maybe the better question is, why is `isExternalModule` slow?

Perhaps the node and webpack resolvers can add a cache there, rather than inside the main plugin?

soryy708: Aren't we doing a bunch of regex matching in `isExternalModule`?

ljharb: generally that's fast, but if it can be sped up, let's!

ljharb:
Unfortunately the fork was deleted, which makes this PR unrecoverable :-/

PR #2613, opened 2022-12-12
[New] `no-absolute-path`: add fixer

adipascu:
Hello, this is my first contribution to this repo. Should I change anything?

### DIFF
--- a/src/rules/no-absolute-path.js
+++ b/src/rules/no-absolute-path.js
@@ -1,3 +1,4 @@
+import path from 'path';
 import moduleVisitor, { makeOptionsSchema } from 'eslint-module-utils/moduleVisitor';
 import { isAbsolute } from '../core/importType';
 import docsUrl from '../docsUrl';
@@ -10,13 +11,26 @@ module.exports = {
       description: 'Forbid import of modules using absolute paths.',
       url: docsUrl('no-absolute-path'),
     },
+    fixable: 'code',
     schema: [ makeOptionsSchema() ],
   },
 
   create(context) {
     function reportIfAbsolute(source) {
       if (isAbsolute(source.value)) {
-        context.report(source, 'Do not import modules using an absolute path');
+        context.report({
+          node: source,
+          message: 'Do not import modules using an absolute path',
+          fix: fixer => {
+            const resolvedContext = context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename();
+            // node.js and web imports work with posix style paths ("/")
+            let relativePath = path.posix.relative(path.dirname(resolvedContext), source.value);
+            if (!relativePath.startsWith('.')) {
+              relativePath = './' + relativePath;
+            }
+            return fixer.replaceText(source, JSON.stringify(relativePath));
+          },
+        });
       }
     }
 
--- a/tests/src/rules/no-absolute-path.js
+++ b/tests/src/rules/no-absolute-path.js
@@ -53,48 +53,74 @@ ruleTester.run('no-absolute-path', rule, {
   invalid: [
     test({
       code: 'import f from "/foo"',
+      filename: '/foo/bar/index.js',
       errors: [error],
+      output: 'import f from ".."',
+    }),
+    test({
+      code: 'import f from "/foo/bar/baz.js"',
+      filename: '/foo/bar/index.js',
+      errors: [error],
+      output: 'import f from "./baz.js"',
     }),
     test({
       code: 'import f from "/foo/path"',
+      filename: '/foo/bar/index.js',
       errors: [error],
+      output: 'import f from "../path"',
     }),
     test({
       code: 'import f from "/some/path"',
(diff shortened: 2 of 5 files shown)


adipascu:
After reading https://github.com/nodejs/node/issues/6049 I think I should have used [path.posix](https://nodejs.org/api/path.html#pathposix).
Are any of these tests ran on MS Windows?

ljharb:
Yes, the travis tests are on Mac and the appveyor ones on Windows.

src/rules/no-absolute-path.js
```diff
@@ -10,13 +11,25 @@ module.exports = {
...
-        context.report(source, 'Do not import modules using an absolute path');
+        context.report({
+          node: source,
+          message: 'Do not import modules using an absolute path',
+          fix: fixer => {
+            const resolvedContext = context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename();
+            let relativePath = path.relative(path.dirname(resolvedContext), source.value);
+            if (!relativePath.startsWith('.')){
+              relativePath = '.' + path.sep + relativePath;
+            }
```
ljharb, line 29:
```suggestion
            if (!relativePath.startsWith('.')) {
              relativePath = path.join('.', relativePath);
            }
```

adipascu: `path.join('.', relativePath)` drops the leading `.`, this is why I chose to do explicit string concatenation.

adipascu:
I've updated the PR to always do posix style paths.
- This seems to be [the way node.js works in general](https://github.com/nodejs/node/issues/6049)
 - [Browser URLs](https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname) seem to only support `/` posix separators
 I hope this will pass windows CI.

adipascu:
I've updated the PR to always do posix style paths.
- This seems to be [the way node.js works in general](https://github.com/nodejs/node/issues/6049)
 - [Browser URLs](https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname) seem to only support `/` posix separators
 I hope this will pass windows CI.
 Edit: sorry for the force pushes, should I do separate commits with any feedback/follow up work?

ljharb:
Force pushes are always fine; if i'm curious about the push diff i can click on "force pushed" in the github timeline :-)

src/rules/no-absolute-path.js
ljharb: ah ok, fair enough

src/rules/no-absolute-path.js
```diff
@@ -10,13 +11,26 @@ module.exports = {
...
       if (isAbsolute(source.value)) {
-        context.report(source, 'Do not import modules using an absolute path');
+        context.report({
+          node: source,
+          message: 'Do not import modules using an absolute path',
+          fix: fixer => {
+            const resolvedContext = context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename();
+            // node.js and web imports work with posix style paths ("/")
+            let relativePath = path.posix.relative(path.dirname(resolvedContext), source.value);
+            if (!relativePath.startsWith('.')){
```
ljharb, line 28:
```suggestion
            if (!relativePath.startsWith('.')) {
```

tests/src/rules/no-absolute-path.js
```diff
@@ -53,48 +55,76 @@ ruleTester.run('no-absolute-path', rule, {
...
+      output: 'var f = require("../../some/path")',
     }),
     // validate amd
     test({
       code: 'require(["/some/path"], function (f) { /* ... */ })',
+      filename: '/foo/bar/index.js',
       options: [{ amd: true }],
       errors: [error],
+      output:
+        'require(["../../some/path"], function (f) { /* ... */ })',
```
ljharb, line 119:
```suggestion
      output: 'require(["../../some/path"], function (f) { /* ... */ })',
```
i smell prettier :-p

tests/src/rules/no-absolute-path.js
```diff
@@ -53,48 +55,76 @@ ruleTester.run('no-absolute-path', rule, {
...
     }),
     test({
       code: 'define(["/some/path"], function (f) { /* ... */ })',
+      filename: '/foo/bar/index.js',
       options: [{ amd: true }],
       errors: [error],
+      output:
+        'define(["../../some/path"], function (f) { /* ... */ })',
```
ljharb, line 127:
```suggestion
      output: 'define(["../../some/path"], function (f) { /* ... */ })',
```

ljharb (approved):
looks good if tests pass

adipascu:
I've accepted the white-space suggestions (and they made separate commits).

I hope this gets squash merged.

Thank you for being receptive to my changes!

adipascu:
CI had network issues, I've triggered a fresh round.

adipascu:
Hi @ljharb, could we merge this in?

Issue #2390, opened 2022-02-25
Rule Suggestion: require-import-type-modifiers


Nokel81:
With the release of typescript 4.5, it is possible to have [type Modifiers on Import Names](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#type-on-import-names).

It would be great if there was a (fixable) rule that could be used to require these.

I am willing to implement this.

ljharb:
It seems better to require them be separate `import type` statements.

Nokel81:
Not if there are only one or two of either kind. But that could be a different option I guess.

ljharb:
It seems unfortunate to have a rule that is only useful for typescript, but perhaps a rule that detects your TS (or flow) version, and can be configured for an order of preferences - ie, between "types as separate import statements", "types mixed with values, but marked with a type modifier", you could prefer one over the other, and it would fall back to the next one if the preferred one wasn't supported, and if neither are supported, the rule would noop.

yume-chan:
[`@typescript-eslint/consistent-type-imports`](https://typescript-eslint.io/rules/consistent-type-imports/) can enforce type-only imports to have a `type` specifier.

[`import/consistent-type-specifier-style`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/consistent-type-specifier-style.md) can choose to ban `import type` or `import { type }`

So bad I need two rules! And the second one hasn't been released yet!

PR #2614, opened 2022-12-14
New typescript flow rule

azyzz228:
Draft version of the new rule proposed at #2390. 

This new rule allows for changing between inline type imports (type modifiers introduced in TS 4.5) and separate type import statements (introduced in TS 3.8). The rule also fixes imports to the preferred option. The rule assumes that user specifies himself which imports are type imports. For example, if there is `enum` that is used as a type, user should import the enum as type. This rule does not check the usage of imports, but rather helps to enforce consistent style for importing types. 

The rule has 2 strict options (inline and separate) and one preferred flow option. In preferred flow option, user passes array of inline and separate in the order of preference. Rule checks if inline is supported, if not, falls back to "separate" as an option. If separate comes first in the array, then resume as if there is strict "separate" option.

Strict option enforces either inline type imports with type modifiers, or separate type imports.

Rule schema for strict options: we pass single string to the rule

```javascript
"import/typescript": [ 0 | 1 | 2, "inline" | "strict" ]
```

### separate option

**Definition**: All type imports must be imported in a separate `import type` statement.

The following patterns are *not* considered warnings:

```javascript
// good1.js

// Separate import type statement
import type { MyType } from "./typescript.ts"
```

```javascript
// good2.js

// Separate import type statement
import type { MyType as Persona, Bar as Foo } from "./typescript.ts"
```

The following patterns are considered warnings:

```javascript
// bad1.js

// type imports must be imported in separate import type statement
import {type MyType,Bar} from "./typescript.ts"

// gets fixed to the following:
import {Bar} from "./typescript.ts"
import type { MyType } from "./typescript.ts"
```

```javascript
// bad2.js

// type imports must be imported in separate import type statement
import {type MyType,type Foo} from "./typescript.ts"

// gets fixed to the following:
import type { MyType, Foo} from "./typescript.ts"
```

```javascript
// bad3.js

// type imports must be imported in separate import type statement
import {type MyType as Persona,Bar,type Foo as Baz} from "./typescript.ts"

// gets fixed to the following:
import {Bar} from "./typescript.ts"\nimport type { MyType as Persona, Foo as Baz } from "./typescript.ts"
```

### inline option

**Definition**: All type imports must be inline imported with type modifiers.

Patterns that do not raise the warning:

```javascript
// good1.js

// no separate import type statement. inline type import exists
import { type MyType } from "./typescript.ts"
```

```javascript
// good2.js

// no separate import type statement. inline type import exists
import { type MyType, Bar, type Foo as Baz } from "./typescript.ts"
```

Patterns are considered warnings and fixed:

```javascript
// bad1.js

// type imports must be imported inline
import type {MyType} from "./typescript.ts"

// gets fixed to the following:
import  {type MyType} from "./typescript.ts"
```

```javascript
// bad1.js

// type imports must be imported inline
import type {MyType, Bar} from "./typescript.ts"

// gets fixed to the following:
import  {type MyType, type Bar} from "./typescript.ts"
```

```javascript
// bad3.js

// type imports must be imported inline
import type {MyType, Bar} from "./typescript.ts"
import {Value} from "./typescript.ts"

// Rule can check if non-type import exists. If yes, then adds inline type imports there. gets fixed to the following:
import {Value, type MyType, type Bar} from "./typescript.ts"
```

Questions:

1) How to handle the cases where there are 2 spaces left when "type" is removed from separate import type. I could not figure out the token for the spaces. Seems like spaces are completely ignored from estree?

```javascript

// option === 'inline'

// input
import type {MyType, Bar} from "./typescript.ts"

//output
import  {type MyType, type Bar} from "./typescript.ts" // there are 2 spaces after import
```

### DIFF
--- a/docs/rules/typescript-flow.md
+++ b/docs/rules/typescript-flow.md
@@ -1,0 +1,24 @@
+This rules helps to define type import patterns for the typescript.
+
+## Rule details
+
+##### rule schema:
+
+```javascript
+"import/prefer-default-export": [
+    ( "off" | "warn" | "error" ),
+	{ "prefer": "none" | "separate" | "modifier" } // default is ???
+]
+```
+### Config Options
+
+There are three avaiable options: `none`, `separate`, `modifier`.
+
+```javascript
+// import-stars.js
+
+// The remote module is not inspected.
+import * from './other-module'
+import * as b from './other-module'
+
+```
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "posttest": "eslint . && npm run update:eslint-docs -- --check",
     "mocha": "cross-env BABEL_ENV=test nyc mocha",
     "tests-only": "npm run mocha tests/src",
+    "debug": "DEBUG=eslint-plugin-import/typescript-flow npm run mocha tests/src/rules/typescript-flow",
     "test": "npm run tests-only",
     "test-compiled": "npm run prepublish && BABEL_ENV=testCompiled mocha --compilers js:babel-register tests/src",
     "test-all": "node --require babel-register ./scripts/testAll",
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -623,6 +623,9 @@ ExportMap.parse = function (path, content, context) {
       // capture declaration
       if (n.declaration != null) {
         switch (n.declaration.type) {
+        case 'TSTypeAliasDeclaration':
+          m.namespace.set(n.declaration.id.name, 'type');
+          break;
         case 'FunctionDeclaration':
         case 'ClassDeclaration':
         case 'TypeAlias': // flowtype with babel-eslint parser
@@ -630,7 +633,6 @@ ExportMap.parse = function (path, content, context) {
         case 'DeclareFunction':
         case 'TSDeclareFunction':
         case 'TSEnumDeclaration':
-        case 'TSTypeAliasDeclaration':
         case 'TSInterfaceDeclaration':
         case 'TSAbstractClassDeclaration':
         case 'TSModuleDeclaration':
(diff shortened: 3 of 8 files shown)


ljharb:
The new rule needs to be added to the shared configs in `src/index.js`, I think.

azyzz228:
Not sure what is causing errors :/. I tried to replicate them on my local machine (node v10, eslint v6), but had many many test cases failing, and completely different error (mine was ELIFECYCLE error). Was wondering maybe you might have any thoughts?

ljharb:
It seems like the tests that are failing are eslint 6 or 7, combined with node < 12. The new TS parser will only be installed on eslint >= 6, and will be v3 on node 10 and 11, and probably v2 on node < 10. On eslint 6, the old TS parser will be present on v20. On node < 8, eslint-import-resolver-typescript will be v1.0.2.

The easiest way to replicate this locally is to run the `tests/dep-time-travel.sh` script yourself, with the same arguments CI uses.

latin-1:
I believe `import { type MyType } from "./typescript.ts"` is not something we really want to have. It will fail if the option `importsNotUsedAsValues` in the `tsconfig.json` is set to `error`.

```ts
import type { T } from "./module"; // ok
import { type T } from "./module"; // error ts(1371)
```

ljharb:
It'd be great to rebase this, so any remaining failing tests can be resolved.

azyzz228:
apologies this PR got abandoned... Would you say there is still a need for this rule? Suggestion for the rule came from here #2390.

ljharb:
There's already a rule for whether you use `import type` or `import { type`, but i don't think there's a rule yet that *forces* you to use a `type` signifier for types or *forces* you not to, so it still seems valuable to me.

PR #2633, opened 2022-12-31
[Docs] add TS import/parsers settings to Readme

erquhart:
Just a line from the `eslint-import-resolver-typescript` readme, which is already linked in a nearby code comment; it took me way too long to figure out critical stuff was in that linked readme. Proposing this as it may help others, not sure if it applies for all typescript uses or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,9 @@ extends:
 # the following lines do the trick
   - plugin:import/typescript
 settings:
+  import/parsers:
+    # Your paths may differ, this example parses all Typescript files within cwd
+    '@typescript-eslint/parser': ['**/.ts', '**/.tsx']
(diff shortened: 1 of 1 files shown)


ljharb:
I'm confused, this should already be covered by https://github.com/import-js/eslint-plugin-import/blob/main/config/typescript.js#L15-L17 which `plugin:import/typescript` provides.

erquhart:
Ah you're right - extending the recommending plugin caused issues and I had to go granular, I assumed extending the typescript plugin would do similarly, didn't realize it brought settings with it. It also looks like I'm the only one that was confused here lol, so I'll just close. Thank you for your work on this amazing tool!

PR #2638, opened 2023-01-02
bump tsconfig paths version

### LINKED ISSUE #2637: json5 dependabot vulnerability due to old tsconfig-paths version
Would be nice if `tsconfig-paths` could be bumped to `^4` so that Github would stop reporting the transitive `json5` dependency as a security vulnerability by using this package. thanks!

osdiab:
closes #2637 , no idea if this breaks anything


ljharb:
It does break things; it's a nonstarter; it's also completely unnecessary and a duplicate of #2447.

PR #2640, opened 2023-01-02
Add a quick note on how unbound imports and --fix

minervabot:
Having unbound imports mixed among the bound ones causes unexpected and incorrect seeming results. I spent several hours trying to fix this problem only to find it was well known!

Since this is only a change to one markdown file, I assume no other tests are required. Hopefully this is helpful.

### DIFF
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -77,6 +77,25 @@ import foo from './foo';
 var path = require('path');
 ```
 
+## Limitations of `--fix`
+
+Unbound imports are assumed to have side effects, and will never be moved/reordered. This can cause other imports to get "stuck" around them, and the fix to fail.
+
+```javascript
+import b from 'b'
+import 'format.css';  // This will prevent --fix from working.
+import a from 'a'
+```
+
+As a workaround, move unbound imports to be entirely above or below bound ones.
+
+```javascript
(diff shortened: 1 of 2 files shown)


docs/rules/order.md
```diff
@@ -80,6 +80,25 @@ import foo from './foo';
...
+
+Unbound imports are assumed to have side effects, and will never be moved/reordered. This can cause other imports to get "stuck" around them, and the fix to fail.
+
+```javascript
+import b from 'b'
+import 'format.css';  // This will prevent --fix from working.
+import a from 'a'
+```
+
+As a workaround, move unbound imports to be entirely above and below bound ones.
```
ljharb, line 93:
```suggestion
As a workaround, move unbound imports to be entirely above or below bound ones.
```

ljharb (approved):
This sounds fine, but in every eslint rule, you can never rely on autofix covering everything - you always have to fix *some* things manually.

ljharb:
@minervabot right - that's why it's fine :-) but perhaps eslint's own documentation would have been a better place to set up proper expectations about autofixing?

ljharb:
@minervabot yes, because that's potentially the case in every eslint rule, including the ones in core.

PR #2641, opened 2023-01-02
[Fix] Update dependency tsconfig-paths

desaisagar:


ljharb:
No, this is a breaking change, and a duplicate of #2447

PR #2644, opened 2023-01-03
Update package.json

anywhereiromy:
Upgrade tsconfig-paths to 4.1.2 https://github.com/dividab/tsconfig-paths/compare/v3.14.1...v4.0.0 to avoid json5 vulnerability https://github.com/advisories/GHSA-9c47-m6qq-7p4h

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,6 @@
     "minimatch": "^3.1.2",
     "object.values": "^1.1.6",
     "resolve": "^1.22.1",
-    "tsconfig-paths": "^3.14.1"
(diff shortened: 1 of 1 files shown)


mikibakaiki (approved):

ljharb:
json5 v1.0.2 has already been updated with this fix, and either way, it's *not a valid vulnerability* for eslint-plugin-import.

As is the case with almost every JS CVE, the best course of action is to *do nothing* until the ecosystem fixes it for you.

This is a duplicate of #2447.

PR #2646, opened 2023-01-03
Bump tsconfig-paths to latest 4.1.2

mjperrone:
[tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) is a dependency of `eslint-plugin-import`. 

`tsconfig-paths` has a dependency on `json5`. 

`json5` recently had a[ security vulnerability published](https://github.com/advisories/GHSA-9c47-m6qq-7p4h), and a patch released on version `2.2.2`.

tsconfig-paths [just updated their json5](https://github.com/dividab/tsconfig-paths/pull/232) dependency, available for version `4.1.2`. 

This PR bumps `eslint-plugin-import`'s dependency on tsconfig-paths, removing the vulnerability.

PR #2647, opened 2023-01-04
[Fix] Vulnerable tsconfig-paths dependency update

dduzgun-security:
### Issue 
Prototype Pollution in package minimist
### Dependency
eslint-plugin-import 
### Path
`eslint-plugin-import > tsconfig-paths > json5 > minimist`
### Patched in
minimist version >=1.2.6
### More info
https://www.npmjs.com/advisories/1067342

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,6 @@
     "minimatch": "^3.1.2",
     "object.values": "^1.1.6",
     "resolve": "^1.22.1",
-    "tsconfig-paths": "^3.14.1"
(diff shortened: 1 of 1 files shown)


ljharb:
This is a breaking change, and a duplicate of #2447, and a duplicate of #2646. Additionally, json5 v1.0.2 fixes the issue, so there's no need to update. Additionally, this is *not a vulnerability* for the way eslint-plugin-import uses json5.

PR #2648, opened 2023-01-04
Updated tsconfig-paths to fix https://github.com/advisories/GHSA-9c47-m6qq-7p4h

dvTB:


NicolasBelliard:
Is there someone that can review this MR?

ljharb:
This is a breaking change, and a duplicate of #2447, and a duplicate of #2646, and a duplicate of #2657. Additionally, json5 v1.0.2 fixes the issue, so there's no need to update. Additionally, this is *not a vulnerability* for the way eslint-plugin-import uses json5.

PR #2506, opened 2022-07-24
[Fix] `order`: leave more space in rankings for consecutive path groups 

Pearce-Ropion:
Resolves #2494 

Creates more space (numerically) between consecutive groups to allow for what would have been overlapping path groups.

The issue occurs when there are enough (9-10) before/after `pathGroups` in consecutive `groups` where the rankings start to overlap with the previous/next group. See the repro in #2494 for one example of this.

I think a better fix would be to rework some of the ranking logic in `convertPathGroupsForRanks`. However, this change was a non-invasive easy workaround that doesn't introduce any regressions (as far as I can tell).

PR #2674, opened 2023-01-13
`import/order` Fix group ranks order when alphabetizing

### LINKED ISSUE #2671: [import/order] Potential regression in `sibling` and `parent` behavior
We have a bit of a complex `import/order` config and are seeing some new weird behavior with ordering on the `sibling` and `parent` groups on `eslint-plugin-import@2.27.4`:

```js
'import/order': [
  'error',
  {
    alphabetize: {
      order: 'asc',
      caseInsensitive: true,
    },
    groups: ['type', 'builtin', 'external', 'internal', 'parent', 'sibling', 'index', 'object'],
    'newlines-between': 'always',
    pathGroups: [
      { pattern: '@Hub', group: 'external', position: 'after' },
      { pattern: '@Hub/**', group: 'external', position: 'after' },
      { pattern: '@Dash/**', group: 'external', position: 'after' },
      { pattern: '@core/**', group: 'external', position: 'after' },
      { pattern: '@routes/**', group: 'external', position: 'after' },
      { pattern: '@ui', group: 'external', position: 'after' },
      { pattern: '@ui/**', group: 'external', position: 'after' },
    ],
    pathGroupsExcludedImportTypes: ['@Hub', '@Dash', '@core', '@routes', '@ui'],
    warnOnUnassignedImports: true,
  },
],
```

Here's what our code looks like on `eslint-plugin-import@2.26.0`:

```js
import React from 'react';

import Flex from '@ui/Flex';
import Menu, { MenuItem, MenuDivider } from '@ui/Menu';

import { isPolymorphic } from '../../utils';

import classes from './style.module.scss';
import './style.scss';
```

And it wants it to be now:

```js
import React from 'react';

import Flex from '@ui/Flex';
import Menu, { MenuItem, MenuDivider } from '@ui/Menu';

import classes from './style.module.scss';
import './style.scss';

import { isPolymorphic } from '../../utils';
```

However if I swap the the order of `parent` and `sibling` in the config to this then the behavior we were seeing on v2.26.0 remains:

```js
groups: ['type', 'builtin', 'external', 'internal', 'sibling', 'parent', 'index', 'object']
```

Pearce-Ropion:
Resolves #2671 

Correctly sorts `groupRanks` in numerical order when alphabetizing. This bug would occur for groups where the `rank` overflowed to double digits (eg. `"10"` is sorted before `"4"`). This is a bug with the original code and not necessarily caused by #2506. There just previously wasn't a case where enough groups were added to get to the double digit ranks.

The test case I made for this is similar to the example case shown in #2671 but it isn't the most general test case. It's just a case that would fail without this change.

I also went ahead and added a description comment to my previous PR's test case since it was missing one

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -334,9 +334,12 @@ function mutateRanksToAlphabetize(imported, alphabetizeOptions) {
     return acc;
   }, {});
 
-  const groupRanks = Object.keys(groupedByRanks);
-
   const sorterFn = getSorter(alphabetizeOptions);
+
+  // sort group keys so that they can be iterated on in order
+  const groupRanks = Object.keys(groupedByRanks).sort(function (a, b) {
+    return a - b;
+  });
 
   // sort imports locally within their group
   groupRanks.forEach(function (groupRank) {
@@ -345,7 +348,7 @@ function mutateRanksToAlphabetize(imported, alphabetizeOptions) {
 
   // assign globally unique rank to each import
   let newRank = 0;
-  const alphabetizedRanks = groupRanks.sort().reduce(function (acc, groupRank) {
+  const alphabetizedRanks = groupRanks.reduce(function (acc, groupRank) {
     groupedByRanks[groupRank].forEach(function (importedItem) {
       acc[`${importedItem.value}|${importedItem.node.importKind}`] = parseInt(groupRank, 10) + newRank;
       newRank += 1;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## [Unreleased]
 
+### Fixed
+- [`order]`: Fix group ranks order when alphabetizing ([#2674], thanks [@Pearce-Ropion])
+
 ## [2.27.4] - 2023-01-11
 
 ### Fixed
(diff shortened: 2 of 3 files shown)


ljharb:
A test case that fails without this fix is definitely necessary, otherwise we'll just break it again in the future :-)

ljharb:
Unfortunately this test case passes on main :-/ do you have one that fails on main?

Pearce-Ropion:
> Unfortunately this test case passes on main :-/ do you have one that fails on main?

Huh...

Pearce-Ropion:
> Unfortunately this test case passes on main :-/ do you have one that fails on main?

Ok, this should be working now. I tested it on `main` too to make sure it failed. I forgot that the only way to get the ranks into the double digits is if you have a lot of `groups`. Thats what I get for trying to cleanup a test case to make it presentable...

src/rules/order.js
```diff
@@ -334,18 +334,21 @@ function mutateRanksToAlphabetize(imported, alphabetizeOptions) {
     return acc;
   }, {});
 
-  const groupRanks = Object.keys(groupedByRanks);
-
   const sorterFn = getSorter(alphabetizeOptions);
 
+  // sort group keys so that they can be iterated on in order
+  const groupRanks = Object.keys(groupedByRanks).sort(function (a, b) {
+    return parseFloat(a) - parseFloat(b);
```
ljharb, line 341:
```suggestion
    return a - b;
```
this should be sufficient

ljharb (approved):
Looks great!

PR #2680, opened 2023-01-16
upgrade minimatch to @5

m-shaka:
minimatch has fixed some bugs so I upgraded it to v5.
This includes breaking changes, and I think some of them are correct behavior as the glob pattern but I'am not sure that all of them are
https://www.digitalocean.com/community/tools/glob?comments=true&glob=app%2F%2A%2A%2F%2A%2A&matches=false&tests=app%2Fa

My main interest is [this bug](https://github.com/isaacs/minimatch/issues/188).

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "has": "^1.0.3",
     "is-core-module": "^2.11.0",
     "is-glob": "^4.0.3",
-    "minimatch": "^3.1.2",
+    "minimatch": "^5.1.4",
     "object.values": "^1.1.6",
     "resolve": "^1.22.1",
     "semver": "^6.3.0",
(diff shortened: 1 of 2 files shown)


ljharb:
Since this is indeed a breaking change, we're highly unlikely to ever merge it unless there's a *very* compelling reason.

How does the linked bug impact this plugin?

m-shaka:
Thank you for your quick response!

For example, I want to loosen rules a bit for testing, but I can't.
```js
module.exports = {
  plugins: ['import'],
  rules: {
    'import/no-internal-modules': [
      'error',
      {
        forbid: ['@src/a/*/**']
      }
    ],
  },
  overrides: [
    {
      files: ['src/**/*.test.ts'],
      rules: {
        'import/no-internal-modules': [
          'error',
          {
            forbid: ['@src/a/*/!(__tests__)/**']
          }
        ],    
      }
    }
  ]
}
```

As a user, it's very reasonable we can use glob expressions correctly.
I know backward compatibility may concern you as a maintainer. I'll close this PR if there is another workaround

ljharb:
I'm quite unfamiliar with inline glob expressions like that - do they work in gitignore?

As such, I have no idea what you're trying to forbid - and allow - with that glob. Can you elaborate?

Issue #2712, opened 2023-02-09
1 high severity issue when we install the plugin (CVE in json5)


messenjer:
Hello,

After installing the plugin, we have : 

```sh
1 high severity vulnerability
```

```sh
# npm audit report

json5  <1.0.2
Severity: high
Prototype Pollution in JSON5 via Parse Method - https://github.com/advisories/GHSA-9c47-m6qq-7p4h
fix available via `npm audit fix`
node_modules/tsconfig-paths/node_modules/json5

1 high severity vulnerability
```

The issue is https://github.com/advisories/GHSA-9c47-m6qq-7p4h in json5

It's fixed in `json5` https://github.com/json5/json5/commit/7774c1097993bc3ce9f0ac4b722a32bf7d6871c8 

This module is used by `tsconfig-path` and fixed in the latest version `4.1.2`  https://github.com/dividab/tsconfig-paths/blob/master/CHANGELOG.md

`Eslint-plugin-import` uses `tsconfig-path` version `3.14.1`, when I read the changelog there is no breaking change between the `3.14.1` and `4.1.2`. https://github.com/dividab/tsconfig-paths/blob/master/CHANGELOG.md 

May be we can update to the `tsconfig-path` version `4.1.2`. What do you think?

Thanks for the module.

Mathieu

ljharb:
1. the vulnerability is in json5, not tsconfig-paths
1. we depend on tsconfig-paths v3, which depends on json5 v1
1. json5 v1 and v2 *both* have the fix published, so all you need to do is update your lockfile
1. this is not actually a valid vulnerability for this plugin anyways, so it's irrelevant
1. tsconfig-paths v4 is *absolutely* a breaking change because it drops, among others, node 4 - see #2447 

This is a duplicate of #2625; a duplicate of #2628; a duplicate of #2626; a duplicate of #2627; a duplicate of #2631; a duplicate of #2632; a duplicate of #2634; a duplicate of #2635; a duplicate of #2636; a duplicate of #2637; a duplicate of #2639; a duplicate of #2642; a duplicate of #2643; a duplicate of #2649; a duplicate of #2655; a duplicate of #2660; a duplicate of #2670; a duplicate of #2692; a duplicate of #2711.

All you need to do is install the plugin fresh, which should update your lockfile to use json5 v1.0.2.

messenjer:
Apologies for the noise, thank you for the information.

Issue #2657, opened 2023-01-10
debug < 3.1.0 Inefficient Regular Expression Complexity vulnerability


sspenst:
CVE-2017-20165

Low severity vulnerability

ljharb:
This is not a vulnerability for this plugin, because you can't DoS your own linting run.

sspenst:
So if I use eslint-plugin-import this warning will stay around forever?

ljharb:
Although, actually, it's debug v4, not v3, that dropped support for node 4, so maybe we can do something about it.

PR #2713, opened 2023-02-09
[Deps] update `tsconfig-paths`

messenjer:
Fix https://github.com/import-js/eslint-plugin-import/issues/2712

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,6 @@
     "object.values": "^1.1.6",
     "resolve": "^1.22.1",
     "semver": "^6.3.0",
-    "tsconfig-paths": "^3.14.1"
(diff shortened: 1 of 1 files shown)


Ollymid (approved):
Not a maintainer but I approve this at allows us to update Json5 and resolve a Dependabot security alert

ljharb:
This is a breaking change, and a duplicate of #2447, and a duplicate of #2646, and a duplicate of #2657. Additionally, json5 v1.0.2 fixes the issue, so there's no need to update. Additionally, this is *not a vulnerability* for the way eslint-plugin-import uses json5.

messenjer:
Apologies for the noise, thank you for the information.

PR #2714, opened 2023-02-10
[utils] [new] `parse`: support flat config

DMartens:
Flat ESLint [no longer supports](https://eslint/eslint/issues/16878) `context.parserPath` which is required for [utils/parse](https://github.com/import-js/eslint-plugin-import/blob/d5fc8b670dc8e6903dbb7b0894452f60c03089f5/utils/parse.js#L60).
This PR uses `context.languageOptions.parse` which has the loaded parser as an alternative.
Some notes:
- Backwards compatible as it only checks if `context.parserPath` is unset
- Currently it does check if `parse` or `parseForESLint` in `context.languageOptions.parser` is a function (I do not if "constructed" context objects are officially supported)
- I added a test to the rule to no-unused-modules for testing with flat eslint rule tester

### DIFF
--- a/utils/parse.js
+++ b/utils/parse.js
@@ -23,10 +23,10 @@ function keysFromParser(parserPath, parserInstance, parsedResult) {
   if (parsedResult && parsedResult.visitorKeys) {
     return parsedResult.visitorKeys;
   }
-  if (/.*espree.*/.test(parserPath)) {
+  if (typeof parserPath === 'string' && /.*espree.*/.test(parserPath)) {
     return parserInstance.VisitorKeys;
   }
-  if (/.*babel-eslint.*/.test(parserPath)) {
+  if (typeof parserPath === 'string' && /.*babel-eslint.*/.test(parserPath)) {
     return getBabelEslintVisitorKeys(parserPath);
   }
   return null;
@@ -51,13 +51,13 @@ function transformHashbang(text) {
 }
 
 exports.default = function parse(path, content, context) {
-
   if (context == null) throw new Error('need context to parse properly');
 
-  let parserOptions = context.parserOptions;
-  const parserPath = getParserPath(path, context);
+  // ESLint in "flat" mode only sets context.languageOptions.parserOptions
+  let parserOptions = (context.languageOptions && context.languageOptions.parserOptions) || context.parserOptions;
+  const parserOrPath = getParser(path, context);
 
-  if (!parserPath) throw new Error('parserPath is required!');
+  if (!parserOrPath) throw new Error('parserPath or languageOptions.parser is required!');
 
   // hack: espree blows up with frozen options
   parserOptions = Object.assign({}, parserOptions);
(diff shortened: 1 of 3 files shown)


utils/parse.js
```diff
@@ -23,10 +23,10 @@ function keysFromParser(parserPath, parserInstance, parsedResult) {
   if (parsedResult && parsedResult.visitorKeys) {
     return parsedResult.visitorKeys;
   }
-  if (/.*espree.*/.test(parserPath)) {
+  if (typeof parserPath === 'string' && /.*espree.*/.test(parserPath)) {
```
ljharb, line 26: if it's not a string wouldn't it fail the test?

utils/parse.js
```diff
@@ -104,18 +104,29 @@ exports.default = function parse(path, content, context) {
...
+  return makeParseReturn(ast, keysFromParser(parserOrPath, parser, undefined));
 };
 
+function getParser(path, context) {
+  const parserPath = getParserPath(path, context);
+  if (parserPath) {
+      return parserPath;
+  }
+  const isFlat = context.languageOptions &&
+                 typeof context.languageOptions.parser === 'object'; // May be a string
```
ljharb, line 125: will this test still work with the approach we've taken to have a config shared between flat and legacy? Meaning, the legacy config object still has `languageOptions`, I just mark it non-enumerable.

utils/parse.js
DMartens: Because the parserPath would parserOrPath, I added the check to make it more understandable.
If wanted I can remove the check.

ljharb: it's likely this way is more performant and clear, i just wanted to confirm that the checks don't change the behavior

utils/parse.js
DMartens: As far as I understand it is that has nothing to do with the read config file, as `context.languageOptions.parserOptions`  is also set for legacy configs ([relevant code](https://github.com/eslint/eslint/blob/5c2a1986bfc8a01952ed3db617973769fa060494/lib/linter/linter.js#L687)). But if the linter runs in "flat" mode it no longer sets the `context.parserPath`, `context.parserOptions` ... .

ljharb: ok - it's not that languageOptions is _set_ that IDs it as the flat config, it's that it's set and has an object `parser` property?

Are we sure it can't be null, only a string or an object?

DMartens: No as I understand it in the current eslint both properties are set: `context.parserOptions` and `context.languageOptions.parserOptions` (same with other properties of languageOptions). But in "flat" mode the top-level properties are not set.

To the type of `languageOptions.parser`: 
- null case I created an extra test case.
- string: is from [@types/eslint](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a7732c024ce77bcee36cea88228cf4bb210f5da7/types/eslint/index.d.ts#L884)
- object: is already an object for the current eslint

ljharb: Let's also add runtime checks that throw errors if the types aren't what we expect.

DMartens: Added runtime checks that `context.languageOptions.parser` either has a function parse or parseForESLint

DMartens:
Added support for `context.languageOptions.parserOptions` (which is preferred over `context.parserOptions`).
Maybe that was the cause of the confusion but I forgot to add this because as I said `context.parserOptions` is not set in "flat" mode.

ljharb (approved):

ljharb:
Did some tweaks for style, but also additional ones to handle that FlatRuleTester isn't available in eslint < 8.

azat-io:
I have a lot of errors like this: `error  Parse errors in imported module 'vite': parserPath or languageOptions.parser is required! (undefined:undefined)  import/namespace`.
I use flat config, can I fix it?

ljharb:
@azat-io meaning, with eslint-module-utils 2.7 you don't see those, but with 2.8 you do?

Issue #2043, opened 2021-05-05
Docs: import/extensions says node resolver resolves file extensions automatically


thernstig:
From https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md

>  For example the node resolver can resolve ./foo/bar to the absolute path /User/someone/foo/bar.js because the .js extension is resolved automatically by default.

Is this true? Node.js requires the extension on files for relative paths: https://nodejs.org/docs/latest-v14.x/api/esm.html#esm_mandatory_file_extensions

ljharb:
It’s true for CJS (or ESM transpiled to CJS), which is the only resolution we deal with currently. That link references node’s native ESM, which isn’t supported here yet and shouldn’t really be used yet anyways.

thernstig:
@ljharb thanks, that makes sense. But how come you say "shouldn't really be used yet anyways"? I decided to convert to ESM and use `--experimental-specifier-resolution=node` to get around not having to add `.js` to all files, and it works good for us.

But the biggest concern is that some packages will require ESM to be upgraded soon. See for example https://github.com/chalk/strip-ansi/releases/tag/v7.0.0 (77 million weekly downloads) and his reasoning https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c - this means we cannot upgrade many of the packages he uses, which is needed at times to get security fixes. So not using ESM is starting to become "not an option" if you want to stay on the latest and greatest.

ljharb:
The experimental flag is likely to go away, and unlikely to ever become default.

Separately, the ecosystem and tooling around native ESM is nowhere near production-ready.

Packages that force consumers to use ESM will simply lose users, no matter how prolific the author is.

thernstig:
> The experimental flag is likely to go away, and unlikely to ever become default.

Meaning that we will be forced to use `.js` on files? I think If I recall correctly I saw some posts by you saying that forcing `.js` on imports is a bad choice, and since you are indeed an authority in the Node ecosystem I listen with big ears 😄

ljharb:
Yes, I think it is a terrible choice, and I would have preferred that that resolution behavior be the default. However, I don't make the decisions, and that doesn't look likely. The "exports" field, at least, relieves package consumers from having to use extensions, but, relative native ESM imports will have to use them, unfortunately.

[...]

PR #2748, opened 2023-03-25
[Docs] Add node ESM behaviour in import/extensions

xM8WVqaG:
This behaviour largely didn't change despite the functionality becoming more widely adopted. Even though the document assumes CJS, It's probably worth including at least a footnote about the resolver behaving differently for ESM.

See: https://github.com/import-js/eslint-plugin-import/issues/2043

### DIFF
--- a/docs/rules/extensions.md
+++ b/docs/rules/extensions.md
@@ -2,7 +2,7 @@
 
 <!-- end auto-generated rule header -->
 
-Some file resolve algorithms allow you to omit the file extension within the import source path. For example the `node` resolver can resolve `./foo/bar` to the absolute path `/User/someone/foo/bar.js` because the `.js` extension is resolved automatically by default. Depending on the resolver you can configure more extensions to get resolved automatically.
+Some file resolve algorithms allow you to omit the file extension within the import source path. For example the `node` resolver (which does not yet support ESM/`import`) can resolve `./foo/bar` to the absolute path `/User/someone/foo/bar.js` because the `.js` extension is resolved automatically by default in CJS. Depending on the resolver you can configure more extensions to get resolved automatically.
 
 In order to provide a consistent use of file extensions across your code base, this rule can enforce or disallow the use of certain file extensions.
 
@@ -170,3 +170,5 @@ import foo from '@/foo';
 ## When Not To Use It
 
 If you are not concerned about a consistent usage of file extension.
+
+In the future, when this rule supports native node ESM resolution, and the plugin is configured to use native rather than transpiled ESM (a config option that is not yet available) - setting this to `always` will have no effect.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Changed
 - [Docs] [`no-duplicates`]: fix example schema ([#2684], thanks [@simmo])
 - [Docs] [`group-exports`]: fix syntax highlighting ([#2699], thanks [@devinrhode2])
+- [Docs] [`extensions`]: reference node ESM behavior ([#2748], thanks [@xM8WVqaG])
 
 ## [2.27.5] - 2023-01-16
 
@@ -1067,6 +1068,7 @@ for info on changes for earlier releases.
 [`memo-parser`]: ./memo-parser/README.md
 
 [#2754]: https://github.com/import-js/eslint-plugin-import/pull/2754
+[#2748]: https://github.com/import-js/eslint-plugin-import/pull/2748
 [#2699]: https://github.com/import-js/eslint-plugin-import/pull/2699
 [#2664]: https://github.com/import-js/eslint-plugin-import/pull/2664
 [#2613]: https://github.com/import-js/eslint-plugin-import/pull/2613
@@ -1853,6 +1855,7 @@ for info on changes for earlier releases.
 [@wKich]: https://github.com/wKich
 [@wschurman]: https://github.com/wschurman
 [@wtgtybhertgeghgtwtg]: https://github.com/wtgtybhertgeghgtwtg
+[@xM8WVqaG]: https://github.com/xM8WVqaG
 [@xpl]: https://github.com/xpl
 [@yordis]: https://github.com/yordis
 [@zloirock]: https://github.com/zloirock


docs/rules/extensions.md
```diff
@@ -170,3 +170,10 @@ import foo from '@/foo';
 ## When Not To Use It
 
 If you are not concerned about a consistent usage of file extension.
+
+You should not configure this rule to be "never" if you are using ECMAScript modules directly in `node`.
```
ljharb, line 174: I don't think this is exactly accurate advice - when the node resolver does eventually support "exports" and ESM, "never" would not apply to anything that required the extension, but would correctly forbid it for anything where the extension was optional.

docs/rules/extensions.md
```diff
@@ -2,7 +2,7 @@
 
 <!-- end auto-generated rule header -->
 
-Some file resolve algorithms allow you to omit the file extension within the import source path. For example the `node` resolver can resolve `./foo/bar` to the absolute path `/User/someone/foo/bar.js` because the `.js` extension is resolved automatically by default. Depending on the resolver you can configure more extensions to get resolved automatically.
+Some file resolve algorithms allow you to omit the file extension within the import source path. For example the `node` CommonJS resolver can resolve `./foo/bar` to the absolute path `/User/someone/foo/bar.js` because the `.js` extension is resolved automatically by default. Depending on the resolver you can configure more extensions to get resolved automatically.
```
ljharb, line 5: it's a node resolver, not a CJS resolver - it's just that it doesn't yet support "exports" or ESM.

ljharb (changes requested):

ljharb:
(Either way, I'd suggest not using native ESM in node, for many reasons, but this isn't the venue to debate that ofc)

docs/rules/extensions.md
xM8WVqaG: To confirm we're on the same page here. My understanding is that `node` in this context refers specifically to NodeJS, is that correct?

NodeJS documents two separate module loaders, one for CommonJS and one for ECMAScript. The ESM loader explicitly says:

> _It does no extension searching. A file extension must be provided when the specifier is a relative or absolute file URL._

https://nodejs.org/api/packages.html#modules-loaders

docs/rules/extensions.md
xM8WVqaG: The documentation for NodeJS has marked _Modules: ECMAScript modules_ as _stable_ since v15.3 - although other parts of the ESM spec are still marked experimental. They've explicitly required the `.js` extension for the ECMAScript loader [since v12](https://nodejs.org/docs/latest-v12.x/api/esm.html#esm_mandatory_file_extensions) (back in 2019).

Do you know that they're actually planning future changes to this part of the `import` behaviour specifically? Having something about the current behaviour in NodeJS would be useful!

docs/rules/extensions.md
ljharb: yes, but "node" describes many node versions over a long period of time, it's not inherently "node latest". It's a node resolver because it matches node's resolution *at one point* and because it intends to match the latest version eventually.

Yes, the ESM loader does no extension searching, but any time it's resolving something in a package's "exports" field, the specifier need not (and often should not!) have an extension. It's only on relative files that users are forced to suffer the pain of specifying extensions.

xM8WVqaG: So this proposed change is just updating the `eslint-plugin-import` documentation to be correct now that NodeJS has more that one resolver, right? NodeJS has had more than one resolver since 2017 and referring to them as the same for this scenario is misleading.

>Yes, the ESM loader does no extension searching, but any time it's resolving something in a package's "exports" field, the specifier need not (and often should not!) have an extension. It's only on relative files that users are forced to suffer the pain of specifying extensions.

Given the rule:

```json
{
  "import/extensions": ["error", "always"],
}
```

This code is correct:

```js
import assert from "assert";

import {aFunc, bFunc} from "../local.js";
```

So `import/extensions` already correctly only enforces extensions on relative files and leaves packages untouched.

ljharb: That's not accurate - node's ESM resolver became unflagged in 2020, not before, and it wasn't until April 2021 that the last non-ESM LTS line went EOL (i was in the node modules working group; i'm quite familiar with the timeline)

xM8WVqaG: Then you'll no doubt also know that NodeJS's ESM resolver was available behind a flag and/or using `.mjs` since [v8 back in 2017](https://nodejs.org/docs/latest-v8.x/api/esm.html). That said, I'm not interested in debating timelines.

I found the `import/extensions` documentation in it's current state to be confusing as it has examples using the `import` syntax and specifically discusses "`node`" as not requiring a file extension for local files which as far as I can tell has never been the case. Because I wasted time trying this rule out and having to revert those changes - and had seen another user already raised this before (#2043) - I thought it might be useful for other users to clarify this.

Feel free to suggest alternative wording, I am happy to rephrase my changes. But I can't tell what you're looking for from your feedback so far.

ljharb: Behind a flag doesn't count :-)

Let me give it a fresh review and see what I can suggest.

ljharb, line 5:
```suggestion
Some file resolve algorithms allow you to omit the file extension within the import source path. For example the `node` resolver (which does not yet support ESM/`import`) can resolve `./foo/bar` to the absolute path `/User/someone/foo/bar.js` because the `.js` extension is resolved automatically by default in CJS. Depending on the resolver you can configure more extensions to get resolved automatically.
```

docs/rules/extensions.md
ljharb, line 174:
```suggestion
In the future, when this rule supports native node ESM resolution (and the plugin is configured to use native, instead of transpiled, ESM, a config option that is not yet available), setting this to `always` will have no effect.
```

docs/rules/extensions.md
```diff
@@ -170,3 +170,10 @@ import foo from '@/foo';
 ## When Not To Use It
 
 If you are not concerned about a consistent usage of file extension.
+
+You should not configure this rule to be "never" if you are using ECMAScript modules directly in `node`.
+
+> A file extension must be provided when using the import keyword to resolve relative or absolute specifiers.
+> Directory indexes (e.g. `'./startup/index.js'`) must also be fully specified.
+
+See [esm#mandatory-file-extensions](https://nodejs.org/api/esm.html#mandatory-file-extensions) for more details.
```
ljharb, line 179:
```suggestion
```

ljharb (commented):
what do you think about this?

xM8WVqaG:
Thanks for your suggestions! This conveys the message I was originally looking for and captures some of the important bits from the conversations in this PR.

docs/rules/extensions.md
```diff
@@ -170,3 +170,6 @@ import foo from '@/foo';
 ## When Not To Use It
 
 If you are not concerned about a consistent usage of file extension.
+
+In the future, when this rule supports native node ESM resolution (and the plugin is configured to use native, instead of transpiled, ESM, a config option that is not yet available), setting this to `always` will have no effect.
```
xM8WVqaG, line 174: On second thought, this sentence clause has quite a few commas. What do you think of:

```suggestion
In the future, when this rule supports native node ESM resolution, and the plugin is configured to use native rather than transpiled ESM (a config option that is not yet available) - setting this to `always` will have no effect.
```

ljharb: sure, that's fine

ljharb (approved):

PR #2754, opened 2023-04-06
[Fix] `consistent-type-specifier-style`: fix accidental removal of comma in certain cases

### LINKED ISSUE #2753: [consistent-type-specifier-style] Fixer breaks adjacent code (syntax error due to missing comma)
Input:

```ts
import { Component, type ComponentProps } from "package-1";
import {
  Component1,
  Component2,
  Component3,
  Component4,
  Component5,
} from "package-2";
```

Actual output (syntax error):

```ts
import { Component  } from "package-1";
import type {ComponentProps} from "package-1";
import {
  Component1
  Component2,
  Component3,
  Component4,
  Component5,
} from "package-2";
```

Expected output:

```diff
 import { Component  } from "package-1";
 import type {ComponentProps} from "package-1";
 import {
-  Component1
+  Component1,
   Component2,
   Component3,
   Component4,
```

bradzacher:
Fixes #2753

When passed the second argument - `getTokenAfter` will keep searching until it finds a token that matches that token - so in this bug it meant it kept scanning to the next import declaration.
Using the single-arg form means it will just grab the next token and we can then check to see if that token is a comma or not (it'll either be a `}` or a `,`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Fixed
 - [`no-duplicates`]: remove duplicate identifiers in duplicate imports ([#2577], thanks [@joe-matsec])
 - TypeScript config: fix resolver extension settings (thanks [@gajus])
+- [`consistent-type-specifier-style`]: fix accidental removal of comma in certain cases ([#2754], thanks [@bradzacher])
 
 ### Changed
 - [Docs] [`no-duplicates`]: fix example schema ([#2684], thanks [@simmo])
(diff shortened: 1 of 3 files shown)


aaronadamsCA:
Ah, that would probably explain a couple other weird things I saw that i thought were just me being clumsy with multi-line editing, e.g. a missing comma between type variables.

Thanks for looking into this!

ljharb (approved):
Thank you!

mathiasose:
Is there any plan for when this there will be a new point release on NPM that includes this fix? It's been merged for a while now, is there anything blocking a release?

PR #2756, opened 2023-04-08
[perf] Improve ExportMap.for performance on larger codebases

leipert:
While looking at a larger code base https://gitlab.com/gitlab-org/gitlab, `ExportMap.for` seems to take a significant amount of time. More than half of it is spend on calculating hashes with `hashObject`.

Digging a little deeper, it seems like we are calling it around 500 thousand times. Each iteration calculates the hash of a context object. This context object is created inside of the `childContext` function and consists of four parts:
- `settings` -> an Object itself
- `parserOptions` -> an Object itself
- `parserPath` -> a String
- `path` -> a String.

Interestingly `settings`, `parserOptions` and `parserPath` rarely do change for us, so calculating their hashes on every iteration seems unnecessary. `hashObject` recursively calculates the hashes of each key / value pair.

So instead of doing:

```js
cacheKey = hashObject({settings, parserOptions, parserPath, path})
```
We could also do:
```js
cacheKey = parserPath +
  hashObject(parserOptions) +
  hashObject(settings) +
  path
```

This would be just as stable as before, although resulting in longer cache keys. `parserPath` and `path` would not need to be hashed, because they are strings and a single character change in them would result in a different cache key.

Furthermore we can memoize the hashes of `parserOptions` and `settings`, in case they didn't change compared. The equality is checked with a simple `JSON.stringify`.

We move this `cacheKey` calculation to `childContext`, adding the cache key to the `context` object. This way, we can fall back to the old calculation inside of `ExportMap.for`, as it is a public interface which consumers might be using.

In our code base the results speak for itself:
- 51.59s spent in `ExportMap.for`, 0ms spent in `childContext`.
    - 16.89s is spent in node:crypto/hash `update` (overall)
- 41.02s spent in `ExportMap.for`, 1.91s spent in `childContext`.
    - Almost no time spent in `hashObject`, actually all calls in
       our flame graph come from other code paths
    - 7.86s is spent in node:crypto/hash `update` (overall)

So on this machine, and project, we are cutting the execution time of `ExportMap.for` in half. On machines, which are hashing slower, the effect might be more pronounced. Similarly machines with less memory, as the `hashObject` function creates a lot of tiny strings.

One side-effect here could be, that the memoization is in-efficient if the `settings` or `parserOptions` change often. (I cannot think of such a scenario, but I am not that versed in the use cases of this plugin.) But even then, the overhead should mainly be the equality checks with `JSON.stringify`.

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -305,7 +305,7 @@ ExportMap.get = function (source, context) {
 ExportMap.for = function (context) {
   const { path } = context;
 
-  const cacheKey = hashObject(context).digest('hex');
+  const cacheKey = context.cacheKey || hashObject(context).digest('hex');
   let exportMap = exportCache.get(cacheKey);
 
   // return cached ignore
@@ -559,7 +559,7 @@ ExportMap.parse = function (path, content, context) {
       if (tsConfigInfo.tsConfigPath !== undefined) {
         // Projects not using TypeScript won't have `typescript` installed.
         if (!ts) { ts = require('typescript'); } // eslint-disable-line import/no-extraneous-dependencies
-  
+
         const configFile = ts.readConfigFile(tsConfigInfo.tsConfigPath, ts.sys.readFile);
         return ts.parseJsonConfigFileContent(
           configFile.config,
@@ -781,12 +781,29 @@ export function recursivePatternCapture(pattern, callback) {
   }
 }
 
+let parserOptionsHash = '';
+let prevParserOptions = '';
+let settingsHash = '';
+let prevSettings = '';
 /**
  * don't hold full context object in memory, just grab what we need.
+ * also calculate a cacheKey, where parts of the cacheKey hash are memoized
  */
 function childContext(path, context) {
   const { settings, parserOptions, parserPath } = context;
+
+  if (JSON.stringify(settings) !== prevSettings) {
+    settingsHash = hashObject({ settings }).digest('hex');
+    prevSettings = JSON.stringify(settings);
+  }
+
+  if (JSON.stringify(parserOptions) !== prevParserOptions) {
+    parserOptionsHash = hashObject({ parserOptions }).digest('hex');
+    prevParserOptions = JSON.stringify(parserOptions);
+  }
+
   return {
+    cacheKey: String(parserPath) + parserOptionsHash + settingsHash + String(path),
     settings,
     parserOptions,
     parserPath,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Fixed
 - [`no-duplicates`]: remove duplicate identifiers in duplicate imports ([#2577], thanks [@joe-matsec])
 - [`consistent-type-specifier-style`]: fix accidental removal of comma in certain cases ([#2754], thanks [@bradzacher])
+- [Perf] `ExportMap`: Improve `ExportMap.for` performance on larger codebases ([#2756], thanks [@leipert])
 
 ### Changed
 - [Docs] [`no-duplicates`]: fix example schema ([#2684], thanks [@simmo])
@@ -1066,6 +1067,7 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#2756]: https://github.com/import-js/eslint-plugin-import/pull/2756
 [#2754]: https://github.com/import-js/eslint-plugin-import/pull/2754
 [#2748]: https://github.com/import-js/eslint-plugin-import/pull/2748
 [#2699]: https://github.com/import-js/eslint-plugin-import/pull/2699
@@ -1737,6 +1739,7 @@ for info on changes for earlier releases.
 [@kylemh]: https://github.com/kylemh
 [@laysent]: https://github.com/laysent
 [@le0nik]: https://github.com/le0nik
+[@leipert]: https://github.com/leipert
 [@lemonmade]: https://github.com/lemonmade
 [@lencioni]: https://github.com/lencioni
 [@leonardodino]: https://github.com/leonardodino


leipert:
This change is quite similar to #2755, but might be a little controversial, because it introduces a new dependency.

I had a look at `fast-deep-equal`: http://unpkg.com/browse/fast-deep-equal@3.1.3/index.js and I think it's simplicity aligns well with our use case here.

src/ExportMap.js
```diff
@@ -781,12 +782,29 @@ export function recursivePatternCapture(pattern, callback) {
...
+    prevSettings = settings;
+  }
+
+  if (!deepEqual(parserOptions, prevParserOptions)){
+    parserOptionsHash = hashObject({ parserOptions }).digest('hex');
+    prevParserOptions = parserOptions;
+  }
+
   return {
+    cacheKey: String(parserPath) + parserOptionsHash + settingsHash + String(path),
```
leipert, line 807: Wasn't too sure, whether we can rely on `parserPath` and `path` being strings.

src/ExportMap.js
```diff
@@ -305,7 +306,7 @@ ExportMap.get = function (source, context) {
 ExportMap.for = function (context) {
   const { path } = context;
 
-  const cacheKey = hashObject(context).digest('hex');
+  const cacheKey = context.cacheKey || hashObject(context).digest('hex');
```
leipert, line 309: Trying not to change the public interface of `ExportMap.for`. In case we did not pre-calculate a cacheKey, it will behave like before.

src/ExportMap.js
```diff
@@ -781,12 +782,29 @@ export function recursivePatternCapture(pattern, callback) {
   }
 }
 
+let parserOptionsHash = '';
+let prevParserOptions = null;
+let settingsHash = '';
+let prevSettings = null;
 /**
  * don't hold full context object in memory, just grab what we need.
+ * also calculate a cashKey, where parts of the cacheKey hash are memoized
```
leipert, line 791: Haha. `cashKey`. I love it.

```suggestion
 * also calculate a cacheKey, where parts of the cacheKey hash are memoized
```

leipert:
Rebased and ready for review 👍 Thank you!

package.json
```diff
@@ -107,6 +107,7 @@
     "doctrine": "^2.1.0",
     "eslint-import-resolver-node": "^0.3.7",
     "eslint-module-utils": "^2.7.4",
+    "fast-deep-equal": "^3.1.3",
```
ljharb, line 110: i'm very hesitant to add this dependency. `deep-equal` is the minimal amount of code required for correctness - i'd prefer to use that package, and if *it* could use some perf optimizations then that'd be great.

src/ExportMap.js
```diff
@@ -781,12 +782,29 @@ export function recursivePatternCapture(pattern, callback) {
...
  */
 function childContext(path, context) {
   const { settings, parserOptions, parserPath } = context;
+
+  if (!deepEqual(settings, prevSettings)){
```
ljharb, line 796:
```suggestion
  if (!deepEqual(settings, prevSettings)) {
```

src/ExportMap.js
```diff
@@ -781,12 +782,29 @@ export function recursivePatternCapture(pattern, callback) {
...
+    settingsHash = hashObject({ settings }).digest('hex');
+    prevSettings = settings;
+  }
+
+  if (!deepEqual(parserOptions, prevParserOptions)){
```
ljharb, line 801:
```suggestion
  if (!deepEqual(parserOptions, prevParserOptions)) {
```

src/ExportMap.js
ljharb: I think we can, but it's fine to stringify them just in case.

ljharb (changes requested):
The utils changes need to be in a separate commit from the main plugin changes, please.

package.json
leipert: @ljharb 

> `deep-equal` is the minimal amount of code required for correctness

I think this is not true, as `deep-equal` [does handle way more cases than needed](https://unpkg.com/deep-equal@2.2.0/index.js), e.g. comparison of ArrayBuffers and whatnot. Furthermore it has  has 43 dependencies: https://npm.anvaka.com/#/view/2d/deep-equal and on my machine is even slower than `assert.deepStrictEqual`:

```
fast-deep-equal x 216,001 ops/sec ±0.37% (92 runs sampled)
fast-equals x 227,635 ops/sec ±0.44% (93 runs sampled)
nano-equal x 144,366 ops/sec ±0.46% (91 runs sampled)
shallow-equal-fuzzy x 114,739 ops/sec ±0.38% (91 runs sampled)
underscore.isEqual x 74,419 ops/sec ±0.46% (92 runs sampled)
lodash.isEqual x 34,833 ops/sec ±0.39% (91 runs sampled)
deep-equal x 96.92 ops/sec ±1.82% (55 runs sampled)
deep-eql x 21,870 ops/sec ±1.87% (89 runs sampled)
ramda.equals x 11,929 ops/sec ±3.34% (88 runs sampled)
util.isDeepStrictEqual x 46,825 ops/sec ±1.26% (88 runs sampled)
assert.deepStrictEqual x 515 ops/sec ±0.75% (88 runs sampled)
```

If we have a look at what `hashObject` / `hashify` is actually capable of hashing / differentiating (Array, 0bject, stuff that is stringifyable with `JSON.stringify`), the [simplicity of `fast-deep-equal`](http://unpkg.com/browse/fast-deep-equal@3.1.3/index.js) actually aligns quite well with the `hashObject` implementation.

- - -

That being said, what do you think of moving the memoization and comparison to a new util, e.g. `hashObjectMemoized` which would be used like this:

```
    cacheKey: String(parserPath) + hashObjectMemoized(parserOptions, 'parserOptions') + hashObjectMemoized(settings, 'settings') + String(path),
```

That way if we ever change the implementation of `hashify`/`hashObject`, we could adjust the `isEqual` function for `hashObjectMemoized` to align with the implementation of the hashing algorithm. For example if support for hashing Sets or Maps is added in the future.

leipert: Another note, I would probably go with:

a) Custom equality implementation that matches `hashObject` needs
b) [`fast-equals`](https://github.com/planttheidea/fast-equals) over `deep-equal` or `fast-deep-equal`, because: 0 dependencies, maintained, future proof and the fastest of the bunch, as seen above and via their [own benchmarks](https://github.com/planttheidea/fast-equals#benchmarks)

src/ExportMap.js
```diff
@@ -781,12 +782,29 @@ export function recursivePatternCapture(pattern, callback) {
...
+let settingsHash = '';
+let prevSettings = null;
 /**
  * don't hold full context object in memory, just grab what we need.
+ * also calculate a cacheKey, where parts of the cacheKey hash are memoized
  */
 function childContext(path, context) {
   const { settings, parserOptions, parserPath } = context;
+
+  if (!deepEqual(settings, prevSettings)) {
```
ljharb, line 796: why not `JSON.stringify(settings) !== JSON.stringify(prevSettings)`?

leipert: @ljharb Pushed a version without `fast-deep-equal` and utilizing `JSON.stringify`.

It unnecessarily creates ~1M objects and therefore seems to take 1.9s instead of 0.6s. But we still improve local execution by ~10s.

leipert: (updated the numbers in the issue description)

ljharb (approved):
Seems good!

src/ExportMap.js
ljharb: I'm surprised JSON.stringify would create any objects at all, but improving execution sounds good.

PR #2763, opened 2023-04-18
Fix invalid YAML in import/parsers example

silverwind:
YAML keys can not start with `@` as that will result in such a parser error:

```
bad indentation of a mapping entry (4:5)

 1 | settings:
 2 |   import/extensions: [.js, .jsx, .ts ...
 3 |   import/parsers:
 4 |     @typescript-eslint/parser: [ .ts ...
---------^
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ directly using webpack, for example:
 # .eslintrc.yml
 settings:
   import/parsers:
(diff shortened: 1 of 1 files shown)


ljharb (approved):
ah, thanks. we should probably change this to json anyways

PR #2778, opened 2023-05-08
[Fix] `extensions`: handle `.` and `..` properly

### LINKED ISSUE #2777: [import/extensions] Doesn't error on . or ..
I have this rule configured like so:
```javascript
'import/extensions': ['error', 'ignorePackages'],
```

With this configuration, the following is flagged as an error:
```javascript
import { Something } from './LocalFolderWithIndexfile'
```

However, the following is not flagged as an error, but I believe it should be:
```javascript
import { Something } from '.'
```

The only difference between the two is where in the file tree the importing file sits. As a workaround, one can supplement this rule with the general `no-restricted-imports` rule like this:
```javascript
'no-restricted-imports': ['error',
  {
    paths: [
      {
        name: '..',
        message: 'Import from ../index.js instead.',
      },
      {
        name: '.',
        message: 'Import from ./index.js instead.',
      },
    ],
  },
],
```

Apologies if this has already been filed. I found it really hard to search for

benasher44:
The issue was `isExternalRootModule` was returning `true` for these, since they had no `/` in them. This fixes that.

Closes #2777

### DIFF
--- a/tests/src/rules/extensions.js
+++ b/tests/src/rules/extensions.js
@@ -268,6 +268,26 @@ ruleTester.run('extensions', rule, {
 
     test({
       code: [
+        'import barjs from "."',
+        'import barjs2 from ".."',
+      ].join('\n'),
+      options: [ 'always' ],
+      errors: [
+        {
+          message: 'Missing file extension "js" for "."',
+          line: 1,
+          column: 19,
+        },
+        {
+          message: 'Missing file extension "js" for ".."',
+          line: 2,
+          column: 20,
+        },
+      ],
+    }),
+
+    test({
+      code: [
         'import barjs from "./bar.js"',
         'import barjson from "./bar.json"',
         'import barnone from "./bar"',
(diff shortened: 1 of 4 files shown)


ljharb (commented):
Can we also add some test cases with the "always" option where `.` and `..` refer to dirs with package.json, and show that these don't error?

ljharb:
Ideally I'd like this PR to either add tests documenting the current behavior (with a TODO comment if it's broken), or, tests matching the desired behavior and fixes for it.

While `ignorePackages` was intended to cover things in `node_modules`, i think it actually makes sense to cover a relative dir with a `package.json` as well. I'm wondering how large a change that would be.

benasher44:
I added tests that show the current behavior (says extensions are required when they shouldn't be)

tests/src/rules/extensions.js
```diff
@@ -594,6 +614,35 @@ ruleTester.run('extensions', rule, {
...
+        },
+      ],
+    }),
+    // TODO: properly ignore packages resolved via relative imports
+    test({
+      code: [
+        'import * as test from ".."',
+      ].join('\n'),
+      filename: testFilePath('./internal-modules/plugins/plugin.js'),
+      options: [ 'ignorePackages' ],
```
benasher44, line 638: I'm not sure about whether these should be `always`. I think with `ignorePackages` it should definitely not error, so it makes for a less ambiguous test case.

In the always + relative case it's a bit more ambiguous to me, since if we assume the relative path is in your project, you might have this set to `always` because you always want someone to choose a file (within the project).

benasher44:
Any tips on what to do about the semver module not found errors from appveyor?

tests/src/rules/extensions.js
ljharb: I'm not sure what to do here, tbh.

ljharb: I'm going to merge this, and please file followup issues and/or PRs :-)

benasher44:
Thank you!

PR #2779, opened 2023-05-08
[New] no-parent-barrel-import: Forbid import from parent barrel file

jonioni:
Resolves: https://github.com/import-js/eslint-plugin-import/pull/2318#issuecomment-1027807460

Forbid import from parent barrel file which often leads to runtime error `cannot read ... of undefined`. 

It resolves the missing cycle check from [`no-self-import`], while being computationally cheap (over [`no-cycle`])

## Rationale

Though it seems that the issue must have been covered by existing rules, there actually isn't one that prevents circular import / runtime error via parent barrel file (`index.js`) efficiently.

- [`no-self-import`]: doesn't check indirect import
- [`no-cycle`]: very costly on large codebase
- [`no-restricted-paths`]: requires fixed patterns and not available as a general rule for every folder

## Related Issue

There is a related rule of a larger scope: #2408 but further adjustment might be needed. Also it seems to be overly restrictive on importing from non-parent folder (forbids `import example2 from '../dir2';` from `dir1/file.js`) which might need more justification.

[`no-self-import`]: ./docs/rules/no-self-import.md
[`no-cycle`]: ./docs/rules/no-cycle.md
[`no-restricted-paths`]: ./docs/rules/no-restricted-paths.md


jonioni:
Hi @ljharb, wondering if you could point out why the tests failed with `context.report is not a function`?

PR #2801, opened 2023-06-23
Added a rule to enforce import path format (relative/absolute)

cnshenj:


docs/rules/path.md
```diff
@@ -0,0 +1,54 @@
...
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+Enforce paths of `require()` / `import` statements to be "absolute" or "relative".
+
+## Rule Details
+This rule enforces module paths in `require()` / `import` statements to be "relative"
+or "absolute" (relative to the project root).
```
ljharb, line 11: The word "absolute" means "relative to `/`", so I very much think this is the wrong term - absolute paths are a horrifically bad practice and should never be used.

How can a path be relative to the project root when there does not exist a generic mechanism for descending from the project root?

docs/rules/path.md
```diff
@@ -0,0 +1,54 @@
...
+  "underSameDirectory": "relative",
+  "other": "relative"
+}
+```
+
+### underSameDirectory
+Example of **incorrect** code for this rule with the `{ "underSameDirectory": "relative" }` options:
+```ts
+// Current file "src/home/foo.ts"
+import bar from 'src/home/bar'; // Imports under the same directory of the current file must be relative to the current file.
```
ljharb, line 34: this code is meaningless because `src` can be aliased to anywhere, and in no way is necessarily the "project root"

ljharb (changes requested):

docs/rules/path.md
cnshenj: Changed "absolute" to "relativeToRoot" (root means project root).
Changed "relative" to "relativeToCurrent".

docs/rules/path.md
cnshenj: I think I can use "resolved" path (by resolver) to determine the location.

ljharb: right, but resolvers aren't limited to only things that are root-relative.

cnshenj: Let's use an example:
- Current file: "/project/home/foo.ts"
- Code: `import bar from 'home/bar'`
- tsconfig: "/project/tsconfig.json" contains `{ "compilerOptions": { "baseUrl": "./" } }`

Analysis:
- No alias. The resolver resolves the import to be "/project/home/bar.ts" (using tsconfig).
  - The import is under the same directory as the current file, so it will use the policy defined by `underSameDirectory`.
- "home" is aliased to "/outside". The resolver resolves the import to be "/outside/bar.ts".
  - The import is **not** under the same directory as the current file, so it will use the policy defined by `other`.
- What if "home" is aliased to "/project/home"? The resolver will resolve the import to be "/project/home/bar.ts".
  - Again, the import is under the same directory as the current file, so it should use the policy defined by `underSameDirectory`.
  - And if rule options has `{ "underSameDirectory": "relativeToCurrent" }`, then this line is invalid. It should be written as `import bar from './bar'`. Even though an alias is defined, the rule enforces the alias can only be used outside of "/project/home". So, files like "/project/away/index.ts" can use the alias.

You can see that the resolver just has to find the full path of the resolved file. The resolved file can be outside of the project root. Also, whether there is an alias doesn't matter, the resolved path will be used by the rule.

ljharb: ok, so you’re saying that anything that resolves to a file inside the project, would be forced to use a path that descends from the project root?

The problem is that resolvers are only one way - given a specifier in code, they provide a file path. There’s no way to know what a root-relative path might look like, because there might be a thousand alternative aliases that are project-relative.

in other words, there simply doesn’t exist the concept you want, which is a universal way for paths to be project-relative, which is also why paths are almost universally *not* that way.

cnshenj: Aliases are defined at the project level. Let's use a symbol `[project]`, not actual path, to represent "project root". If you define an alias to be `alias1` -> `/outside/of/project/root`, then `alias1` means `[project]/alias1`. Thus, `import from "alias1"` is considered to be "relative to project root".

That being said. I'll close this PR because it is not a perfect solution for the problem that I'm trying to solve. Thanks for reviewing.

ljharb: The challenge is that a) there isn't a generic/universal "way" to reference the project root, and b) that "hell" is actually a good thing imo - it's a strong signal that you need to extract more things into separate packages.

cnshenj: > The challenge is that a) there isn't a generic/universal "way" to reference the project root, and b) that "hell" is actually a good thing imo - it's a strong signal that you need to extract more things into separate packages.

a) TypeScript projects have a universal way: `baseUrl`. JavaScript projects can use the provided option `root` to work with *any* way the project chooses to use - it doesn't need to be universal.
b) It is a signal, but not necessarily "strong". Project contributors may or may not extract a separate package after evaluating all pros/cons.

PR #2810, opened 2023-06-28
[Deps] update semver

### LINKED ISSUE #2809: Update semver version  ">=7.5.2"
Versions of the package semver before 7.5.2 are vulnerable to Regular Expression Denial of Service (ReDoS) via the function new Range, when untrusted user data is provided as a range.

Vhndaree:
Fix #2809


ljharb:
See my comment in the issue; this doesn’t fix anything both because there’s nothing to fix, and because it breaks things in older nodes.

Issue #2808, opened 2023-06-27
Upgrade to semver v7


beanow-at-crabnebula:
Semver disclosed this vulnerability: https://github.com/advisories/GHSA-c2qf-rxjj-qqgw

And there are *no plans to backport the fix* to v5/v6. https://github.com/npm/node-semver/issues/576
This plugin is one of my transient dependencies depending on v6.

```
$ pnpm why semver@^6

devDependencies:
eslint-config-standard-with-typescript 34.0.1
├─┬ eslint-config-standard 17.0.0
│ └─┬ eslint-plugin-import 2.27.5 peer
│   └── semver 6.3.0
└─┬ eslint-plugin-import 2.27.5 peer
  └── semver 6.3.0
```

While it seems some community members offer to backport the fix, evidently the maintainers are reluctant to support older versions, given v7 was released in 2019. For future sanity of this plugin I'd encourage using the latest major version :laughing:

ljharb:
Like most CVEs, it's a false positive. We're not using `new Range` nor are we doing anything that wouldn't be a self-attack (ie, not an attack).

We can't ever bump the semver version because v7 drops support for engines we support, so unless the fix is backported to v6, it'll just have to remain a false positive.

Duplicate of #2803. Duplicate of #2804. Duplicate of #2805. Duplicate of #2806. Duplicate of #2807.

Issue #2807, opened 2023-06-27
semver vulnerable to Regular Expression Denial of Service


stantalid:
There is an [advisory](https://github.com/advisories/GHSA-c2qf-rxjj-qqgw) for the `semver` package that is a dependency.

Could you please update the dependecy to solve the issue?

ljharb:
Like most CVEs, it's a false positive. We're not using `new Range` nor are we doing anything that wouldn't be a self-attack (ie, not an attack).

We can't ever bump the semver version because v7 drops support for engines we support, so unless the fix is backported to v6, it'll just have to remain a false positive.

Duplicate of #2803. Duplicate of #2804. Duplicate of #2805. Duplicate of #2806.

Issue #2806, opened 2023-06-26
Upgrade `semver` to fix vulnerability


brianlenz:
There is a vulnerability with `semver` < `7.5.2`:

https://github.com/advisories/GHSA-c2qf-rxjj-qqgw

The solution is to upgrade `semver` to `7.5.3` which includes a patch to fix the vulnerability.

ljharb:
Like most CVEs, it's a false positive. We're not using `new Range` nor are we doing anything that wouldn't be a self-attack (ie, not an attack).

We can't ever bump the semver version because v7 drops support for engines we support, so unless the fix is backported to v6, it'll just have to remain a false positive.

Duplicate of #2803. Duplicate of #2804. Duplicate of #2805.

PR #2812, opened 2023-07-02
Update package.json

CaptThomas:


CaptThomas:
Hey reviewers, thanks for a great tool. This PR is designed to fix CVE-2022-25883. I tested this locally using Windows 11 x64 using node v20.3.1 and npm 9.7.2 and it worked as expected.

ljharb:
Like most CVEs, it's a false positive. We're not using `new Range` nor are we doing anything that wouldn't be a self-attack (ie, not an attack).

We can't ever bump the semver version because v7 drops support for engines we support, so unless the fix is backported to v6, it'll just have to remain a false positive.

Duplicate of #2803. Duplicate of #2804. Duplicate of #2805. Duplicate of #2806. Duplicate of #2807. Duplicate of #2808. Duplicate of #2809.

Issue #2776, opened 2023-05-03
Resolve .jsx? to .tsx? and force the .jsx? extensions


slavafomin:
Hello!

Thank you for this great library!

However, I can't really figure out how to configure it in ESM TypeScript React project.

I would like the ESLint to report (and autofix) missing '.js' and '.jsx' extensions, however, it forces me to use the `.ts(x)` extensions instead.

Thanks!

ljharb:
Are the files actually present on disk with those extensions?

If not, this is a duplicate of #2729 and should be closed.

karlhorky:
@ljharb  I believe that the inability of `import/extensions` to map `.js` imports to `.ts` files (and `.jsx` to `.tsx`, etc)  is the same issue as @phryneas clarified over here:

- https://github.com/import-js/eslint-import-resolver-typescript/issues/233#issuecomment-1627833938

Seems like there's also a PR open for that here:

- https://github.com/import-js/eslint-plugin-import/pull/2813

Issue #2729, opened 2023-03-01
Resolving .js file extensions to .ts file extensions in imports


kylerush:
I love this project. I'm very excited to get it up and running.

The problem I'm running into is that my example.ts file has an import in it like this:

```
import { environmentConfig } from "../src/environment-config.js";
```

The file that actually points to is "../src/environment-config.ts".

`ts-node-esm example.ts` seems to handle the module resolution for me magically; it just works. Additionally, when I build from project with typescript, it also just works because the .ts files get changed to .js, so the ".js" file extension in the import paths points to the right file when I execute the file with Node: `node example.js`.

It seems ideal to me to keep this configuration. However, I cannot get the eslint-plugin-import to resolve ".js" to ".ts" in my imports. I've spent a few hours on it, looking for resolvers on NPM that might do this and trying to write my own resolver. I didn't find anything on NPM and in writing my own resolver, I realized that module resolution is a bit more complex than I anticipated. I also looked at `eslint-import-resolver-typescript`, but that seems to be created to support .ts file extensions in your imports, which I do not have. I've also tried searching the issues on this repo and was not able to find anything related.

I am wondering if there is something simple that I am missing to achieve this. I am not particularly knowledgable about module resolution, or the best way to make all of this work. If possible, any thoughts/help would be greatly appreciated. Thank you very much.

Here is my .eslintrc.cjs:

```js
module.exports = {
  root: true,
  parser: "@typescript-eslint/parser",
  parserOptions: {
    tsconfigRootDir: __dirname,
    project: ["./tsconfig.json"],
  },
  plugins: ["@typescript-eslint", "prettier", "import"],
  extends: [
    "eslint:recommended",
    "plugin:@typescript-eslint/recommended",
    "plugin:@typescript-eslint/recommended-requiring-type-checking",
    "plugin:@typescript-eslint/strict",
    "prettier",
    "plugin:import/recommended",
  ],
  env: {
    node: "true",
  },
};

```

terenceodonoghue:
I'm using `eslint-import-resolver-typescript` in an ESM project - the below config in my `.eslintrc.cjs` matches the `.js` in the import statement to the `.ts` on the underlying filesystem and resolves the error.

```javascript
  settings: {
    'import/resolver': {
      typescript: {
        alwaysTryTypes: true,
      },
    },
  },
```

connorjs:
FWIW, I had the following which worked for `.js` extensions.

```
"import/resolver:": {
[...]

PR #2813, opened 2023-07-03
`extensions`: use the `extensionAlias` option of `import-resolver-typescript`

phryneas:
TypeScript with `moduleResolution: "node16"` requires imports like `import { foo } from './tsFile.js'` while on the disk, there is actually a `./tsFile.ts`.  
The idea here is that a module import needs a file extension and that TypeScript will not change the import statement during transpilation.

Currently, this does not work with the `import/extensions` rule if the `eslint-import-resolver-typescript` is set up - the resolver will return the original file name with the `.ts` extension, and the eslint rule will error since `"js" !== "ts"`.

This PR modifies the `import/extensions` rule to take the `extensionAlias` setting into account if the `typescript` resolver is configured.

Related issues: #2729 and #2776

### DIFF
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -165,7 +165,8 @@ module.exports = {
         context,
       ) || isScoped(importPath);
 
-      if (!extension || !importPath.endsWith(`.${extension}`)) {
+      const validExtensions = getValidExtensionFor(context, importPath, extension);
+      if (!extension || !validExtensions.some((extension) => importPath.endsWith(`.${extension}`))) {
         // ignore type-only imports and exports
         if (node.importKind === 'type' || node.exportKind === 'type') { return; }
         const extensionRequired = isUseOfExtensionRequired(extension, isPackage);
@@ -190,3 +191,39 @@ module.exports = {
     return moduleVisitor(checkFileExtension, { commonjs: true });
   },
 };
+
+/**
+ * Taken from `eslint-import-resolver-typescript`.
+ * This could be imported from current versions of that plugin,
+ * but this project still depends on an older version.
+ * Also, importing it would add a dependency, or at least an
+ * optional peer dependency - copying the code seems like the
+ * more sane option.
+ * [LICENSE](https://github.com/import-js/eslint-import-resolver-typescript/blob/71b23a206514842fef70a99220e5ffb1d6da2a0e/LICENSE)
+ */
+const defaultExtensionAlias = {
+  '.js': [
+    '.ts',
+    // `.tsx` can also be compiled as `.js`
+    '.tsx',
+    '.d.ts',
+    '.js',
+  ],
+  '.jsx': ['.tsx', '.d.ts', '.jsx'],
+  '.cjs': ['.cts', '.d.cts', '.cjs'],
+  '.mjs': ['.mts', '.d.mts', '.mjs'],
+};
+
+function getValidExtensionFor(context, importPath, resolvedExtension) {
+  let extensionAlias = {};
+  if (context.settings['import/resolver']  && context.settings['import/resolver'].typescript) {
+    extensionAlias = context.settings['import/resolver'].typescript.extensionAlias || defaultExtensionAlias;
+  }
+
+  const importedExtension = path.extname(importPath);
+  if (importedExtension in extensionAlias) {
+    return extensionAlias[importedExtension].map((ext) => ext.substring(1));
+  }
+  return [ resolvedExtension ];
+}
+
--- a/tests/files/typescript-tsx.tsx
+++ b/tests/files/typescript-tsx.tsx
@@ -1,0 +1,1 @@
+export const foo = <div>Foo</div>;
--- a/tests/files/typescript-with-index/index.ts
+++ b/tests/files/typescript-with-index/index.ts
@@ -1,0 +1,1 @@
+export const foo = "foo";
--- a/tests/src/rules/extensions.js
+++ b/tests/src/rules/extensions.js
@@ -3,6 +3,15 @@ import rule from 'rules/extensions';
 import { getTSParsers, test, testFilePath, parsers } from '../utils';
 
 const ruleTester = new RuleTester();
+const ruleTesterWithTypeScriptImports = new RuleTester({
+  settings: {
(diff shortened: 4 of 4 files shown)


src/rules/extensions.js
```diff
@@ -190,3 +191,39 @@ module.exports = {
...
+
+/**
+ * Taken from `eslint-import-resolver-typescript`.
+ * This could be imported from current versions of that plugin,
+ * but this project still depends on an older version.
+ * Also, importing it would add a dependency, or at least an
+ * optional peer dependency - copying the code seems like the
+ * more sane option.
+ * [LICENSE](https://github.com/import-js/eslint-import-resolver-typescript/blob/71b23a206514842fef70a99220e5ffb1d6da2a0e/LICENSE)
+ */
```
ljharb, line 203: is it that we depend on an older *major* of that plugin? or that we depend on an older minor or patch?

Probably a better option would be extracting the code into its own package, that both we and eslint-import-resolver-typescript use. @JounQin, thoughts?

src/rules/extensions.js
```diff
@@ -190,3 +191,39 @@ module.exports = {
...
+};
+
+function getValidExtensionFor(context, importPath, resolvedExtension) {
+  let extensionAlias = {};
+  if (context.settings['import/resolver']  && context.settings['import/resolver'].typescript) {
+    extensionAlias = context.settings['import/resolver'].typescript.extensionAlias || defaultExtensionAlias;
+  }
+
+  const importedExtension = path.extname(importPath);
+  if (importedExtension in extensionAlias) {
```
ljharb, line 224: `extensionAlias` isn't a null object, so this would pass for `foo.toString`, i think? let's use `has` here instead.

src/rules/extensions.js
```diff
@@ -190,3 +191,39 @@ module.exports = {
...
+    return extensionAlias[importedExtension].map((ext) => ext.substring(1));
```
ljharb, line 225:
```suggestion
    return extensionAlias[importedExtension].map((ext) => ext.slice(1));
```

src/rules/extensions.js
```diff
@@ -190,3 +191,39 @@ module.exports = {
...
+  }
+  return [ resolvedExtension ];
```
ljharb, line 227:
```suggestion
  return [resolvedExtension];
```

src/rules/extensions.js
phryneas: `eslint-plugin-import` doesn't have any `dependency` on `eslint-import-resolver-typescript` yet, but a `devDependency` on `1.0.0 || 1.1.0` - the current version is `3.5.5`.  
I didn't want to add a new dependency here just for that one config object, but I'm happy with whatever solution we end up with here :)

src/rules/extensions.js
phryneas: `has` should only work on a `Map`. Do you mean
```suggestion
  if (Object.hasOwn(extensionAlias, importedExtension)) {
```
?

src/rules/extensions.js
ljharb: ah, true enough. adding a dependency is fine, but perhaps the resolver's not ideal to add. hopefully an extracted package is viable.

src/rules/extensions.js
phryneas: I've taken this straight from line 159:
```js
      const extension = path.extname(resolvedPath || importPath).substring(1);
```
Should I change it in both places so we are consistent?

src/rules/extensions.js
phryneas: If that doesn't work, we could go for an `peerDependency` with `peerDependenciesMeta` `optional: true`. 
If the package is there, use it, if it isn't we should avoid the relevant codeblock anyways.

ljharb: I don't want to rely on optional peers; any npm client older than that feature would treat it as required.

src/rules/extensions.js
ljharb: i mean `require('has')`, since we support environments without `Object.hasOwn`.

src/rules/extensions.js
ljharb: ah, yep, that'd be great, thanks :-)

src/rules/extensions.js
phryneas: Good point. Then let's see what @JounQin has to say :)

phryneas:
Little heads up: It's already late here, and tomorrow is public holiday, so I'll probably add the requested changes on Wednesday.

src/rules/extensions.js
JounQin: Should it be a setting of import plugin to for using without TypeScript resolver? For example, could webpack resolver resolves same path?

phryneas: I think that's already the `settings['import/resolver'].typescript` setting? 
If that's set, the import plugin will use the TS resolver, otherwise not.

JounQin: https://webpack.js.org/configuration/resolve/#resolveextensionalias

I mean the webpack resolver could also resolve different extensions if it's working correctly? Then only relying on the typescript resolver's setting seems incorrect to me.

phryneas: Maybe? I don't know what that one does, or if anyone ever had this problem with it.  
The problem at hand is that this is something TypeScript expects with the `node16` target, for at least the last 10 releases, and it is not supported by the `imports/extensions` rule. There doesn't seem any way of asking the resolver "is this a valid replacement file extension for this other extension", since the resolver is pretty much hidden from the rule, so I don't know a way of implementing this better.

ljharb:
I've rebased this; if all tests pass, then the only open question seems to be whether we can extract this logic into a package that is shared by both this plugin and the typescript resolver (cc @JounQin)

beamery-tomht:
Are there any updates on this?

ljharb:
You wouldn’t have to, if the extracted package does, it’d work for both.

niksy:
Anything we can do to get this merged?

ljharb:
Still waiting to figure out if we can share a common package (matching this plugin's version support).

PR #2815, opened 2023-07-03
Add "eslint-plugin" to the list of keywords in package.json

Dremora:
Most plugins on npm seem to specify both [eslintplugin](https://www.npmjs.com/search?q=keywords:eslintplugin) and [eslint-plugin](https://www.npmjs.com/search?q=keywords:eslint-plugin) keywords. I have just looked up the list of top plugins using the latter, and got surprised that `eslint-plugin-import` was nowhere to be found.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   "keywords": [
     "eslint",
     "eslintplugin",
+    "eslint-plugin",
     "es6",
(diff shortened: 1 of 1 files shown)


ljharb (approved):
sure, why not

PR #2817, opened 2023-07-05
[New] `extensions`: add the `checkTypeImports` option

phryneas:
Currently, the patterns 
```ts
import type { Foo } from './foo';

export type { Foo } from './foo';
```

are ignored by the `import/extensions` rule.

This adds the `checkTypeImports` option to that rule, so these types of imports can be checked on an opt-in basis.

### DIFF
--- a/docs/rules/extensions.md
+++ b/docs/rules/extensions.md
@@ -56,6 +56,8 @@ For example, `["error", "never", { "svg": "always" }]` would require that all ex
 In that case, if you still want to specify extensions, you can do so inside the **pattern** property.
 Default value of `ignorePackages` is `false`.
 
+By default, `import type` and `export type` style imports/exports are ignored. If you want to check them as well, you can set the `checkTypeImports` option to `true`.
+
 ### Exception
 
 When disallowing the use of certain extensions this rule makes an exception and allows the use of extension when the file would not be resolvable without extension.
@@ -102,6 +104,14 @@ import Component from './Component';
 import express from 'express/index';
 
 import * as path from 'path';
+```
+
+The following patterns are considered problems when the configuration is set to "never" and the option "checkTypeImports" is set to `true`:
+
+```js
+import type { Foo } from './foo.ts';
+
+export type { Foo } from './foo.ts';
 ```
 
 The following patterns are considered problems when configuration set to "always":
@@ -165,6 +175,14 @@ import baz from 'foo/baz.js';
 import express from 'express';
 
 import foo from '@/foo';
+```
+
+The following patterns are considered problems when the configuration is set to "always" and the option "checkTypeImports" is set to `true`:
+
+```js
+import type { Foo } from './foo';
+
+export type { Foo } from './foo';
 ```
 
 ## When Not To Use It
--- a/src/rules/extensions.js
+++ b/src/rules/extensions.js
@@ -14,6 +14,7 @@ const properties = {
   type: 'object',
   properties: {
     pattern: patternProperties,
+    checkTypeImports: { type: 'boolean' },
     ignorePackages: { type: 'boolean' },
   },
 };
@@ -35,7 +36,7 @@ function buildProperties(context) {
     }
 
     // If this is not the new structure, transfer all props to result.pattern
-    if (obj.pattern === undefined && obj.ignorePackages === undefined) {
+    if (obj.pattern === undefined && obj.ignorePackages === undefined && obj.checkTypeImports === undefined) {
       Object.assign(result.pattern, obj);
       return;
     }
@@ -48,6 +49,10 @@ function buildProperties(context) {
     // If ignorePackages is provided, transfer it to result
     if (obj.ignorePackages !== undefined) {
       result.ignorePackages = obj.ignorePackages;
+    }
+
+    if (obj.checkTypeImports !== undefined) {
+      result.checkTypeImports = obj.checkTypeImports;
     }
   });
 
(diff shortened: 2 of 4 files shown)


ljharb:
Is this something that's more than purely style? At runtime, the types are erased, so does it matter what extensions are in the specifier?

phryneas:
If you author a library, it is vital. If your types are missing the file extension (or have a `.ts` instead of a `.js` extension), consumers with the `tsconfig` setting  `moduleResolution: "node16"` will not be able to use your library.

ljharb:
That seems like a flaw in TS's `moduleResolution` then - it's not reasonable to expect your entire dep graph to update its specifier style just to make types work.

phryneas:
Honestly: can we please not have this discussion about something we cannot change? In the end, `import type` is something the TS team invented, and this is how they decided it should work. 

This is behaviour of TypeScript at least from version 4.7 to 5.2, and even if TypeScript would suddenly change their mind, as library authors we need to write code that works with these TypeScript versions, too. 

As you can see [here](https://arethetypeswrong.github.io/?p=%40apollo%2Fclient%403.8.0-beta.3), a library omitting the extension in their types will end up with an "Internal resolution error (2)" if consumed in ESM mode. [here](https://arethetypeswrong.github.io/?p=%40apollo%2Fclient%403.8.0-rc.1) you can see that we fixed it. [This](https://github.com/apollographql/apollo-client/pull/10994) is the PR that we did for that, and we would like to use this ESLint plugin to make sure that we keep our code compliant - isn't that pretty much the purpose of this plugin?

ljharb:
That seems reasonable, but it's still worth pursuing a fix in TS so that the lint rule isn't *needed* in the long run - have you filed an issue there?

phryneas:
There is ongoing discussion in many issues - I think their general argument is that it's required for all other imports to stay ESM-compatible, so it should for those, too. 

There is movement to allow it together with `allowImportingTsExtensions ` over in https://github.com/microsoft/TypeScript/pull/54746, but I think it's more about `.ts` instead of `.js`, not about no extension at all. There is a lot of context in [this gist](https://gist.github.com/andrewbranch/79f872a8b9f0507c9c5f2641cfb3efa6) by Andrew Branch, and I'm very sure they thought about that twice.

Keep in mind, this is not about CJS packages being consumed by ESM, but about how the types of ESM packages consumed by ESM should look like, so they are striving for some kind of new standard here either way and "old ecusystem" won't be too affected by it, since that never shipped ESM in the first place.

phryneas:
🤷‍♂️ Please don't shoot the messenger. 😕

docs/rules/extensions.md
```diff
@@ -167,6 +169,14 @@ import express from 'express';
 import foo from '@/foo';
 ```
 
+The following patterns are considered problems when the option "checkTypeImports" is set to `true`:
+
+```js
+import type { Foo } from './foo';
+
+export type { Foo } from './foo';
+```
```
ljharb, line 178: Let's add a similar section under "never".

src/rules/extensions.js
```diff
@@ -167,7 +172,7 @@ module.exports = {
 
       if (!extension || !importPath.endsWith(`.${extension}`)) {
         // ignore type-only imports and exports
-        if (node.importKind === 'type' || node.exportKind === 'type') { return; }
+        if (props.checkTypeImports !== true && (node.importKind === 'type' || node.exportKind === 'type')) { return; }
```
ljharb, line 175:
```suggestion
        if (!props.checkTypeImports && (node.importKind === 'type' || node.exportKind === 'type')) { return; }
```

tests/src/rules/extensions.js
```diff
@@ -640,6 +640,24 @@ describe('TypeScript', () => {
...
+          }),
+          test({
+            code: 'export type { MyType } from "./typescript-declare";',
+            errors: ['Missing file extension for "./typescript-declare"'],
+            options: [
+              'always',
+              { ts: 'never', tsx: 'never', js: 'never', jsx: 'never', checkTypeImports: true },
+            ],
+            parser,
+          }),
```
ljharb, line 660: Let's also add similar tests with it set to `always`

ljharb (changes requested):

phryneas:
Looking at those "opposite" tests, this will need to have #2813 merged in first so I can use `ruleTesterWithTypeScriptImports` in the tests - right now it does not resolve extensions correctly for `.ts`, so the `{ ts: 'always', tsx: 'always', js: 'always', jsx: 'always', checkTypeImports: true }` line pretty much boils down to `{ checkTypeImports: true }` for imports without extension.

I'll wait for that other PR and then rework the test for this one to include the correct parser.

ljharb:
The `ruleTesterWithTypeScriptImports` part is pretty simple; I think you can pull that into this PR. If you leave it in a separate commit I can preserve authorship.

phryneas:
I've added the comment to the doc you requested and some additional tests with `ruleTesterWithTypeScriptImports`.

As for code ownership - honestly, I don't care much about that, I have too many open PRs. Feel free to just edit away in this :)

tests/src/rules/extensions.js
phryneas: See https://github.com/import-js/eslint-plugin-import/pull/2817/commits/6ac3d03217b80c068fc7846f77410ce5976f943f

docs/rules/extensions.md
phryneas: See https://github.com/import-js/eslint-plugin-import/pull/2817/commits/6ac3d03217b80c068fc7846f77410ce5976f943f

phryneas:
Seems like the CI errors are unrelated to this PR.

alexgwolff:
Hi, any update about this?

luisfernandomoraes:
The checkTypeImports option is a great idea. It ensures that type imports get the same extension checks as regular imports, which is super useful for keeping code consistent and clean, especially in TypeScript projects with strict settings.

ljharb (approved):

PR #2826, opened 2023-07-10
Update SECURITY.md

MarketingIQ:


ljharb (changes requested):
why would i want to delete the security policy?

PR #2832, opened 2023-07-18
[Fix] guard against empty parent

laurens-dg:
Without this fix I get a crash on my current linting set-up.
Not sure what is causing the parent to be empty, might be a bug in the parser, but I ran into this updating everything to the latest versions.


ljharb (changes requested):
We’d need a regression test to land this - can you figure out which code is crashing?

laurens-dg:
Absolutely, but I don't know  how to go about this, I get this crash for parents with type `ExportNamedDeclaration` I've got about 20 of those in my project.

The crash looks like this:
ESLint: 8.45.0

TypeError: Cannot read properties of undefined (reading 'indexOf')
Occurred while linting /foo/bar/baz.tsx:18
Rule: "import/newline-after-import"
    at checkImport (/foo/bar/node_modules/eslint-plugin-import/lib/rules/newline-after-import.js:155:40)

ljharb:
What's on line 18 of that tsx file?

laurens-dg:
`export import Foo = ns.baz.bar.Foo;`

We have a bunch of namespaced classes which we rename and re-export in this way.

ljharb:
woof, that is some gnarly TS syntax :-) i know about `export =` and `import =` but not `export import =`. thanks - it shouldn't be too tricky to come up with a test case based on that.

ljharb:
Although in this case, the rule should presumably be enforcing a newline after the `export import =` line, not ignoring it?

laurens-dg:
To be honest, I don't know.
I also believe this used to work but it might be a recent change to the @typescript-eslint/parser that causes this issue.
[A bit further in the code](https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/newline-after-import.js#L161) this case is considered and skipped.

ljharb:
aha, #1830 indeed attempts to ignore them, treating them as exports rather than imports.

In this case, it sounds like the right bug fix, as long as there's a regression test.

ljharb (approved):

laurens-dg:
Thanks for the help and feedback in getting this merged!

PR #2835, opened 2023-07-19
[Fix] `no-duplicates`: Prefer combined type and regular imports when using `prefer-inline`

### LINKED ISSUE #2834: [import/no-duplicates] (prefer-inline) Does not detect across type and regular imports
I misread the [no-duplicates](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md) doc to mean that when using the "prefer-inline" option, the following would be detected as an error:
```ts
import type { Type } from "mod";
import { value } from "mod";
```
And would fix to:
```ts
import { type Type, value } from "mod";
```

I realize now that this is not what is stated in the doc, but I still want this behavior. It is possible to implement, and from my perspective is probably what those using `prefer-inline` actually want. But if not, perhaps could be implemented with a new option, e.g. `prefer-combined-type-imports`

benkrejci:
…refer-inline

Fixes #2834

I would expect that when using [no-duplicates](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md) with "prefer-inline" option, the following would be detected as an error:

```ts
import type { Type } from "mod";
import { value } from "mod";
```

And would fix to:

```ts
import { type Type, value } from "mod";
```


ljharb (approved):
Looks great!

PR #2842, opened 2023-07-26
[import/no-unused-modules] Report error on export specifier node

Chamion:
I noticed when an export declaration has multiple specifiers of which only a subset is unused the rule reports the error on the whole statement. In my code editor this draws a red squiggly under the whole statement. I can read which specifier is unused from the message of the reported error. Wouldn't it be nicer if the error was reported only on the specifier that is unused? That way I wouldn't necessarily need to even read and can let the red squiggly lead my monkey brain.

Consider the following export declaration

```js
export { foo, bar } from './foo';
```

Assume `foo` is used, `bar` is unused. Usually the fix would be to remove `bar`. I'd expect the reported error to point to that node as it would make it slightly more intuitive to attack it directly.

I'd imagine this change isn't breaking as the same errors still get reported for the same reasons, just pointing to different nodes.

I couldn't think of a reason one would want the report on the parent declaration instead but if there is one please educate me.

I changed the test assertions that have to do with a declaration that has multiple specifiers to also assert a specific location for the reported errors. I changed the node the errors are reported on to make the tests pass again. I then reverted the test changes because CI disagreed even though they passed locally :shrug:.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`order`]: partial fix for [#2687] (thanks [@ljharb])
 - [`no-duplicates`]: Detect across type and regular imports ([#2835], thanks [@benkrejci])
 - [`extensions`]: handle `.` and `..` properly ([#2778], thanks [@benasher44])
- - [`no-unused-modules`]: improve schema (thanks [@ljharb])
+- [`no-unused-modules`]: improve schema (thanks [@ljharb])
+- [`no-unused-modules`]: report error on binding instead of parent export ([#2842], thanks [@Chamion])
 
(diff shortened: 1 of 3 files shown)


ljharb:
I am very surprised that test changes wouldn't be required - can we remove the revert, so i can look into the failures?

Chamion:
> can we remove the revert, so i can look into the failures?

The history of the failed [test runs](https://github.com/import-js/eslint-plugin-import/actions/runs/5669722288/job/15363114272) is still available.

Chamion:
The tests failed on `endLine` so I knew `line` had passed. I tried removing just the end assertions and that seems to work.

I still don't know why it didn't work but this is almost as good in terms of test strictness.

ljharb (approved):

PR #2845, opened 2023-08-01
Document fix-type of no-relative-packages

eps1lon:
Fix types aren't obvious so documenting them avoids having to find it out via trial&error or code spelunking.

### DIFF
--- a/docs/rules/no-relative-packages.md
+++ b/docs/rules/no-relative-packages.md
@@ -1,6 +1,6 @@
 # import/no-relative-packages
 
-🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix) using [`--fix-type suggestion`](https://eslint.org/docs/latest/use/command-line-interface#--fix-type).
 
 <!-- end auto-generated rule header -->
 


docs/rules/no-relative-packages.md
```diff
@@ -1,6 +1,6 @@
 # import/no-relative-packages
 
-🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix) using [`--fix-type suggestion`](https://eslint.org/docs/latest/use/command-line-interface#--fix-type).
```
ljharb, line 3: while this seems great, the header of the docs are autogenerated, and `npm run update:eslint-docs` overwrites this.

@bmish, is this something that can be supported by eslint-doc-generator?

bmish: Since ESLint has a `--fix-type` option, I agree it's a good idea to expose the type that needs to be passed to the option in the doc somewhere.

If we want to add this to the fixable notice, I would want to note that it's optional to specify the fix type:

> 🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix), optionally using [`--fix-type suggestion`](https://eslint.org/docs/latest/use/command-line-interface#--fix-type).

There's also an optional rule doc notice value of `type` (`--rule-doc-notice type`) that could be useful for this purpose. When this notice is enabled, it will show the following for a `suggestion` type rule, and this could be updated to specify the type string as well:

Before:

> 📖 This rule identifies potential improvements.

After:

> 📖 This rule is of type `suggestion` and identifies potential improvements.

https://github.com/bmish/eslint-doc-generator#column-and-notice-types

Filed an issue to track this improvement: https://github.com/bmish/eslint-doc-generator/issues/469

Issue #2846, opened 2023-08-02
dep item resolve@^1.22.3 is a bug version and should not be used


EAGzzyCSL:
> see issue https://github.com/browserify/resolve/issues/297

eslint-plugin-import declared `resolve@^1.22.3` in dependencies（ commit: https://github.com/import-js/eslint-plugin-import/commit/261ee3aa7666edb0131e08db3d3d3428b013df42 ） will cause project which depend on eslint-plugin-import installed    `resolve@1.22.3` rather than 1.22.2 and occur error when build with Webpack or other bundle tools. 

<img width="1206" alt="image" src="https://github.com/import-js/eslint-plugin-import/assets/6220109/2763b14e-08bb-4af8-88ee-e00209fdae2e">

ljharb:
It's not buggy, to be clear, it's just that it breaks bundlers.

v1.22.4 will be published soon, which will address the problem without any change needed here.

ljharb:
v1.22.4 is published.

PR #2847, opened 2023-08-02
[Fix] avoid to use resolve@^1.22.3

EAGzzyCSL:
fix https://github.com/import-js/eslint-plugin-import/issues/2846

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "object.fromentries": "^2.0.6",
     "object.groupby": "^1.0.0",
     "object.values": "^1.1.6",
-    "resolve": "^1.22.3",
+    "resolve": "^1.22.2",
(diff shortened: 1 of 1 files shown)


ljharb:
I’m not going to be going this route; when i have time (I’m traveling atm) I’ll publish a resolve at v1.22.4 and then nothing will need to change here.

EAGzzyCSL:
> I’m not going to be going this route; when i have time (I’m traveling atm) I’ll publish a resolve at v1.22.4 and then nothing will need to change here.

I just realized you are both maintainer of resolve and eslint-plugin-import，it's awesome.

I will close this PR，have a nice trip!

ljharb:
v1.22.4 is published.

PR #2850, opened 2023-08-05
[Docs] remove duplicate fixable notices in docs

bmish:
I noticed this issue randomly.

The automatic fixable notices are already handled by eslint-doc-generator from:
* https://github.com/import-js/eslint-plugin-import/pull/2582.


ljharb (approved):
Thanks, great catch

PR #2851, opened 2023-08-07
[New] TypeScript config: add .cts and .mts extensions

Zamiell:
This PR is pretty self explanatory.
`.cts` and `.mts` are common file extensions used in projects that are not fully CJS or not fully ESM.
So I added those to the config.
I also fixed some of the grammar and made the comments wrap to 100 characters.

### DIFF
--- a/config/typescript.js
+++ b/config/typescript.js
@@ -1,19 +1,22 @@
 /**
- * Adds `.jsx`, `.ts` and `.tsx` as an extension, and enables JSX/TSX parsing.
+ * This config:
+ * 1) adds `.jsx`, `.ts`, `.cts`, `.mts`, and `.tsx` as an extension
+ * 2) enables JSX/TSX parsing
  */
 
 // Omit `.d.ts` because 1) TypeScript compilation already confirms that
 // types are resolved, and 2) it would mask an unresolved
 // `.ts`/`.tsx`/`.js`/`.jsx` implementation.
-const allExtensions = ['.ts', '.tsx', '.js', '.jsx'];
+const typeScriptExtensions = ['.ts', '.cts', '.mts', '.tsx'];
+
+const allExtensions = [...typeScriptExtensions, '.js', '.jsx'];
 
 module.exports = {
-
   settings: {
     'import/extensions': allExtensions,
     'import/external-module-folders': ['node_modules', 'node_modules/@types'],
     'import/parsers': {
-      '@typescript-eslint/parser': ['.ts', '.tsx'],
+      '@typescript-eslint/parser': typeScriptExtensions,
     },
     'import/resolver': {
       node: {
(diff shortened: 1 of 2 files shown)


config/typescript.js
```diff
@@ -25,7 +30,7 @@ module.exports = {
   rules: {
     // analysis/correctness
 
-    // TypeScript compilation already ensures that named imports exist in the referenced module
+    // TypeScript compilation already ensures that named imports exist in the referenced module.
```
ljharb, line 33:
```suggestion
    // TypeScript compilation already ensures that named imports exist in the referenced module
```

config/typescript.js
```diff
@@ -1,19 +1,24 @@
...
 
-// Omit `.d.ts` because 1) TypeScript compilation already confirms that
-// types are resolved, and 2) it would mask an unresolved
-// `.ts`/`.tsx`/`.js`/`.jsx` implementation.
-const allExtensions = ['.ts', '.tsx', '.js', '.jsx'];
+/**
+ * We omit `.d.ts` because:
+ * 1) TypeScript compilation already confirms that types are resolved
+ * 2) it would mask an unresolved `.ts`/`.tsx`/`.js`/`.jsx` implementation
+ */
```
ljharb, line 11: please revert the comment changes; make sure you're not running prettier on a project that doesn't use it :-)

ljharb (changes requested):

config/typescript.js
Zamiell: I didn't use Prettier. I intentionally turned the comment into a JSDoc as a "fix", following the general heuristic of having comments that are next to a variable always be JSDoc so that they are properly "attached" to the variable. In VSCode, when you hover over the variable elsewhere in the code, it will show the comment, but when you use a line comment like this, you don't get this functionality, so using JSDoc is the more "correct" choice. Furthermore, it's a bit confusing to use both styles of comments in the file; why not choose one and stick with it?

ljharb (approved):
Seems reasonable.

ljharb:
I'll rebase and land this once we get more travis-ci credits.

JounQin:
AFAIK, `.cjs`/`.cts` and `.mjs`/`.mts` must be used specifically, want means `'import/extensions': allExtensions` is wrong configed now. `import './x'` should not try to resolve `./x.cts`/`.x.mts` at all.

ljharb:
@JounQin is that always? or only when verbatimModuleSyntax: true?

JounQin:
@ljharb I don't think it's the same thing here.

ljharb:
With TS transpiled to CJS, i would expect extensions to be looked up.

JounQin:
No, the output extension will not change, although this feature is requested for a long time.

ljharb:
Gotcha. Does that mean this PR should be reverted? If it's not that straightforward, a PR to fix it would be appreciated.

JounQin:
Yes, it **must** be reverted, I'll raise a PR tomorrow.

Hope you can pin me with TypeScript related changes. 🩷

ljharb:
That's very reasonable; will do.

PR #2870, opened 2023-08-31
Add support for eslint flat config

OlivierZal:


OlivierZal:
@ljharb, here is the PR to make https://github.com/airbnb/javascript/pull/2818 work.

config/flat.js
```diff
@@ -0,0 +1,9 @@
+const * as importPlugin = require('../src/index');
```
ljharb, line 1:
```suggestion
const importPlugin = require('../src/index');
```

ljharb:
this may also need to be coordinated with #2829.

OlivierZal:
@ljharb, am I suppose to solve the failing tests?

ljharb:
@bmish this PR adds a new entry point; what needs to be done for the docs generator to recognize it?

OlivierZal:
Maybe because there is no `rules` in this entrypoint?

OlivierZal:
https://github.com/bmish/eslint-doc-generator/blob/d27c509ef0dda1cdf5dc911d9eb6bec57556e60a/lib/package-json.ts#L64

OlivierZal:
@ljharb, I think it's because the plugin value is not string: https://github.com/bmish/eslint-doc-generator/blob/d27c509ef0dda1cdf5dc911d9eb6bec57556e60a/lib/package-json.ts#L57

OlivierZal:
https://github.com/bmish/eslint-doc-generator/blob/d27c509ef0dda1cdf5dc911d9eb6bec57556e60a/lib/package-json.ts#L64

ljharb:
hm, maybe it's not yet compatible with flat config?

OlivierZal:
That's my guess 😢

ljharb:
um, I’m confused, why did you delete your forks, making these PRs unrecoverable??

bmish:
I have an issue related to this: https://github.com/bmish/eslint-doc-generator/issues/376. Any thoughts on what needs to be done to support flat config?

ljharb:
Seems like it relies on the plugin name in an entry point being a string.

Issue #2853, opened 2023-08-08
[import/extensions][TypeScript] Suggestion: set `rules` for `.ts`, `.tsx`, `.mts`, `.cts` to `'never'` in `import/typescript` config


OlivierZal:
I suggest adding this rule to the `import/typescript` config (`<eslint-plugin-import>.configs.typescript.rules`):

```javascript
'import/extensions': [
  'error',
  ignorePackages,
  {
    ts: 'never',
    tsx: 'never',
    mts: 'never',
    cts: 'never',
  },
],
```

since explicit `.ts+` extensions are not allowed in TypeScript (except *as on option* in the cases of `noEmit` or `emitDeclarationOnly`, which are not the dominant `tsconfigs` – I guess).

So `eslint` would be prevented from raising an error in the case of an `import/extensions` rule from another eslint config, which usually rules `error` with `never` only for `.js+` extensions.

@ljharb FYI, don't hesitate to let know what you think.

ljharb:
That sounds reasonable to me, but we'd need to use overrides so it only applies in TS files, and not in JS files.

OlivierZal:
```javascript
"overrides": [
    {
      "files": ["**/*.ts", ""**/*.tsx", "**/*.mts", "**/*.cts"],
      "rules": {
        "import/extensions": [
          "error",
          ignorePackages,
          {
            js: "ignorePackages",
            jsx: "ignorePackages",
            mis: "ignorePackages",
            cjs: "ignorePackages",
            ts: "never",
            tsx: "never",
            mts: "never",
            cts: "never",
          },
       ],
    ]
  }
}
```

OlivierZal:
@ljharb, I don't think it's a good thing to do anymore since TypeScript 5.7 now allows to use .ts (and equivalent) extensions thanks to this [tsconfig option](https://www.typescriptlang.org/tsconfig/#rewriteRelativeImportExtensions).

That's why I'm closing the issue.

PR #2875, opened 2023-09-04
Update `import/extensions` rules in `typescript` config

OlivierZal:


OlivierZal:
Related to https://github.com/import-js/eslint-plugin-import/issues/2853

config/typescript.js
```diff
@@ -7,10 +7,16 @@
...
 // `.ts`/`.tsx`/`.js`/`.jsx` implementation.
-const typeScriptExtensions = ['.ts', '.cts', '.mts', '.tsx'];
+const typeScriptKeys = ['ts', 'cts', 'mts', 'tsx'];
+const typeScriptExtensions = typeScriptKeys.map((key) => `.${key}`);
+const typeScriptFiles = typeScriptExtensions.map((ext) => `**/*${ext}`);
 
 const allExtensions = [...typeScriptExtensions, '.js', '.jsx'];
 
+const typeScriptRules = Object.fromEntries(
+  typeScriptKeys.map((key) => [key, 'never']),
```
ljharb, line 17: Doesn’t this depend on your resolution settings in tsconfig?

OlivierZal: In the most recent versions of TypeScript we can configure in the `tsconfig` if we want `ts` extensions to be allowed or not while importing (https://www.typescriptlang.org/tsconfig#allowImportingTsExtensions). 

That's why it's good to deactivate this eslint rule for `ts` extensions, so the tsconfig can raise an error or not without conflicting with it.

ljharb: then it shouldn't be setting it to "never", should it? We don't have an "ignore" setting, so i don't think adding this makes much sense.

OlivierZal: You're right, but `never` is the default and main use of typescript (compilation), the other use is type-checking-only. 

In my opinion a user setting the eslint-import-plugin typescript config wouldn't like to constantly have an eslint issue although he respects the typescript rules.

ljharb: I agree - which is why the override here should be _disabling_ the rule, not configuring it.

ljharb: however, what does the typescript-eslint plugin disable? is this rule not already on the list?

OlivierZal: I don't see it configured: https://github.com/typescript-eslint/typescript-eslint/tree/d2973cca391eeb763d72c5ef2f49cdafe4ffc216/packages/eslint-plugin/src/configs

OlivierZal: They just make a recommendation of use or not, indeed depending on `modeResolution` (https://github.com/typescript-eslint/typescript-eslint/blob/d2973cca391eeb763d72c5ef2f49cdafe4ffc216/docs/linting/troubleshooting/Performance.mdx#L94)

ljharb: ok - it sounds like unless we read the tsconfig here, we can't actually configure OR disable this rule, and safely cover all use cases?

OlivierZal: You're right.

OlivierZal: > I agree - which is why the override here should be _disabling_ the rule, not configuring it.

Disabled by the new commit

ljharb: right, but per the typescript-eslint recommendation, we can't just disable it either.

OlivierZal: Ok, so should depend on `modeResolution`

OlivierZal: @ljharb, and what about a new entry point `typescript-node16` importing current `typescript` config and adding `overrides` to it?

ljharb: a new entry point may be reasonable, but calling it "node16" implies you should be using it if you're on node 16+, which isn't the case.

instead could we use tsconfig-paths to read the tsconfig and dynamically set the right thing?

OlivierZal: I'm not familiar with but I can try. I need to have in mind that sometimes a tsconfig extends another one beforehand.

ljharb: True, but we don't support that since we're stuck on tsconfig-paths v3, and they didn't backport that feature from v4. It would be automatically handled if that changes, so you don't have to worry about it.

ljharb:
@OlivierZal are you no longer interested in completing this PR?

OlivierZal:
Hi @ljharb, you're right I should have explained why.

Actually I don't really remember why I needed it 😅
I think it was when I use the airbnb config in a `.eslintrc` json, but now I customize my config directly in JS.

However I don't think it's a good thing to do anymore since TypeScript 5.7 now allows to use `.ts` (and equivalent) extensions thanks to this [tsconfig option](https://www.typescriptlang.org/tsconfig/#rewriteRelativeImportExtensions).

PR #2877, opened 2023-09-06
Support 'node:' prefix in import/require statements. 

GoldStrikeArch:
Added plugin 'prefer-node-builtin-imports'. 
The plugin is based on 'prefer-node-protocol' rule from 'eslint-plugin-unicorn' plugin. 
This plugin has 0 dependencies and is very minimalistic. All test cases are copied from [eslint-plugin-unicorn_test](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/prefer-node-protocol.mjs)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,7 @@
     "plugins": [
         "eslint-plugin",
         "import",
+        "prefer-node-builtin-imports",
     ],
     "extends": [
         "eslint:recommended",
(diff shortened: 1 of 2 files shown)


ljharb (changes requested):
This would break in a bunch of the node versions we support, and philosophically I prefer to never, ever use the `node:` protocol.

Issue #2882, opened 2023-09-23
newline-after-imports: new option exactCount bug with following comments


FFdhorkin:
I checked out the main branch to try out the new [`exactCount`](https://github.com/import-js/eslint-plugin-import/pull/1933) option for `newline-after-imports`. This does not work properly if there are comments following an import.

Regardless of whether I use
`'import/newline-after-import': ['error', { count: 1, considerComments: false, exactCount: true }],`
or
`'import/newline-after-import': ['error', { count: 1, considerComments: true, exactCount: true }],`

The following code snippets produce an error:
```ts
import _ from 'lodash';

/**
 * A comment is here
 */
```
```ts
import _ from 'lodash';

// A comment is here
```

> Expected 1 empty line after import statement not followed by another import

PR #2884, opened 2023-09-26
[Fix] `newline-after-import`: fix `exactCount` with `considerComments` false positive, when there is a leading comment

kinland:
This addresses #2882 and adds additional test cases to cover the combination of `exactCount` and `considerComments`

### DIFF
--- a/tests/src/rules/newline-after-import.js
+++ b/tests/src/rules/newline-after-import.js
@@ -143,6 +143,31 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
       options: [{ count: 1, exactCount: true }],
     },
     {
+      code: `import foo from 'foo';\n\n// Some random comment\nvar bar = 'bar';`,
+      parserOptions: { ecmaVersion: 2015, sourceType: 'module' },
+      options: [{ count: 2, exactCount: true }],
+    },
+    {
+      code: `import foo from 'foo';\n// Some random comment\nvar bar = 'bar';`,
+      parserOptions: { ecmaVersion: 2015, sourceType: 'module' },
+      options: [{ count: 1, exactCount: true }],
+    },
+    {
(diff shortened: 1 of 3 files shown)


kinland:
Sorry about the force-pushes... I'm a bit sleep deprived

kinland:
I found an edge case that I missed... will push a new commit when I figure out how to fix it.

tests/src/rules/newline-after-import.js
```diff
@@ -133,15 +133,30 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
...
     },
     {
-      code: `import foo from 'foo';\n\n\nvar bar = 'bar';`,
+      code: `import foo from 'foo';\n\n// Some random comment\nvar bar = 'bar';`,
       parserOptions: { ecmaVersion: 2015, sourceType: 'module' },
       options: [{ count: 2, exactCount: true }],
     },
     {
-      code: `import foo from 'foo';\n\nvar bar = 'bar';`,
+      code: `import foo from 'foo';\n// Some random comment\nvar bar = 'bar';`,
```
ljharb, line 141: instead of changing existing test cases, let's add new ones :-)

ljharb (changes requested):

tests/src/rules/newline-after-import.js
kinland: as you wish! new commit just pushed

ljharb (approved):
LGTM assuming the new tests fail without the change, and pass with them. Thanks!

reosarevok:
Thanks a lot! Clearly my one added test wasn't good enough when rebasing the previous PR, sorry.

kinland:
> Thanks a lot! Clearly my one added test wasn't good enough when rebasing the previous PR, sorry.

Honestly, I can't fault you for that. I didn't consider that a comment *before* an import would break my code, either. These things get very complex quickly.

Issue #2722, opened 2023-02-20
[import/order] Breaking change behavior in v2.27


StefanoMagrassi:
In version 2.27.0 it was introduced a [new strategy](https://github.com/import-js/eslint-plugin-import/compare/v2.26.0...v2.27.0#diff-2825fcc427c846fe18d501930e1a3d83be722027f1f643801afdcc2c255dd4a8R285) which splits the imports paths in segments and compares them independently.

However, this leads to different results for the same inputs in some cases, for example:

```ts
import {screen} from '@testing-library/react';
import {checkDropdown} from '@testing/form';
```
this is the "right" order as by v2.26 because:
```ts
compareString('@testing-library/react', '@testing/form') === -1
```
but in v2.27 it is reported as error because:
```ts
compareString('@testing-library', '@testing') === 1
```
I don't know if the current behavior is the right one, but it is definitely a breaking change

ljharb:
I agree that it was an unintentional change, and also that the one that’s right in 2.26 is the right one. (i think the change you’re referencing was #2396)

StefanoMagrassi:
> (i think the change you’re referencing was #2396)

yep

7iomka:
Any news on this?

ljharb:
No, not yet, but it'd be great to get a PR up to fix it.

7iomka:
> No, not yet, but it'd be great to get a PR up to fix it.

Just revert #2396

ljharb:
@7iomka it may not be that simple; there may be some fixes in it that we'd need to keep.

DaveTryon:
This is still broken in 2.28. We're hitting it in https://github.com/microsoft/accessibility-insights-web/pull/6846. `git bisect` confirms that it broke in #2396 

Here's a quick unit test to use as a reference:

```
    // Option alphabetize: {order: 'asc'} when folders and paths start with the same substring
    test({
      code: `
        import a from "foo/foobar-file";
        import b from "foo/foobar/file";
      `,
      options: [{ alphabetize: { order: 'asc' } }],
    }),
```

tw3:
PR: https://github.com/import-js/eslint-plugin-import/pull/3196

PR #2885, opened 2023-09-26
[Fix] `order`: do not compare first path segment for relative paths

mihkeleidast:
Should fix the regression in #2682.

The gist of the fix is that for relative paths, siblings always start with "." and parents always start with "..".  When comparing them, it will always result in "." being first, which was the bug. The sorting based on subpath segments should start from the second path segment, when there is a possibility that one of them is different.

### DIFF
--- a/tests/src/rules/order.js
+++ b/tests/src/rules/order.js
@@ -169,6 +169,34 @@ ruleTester.run('order', rule, {
         ['sibling', 'parent', 'external'],
       ] }],
     }),
+    // Grouping import types and alphabetize
+    test({
+      code: `
+        import async from 'async';
+        import fs from 'fs';
+        import path from 'path';
+
+        import index from '.';
+        import relParent3 from '../';
+        import relParent1 from '../foo';
+        import sibling from './foo';
+      `,
+      options: [{ groups: [
+        ['builtin', 'external'],
+      ], alphabetize: { order: 'asc', caseInsensitive: true } }],
+    }),
+    test({
+      code: `
+      import { fooz } from '../baz.js'
+      import { foo } from './bar.js'
+      `,
+      options: [{
+        alphabetize: { order: 'asc', caseInsensitive: true },
+        groups: ['builtin', 'external', 'internal', ['parent', 'sibling', 'index'], 'object'],
+        'newlines-between': 'always',
+        warnOnUnassignedImports: true,
+      }],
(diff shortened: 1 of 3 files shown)


tests/src/rules/order.js
```diff
@@ -169,6 +169,22 @@ ruleTester.run('order', rule, {
...
+    test({
+      code: `
+        import async from 'async';
+        import fs from 'fs';
+        import path from 'path';
+
+        import index from '.';
+        import relParent3 from '../';
+        import relParent1 from '../foo';
+        import sibling from './foo';
```
Pearce-Ropion, line 182: This test confuses me. The default group order is has `index` last so why is it first here?

mihkeleidast: I think this configuration ends up so that all other groups are grouped, so the imports are mingled together. Then they are sorted alphabetically, which puts this first.

That's also how it worked in 2.26.0, which is why I added it here. Ideally I would also sort it last, as that would be most logical - for that we can use either pathGroups and distinctGroups options, or your other PR.

ljharb:
Can you add the exact test case from the OP of #2682?

mihkeleidast:
@ljharb added! This was a good suggestion, as that particular case needed additional handling.

mihkeleidast:
Not sure what is going on with the failing actions, do not seem related to my changes?

ljharb:
Likely not; i'll rerun.

acidoxee:
Hi everyone, do you reckon this PR has a reason to be blocked? Like a lot of people, I've been sticking with v2.26.0 of this plugin since https://github.com/import-js/eslint-plugin-import/issues/2682 has been a thing, and I'd very much like to get back to being up-to-date and with imports in the right order 😉

hood:
Any news here? MR looks fine, any blockers I can help get rid of?

mihkeleidast:
> Any news here? PR looks fine, any blockers I can help getting rid of?

I think while my fix does work for some of the reported issues, some would still not be fixed by this, e.g. https://github.com/import-js/eslint-plugin-import/issues/2722

I could not fix the other issues, so gave up as it does not affect my usecases as much. But probably would help if the fix would resolve all of the different issues that have been reported.

ljharb (approved):
LGTM as long as existing tests pass :-)

PR #2886, opened 2023-09-26
fix: changed resolvedPath check to check if its not a string

prisis:
The "eslint-module-utils/resolve" returns 3 different types

- the full module filesystem path
- null if package is core
- undefined if not found

if null is returned from the resolve function the path.relative throws a error that i expects a string

> TypeError [ERR_INVALID_ARG_TYPE]: The "to" argument must be of type string. Received null

Data passed into the resolve function.

```
resolve("../dist/postinstall", {
  id: 'import/no-useless-path-segments',
  options: [ { commonjs: true, noUselessIndex: true } ],
  report: [Function: report]
});
```

### DIFF
--- a/src/rules/no-useless-path-segments.js
+++ b/src/rules/no-useless-path-segments.js
@@ -115,7 +115,7 @@ module.exports = {
       }
 
       // Path is not existing --> Return directly (following code requires path to be defined)
-      if (resolvedPath === undefined) {
+      if (typeof resolvedPath !== "string") {
         return;
       }
 


ljharb (changes requested):
Can you provide a test case that fails without this change?

prisis:
Did dive a bit deeper into this issue, looks like this is the core of it, https://github.com/jsenv/core/blob/main/packages/independent/eslint-import-resolver/src/resolver.js#L282

ljharb:
That looks like a vendored and modified copy of one of this repo's packages? I'm very confused about that, so please keep me posted.

prisis:
Linking: https://github.com/jsenv/core/issues/379

prisis:
Will close this PR, the issue was fixed in the resolver package. Thank you for your time!

PR #2917, opened 2023-11-04
Autogenerate many rule options list with eslint-doc-generator

bmish:
I'm trying out this new feature I implemented that generates the lists of rule options from the rule schemas. More info:
* https://github.com/bmish/eslint-doc-generator/pull/481

This requires quite a bit of cleanup including:

* Moving option info into the schema as much as possible
* Reorganizing rule docs so that the options list can fit nicely

I noticed that many rules do not even mention their options in their docs today, or do so incompletely.

Since there are so many rules with options, I haven't quite fully filled out and perfected all the schemas/docs yet, but it's a start.


docs/rules/dynamic-import-chunkname.md
```diff
@@ -11,6 +11,15 @@ This rule enforces naming of webpack chunks in dynamic imports. When you don't e
 This rule runs against `import()` by default, but can be configured to also run against an alternative dynamic-import function, e.g. 'dynamicImport.'
 You can also configure the regex format you'd like to accept for the webpackChunkName - for example, if we don't want the number 6 to show up in our chunk names:
 
+<!-- begin auto-generated rule options list -->
+
+| Name                     | Type     |
+| :----------------------- | :------- |
+| `importFunctions`        | String[] |
+| `webpackChunknameFormat` | String   |
```
ljharb, line 19: `string` should be lowercase, i think

docs/rules/max-dependencies.md
```diff
@@ -10,14 +10,14 @@ Importing multiple named exports from a single module will only count once (e.g.
...
-```js
-"import/max-dependencies": ["error", {
-  "max": 10,
-  "ignoreTypeImports": false,
-}]
-```
+| Name                | Type    | Default |
+| :------------------ | :------ | :------ |
+| `ignoreTypeImports` | Boolean | `false` |
+| `max`               | Number  | `10`    |
```
ljharb, line 18: same with all the primitive types

ljharb (commented):
Looks like a great start :-)

docs/rules/dynamic-import-chunkname.md
bmish: Do you know of any related example / precedent for displaying variable/data types one way or the other, or have any other reasoning behind your preference? I'm open to considering a change.

The current reasoning is that arbitrary, user-defined values like the name of an option are displayed in backticks, whereas values from a fixed set like variable types are simply capitalized.

bmish: Here's a comprehensive example that shows how we format different kinds of values: https://github.com/bmish/eslint-doc-generator/blob/main/docs/examples/eslint-plugin-test/docs/rules/no-foo.md#options

ljharb: `String` means an object - the constructor - and `string` means a primitive. It's what `typeof` produces for primitives as well.

TypeScript follows this convention as well.

PR #2919, opened 2023-11-07
[Fix] `no-extraneous-dependencies`: ignore `export type { ... } from '...'` when `includeTypes` is `false`

Pandemic1617:
## Summary

The `no-extraneous-dependencies` rule still checks a the dependency of `export type { T } from 'a-dev-dependency'` even when it is supposed to exclude type only imports

## Example
This triggers an error
```typescript
export type { T } from "a-dev-dependency";
``` 

This doesn't
```typescript
import type { T } from "a-dev-dependency"; 
export type { T };
``` 

`package.json`
```json
{
    "devDependencies": {
        "a-dev-dependency": "*"
    }
}
```
rule:
```json
{
    "devDependencies": false,
    "includeTypes": false // false by default
 }
```

## Fix

Check if the export statement's kind is `type` and if so, treat it like an `import type` statement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


ljharb (approved):
Thanks, looks good!

PR #2920, opened 2023-11-10
remove array-includes

dulmandakh:


ljharb (changes requested):
This would be a breaking change, and we’re definitely not dropping support for older node versions.

dulmandakh:
@ljharb you can create a stable branch, and accept breaking changes into the main, then release a major version with breaking changes.

why support a long ago EOL Node.js versions?

ljharb:
I could, but i have no desire to do that.

Breaking changes are the most costly thing to inflict on the ecosystem, and there’s still people using those versions. Node’s EOL status is irrelevant to whether the ecosystem should support it.

dulmandakh:
IMHO, we could reduce dependencies and make good use of features added in newer Node.js versions. Dependency reduction will reduce bandwidth, storage requirements thus make this library greener, also improve developer experience.

ljharb:
I could, but more dependencies is better, and there are no features added in newer node versions that I need and can't take advantage of through packages.

Bandwith and storage is npm's concern, as it's simply not sustainable for individual authors to try to fix that problem, and developer experience is largely unaffected since installs happen rarely and are cached, and disk space is infinite and free.

Either way, for future reference, to submit an unsolicited PR with a breaking change to a repo without even an explanation is quite rude.

PR #2931, opened 2023-11-26
Feature: no duplicates require

### LINKED ISSUE #2925: Disallow duplicate require
I have following code : 

```
const { addCellToRow } = require("./cell-utils.js");
const { setCellValue, findCell, createCell } = require("./cell-utils.js");
```

Both require are using the same path, and should thus be grouped.

I've tried to enable this in my .eslintrc : 

    "import/no-duplicates": ["error", {"prefer-inline": true}],

But it didn't have any effect.

Are there any rules that can catch this ?

Pandemic1617:
implements detection of duplicate `require` imports

resolves #2925 

Note that auto fix isn't implemented

### DIFF
--- a/src/rules/no-duplicates.js
+++ b/src/rules/no-duplicates.js
@@ -9,6 +9,10 @@ try {
   typescriptPkg = require('typescript/package.json'); // eslint-disable-line import/no-extraneous-dependencies
 } catch (e) { /**/ }
 
+function getSourceFromNode(node) {
+  return node.callee ? node.arguments[0] : node.source;
+}
+
 function checkImports(imported, context) {
   for (const [module, nodes] of imported.entries()) {
     if (nodes.length > 1) {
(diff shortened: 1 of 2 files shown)


ljharb (changes requested):
Can we add some tests with a function that's called "require" but isn't the global one?

Pandemic1617:
> Can we add some tests with a function that's called "require" but isn't the global one?

```js
const require = (x) => {};

const a = require("./a.js");
```
This require would be detected as the global require even though it isn't.

This is indeed a limitation of my implementation but I thought it would be acceptable since the following rules also have the same limitation:
- `no-dynamic-require`
- `no-amd`
- `max-dependencies`
- `no-self-import`
- `no-relative-packages`

ljharb:
Hmm - are there no rules that *dont* have the limitation?

Pandemic1617:
I am not aware of any and even the `moduleVisitor` util has that limitation

PR #2952, opened 2024-01-12
[Fix] `newline-after-import`: fix `considerComments` option when require

### LINKED ISSUE #2905: [import/newline-after-import] `considerComments` not working with `require` statements
The option `considerComments` doesn't seem to work with `require` statements. For example, with the following configuration:

```js
"import/newline-after-import": ["warn", { count: 2, considerComments: true }]
```

These are considered valid:

```js
const fs = require("fs");

// Test comment
const FOO = true;
```
```js
const fs = require("fs");
/**
 * Test comment
 */
const FOO = true;
```

While they should be:
```js
const fs = require("fs");


// Test comment
const FOO = true;
```
```js
const fs = require("fs");


/**
 * Test comment
 */
const FOO = true;
```

It works as it should with `import` statements.

developer-bandi:
fix #2905 

- new var in test file  use multiple space when require
- fix or remove invalid test cases
- add feature considerComments is applied normally when use require

### DIFF
--- a/src/rules/newline-after-import.js
+++ b/src/rules/newline-after-import.js
@@ -124,7 +124,7 @@ module.exports = {
       }
     }
 
-    function commentAfterImport(node, nextComment) {
+    function commentAfterImport(node, nextComment, type) {
       const lineDifference = getLineDifference(node, nextComment);
       const EXPECTED_LINE_DIFFERENCE = options.count + 1;
 
@@ -140,7 +140,7 @@ module.exports = {
             line: node.loc.end.line,
             column,
           },
-          message: `Expected ${options.count} empty line${options.count > 1 ? 's' : ''} after import statement not followed by another import.`,
+          message: `Expected ${options.count} empty line${options.count > 1 ? 's' : ''} after ${type} statement not followed by another ${type}.`,
           fix: options.exactCount && EXPECTED_LINE_DIFFERENCE < lineDifference ? undefined : (fixer) => fixer.insertTextAfter(
             node,
             '\n'.repeat(EXPECTED_LINE_DIFFERENCE - lineDifference),
@@ -178,7 +178,7 @@ module.exports = {
       }
 
       if (nextComment && typeof nextComment !== 'undefined') {
-        commentAfterImport(node, nextComment);
+        commentAfterImport(node, nextComment, 'import');
       } else if (nextNode && nextNode.type !== 'ImportDeclaration' && (nextNode.type !== 'TSImportEqualsDeclaration' || nextNode.isExport)) {
         checkForNewLine(node, nextNode, 'import');
       }
@@ -215,8 +215,18 @@ module.exports = {
               || !containsNodeOrEqual(nextStatement, nextRequireCall)
             )
           ) {
+            let nextComment;
+            if (typeof statementWithRequireCall.parent.comments !== 'undefined' && options.considerComments) {
+              const endLine = node.loc.end.line;
+              nextComment = statementWithRequireCall.parent.comments.find((o) => o.loc.start.line >= endLine && o.loc.start.line <= endLine + options.count + 1);
+            }
 
-            checkForNewLine(statementWithRequireCall, nextStatement, 'require');
+            if (nextComment && typeof nextComment !== 'undefined') {
+
+              commentAfterImport(statementWithRequireCall, nextComment, 'require');
+            } else {
+              checkForNewLine(statementWithRequireCall, nextStatement, 'require');
+            }
           }
         });
       },
--- a/tests/src/rules/newline-after-import.js
+++ b/tests/src/rules/newline-after-import.js
@@ -8,6 +8,7 @@ import { getTSParsers, parsers, testVersion } from '../utils';
 const IMPORT_ERROR_MESSAGE = 'Expected 1 empty line after import statement not followed by another import.';
 const IMPORT_ERROR_MESSAGE_MULTIPLE = (count) => `Expected ${count} empty lines after import statement not followed by another import.`;
 const REQUIRE_ERROR_MESSAGE = 'Expected 1 empty line after require statement not followed by another require.';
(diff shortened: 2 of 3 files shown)


src/rules/newline-after-import.js
```diff
@@ -140,7 +140,7 @@ module.exports = {
             line: node.loc.end.line,
             column,
           },
-          message: `Expected ${options.count} empty line${options.count > 1 ? 's' : ''} after import statement not followed by another import.`,
+          message: `Expected ${options.count} empty line${options.count > 1 ? 's' : ''} after ${type} statement not followed by another ${type}.`,
```
ljharb, line 143: can we make `type` (and while we're at it, the count and the pluralization) part of the `data` instead of interpolated in?

it'd also be fine to do this as a followup.

tests/src/rules/newline-after-import.js
```diff
@@ -202,7 +203,7 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
       options: [{ count: 4, exactCount: true }],
     },
     {
-      code: `var foo = require('foo-module');\n\n\n\n// Some random comment\nvar foo = 'bar';`,
+      code: `var foo = require('foo-module');\n\n\n\n\n// Some random comment\nvar foo = 'bar';`,
```
ljharb, line 206: the "count" is set to 4, so it's surprising to me that this would have 5 newlines. can you explain?

tests/src/rules/newline-after-import.js
```diff
@@ -852,14 +861,26 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
...
+      code: `var foo = require('foo-module');\n\n/**\n * Test comment\n */\nvar foo = 'bar';`,
+      output: `var foo = require('foo-module');\n\n\n/**\n * Test comment\n */\nvar foo = 'bar';`,
       errors: [{
         line: 1,
         column: 1,
-        message: 'Expected 2 empty lines after require statement not followed by another require.',
+        message: REQUIRE_ERROR_MESSAGE_MULTIPLE(2),
       }],
-      parserOptions: { ecmaVersion: 2015 },
+      parserOptions: { ecmaVersion: 2015, sourceType: 'module' },
```
ljharb, line 882: why is sourceType module needed here? this is CJS, not ESM.
```suggestion
      parserOptions: { ecmaVersion: 2015 },
```

developer-bandi: this is my mistake. I will remove the module option.

tests/src/rules/newline-after-import.js
developer-bandi: If we print code above, it will look like this:
```javascript
var foo = require('foo-module');




// Some random comment
var foo = 'bar';
```
In other words, five newlines are needed to create four valid blank lines for the option.

src/rules/newline-after-import.js
developer-bandi: I'm not sure I understand exactly. Is it correct to say that create a function that params are count ,type and return the created message and apply it to other messages? below is example code

```
function example(count, type){
    return  `Expected ${options.count} empty line${options.count > 1 ? 's' : ''} after ${type} statement not followed by another ${type}.`,
}
....
messages:example(options.count,type)

```

tests/src/rules/newline-after-import.js
ljharb: aha, thanks, that makes sende

src/rules/newline-after-import.js
developer-bandi: @ljharb If my understanding is correct, I will proceed in the follow-up PR.

ljharb: No, not a function - using eslint's own message interpolation format; no function required. A followup is fine.

tests/src/rules/newline-after-import.js
```diff
@@ -852,14 +861,26 @@ ruleTester.run('newline-after-import', require('rules/newline-after-import'), {
       parserOptions: { ecmaVersion: 2015, considerComments: true, sourceType: 'module' },
     },
     {
-      code: `const foo = require('foo');\n\n\n// some random comment\nconst bar = function() {};`,
-      options: [{ count: 2, exactCount: true, considerComments: true }],
```
ljharb, line 856: can we keep this existing test case, to ensure it's not a breaking change?

src/rules/newline-after-import.js
developer-bandi: Ah, I understand. Are you referring to the code format below? (example is `consistent-type-specifier-style` rule)

```javascript
context.report({
....
            message: 'Prefer using a top-level {{kind}}-only import instead of inline {{kind}} specifiers.',
            data: {
              kind: kind.join('/'),
            },
          }
 ....
```

tests/src/rules/newline-after-import.js
developer-bandi: This test case is valid, so if we want to keep it, I think the test case should be included in the valid array. Is it okay to do this?

src/rules/newline-after-import.js
ljharb: yes, exactly that.

developer-bandi: Thank you. I will proceed with the follow-up work as mentioned above.

tests/src/rules/newline-after-import.js
ljharb: Sure, that's fine - moving the test shows it's a bug fix.

developer-bandi: thanks, I applied it.

Arkellys:
Any news about this?

developer-bandi:
@Arkellys for merge, i wait code review. @ljharb Could you please take a look?

ljharb (approved):
LGTM, thanks

Issue #2132, opened 2021-06-19
Rule `import/no-unresolved` raise false positive when there is no `main` field in the `package.json`


macabeus:
**What did you do? Please include the actual source code causing the issue, as well as the command that you used to run ESLint.**

```js
import GBAEmulator, { GbaContext } from 'react-gbajs'
```
[_Full code_](https://github.com/macabeus/klo-gba.js/blob/master/brush/src/components/Emulator/index.js)

```bash
./node_modules/.bin/eslint .
```

**What did you expect to happen?**
Should raise no error. This package is present in the `node_modules`.

An important note:
Since it's only for the browser, it has the field `browser` instead of `main` in its `package.json` ([link](https://github.com/macabeus/react-gbajs/blob/87ef927edbfd24adc1370f95d650edc32570ac83/package.json#L5)). I made it following [the NPM documentation](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#browser).
Despite that, I noticed that by adding the field `main` in its `package.json`, the error suppress - but it doesn't look correct.

**What actually happened? Please copy-paste the actual, raw output from ESLint.**
```
/Users/macabeus/ApenasMeu/klo-gba.js/brush/src/components/Emulator/index.js
  2:41  error  Unable to resolve path to module 'react-gbajs'  import/no-unresolved

/Users/macabeus/ApenasMeu/klo-gba.js/brush/src/components/KloGbaSidebar/Hacks/index.js
  7:28  error  Unable to resolve path to module 'react-gbajs'  import/no-unresolved

/Users/macabeus/ApenasMeu/klo-gba.js/brush/src/hooks/useGbaSaveRestoreState.js
  2:28  error  Unable to resolve path to module 'react-gbajs'  import/no-unresolved

/Users/macabeus/ApenasMeu/klo-gba.js/brush/src/index.js
  8:29  error  Unable to resolve path to module 'react-gbajs'  import/no-unresolved

/Users/macabeus/ApenasMeu/klo-gba.js/brush/src/providers/VisionProvider.js
  4:28  error  Unable to resolve path to module 'react-gbajs'  import/no-unresolved

✖ 5 problems (5 errors, 0 warnings)

✨  Done in 4.73s.
```

**Steps to reproduce this issue:**

1. Clone this repository: https://github.com/macabeus/klo-gba.js
1. Run `yarn` on its root
1. `cd  brush`
2. `yarn run lint`

ljharb:
[...]

PR #2957, opened 2024-01-22
[Fix/New] Node resolver: Try to use `require.resolve` when suitable

akx:
This PR proposes to have the Node resolver use [`require.resolve`](https://nodejs.org/api/modules.html#requireresolverequest-options) when possible; among other things, this enables support for packages that only have `exports.main` (when the underlying Node.js version supports them) (see https://github.com/import-js/eslint-plugin-import/issues/2132).

I don't know if there are subtleties to this, but I feel like this could work.

### DIFF
--- a/resolvers/node/index.js
+++ b/resolvers/node/index.js
@@ -60,6 +60,19 @@ exports.resolve = function (source, file, config) {
     return { found: true, path: null };
   }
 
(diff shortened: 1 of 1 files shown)


akx:
Rebased to fix conflicts from 2d38b3367e90df2eabd6c6aad0a05d6a1fc96734.

PR #2963, opened 2024-02-10
fix: also delete parserOptions.EXPERIMENTAL_useProjectService

### LINKED ISSUE #2962: When clearing typed linting options, also delete parserOptions.EXPERIMENTAL_useProjectService
Following https://github.com/typescript-eslint/typescript-eslint/issues/8424#issuecomment-1937117103: right now, `eslint-plugin-import` deletes the two known `parserOptions` from [typescript-eslint's typed linting](https://typescript-eslint.io/linting/typed-linting):

https://github.com/import-js/eslint-plugin-import/blob/7a21f7e10f18c04473faadca94928af6b8e28009/utils%2Fparse.js#L80-L85

...but we added an `EXPERIMENTAL_useProjectService` too. Folks using it are experiencing the performance issues from #1408 again.

JoshuaKGoldberg:
Fixes #2962.

### DIFF
--- a/utils/parse.js
+++ b/utils/parse.js
@@ -81,6 +81,7 @@ exports.default = function parse(path, content, context) {
   // "project" or "projects" in parserOptions. Removing these options means the parser will
   // only parse one file in isolate mode, which is much, much faster.
   // https://github.com/import-js/eslint-plugin-import/issues/1408#issuecomment-509298962
+  delete parserOptions.EXPERIMENTAL_useProjectService;
   delete parserOptions.project;
   delete parserOptions.projects;
 
--- a/utils/CHANGELOG.md
+++ b/utils/CHANGELOG.md
@@ -5,6 +5,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
+### Fixed
+- `parse`: also delete `parserOptions.EXPERIMENTAL_useProjectService` ([#2963], thanks [@JoshuaKGoldberg])
+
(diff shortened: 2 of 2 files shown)


utils/parse.js
```diff
@@ -81,6 +81,7 @@ exports.default = function parse(path, content, context) {
   // "project" or "projects" in parserOptions. Removing these options means the parser will
   // only parse one file in isolate mode, which is much, much faster.
   // https://github.com/import-js/eslint-plugin-import/issues/1408#issuecomment-509298962
+  delete parserOptions.EXPERIMENTAL_useProjectService;
   delete parserOptions.project;
   delete parserOptions.projects;
```
JoshuaKGoldberg, line 86: We'll eventually add a new name for the option like `projectService` or something like it. Which would increase this list to four properties.

An alternative to hardcoding names in this plugin would be https://github.com/typescript-eslint/typescript-eslint/issues/8428: importing a shared utility we provide. Is that something you'd find convenient? Either as a part of one of our packages _(easiest for us)_ or as a standalone thing _(we'd have to discuss)_?

JounQin: It's late to add a util for compatibility, what means deleting manually is required for old `@typescript-eslint` versions.

JounQin (approved):

utils/parse.js
ljharb: For new ones, though, a conditional require would be just fine.

ljharb: sure, sounds good

ljharb (approved):

PR #2982, opened 2024-03-13
make exportmap procedures static instead of monkeypatching exportmap

soryy708:
The following procedures:

```
ExportMap.get
ExportMap.for
ExportMap.parse
```

are conceptually static methods of `ExportMap`, but instead of using the `static` keyword they are monkey-patched onto the `ExportMap` class

```
ExportMap.get = function (...) {...}
```

which is surprising.

Lets make them static methods proper.

---

Tip for the reviewer/maintainer:

Use the "Hide whitespace" feature of GitHub's diff.

image

### DIFF
--- a/src/ExportMap.js
+++ b/src/ExportMap.js
@@ -26,6 +26,97 @@ const log = debug('eslint-plugin-import:ExportMap');
 const exportCache = new Map();
 const tsconfigCache = new Map();
 
+/**
+ * parse docs from the first node that has leading comments
+ */
+function captureDoc(source, docStyleParsers, ...nodes) {
+  const metadata = {};
+
+  // 'some' short-circuits on first 'true'
+  nodes.some((n) => {
+    try {
+
+      let leadingComments;
+
+      // n.leadingComments is legacy `attachComments` behavior
+      if ('leadingComments' in n) {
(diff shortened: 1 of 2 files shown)


src/ExportMap.js
```diff
@@ -296,442 +732,6 @@ function captureTomDoc(comments) {
 
 const supportedImportTypes = new Set(['ImportDefaultSpecifier', 'ImportNamespaceSpecifier']);
```
ljharb, line 733: this (and i think the others) is used inside parse, so let's move all such things higher than the class definition?

or, is there's a conceptual cycle here so there's no source position that can fully satisfy "define before use"?

soryy708: I wasn't aware that we have a convention of "define before use".
CI passed, so it's not enforced in lint. If we have that convention, then I believe we ought to enforce it in lint.

ljharb: it's enforced for functions - but i guess the eslint core rule doesn't apply to const/lets.

ljharb (approved):
thanks!

ljharb:
you're right that this repo doesn't enforce that rule for some reason; i'll fix it shortly :-)

PR #2983, opened 2024-03-13
resolve lint warnings

soryy708:
When running the tests / ci, ESLint complains about warnings related to the `no-console` rule.

The places where we have that console make sense to have console, so it's a false-positive.

Lets mark those specific places to be ignored by eslint, to reduce noise from the linter output.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -237,6 +237,16 @@
         },
         {
             "files": [
+                "resolvers/webpack/index.js",
+                "resolvers/webpack/test/example.js",
+                "utils/parse.js",
+            ],
+            "rules": {
+                "no-console": "off",
+            },
+        },
+        {
+            "files": [
                 "resolvers/*/test/**/*",
             ],
             "env": {


utils/parse.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+/* eslint-disable no-console */
```
ljharb, line 3: let's use overrides in the eslintrc instead of inline comments for these

ljharb: no - eslintrc's "overrides" can apply to specific globs of files, so you can indicate that `utils/*`, or `utils/parse.js`, etc, should have the no-console rule set to "off".

soryy708: TIL.

How's this?

- https://github.com/import-js/eslint-plugin-import/pull/2983/commits/d6907f795fcd7732d3c866d1646929316d796fcc

ljharb: looks great, thanks!

ljharb (approved):

PR #2985, opened 2024-03-18
separate export map from its builder

soryy708:
`ExportMap` is currently a very large file that is hard to follow, and therefore hard to maintain.

I identified that its static methods, which are responsible for building instances of `ExportMap`, can be generalized to the [builder software design pattern](https://en.wikipedia.org/wiki/Builder_pattern).

Lets make the separation, to make it clear that instances of `ExportMap` aren't coupled to any of the globals or procedures that are currently in `ExportMap.js`.

It's also good to rename the files and classes appropriately.

### DIFF
--- a/tests/src/core/getExports.js
+++ b/tests/src/core/getExports.js
@@ -4,7 +4,7 @@ import sinon from 'sinon';
 import eslintPkg from 'eslint/package.json';
 import typescriptPkg from 'typescript/package.json';
 import * as tsConfigLoader from 'tsconfig-paths/lib/tsconfig-loader';
-import ExportMap from '../../../src/ExportMap';
+import ExportMapBuilder from '../../../src/exportMapBuilder';
 
 import * as fs from 'fs';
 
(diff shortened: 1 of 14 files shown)


tests/src/core/getExports.js
```diff
@@ -4,7 +4,7 @@ import sinon from 'sinon';
 import eslintPkg from 'eslint/package.json';
 import typescriptPkg from 'typescript/package.json';
 import * as tsConfigLoader from 'tsconfig-paths/lib/tsconfig-loader';
-import ExportMap from '../../../src/ExportMap';
+import ExportMap from '../../../src/exportMapBuilder';
```
ljharb, line 7:
```suggestion
import Exports from '../../../src/exportMapBuilder';
```
should this be named "Exports" or "ExportMapBuilder"?

ljharb (approved):
Assuming that the ExportMap class is moved without changes, this LGTM!

tests/src/core/getExports.js
soryy708: Fair.

- https://github.com/import-js/eslint-plugin-import/pull/2985/commits/cc1b19e6417c5f28ba2ade6db6477a24a8f42b3b

ljharb (approved):

PR #2987, opened 2024-03-19
[Tests] appveyor -> GHA (run tests on Windows in both pwsh and WSL + Ubuntu)

joeyguerra:
Converting the `appveyor` pipeline to GitHub Actions.


.eslintrc
```diff
@@ -265,4 +265,13 @@
...
         },
     ],
+    "ignorePatterns": [
+        "tests/files/broken-trampoline.js",
+        "tests/files/cycles/flow-typeof.js",
+        "tests/files/default-export-from.js",
+        "tests/files/flowtypes.js",
+        "tests/files/issue-195/Endpoints.js",
+        "tests/files/named-trampoline.js",
+        "tests/files/trampoline.js",
```
ljharb, line 275: fine for this PR, but what were the linting errors? if it's not impractical, i'd rather disable unimportant rules to ensure that files can still be linted for parse correctness

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,175 @@
...
+      if: matrix.configuration == 'wsl'
+      run: |
+        export ESLINT_VERSION=7
+        sudo apt update -y
+        sudo apt-get install curl -y
+        curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/master/install.sh | bash
+        export NVM_DIR="$HOME/.nvm"
+        [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
+        [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
+        nvm install ${{ matrix.node-version }}
```
ljharb, line 38: if we're going to use nvm anyways, can we use the ljharb/actions/node/install action instead for WSL?

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,175 @@
...
+          npm install @typescript-eslint/parser@4
+        fi
+        if [ ESLINT_VERSION -eq 8 ]; then
+          npm i --no-save eslint-plugin-import@'.' -f
+          npm run build
+        fi
+
+        git config core.symlinks true
+        git reset --hard
+        npm config set package-lock false
```
ljharb, line 71: this should already be set via .npmrc

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,175 @@
...
+            pushd $resolver
+            npm install
+            npm ls nyc > /dev/null
+            if [ $? -eq 0 ]; then
+              npm install nyc@latest --no-save
+            fi
+            popd
+          fi  # Close the if statement here
+        done
+        npm config set package-lock true
```
ljharb, line 84: meaning that this isn't needed either - i don't use lockfiles in packages

.nycrc
```diff
@@ -14,6 +14,7 @@
 		"resolvers/*/test",
 		"scripts",
 		"memo-parser",
-		"lib"
+		"lib",
+		"utils"
```
ljharb, line 18: why would we want to exclude utils from coverage?

tests/src/core/resolve.js
```diff
@@ -113,7 +113,7 @@ describe('resolve', function () {
   });
 
   it('finds resolvers from the source files rather than eslint-module-utils', function () {
-    const testContext = utils.testContext({ 'import/resolver': { foo: {} } });
+    const testContext = utils.testContext({ 'import/resolver': { './foo-bar-resolver-v2': {} } });
```
ljharb, line 116: why these changes?

.eslintrc
joeyguerra: here are some samples:

```sh
(node:34323) DeprecationWarning: "first" rule is using `context.getDeclaredVariables()`, which is deprecated and will be removed in ESLint v9. Please use `sourceCode.getDeclaredVariables()` instead.
(Use `node --trace-deprecation ...` to show where the warning was created)

File: .github/workflows/native-wsl.yml (hunk 2 above)
Comment (joeyguerra) [reply .github/workflows/native-wsl.yml:38]: Yes. But this is more transparent, which makes the whole easier to understand.

Comment (joeyguerra) [reply .github/workflows/native-wsl.yml:38]: It won't work because nvm won't be available across runs. That's why the whole script is in a single `run` instead of broken out into multiple runs. It looks like a limitation in the Vampire/setup-wsl@v3 action.

File: .github/workflows/native-wsl.yml (hunk 3 above)
Comment (joeyguerra) [reply .github/workflows/native-wsl.yml:71]: I don't see it in the root level `.npmrc` file.

File: .nycrc (hunk 5 above)
Comment (joeyguerra) [reply .nycrc:18]: Some of the runs were blowing up on the typescript files. I'll remove it and push and see if it still happens.

File: tests/src/core/resolve.js (hunk 6 above)
Comment (joeyguerra) [reply tests/src/core/resolve.js:116]: It looks like the test was incorrectly setup. At one point during these changes, these tests were failing.

File: .github/workflows/native-wsl.yml (hunk 4 above)
Comment (joeyguerra) [reply .github/workflows/native-wsl.yml:84]: I believe what's happening here is that there's mixed/incorrect versions. Here's the error happening in `posttest`

```sh
Run npm run posttest
  npm run posttest
  shell: /usr/bin/bash -e {0}

> eslint-plugin-import@2.29.1 posttest
> eslint . && npm run update:eslint-docs -- --check && markdownlint "**/*.md"


/home/runner/work/eslint-plugin-import/eslint-plugin-import/tests/src/core/resolve.js
  116:95  error  A space is required before '}'  object-curly-spacing

  1 error and 0 warnings potentially fixable with the `--fix` option.

Error: Process completed with exit code 1.
```

Setting the `package-lock` to false prior to installing in this run fixes that error. I set it back to true just because I want to "reset" what is down at the beginning.

.nycrc
joeyguerra: Here's the error:

```sh
> eslint-plugin-import@2.29.1 tests-only /mnt/d/a/eslint-plugin-import/eslint-plugin-import
> npm run mocha tests/src


> eslint-plugin-import@2.29.1 mocha /mnt/d/a/eslint-plugin-import/eslint-plugin-import
> cross-env BABEL_ENV=test nyc mocha "tests/src"

/mnt/d/a/eslint-plugin-import/eslint-plugin-import/utils/declaredScope.d.ts: Unexpected token, expected ] (6:15)
```

joeyguerra: Here's the same error showing up in the [Appveyor pipeline](https://ci.appveyor.com/project/ljharb/eslint-plugin-import/builds/49398655/job/r9l74p4cfvb7gmat)

.github/workflows/native-wsl.yml
ljharb: are you sure? i use it across runs with that action on hundreds of repos.

.github/workflows/native-wsl.yml
ljharb: https://github.com/import-js/eslint-plugin-import/blob/c77c1a6ace311f99ed25878768c2dbee49c0512a/.npmrc#L1

.nycrc
ljharb: i think this is maybe because of a newer version of mocha defaulting to include `.ts` files? if so, maybe we can exclude them with mocha.opts?

tests/src/core/resolve.js
ljharb: they pass currently tho (just not on windows), so i'd expect them to continue to pass

.github/workflows/native-wsl.yml
ljharb: that's very confusing since it should be false for the entire repo, at all times :-/

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,175 @@
...
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    defaults:
+        run:
+          shell: ${{ matrix.configuration == 'wsl' && 'wsl-bash {0}' || 'pwsh' }}
+    strategy:
+      matrix:
+        os: [windows-latest]
+        node-version: [16, 14, 12, 10]
```
ljharb, line 14: where's node 8 and below?

.eslintrc
ljharb: those are just warnings, i don't think they'll fail the build?

.github/workflows/native-wsl.yml
joeyguerra: Yes. I was 100% sure. But now I'm questioning all of my life choices. Will you please send me some examples I can compare to?

.github/workflows/native-wsl.yml
joeyguerra: I forgot to add 8 back. But 6 and 4 were commented out in the `appveyor.yml`.

joeyguerra: What versions should be tested? why not include 20, 19, 18, 17, 11, and 9 as well?

tests/src/core/resolve.js
joeyguerra: I reverted the changes. I don't remember what configuration I was in, but these tests were failing and when I looked at the testContext code, it made sense to me why it was failing. The resulting require module name was something like `node_modules/eslint-import-utils-foo` and both the test description and assertion expect the path to be from the "source files". So I changed it and it passed. But now that I run the tests, I don't see a failure error in the logs.

.github/workflows/native-wsl.yml
joeyguerra: I tried using the `ljharb/actions/node/install@main` action again. Here's the [pipeline results](https://github.com/joeyguerra/eslint-plugin-import/actions/runs/8360381854/job/22885920314).

I think there's several parameters involved that are not working together correctly.

- The runner is a Windows runner. I think the `ljharb/actions/node/install@main` action is installing a node version for Windows, not Ubuntu
- The `Vampire/setup-wsl@v3` action is setting up WSL and installing Ubuntu
- Every `run` runs a new PowerShell (or cmd) "terminal" session and `Vampire/setup-wsl@v3` wraps the `run` scripts in a call to `wsl.BAT`, so previous `runs` that change "state" are not getting persisted for the next "session"

I'll tinker a bit more, but it's zapping my energy and motivation because I keep hitting different errors.

joeyguerra:
I've confirmed that separate runs using `Vampire/setup-wsl` are disconnected in that you can't use `npm` in one run when Node is installed in another one. In addition, `nvm` is unavailable in a run when using `ljharb/actions/node/install` in a previous run. Thus necessitating the whole script be done in 1 run for the WSL configuration.

I've also taken a different strategy to address the other problems (typically unexpected token errors). Which results in passing tests without having to ignore checks in the `utils/` code.

@ljharb would you mind taking another look at this PR?

.github/workflows/native-wsl.yml
ljharb: ahh, interesting, ok. certainly my personal action doesn't usually run on windows so it might not work properly.

.github/workflows/native-wsl.yml
ljharb: Ideally we should - at least the evens.

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,153 @@
...
+    - name: Do everything here because npm is not found when running in separate steps
+      if: matrix.configuration == 'wsl'
+      run: |
+        curl --version
+        export ESLINT_VERSION=7
+        export TRAVIS_NODE_VERSION=${{ matrix.node-version }}
+        curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/master/install.sh | bash
+        export NVM_DIR="$HOME/.nvm"
+        [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
+        [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
```
ljharb, line 31:
```suggestion
```
no need for shell completions in CI :-)

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,153 @@
...
+        nvm install ${{ matrix.node-version }}
+        nvm use ${{ matrix.node-version }}
+        nvm install-latest-npm
```
ljharb, line 34:
```suggestion
        nvm install --latest-npm ${{ matrix.node-version }}
```
install auto-uses.

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,153 @@
...
+
+        if [ ${{ matrix.node-version }} -ge 4 -a ${{ matrix.node-version }} -lt 6 ]; then
+          npm install eslint@4 --no-save --ignore-scripts
+          npm install
+          npm install eslint-import-resolver-typescript@1.0.2 --no-save
+          npm uninstall @angular-eslint/template-parser --no-save
+          npm uninstall @typescript-eslint/parser --no-save
```
ljharb, line 41:
```suggestion
          npm uninstall @angular-eslint/template-parser @typescript-eslint/parser --no-save
```

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,153 @@
...
+          npm install eslint-import-resolver-typescript@1.0.2 --no-save
+          npm uninstall @angular-eslint/template-parser --no-save
+          npm uninstall @typescript-eslint/parser --no-save 
+        fi
+        if [ ${{ matrix.node-version }} -ge 6 -a ${{ matrix.node-version }} -lt 7 ]; then
+          npm install eslint@5 --no-save --ignore-scripts
+          npm install
+          npm install eslint-import-resolver-typescript@1.0.2 --no-save
+          npm uninstall @angular-eslint/template-parser --no-save
+          npm install @typescript-eslint/parser@3 --no-save
```
ljharb, line 48:
```suggestion
          npm uninstall @angular-eslint/template-parser --no-save
          npm install eslint-import-resolver-typescript@1.0.2 @typescript-eslint/parser@3 --no-save 
```
assuming this still works?

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,153 @@
...
+        ESLINT_VERSION: 7
+        TRAVIS_NODE_VERSION: ${{ matrix.node-version }}
+      run: |
+        npm run copy-metafiles
+        bash ./tests/dep-time-travel.sh 2>&1
+    - name: Config git to allow symlinks
+      if: matrix.configuration == 'native'
+      run: |
+        git config core.symlinks true
+        git reset --hard
```
ljharb, line 109: should this be done before the checkout action?

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,153 @@
...
+        $env:RESOLVERS = [string]::Join(";", $resolvers);
+        foreach ($resolver in $resolvers) {
+            Set-Location -Path $resolver.Trim('"')
+            npm test
+            Set-Location -Path $PSScriptRoot
+        }
+    - name: Upload Codecov
+      uses: codecov/codecov-action@v4
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
```
ljharb, line 149:
```suggestion
      uses: codecov/codecov-action@v3.1.5
      with:
```
i want to stick with tokenless, and this is the only tokenless version that uses node20 instead of node16

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,153 @@
...
+        file: ./coverage/coverage-final.json # optional
+        flags: unittests # optional
+        name: codecov-umbrella # optional
+        fail_ci_if_error: true # optional (default = false)
```
ljharb, line 153:
```suggestion

```

.github/workflows/native-wsl.yml
```diff
@@ -0,0 +1,153 @@
...
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    defaults:
+        run:
+          shell: ${{ matrix.configuration == 'wsl' && 'wsl-bash {0}' || 'pwsh' }}
+    strategy:
+      matrix:
```
ljharb, line 12:
```suggestion
      fail-fast: false
      matrix:
```

ljharb (changes requested):
Thanks, if the tests pass then I'm happy with this approach - I can consolidate things into reusable scripts later :-)

.github/workflows/native-wsl.yml
joeyguerra: I don't know what its purpose is in the first place. I copied over from the `appveyor` pipeline. I'll remove it and rerun and "see if anything breaks".

ljharb (approved):

PR #2989, opened 2024-03-25
fix lint errors in export map builder

soryy708:
[`exportMapBuilder.js`](https://github.com/import-js/eslint-plugin-import/blob/38f8d25ef710747fe318d99b3f42e70bd1efc0f4/src/exportMapBuilder.js) currently has [ESLint "no-use-before-define" rule suppressed](https://github.com/import-js/eslint-plugin-import/blob/38f8d25ef710747fe318d99b3f42e70bd1efc0f4/.eslintrc#L232).

Lets fix the lint errors & un-suppress it for the file.

### DIFF
--- a/src/exportMapBuilder.js
+++ b/src/exportMapBuilder.js
@@ -118,6 +118,100 @@ const availableDocStyleParsers = {
 
 const supportedImportTypes = new Set(['ImportDefaultSpecifier', 'ImportNamespaceSpecifier']);
 
+let parserOptionsHash = '';
+let prevParserOptions = '';
+let settingsHash = '';
+let prevSettings = '';
+/**
+ * don't hold full context object in memory, just grab what we need.
+ * also calculate a cacheKey, where parts of the cacheKey hash are memoized
+ */
+function childContext(path, context) {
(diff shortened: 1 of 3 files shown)


src/exportMapBuilder.js
```diff
@@ -201,6 +322,19 @@ export default class ExportMapBuilder {
...
+    }
+
+    /**
+     * The creation of this closure is isolated from other scopes
+     * to avoid over-retention of unrelated variables, which has
+     * caused memory leaks. See #1266.
+     */
+    function thunkFor(p, context) {
+      return () => ExportMapBuilder.for(childContext(p, context));
+    }
```
ljharb, line 336: putting this in the class means it's redefined every time `parse` is called. I'm fine with an eslint override here rather than affecting runtime performance

soryy708: Ok.

- https://github.com/import-js/eslint-plugin-import/pull/2989/commits/4d276ef4a115c9bcc2fdc7f4d0488a2c3291b5d0

ljharb (approved):

Issue #2866, opened 2023-08-28
Error while loading rule 'import/no-unused-modules': Cannot read properties of undefined (reading 'get')


magic-m-johnson:
was closed here, but is still randomly occurring today:

https://github.com/import-js/eslint-plugin-import/issues/2388

ljharb:
Can you provide more info? The code being linted, or a repro repo?

iz-podpolja:
This occurs for us only when run through LSP, I cannot provide a full repro, but the following patch fixes it locally:
```diff
diff --git a/lib/rules/no-unused-modules.js b/lib/rules/no-unused-modules.js
index 098f72a5340a706e0652a272e96d473b6abc6051..d248367f5c4ab68d69806a762b8a90cdeb11e43b 100644
--- a/lib/rules/no-unused-modules.js
+++ b/lib/rules/no-unused-modules.js
@@ -525,7 +525,7 @@ module.exports = {
           exports = exportList.get(file);
 
           // special case: export * from
-          var exportAll = exports.get(EXPORT_ALL_DECLARATION);
+          var exportAll = exports && exports.get(EXPORT_ALL_DECLARATION);
           if (typeof exportAll !== 'undefined' && exportedValue !== IMPORT_DEFAULT_SPECIFIER) {
             if (exportAll.whereUsed.size > 0) {
               return;
@@ -533,7 +533,7 @@ module.exports = {
           }
 
           // special case: namespace import
-          var namespaceImports = exports.get(IMPORT_NAMESPACE_SPECIFIER);
+          var namespaceImports = exports && exports.get(IMPORT_NAMESPACE_SPECIFIER);
           if (typeof namespaceImports !== 'undefined') {
             if (namespaceImports.whereUsed.size > 0) {
               return;
@@ -543,7 +543,7 @@ module.exports = {
           // exportsList will always map any imported value of 'default' to 'ImportDefaultSpecifier'
           var exportsKey = exportedValue === DEFAULT ? IMPORT_DEFAULT_SPECIFIER : exportedValue;
 
-          var exportStatement = exports.get(exportsKey);
+          var exportStatement = exports && exports.get(exportsKey);
 
           var value = exportsKey === IMPORT_DEFAULT_SPECIFIER ? DEFAULT : exportsKey;
```

ljharb:
What's LSP?

Thanks, that helps; I'll take a look (altho I'm still confused why `exports` there would be falsy)

ljharb:
[...]

PR #2990, opened 2024-03-27
fix: check exports exist before access to it

headfire94:
while this issue exist https://github.com/import-js/eslint-plugin-import/issues/2866, this fix helps to prevent eslint from crashing.

### DIFF
--- a/src/rules/no-unused-modules.js
+++ b/src/rules/no-unused-modules.js
@@ -534,7 +534,7 @@ module.exports = {
       }
 
       // special case: export * from
-      const exportAll = exports.get(EXPORT_ALL_DECLARATION);
+      const exportAll = exports && exports.get(EXPORT_ALL_DECLARATION);
       if (typeof exportAll !== 'undefined' && exportedValue !== IMPORT_DEFAULT_SPECIFIER) {
         if (exportAll.whereUsed.size > 0) {
           return;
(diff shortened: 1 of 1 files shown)


src/rules/no-unused-modules.js
```diff
@@ -534,15 +534,15 @@ module.exports = {
       }
 
       // special case: export * from
-      const exportAll = exports.get(EXPORT_ALL_DECLARATION);
+      const exportAll = exports && exports.get(EXPORT_ALL_DECLARATION);
```
JounQin, line 537: Why not just return at https://github.com/import-js/eslint-plugin-import/pull/2990/files#diff-7d2908b010ec724c0f3a0d7615cb91f4673d765d52c6ba6607501e23470f2966R533

headfire94: I think we still need to report from line 567 to be shown:

```
context.report(
          node,
          `exported declaration '${value}' not used within other modules`,
        );
``` 
At the end

JounQin: The error console says the file has no exports then why it should report any export error? Or that just means an internal bug?

headfire94: The scenario is following:
1. create new file and define export it in
2. see eslint error that export is not used anywhere.

Without this PR we see eslint crash and no longer provide other checks and autofixes.
I assume this is internal bug mentioned in console warning, file *has* export, so this should be fixed somehow. till then this fix works as well

ljharb (changes requested):
We can’t fix a bug without a regression test. Can you add one?

headfire94:
@ljharb 

1. Use webstorm
2. create new js file and define variable and export it 'export default someVar'.
3. see eslint error that export is not used anywhere.

ljharb:
I understand the repro, but we need a *code* regression test.

headfire94:
Is there an example in this repo of regression test with dynamically added code to mirror for this case?

ljharb:
Not that I know of. Unfortunately this is a really difficult failure case to reproduce in tests.

PR #2993, opened 2024-03-28
Make it clear how to install the plugin

jwbth:
The markup in README.md is misleading, as it puts several alternatives into one block of code while not making it clear where the alternatives begin and end, forcing the reader to think hard about it. The pull request splits the block of code into three isolated ones, preceding each one with a comment and organizing them in a list.


ljharb (commented):
Thanks! Let's convert these to json instead of yaml, while we're at it.

ljharb (approved):
went ahead and rebased, and converted most of the yaml examples to eslintrc

PR #2997, opened 2024-04-07
ci: cancel in-progress runs when new changes are pushed up

G-Rath:
I personally think this is a good idea especially given this projects CI involves over 250+ jobs but I can understand if someone has an alternative workflow that makes it undesirable so feel free to close if you're not a fan 🙂 

Related docs: https://docs.github.com/en/actions/using-jobs/using-concurrency

### DIFF
--- a/.github/workflows/node-4+.yml
+++ b/.github/workflows/node-4+.yml
@@ -2,6 +2,10 @@ name: 'Tests: node.js'
 
 on: [pull_request, push]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 permissions:
   contents: read
 
--- a/.github/workflows/native-wsl.yml
+++ b/.github/workflows/native-wsl.yml
@@ -2,6 +2,10 @@ name: Native and WSL
 
 on: [push, pull_request]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   build:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -2,6 +2,10 @@ name: 'Tests: packages'
 
 on: [pull_request, push]
 
+concurrency:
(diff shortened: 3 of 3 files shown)


ljharb:
On branch pushes, it's critical that a new push not cancel an existing run. I'm fine with this change if it can be restricted to PR runs.

G-Rath:
hmm I'm not sure if that's possible, at least with `concurrency` - you should be able to include the trigger type in the grouping to ensure e.g. `pull_request` triggered workflows won't cancel `push` ones, but that would still mean in theory you could cancel a `push`-based workflow; though I'm not sure if that would actually be possible if its keyed to `ref` as well because that should change on every push? 🤔

Could you expand on why its critical that branch pushes are not cancelled by pushes to the same branch? is it because of the release process, or a way of working, or something different entirely?

ljharb:
Imagine i land two pushes to master/main and the second fails. If the first one is cancelled, how can i quickly know which one introduced the failure?

G-Rath:
pfft ok all the writing for nothing - [you can make `cancel-in-progress` an expression](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-on-specific-branches):

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ !contains(github.ref, 'release/')}}
```

ljharb:
I’d like it to only apply to PRs, and to not apply to *any* pushes to any branch, to be clear.

G-Rath:
@ljharb should be good to go now

.github/workflows/node-4+.yml
```diff
@@ -2,6 +2,11 @@ name: 'Tests: node.js'
 
 on: [pull_request, push]
 
+concurrency:
+  # Pushing new changes to a branch will cancel any in-progress CI runs of this workflow
```
ljharb, line 6: i think the comments need updating

ljharb (approved):
LGTM

.github/workflows/node-4+.yml
G-Rath: Mmhm yup good point - do you actually want to keep it at all?

ljharb: i'm fine removing the comment :-)

Issue #2937, opened 2023-12-11
improve no-cycle rule performance by leveraging strongly connected components


soryy708:
I see that the [`no-cycle`](https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/no-cycle.js) uses [`ExportMap`](https://github.com/import-js/eslint-plugin-import/blob/main/src/ExportMap.js) to get a cached dependency graph, and then run BFS to find cycles.

... Are we running BFS, on the entire dependency graph, for every file we lint?

I think we can improve performance in this way:

- When a file changes, run [Tarjan's SCC algorithm](https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm) to find SCCs in linear time
- To check if any two files have a cycle, check if they belong to the same SCC (can be very fast, depending on data structure)
- If the files are not in the same SCC, we don't need to search for cycles, because there can't be any
- To get the import chain to report the directed cycle, search only the SCC (as opposed to the entire dependency graph)

An [SCC](https://en.wikipedia.org/wiki/Strongly_connected_component) means that every node that belongs to it has a path to all other nodes that belong to it. Therefore if two nodes belong to the same SCC, they have a path to each other, therefore there is a directed cycle = circular dependency between them.
(The difference between an SCC and a directed cycle is that an SCC may contain at least one directed cycle)

By the way, I'm not sure that BFS is the right algorithm for cycle detection, consider DFS instead.

[Stackoverflow: Why DFS and not BFS for finding cycle in graphs](https://stackoverflow.com/questions/2869647/why-dfs-and-not-bfs-for-finding-cycle-in-graphs#:~:text=In%20DFS%20we%20take%20one,out%20if%20it%20has%20cycle.)

ljharb:
I have no attachment to any particular algorithm; as long as it passes tests and is robust and correct, whatever will perform best is great. I'd be happy to review a PR.

lukaselmer:
Sounds interesting, let's try it!

soryy708:
@benmosher I think you were the original contributor of the `no-cycle` rule?

soryy708:
The SCC optimization has been released, try v `>=2.30.0`

PR #2998, opened 2024-04-08
optimize no-cycle rule using strongly connected components

soryy708:
Lets make the `no-cycle` rule faster by not running many unnecessary BFSes.

For each dependency graph (aka `ExportMap`) we can run [Tarjan's SCC](https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm) once (which is a derivative of DFS = O(n))

That saves us a lot of work because we run a linear-complexity algorithm once, as opposed to for each linted file (which turned us O(n^2))

https://github.com/import-js/eslint-plugin-import/issues/2937

### DIFF
--- a/src/rules/no-cycle.js
+++ b/src/rules/no-cycle.js
@@ -5,6 +5,7 @@
 
 import resolve from 'eslint-module-utils/resolve';
 import ExportMapBuilder from '../exportMap/builder';
+import StronglyConnectedComponentsBuilder from '../scc';
 import { isExternalModule } from '../core/importType';
 import moduleVisitor, { makeOptionsSchema } from 'eslint-module-utils/moduleVisitor';
 import docsUrl from '../docsUrl';
@@ -47,6 +48,11 @@ module.exports = {
         type: 'boolean',
         default: false,
       },
+      disableScc: {
+        description: 'When true, don\'t calculate a strongly-connected-components graph. SCC is used to reduce the time-complexity of cycle detection, but adds overhead.',
+        type: 'boolean',
+        default: false,
+      },
     })],
   },
 
@@ -61,6 +67,8 @@ module.exports = {
       resolve(name, context),
       context,
     );
+
+    const scc = options.disableScc ? {} : StronglyConnectedComponentsBuilder.get(myPath, context);
 
     function checkSourceValue(sourceNode, importer) {
       if (ignoreModule(sourceNode.value)) {
@@ -98,6 +106,16 @@ module.exports = {
         return;  // no-self-import territory
       }
 
+      /* If we're in the same Strongly Connected Component,
+       * Then there exists a path from each node in the SCC to every other node in the SCC,
+       * Then there exists at least one path from them to us and from us to them,
+       * Then we have a cycle between us.
+       */
+      const hasDependencyCycle = options.disableScc || scc[myPath] === scc[imported.path];
+      if (!hasDependencyCycle) {
+        return;
+      }
+
       const untraversed = [{ mget: () => imported, route: [] }];
       function detectCycle({ mget, route }) {
         const m = mget();
@@ -106,6 +124,9 @@ module.exports = {
         traversed.add(m.path);
 
         for (const [path, { getter, declarations }] of m.imports) {
+          // If we're in different SCCs, we can't have a circular dependency
+          if (!options.disableScc && scc[myPath] !== scc[path]) { continue; }
+
           if (traversed.has(path)) { continue; }
           const toTraverse = [...declarations].filter(({ source, isOnlyImportingTypes }) => !ignoreModule(source.value)
             // Ignore only type imports
--- a/src/scc.js
+++ b/src/scc.js
@@ -1,0 +1,86 @@
+import calculateScc from '@rtsao/scc';
+import { hashObject } from 'eslint-module-utils/hash';
+import resolve from 'eslint-module-utils/resolve';
+import ExportMapBuilder from './exportMap/builder';
+import childContext from './exportMap/childContext';
+
+let cache = new Map();
+
+export default class StronglyConnectedComponentsBuilder {
+  static clearCache() {
+    cache = new Map();
+  }
+
+  static get(source, context) {
+    const path = resolve(source, context);
+    if (path == null) { return null; }
+    return StronglyConnectedComponentsBuilder.for(childContext(path, context));
+  }
+
+  static for(context) {
+    const cacheKey = context.cacheKey || hashObject(context).digest('hex');
+    if (cache.has(cacheKey)) {
+      return cache.get(cacheKey);
+    }
+    const scc = StronglyConnectedComponentsBuilder.calculate(context);
+    cache.set(cacheKey, scc);
+    return scc;
+  }
+
+  static calculate(context) {
+    const exportMap = ExportMapBuilder.for(context);
+    const adjacencyList = this.exportMapToAdjacencyList(exportMap);
+    const calculatedScc = calculateScc(adjacencyList);
+    return StronglyConnectedComponentsBuilder.calculatedSccToPlainObject(calculatedScc);
+  }
+
+  /** @returns {Map<string, Set<string>>} for each dep, what are its direct deps */
+  static exportMapToAdjacencyList(initialExportMap) {
+    const adjacencyList = new Map();
+    // BFS
+    function visitNode(exportMap) {
+      if (!exportMap) {
+        return;
+      }
+      exportMap.imports.forEach((v, importedPath) => {
+        const from = exportMap.path;
+        const to = importedPath;
+
+        // Ignore type-only imports, because we care only about SCCs of value imports
+        const toTraverse = [...v.declarations].filter(({ isOnlyImportingTypes }) => !isOnlyImportingTypes);
+        if (toTraverse.length === 0) { return; }
+
+        if (!adjacencyList.has(from)) {
+          adjacencyList.set(from, new Set());
+        }
+
+        if (adjacencyList.get(from).has(to)) {
+          return; // prevent endless loop
+        }
+        adjacencyList.get(from).add(to);
+        visitNode(v.getter());
+      });
+    }
+    visitNode(initialExportMap);
+    // Fill gaps
+    adjacencyList.forEach((values) => {
+      values.forEach((value) => {
+        if (!adjacencyList.has(value)) {
+          adjacencyList.set(value, new Set());
+        }
+      });
+    });
+    return adjacencyList;
+  }
+
+  /** @returns {Record<string, number>} for each key, its SCC's index */
+  static calculatedSccToPlainObject(sccs) {
+    const obj = {};
+    sccs.forEach((scc, index) => {
+      scc.forEach((node) => {
+        obj[node] = index;
+      });
+    });
+    return obj;
+  }
+}
--- a/tests/src/rules/no-cycle.js
+++ b/tests/src/rules/no-cycle.js
@@ -17,7 +17,7 @@ const testVersion = (specifier, t) => _testVersion(specifier, () => Object.assig
 
 const testDialects = ['es6'];
 
-ruleTester.run('no-cycle', rule, {
+const cases = {
   valid: [].concat(
     // this rule doesn't care if the cycle length is 0
     test({ code: 'import foo from "./foo.js"' }),
@@ -290,4 +290,30 @@ ruleTester.run('no-cycle', rule, {
       ],
     }),
   ),
+};
+
+ruleTester.run('no-cycle', rule, {
+  valid: flatMap(cases.valid, (testCase) => [
+    testCase,
+    {
+      ...testCase,
+      code: `${testCase.code} // disableScc=true`,
+      options: [{
+        ...testCase.options && testCase.options[0] || {},
+        disableScc: true,
+      }],
+    },
+  ]),
+
+  invalid: flatMap(cases.invalid, (testCase) => [
+    testCase,
+    {
+      ...testCase,
+      code: `${testCase.code} // disableScc=true`,
+      options: [{
+        ...testCase.options && testCase.options[0] || {},
+        disableScc: true,
+      }],
+    },
+  ]),
 });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ### Fixed
 - [`no-extraneous-dependencies`]: allow wrong path ([#3012], thanks [@chabb])
+- [`no-cycle`]: use scc algorithm to optimize ([#2998], thanks [@soryy708])
 
 ### Changed
 - [Docs] `no-extraneous-dependencies`: Make glob pattern description more explicit ([#2944], thanks [@mulztob])
@@ -1123,6 +1124,7 @@ for info on changes for earlier releases.
 [#3012]: https://github.com/import-js/eslint-plugin-import/pull/3012
 [#3011]: https://github.com/import-js/eslint-plugin-import/pull/3011
 [#3004]: https://github.com/import-js/eslint-plugin-import/pull/3004
+[#2998]: https://github.com/import-js/eslint-plugin-import/pull/2998
 [#2991]: https://github.com/import-js/eslint-plugin-import/pull/2991
 [#2989]: https://github.com/import-js/eslint-plugin-import/pull/2989
 [#2987]: https://github.com/import-js/eslint-plugin-import/pull/2987
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8"
   },
   "dependencies": {
+    "@rtsao/scc": "^1.1.0",
     "array-includes": "^3.1.8",
     "array.prototype.findlastindex": "^1.2.5",
     "array.prototype.flat": "^1.3.2",
--- a/tests/src/scc.js
+++ b/tests/src/scc.js
@@ -1,0 +1,179 @@
+import sinon from 'sinon';
+import { expect } from 'chai';
+import StronglyConnectedComponentsBuilder from '../../src/scc';
+import ExportMapBuilder from '../../src/exportMap/builder';
+
+function exportMapFixtureBuilder(path, imports, isOnlyImportingTypes = false) {
+  return {
+    path,
+    imports: new Map(imports.map((imp) => [imp.path, { getter: () => imp, declarations: [{ isOnlyImportingTypes }] }])),
+  };
+}
+
+describe('Strongly Connected Components Builder', () => {
+  afterEach(() => ExportMapBuilder.for.restore());
+  afterEach(() => StronglyConnectedComponentsBuilder.clearCache());
+
+  describe('When getting an SCC', () => {
+    const source = '';
+    const context = {
+      settings: {},
+      parserOptions: {},
+      parserPath: '',
+    };
+
+    describe('Given two files', () => {
+      describe('When they don\'t value-cycle', () => {
+        it('Should return foreign SCCs', () => {
+          sinon.stub(ExportMapBuilder, 'for').returns(
+            exportMapFixtureBuilder('foo.js', [exportMapFixtureBuilder('bar.js', [])]),
+          );
+          const actual = StronglyConnectedComponentsBuilder.for(source, context);
+          expect(actual).to.deep.equal({ 'foo.js': 1, 'bar.js': 0 });
+        });
+      });
+
+      describe('When they do value-cycle', () => {
+        it('Should return same SCC', () => {
+          sinon.stub(ExportMapBuilder, 'for').returns(
+            exportMapFixtureBuilder('foo.js', [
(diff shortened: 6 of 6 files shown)


soryy708:
Why is `socket-security` failing this?
The changes in the dependencies it found don't make sense.
The library I added `@rtsao/scc` has no dependencies at all.
Here's their `package.json`: https://github.com/rtsao/scc/blob/1120edc92040b0ca748ab62a882a98565c85deed/package.json

ljharb:
Good question. I think it's that this is the first new PR it's been run on since i installed it?

src/rules/no-cycle.js
```diff
@@ -27,6 +22,7 @@ module.exports = {
       maxDepth: {
         anyOf: [
           {
+            // This is a nop, but we can keep it for backwards-compatibility
```
ljharb, line 25: it needs to not be a no-op - if there's a flaw at a depth greater than this value, we can't report it.

tests/src/rules/no-cycle.js
```diff
@@ -178,69 +178,69 @@ ruleTester.run('no-cycle', rule, {
...
       test({
         code: `import("./${testDialect}/depth-three-star")`,
-        errors: [error(`Dependency cycle via ./depth-two:1=>./depth-one:1`)],
+        errors: [error(`Dependency cycle detected.`)],
         parser: parsers.BABEL_OLD,
       }),
       test({
         code: `import("./${testDialect}/depth-three-indirect")`,
-        errors: [error(`Dependency cycle via ./depth-two:1=>./depth-one:1`)],
+        errors: [error(`Dependency cycle detected.`)],
```
ljharb, line 232: hm, this seems like a pretty big expressivity loss in the error messages. is there no way to get this info from the SCC graph?

src/rules/no-cycle.js
soryy708: Define "flaw"?

ljharb: like if there's a cycle beyond that depth, we currently don't report it, and we need to continue not reporting it

tests/src/rules/no-cycle.js
soryy708: The reported path is the shortest. We don't have that in the SCC.
In the SCC we have all of the files that have a path to each other, which may contain multiple cycles.

To remain backwards compatible, we need to run BFS to find the shortest cycle. That will have a performance cost, but we're still going to be more efficient than without SCC, since we only need to do this BFS if we know that there's a cycle. No BFS if there's no cycle.

ljharb: Could we just pick one of the paths? or, include them all in the message?

src/rules/no-cycle.js
soryy708: If we insist on backwards-compatibility, yes.

But why would users want to ignore dependency cycles that are very big? I think this was originally supposed to be a performance workaround to prevent `no-cycle` from taking a very long time in large code-bases?

tests/src/rules/no-cycle.js
ljharb: "Different from the current message" is fine, if it's better :-)

The particular case I'm thinking about is when the cycle is a much, much larger circle - when a seemingly innocuous change ends up with a cycle far away from you, and it's good if the message can point you to the source of the problem.

src/rules/no-cycle.js
ljharb: We do; breaking changes are harmful.

I agree it exists as a perf workaround and that ideally nobody should need to use it after this PR ships, which is great! but it could also be used to silence cycle warnings in dependencies, or unrelated parts of the codebase.

tests/src/rules/no-cycle.js
soryy708: I think the best UX will be to keep the shortest path in the error message as it was, though it will have a perf cost due to the BFS as discussed.

Wdyt about adding a new config option to let the user decide if they prefer expressivity/precision vs speed?

ljharb: I suppose we could do that - but i'm not sure the performance cost of a failure case matters? when there's no cycles, it would be fast - it'd only be when there is a cycle that there'd be a cost to compute the shortest path

src/rules/no-cycle.js
soryy708: 👍🏻

tests/src/rules/no-cycle.js
soryy708: Here I add a new option to skip the cycle-route BFS:

- https://github.com/import-js/eslint-plugin-import/pull/2998/commits/ebc2e80f4dd45d4f2592fede491bcf173014e093

Feel free to propose a better name.

Note that this lacks test coverage, and when on will definitely report cycles that it wouldn't when off.

ljharb: maybe i'm confused.

I'm saying that we should always do the fast approach. Then, only if there's a cycle detected, we can do the slow thing to come up with a helpful error message.

soryy708: Right. The way I measured the 323% increase is:

`"rules": ["import/no-cycle", { noRoute: true  }]` -> 1152 ms
`"rules": ["import/no-cycle", { noRoute: false }]` -> 4876 ms

ljharb: but in the case where there are no cycles, the perf is the same, yes?

soryy708: Yes, when there is no cycle, we don't have any additional work to do, so we take the exact same code path (early exit at https://github.com/import-js/eslint-plugin-import/pull/2998/files#diff-4b652cd5e29e44f86a5f087e0f8eec94bd9b6ef723d86dccd60656a044899bccR116).

src/scc.js
```diff
@@ -0,0 +1,91 @@
...
+let exportMapBuilder = ExportMapBuilder;
+
+export default class StronglyConnectedComponentsBuilder {
+  static injectExportMapBuilder(value) {
+    exportMapBuilder = value;
+  }
+
+  static resetExportMapBuilder() {
+    exportMapBuilder = ExportMapBuilder;
+  }
```
ljharb, line 17: is there a better way to do this than manual injection methods? even a public property mocked with sinon would be preferable.

src/scc.js
```diff
@@ -0,0 +1,91 @@
...
+    function visitNode(exportMap) {
+      if (!exportMap) {
+        return;
+      }
+      exportMap.imports.forEach((v, importedPath) => {
+        const from = exportMap.path;
+        const to = importedPath;
+
+        if (!adjacencyList.has(from)) {
+          adjacencyList.set(from, new Set());
```
ljharb, line 59: would it be faster to defer creation of these new sets until they're needed?

tests/src/rules/no-cycle.js
```diff
@@ -290,4 +290,25 @@ ruleTester.run('no-cycle', rule, {
...
+
+ruleTester.run('no-cycle', rule, {
+  valid: cases.valid.map((testCase) => [testCase, {
+    ...testCase,
+    code: `${testCase.code} // skipErrorMessagePath=true`,
+    options: [{
+      ...testCase.options && testCase.options[0] || {},
+      skipErrorMessagePath: true,
+    }],
+  }]).flat(1),
```
ljharb, line 303: we can't rely on `.flat` existing, but i think you csn do this with `[].concat()` instead

tests/src/rules/no-cycle.js
```diff
@@ -290,4 +290,25 @@ ruleTester.run('no-cycle', rule, {
...
+
+  invalid: cases.invalid.map((testCase) => [testCase, {
+    ...testCase,
+    code: `${testCase.code} // skipErrorMessagePath=true`,
+    options: [{
+      ...testCase.options && testCase.options[0] || {},
+      skipErrorMessagePath: true,
+    }],
+    errors: [error(`Dependency cycle detected.`)],
+  }]).flat(1),
```
ljharb, line 313: same here

src/scc.js
soryy708: What exactly would you like made better?
How is overriding a public property with a mock better?
Dependency injection in the constructor is a well known design pattern, and it's good to not surprise the reader.

src/scc.js
soryy708: They are deferred already.
`exportMapToAdjacencyList` is called from `calculate` which is called from `for` which is called from `get`, which is called from `rules/no-cycle`.
We avoid doing unnecessary work by `for` being memoized (leveraging caching), and when we're done with the dependency graph we need these sets immediately to run the SCC algorithm.

tests/src/rules/no-cycle.js
soryy708: In production code, sure - but this is test code.
I can replace it with `[].concat()` manually but I believe it's better to use a polyfill for this.
If we don't go with polyfill then we ought to add an ESLint rule to error when using `Array.flat`.

src/scc.js
ljharb: It's not an idiomatic pattern in JS, though, and adding it's better because adding extra code just for tests is a huge antipattern.

tests/src/rules/no-cycle.js
ljharb: installing a polyfill would invalidate the tests, so that's not a good idea.
we can import `array.prototype.flat`, certainly, but `[].concat` is pretty idiomatic - either way is fine.

it's not possible to add an eslint rule that can reliably warn against instance method usage without type information.

src/scc.js
soryy708: > adding extra code just for tests is a huge antipattern.

That's a surprising statement. I'd like to present an opposing view, where adding extra code just for tests is a good thing, explained well by Robert C Martin in his book Clean Architecture on pages 251-252:

> The extreme isolation of tests, combined with the fact that they are not usually deployed, often causes developers to think that tests fall outside of the design of the system. This is a catastrophic point of view. Tests that are not well integrated into the design of the system tend to be fragile, and they make the system rigid and difficult to change.
> The solution is to design for testability. Don't depend on volatile things.
> The way to accomplish this goal is to create a specific API that the tests can use [...]
> The purpose of the testing API is to decouple the tests from the application. The goal is to decouple the structure of the tests from the structure of the application.

tests/src/rules/no-cycle.js
soryy708: > it's not possible to add an eslint rule that can reliably warn against instance method usage without type information

False-positives would be possible in theory, you're correct that it won't be reliable. But how bad would it be practically speaking?

src/scc.js
ljharb: https://qntm.org/clean might be a helpful read.

tests/src/rules/no-cycle.js
ljharb: It'd add complication without much value; the tests will already fail as-is due to the function being unavailable.

src/scc.js
soryy708: Clean Code is a different book.

Does the blog you linked to say anything about adding extra code just for tests?

ljharb: That's totally a fair point - in that both approaches are fully globally mutable - but as a property, there's no code that goes unexecuted in production, which is still better than the current situation.

A tight coupling imo is preferable to test-only production code (neither are ideal).

soryy708: Got rid of injection by stubbing the imported dependency using sinon:

- https://github.com/import-js/eslint-plugin-import/pull/2998/commits/f64a1232edb46f12725cc466b0e4eb4f1237ec7a

tests/src/rules/no-cycle.js
soryy708: replaced `array.flat()` with `[].concat(...array)`

- https://github.com/import-js/eslint-plugin-import/pull/2998/commits/a6b26c461e6f14b40257aaa769128aeb3203c9f3

tests/src/rules/no-cycle.js
soryy708: replaced `array.flat()` with `[].concat(...array)`

- https://github.com/import-js/eslint-plugin-import/pull/2998/commits/a6b26c461e6f14b40257aaa769128aeb3203c9f3

src/rules/no-cycle.js
```diff
@@ -62,6 +68,8 @@ module.exports = {
       context,
     );
 
+    const scc = StronglyConnectedComponentsBuilder.get(myPath, context);
```
kuoruan, line 71: Can this row and line 114-117 move after line 119?
when skipErrorMessagePath is false, there is no need to do scc check.

ljharb: the scc check is how we know if there's cycles or not.

src/scc.js
```diff
@@ -0,0 +1,82 @@
...
+import { hashObject } from 'eslint-module-utils/hash';
+import resolve from 'eslint-module-utils/resolve';
+import ExportMapBuilder from './exportMap/builder';
+import childContext from './exportMap/childContext';
+
+let cache = new Map();
+
+export default class StronglyConnectedComponentsBuilder {
+  static clearCache() {
+    cache = new Map();
```
ljharb, line 11:
```suggestion
    cache.clear();
```
this might be less GC intensive? i don't feel strongly tho

tests/src/rules/no-cycle.js
```diff
@@ -290,4 +290,25 @@ ruleTester.run('no-cycle', rule, {
       ],
     }),
   ),
+};
+
+ruleTester.run('no-cycle', rule, {
+  valid: [].concat(...cases.valid.map((testCase) => [testCase, {
```
ljharb, line 296: concat autoflattens arrays, so the spread isn't needed - which makes it much simpler and faster :-)
```suggestion
  valid: [].concat(cases.valid.map((testCase) => [testCase, {
```

tests/src/rules/no-cycle.js
```diff
@@ -290,4 +290,25 @@ ruleTester.run('no-cycle', rule, {
...
+    ...testCase,
+    code: `${testCase.code} // skipErrorMessagePath=true`,
+    options: [{
+      ...testCase.options && testCase.options[0] || {},
+      skipErrorMessagePath: true,
+    }],
+  }])),
+
+  invalid: [].concat(...cases.invalid.map((testCase) => [testCase, {
```
ljharb, line 305:
```suggestion
  invalid: [].concat(cases.invalid.map((testCase) => [testCase, {
```

ljharb (commented):
Looks great! I'll look at failing tests next

tests/src/rules/no-cycle.js
soryy708: I tried without spread and the tests didn't work for me?

src/scc.js
soryy708: Maybe. `clearCache` is currently used only by tests, so I think it doesn't matter.

tests/src/rules/no-cycle.js
ljharb: oh hm, yeah i think it's trickier than that - keep the spreads for now.

pumano:
would love to see that optimization!

jonahallibone:
Can't wait for this to be merged!

pumano:
@ljharb any news on this bro?

soryy708:
There are 2 pathologies with SCC right now:

- It keeps track of type imports, but should only keep value imports. This prevents us from exiting early in no-cycle when the cycle is types only.
- It traverses ignored modules (`isExternalModule`), which slows it down when there are many node_modules.

src/scc.js
soryy708: How do we move forward with this? Do we make the change or do we resolve this conversation as-is?

src/rules/no-cycle.js
soryy708: How do we move forward with this? Can we resolve this conversation?

soryy708:
There's a pathology with SCC right now:

- It traverses ignored modules (isExternalModule), which slows it down

ljharb (approved):
I see two uncovered lines - can we add test cases to cover them?

Otherwise, this LGTM assuming all existing tests pass!

soryy708:
About the uncovered lines,

I see them in codecov. The 1st is early return marked "no-self-import territory", so unrelated to my changes.

The other is early return when files don't have the same SCC.
Unfortunately it can never happen in the tests since ExportMap is null for some reason, so both files are in an undefined scc.

src/scc.js
```diff
@@ -0,0 +1,86 @@
...
+        // Ignore type-only imports, because we care only about SCCs of value imports
+        const toTraverse = [...v.declarations].filter(({ isOnlyImportingTypes }) => !isOnlyImportingTypes);
+        if (toTraverse.length === 0) { return; }
+
+        if (!adjacencyList.has(from)) {
+          adjacencyList.set(from, new Set());
+        }
+
+        if (adjacencyList.get(from).has(to)) {
+          return; // prevent endless loop
```
ljharb, line 58: if this line can’t happen, there’s no loop to prevent, so should this check be removed?

soryy708: The fact that it shouldn't happen doesn't mean that it can't.
I don't remember why I put it in here, either I actually experienced an endless loop (in which case it can happen), or it protects against regressions on code mutation

soryy708:
After all these changes, I decided to measure performance impact.
I ran lint on a big TypeScript codebase with many circular dependencies, and looked at total command execution time (as reported by yarn).

Without SCC (disableScc: true):

367 seconds (6:07 minutes)

With SCC:

133 seconds (2:13 minutes)

(I made sure it isn't using eslint cache)

So we're talking about a 63% decrease in run time.

JounQin:
Notice that we discovered performance downgrade at https://github.com/un-ts/eslint-plugin-import-x/issues/113 when we ported this "optimization", I don't quite sure whether there were some improvements after we ported.

cc @SukkaW

soryy708:
> Notice that we discovered performance downgrade at [un-ts#113](https://github.com/un-ts/eslint-plugin-import-x/issues/113) when we ported this "optimization", I don't quite sure whether there were some improvements after we ported.
> 
> cc @SukkaW

Thanks, I have already seen this. What I did about it from when it was ported to the community fork:

- Simplified things by ripping out "skip error message path" option. This means less things can go wrong
- Changed SCC to differentiate between value and type imports
- Changed BFS to look at SCC
- Added option to disable SCC via config

In summary, this PR diverged significantly from the port discussed, in a way which should improve performance, invalidating prior benchmarks. A workaround has been built in to fall-back on in case of performance regression, without rolling-back.

SukkaW:
> > Notice that we discovered performance downgrade at [un-ts#113](https://github.com/un-ts/eslint-plugin-import-x/issues/113) when we ported this "optimization", I don't quite sure whether there were some improvements after we ported.
> > cc @SukkaW
> 
> Thanks, I have already seen this. What I did about it from when it was ported to the community fork:
> 
> * Simplified things by ripping out "skip error message path" option. This means less things can go wrong
> * Changed SCC to differentiate between value and type imports
> * Changed BFS to look at SCC
> * Added option to disable SCC via config
> 
> In summary, this PR diverged significantly from the port discussed, in a way which should improve performance, invalidating prior benchmarks. A workaround has been built in to fall-back on in case of performance regression, without rolling-back.

It was only after backporting the PR that I realized this was still an early experiment and a PoC.

I'm considering using [`@newdash/graphlib`](https://www.npmjs.com/package/@newdash/graphlib) (which powers [`cycle-import-check`](https://www.npmjs.com/package/cycle-import-check)) to reimplement the `no-cycle` rule. With an existing graph library, we can effortlessly identify any "circle" within the module graph. However, without appending extra information, we lose track of the AST node.

src/rules/no-cycle.js
```diff
@@ -98,6 +106,16 @@ module.exports = {
         return;  // no-self-import territory
       }
 
+      /* If we're in the same Strongly Connected Component,
+       * Then there exists a path from each node in the SCC to every other node in the SCC,
+       * Then there exists at least one path from them to us and from us to them,
+       * Then we have a cycle between us.
+       */
+      const hasDependencyCycle = scc[myPath] === scc[imported.path];
```
ljharb, line 114: when `disableScc` is true, won't this always be `false`, meaning it will always early-return on the next line?

soryy708: When `disableScc` is `true`, `scc` will be `{}`, so `scc[whatever]` will be `undefined`.
`undefined === undefined` is always true.

ljharb: even if the path is `'toString'`?

soryy708: Fine. https://github.com/import-js/eslint-plugin-import/pull/2998/commits/9cf7dbc25dc4d85ad154b3cec0542843134f2102

src/scc.js
ljharb: maybe we could add a console.warn or something, that logs the code that causes it, so people can file an issue?

soryy708: Reading it again, I think this is just the exit condition for the recursive node visiting (BFS). We have a similar condition in https://github.com/import-js/eslint-plugin-import/blob/19dbc33ecf09c774db35f362b05caaec027d2e18/src/rules/no-cycle.js#L109

ljharb (approved):
awesome, LGTM once tests pass

soryy708:
Should be included in v 2.30.0

PR #3004, opened 2024-04-24
[dynamic-import-chunkname] Disallow chunk name when eager mode is enabled

amsardesai:
The dynamic-import-chunkname rule is useful for catching missing chunk names for imports, however much of our imports use `webpackMode: "eager"` to disable creating a separate chunk (i.e for lazy evaluation). Adding `webpackChunkName` anyway can force webpack to generate a chunk, defeating the purpose of the eager mode. This PR changes the rule to disallow chunk names when `webpackMode: "eager"` is set.

Related issue: https://github.com/import-js/eslint-plugin-import/issues/1904

### DIFF
--- a/src/rules/dynamic-import-chunkname.js
+++ b/src/rules/dynamic-import-chunkname.js
@@ -27,6 +27,7 @@ module.exports = {
         },
       },
     }],
+    hasSuggestions: true,
   },
 
   create(context) {
@@ -36,8 +37,10 @@ module.exports = {
 
     const paddedCommentRegex = /^ (\S[\s\S]+\S) $/;
     const commentStyleRegex = /^( ((webpackChunkName: .+)|((webpackPrefetch|webpackPreload): (true|false|-?[0-9]+))|(webpackIgnore: (true|false))|((webpackInclude|webpackExclude): \/.*\/)|(webpackMode: ["'](lazy|lazy-once|eager|weak)["'])|(webpackExports: (['"]\w+['"]|\[(['"]\w+['"], *)+(['"]\w+['"]*)\]))),?)+ $/;
-    const chunkSubstrFormat = ` webpackChunkName: ["']${webpackChunknameFormat}["'],? `;
+    const chunkSubstrFormat = `webpackChunkName: ["']${webpackChunknameFormat}["'],? `;
     const chunkSubstrRegex = new RegExp(chunkSubstrFormat);
+    const eagerModeFormat = `webpackMode: ["']eager["'],? `;
+    const eagerModeRegex = new RegExp(eagerModeFormat);
 
     function run(node, arg) {
       const sourceCode = context.getSourceCode();
@@ -54,6 +57,7 @@ module.exports = {
       }
 
       let isChunknamePresent = false;
+      let isEagerModePresent = false;
 
       for (const comment of leadingComments) {
         if (comment.type !== 'Block') {
@@ -92,12 +96,55 @@ module.exports = {
           return;
         }
 
+        if (eagerModeRegex.test(comment.value)) {
+          isEagerModePresent = true;
+        }
+
         if (chunkSubstrRegex.test(comment.value)) {
           isChunknamePresent = true;
         }
       }
 
-      if (!isChunknamePresent && !allowEmpty) {
+      if (isChunknamePresent && isEagerModePresent) {
+        context.report({
+          node,
+          message: 'dynamic imports using eager mode do not need a webpackChunkName',
+          suggest: [
+            {
+              desc: 'Remove webpackChunkName',
+              fix(fixer) {
+                for (const comment of leadingComments) {
+                  if (chunkSubstrRegex.test(comment.value)) {
+                    const replacement = comment.value.replace(chunkSubstrRegex, '').trim().replace(/,$/, '');
+                    if (replacement === '') {
+                      return fixer.remove(comment);
+                    } else {
+                      return fixer.replaceText(comment, `/* ${replacement} */`);
+                    }
+                  }
+                }
+              },
+            },
+            {
+              desc: 'Remove webpackMode',
+              fix(fixer) {
+                for (const comment of leadingComments) {
+                  if (eagerModeRegex.test(comment.value)) {
+                    const replacement = comment.value.replace(eagerModeRegex, '').trim().replace(/,$/, '');
+                    if (replacement === '') {
+                      return fixer.remove(comment);
+                    } else {
+                      return fixer.replaceText(comment, `/* ${replacement} */`);
+                    }
+                  }
+                }
+              },
+            },
+          ],
+        });
+      }
+
+      if (!isChunknamePresent && !allowEmpty && !isEagerModePresent) {
         context.report({
           node,
           message:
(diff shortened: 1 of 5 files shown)


ljharb:
When would you not want this option enabled?

amsardesai:
Never, really. I'll just make it the default.

ljharb:
Is there anyone who this could break? I assume not, because the only complaint would be that it _fails_ to warn on something?

amsardesai:
Exactly, this only suppresses existing warnings in a narrow case, and doesn't introduce new warnings.

ljharb (approved):
Thanks, this is great!

One more enhancement we may want to consider before landing this: when webpackMode is eager, and a useless name is present, could we give a suggestion to remove the name?

amsardesai:
@ljharb Yeah that makes sense to me, just updated the PR to warn when chunk name and eager mode are both set.

src/rules/dynamic-import-chunkname.js
```diff
@@ -92,12 +94,24 @@ module.exports = {
...
       }
 
-      if (!isChunknamePresent && !allowEmpty) {
+      if (isChunknamePresent && isEagerModePresent) {
+        context.report({
+          node,
+          message:
+            `dynamic imports should not have both webpackChunkName and webpackMode: "eager"`,
+        });
+      }
```
ljharb, line 112:
```suggestion
      if (isChunknamePresent && isEagerModePresent) {
        context.report({
          node,
          message: 'dynamic imports should not have both webpackChunkName and webpackMode: "eager"`\',
        });
      }
```

also is there any reason why someone might still want a chunk name?

amsardesai: I don't think having a chunk name with eager mode makes sense, since eager mode explicitly disables creating a new chunk, so the module is just lazily evaluated instead of lazily loaded+evaluated. Adding a chunk name anyways basically reverts back to being `webpackMode: 'lazy'` and generates a new chunk. A warning would help users catch unintentional additional chunks when using `webpackMode: 'eager'`

ljharb: ok, so having both is always a mistake. thanks!

ljharb: altho actually - that means that removing the name would quietly change the mode from lazy to eager - i think this should be a suggestion to *either* remove the name, *or* remove/change the webpackMode, to be safe.

ljharb:
@amsardesai sorry if i wasn't clear; a "suggestion" isn't prose giving the user a hint, it's an actual eslint feature where the editor will prompt them with multiple autofixes they have to explicitly choose. For an example, look for a rule with `hasSuggestions: true` in its schema.

amsardesai:
Ah, I didn't realize you meant fixers too. Let me do that.

src/rules/dynamic-import-chunkname.js
```diff
@@ -36,8 +37,10 @@ module.exports = {
 
     const paddedCommentRegex = /^ (\S[\s\S]+\S) $/;
     const commentStyleRegex = /^( ((webpackChunkName: .+)|((webpackPrefetch|webpackPreload): (true|false|-?[0-9]+))|(webpackIgnore: (true|false))|((webpackInclude|webpackExclude): \/.*\/)|(webpackMode: ["'](lazy|lazy-once|eager|weak)["'])|(webpackExports: (['"]\w+['"]|\[(['"]\w+['"], *)+(['"]\w+['"]*)\]))),?)+ $/;
-    const chunkSubstrFormat = ` webpackChunkName: ["']${webpackChunknameFormat}["'],? `;
+    const chunkSubstrFormat = `webpackChunkName: ["']${webpackChunknameFormat}["'],? `;
```
ljharb, line 40: how come this one no longer has a leading space, but still has a trailing space?

amsardesai: I'm using this regex in a `replace()` operation to remove it, and I didn't want the preceding space to be removed in a comma-separated list of webpack options, it could lead to the format being incorrect after applying the suggestion (i.e there could be no space after comma). The existence of the space is already checked by the lint rule using the regex right above this.

amsardesai:
Hey @ljharb, would love to know if this can be merged now, or if any other changes need to be made!

ljharb:
@amsardesai ok, i've rebased this and it's probably ready to go :-) just wondering about some of the remaining test _changes_ (not the additions, those make sense) - some are lazy → eager, and some are eager → lazy. Can you go through and comment in the PR on those lines, and explain what motivated the change? i suspect you have a convincing explanation for them all but this way it'll be documented for me in the future :-)

tests/src/rules/dynamic-import-chunkname.js
```diff
@@ -354,7 +355,6 @@ ruleTester.run('dynamic-import-chunkname', rule, {
     },
     {
       code: `import(
-        /* webpackChunkName: "someModule" */
         /* webpackMode: "eager" */
```
amsardesai, line 358: Removed the chunk name since it's not allowed anymore when eager mode is set

tests/src/rules/dynamic-import-chunkname.js
```diff
@@ -412,7 +412,7 @@ ruleTester.run('dynamic-import-chunkname', rule, {
         /* webpackPrefetch: true */
         /* webpackPreload: true */
         /* webpackIgnore: false */
-        /* webpackMode: "eager" */
+        /* webpackMode: "lazy" */
```
amsardesai, line 415: This test isn't checking for eager mode functionality, it's checking whether other modes are valid, so changing this to "lazy" so it doesn't interfere with the eager mode conditions.

tests/src/rules/dynamic-import-chunkname.js
```diff
@@ -1242,7 +1269,7 @@ context('TypeScript', () => {
         {
           code: `import(
             /* webpackChunkName: "someModule" */
-            /* webpackMode: "eager" */
+            /* webpackMode: "lazy" */
```
amsardesai, line 1272: This test would fail if it was eager mode, changing it to lazy so it's more correct.

tests/src/rules/dynamic-import-chunkname.js
```diff
@@ -1299,13 +1326,21 @@ context('TypeScript', () => {
             /* webpackPrefetch: true */
             /* webpackPreload: true */
             /* webpackIgnore: false */
-            /* webpackMode: "eager" */
+            /* webpackMode: "lazy" */
```
amsardesai, line 1329: Ditto here, this test is checking for webpack comment correctness, not eager mode, so changing this to "lazy"

amsardesai (commented):
Also updated some tests to be a bit more thorough!

amsardesai:
Hey @ljharb, would love to know when you will put out a new release - I'm hoping to use the changes in this PR in our repo.

ljharb:
I was hoping to get flat config support landed before cutting a release; I'll plan to cut one within the next couple of weeks regardless.

amsardesai:
Thanks for the update!

amsardesai:
Hey @ljharb, just pinging again to check if you will release soon, since it's been 8 months since the last release.

ljharb:
@amsardesai i'll probably have to cut a release without flat config support, sadly, so i'd expect one in the next few weeks.

amsardesai:
thank you!

PR #3008, opened 2024-05-20
bump minimum node version to 16

rotu:
eslint-plugin-import declares support for extremely broad versions of node. Updated to remove support for out-of-support node versions (node 16 is the oldest currently supported version, reaching end of life next month)


ljharb (changes requested):
Absolutely not. This would be a breaking change, and platform support has *zero bearing whatsoever* on what node versions a library should support.

rotu:
> Absolutely not. This would be a breaking change, and platform support has _zero bearing whatsoever_ on what node versions a library should support.

Thanks for the feedback. What node versions must still be supported by this project and why?

ljharb:
We support down to eslint 2, and we support down to node 4 (arguably we should support down to node 0.10, but we haven't prioritized that and nobody's asked for it)

Additionally, breaking changes are the most costly thing any package can impose on the ecosystem, and "dropping a node version" has a very small amount of benefit, so there'd need to be actually good reasons to do a breaking change before dropping a node version happened.

Issue #2694, opened 2023-01-25
[import/no-unused-modules] Add an option to ignore typescript types


demedos:
Hi, when using typescript it might be useful to add an option to ignore exported but unused types (including interfaces and enumerators) since it is common to export a component along with its props and let the consumer create derivative types for example.

This is something that has been implemented in [ts-unused-exports](https://github.com/pzavolinsky/ts-unused-exports) via the `allowUnusedEnums` and `allowUnusedTypes` options.

Looking through the code I found out that typescript types are being handled already, see [src/rules/no-unused-modules.js#L89](https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/no-unused-modules.js#L89).

ljharb:
Is this something that `@typescript-eslint/plugin` might already cover?

demedos:
Nope, there's nothing such as unused exports

ciiqr:
Might be a bit more effort than just ignoring types entirely, but it would be cool if this could mark types referenced from used exports as used (ie. in the component/props example, the props type could be marked as used if the component itself is used). Might still want an option to enable this, though it seems like it would allow us to preserve some of the safety of this rule.

silverwind:
I think adding a `allowUnusedTypeExports` to `no-unused-modules` which defaults to `false` should be able to solve it, at least for me.

Which things shall not trigger it? Just `export type` and `export interface`, or is there more?

PR #3011, opened 2024-05-28
[New] `no-unused-modules`: Add `ignoreUnusedTypeExports` option

silverwind:
Fixes: https://github.com/import-js/eslint-plugin-import/issues/2694

### DIFF
--- a/docs/rules/no-unused-modules.md
+++ b/docs/rules/no-unused-modules.md
@@ -29,8 +29,9 @@ This rule takes the following option:
 
  - **`missingExports`**: if `true`, files without any exports are reported (defaults to `false`)
  - **`unusedExports`**: if `true`, exports without any static usage within other modules are reported (defaults to `false`)
- - `src`: an array with files/paths to be analyzed. It only applies to unused exports. Defaults to `process.cwd()`, if not provided
- - `ignoreExports`: an array with files/paths for which unused exports will not be reported (e.g module entry points in a published package)
+ - **`ignoreUnusedTypeExports`**: if `true`, TypeScript type exports without any static usage within other modules are reported (defaults to `false` and has no effect unless `unusedExports` is `true`)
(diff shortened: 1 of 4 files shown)


silverwind:
Will tweak tests a bit.

ljharb:
This seems great, except that it's best when boolean arguments default to false. Can you come up with an alternative name for it so it can default to false?

silverwind:
> This seems great, except that it's best when boolean arguments default to false. Can you come up with an alternative name for it so it can default to false?

Hmm I could invert the name and function but thought that the current name matches nicely into the existing options `missingExports` and `unusedExports`.

silverwind:
Inverted it, so now it's `ignoreUnusedTypeExports` defaulting to `false`.

ljharb:
Your original option name definitely did match better, but i prioritize booleans defaulting to false over option names :-) thanks!

docs/rules/no-unused-modules.md
```diff
@@ -29,6 +29,7 @@ This rule takes the following option:
 
  - **`missingExports`**: if `true`, files without any exports are reported (defaults to `false`)
  - **`unusedExports`**: if `true`, exports without any static usage within other modules are reported (defaults to `false`)
+ - **`ignoreUnusedTypeExports`**: if `true`, type exports without any static usage within other modules are reported (defaults to `false` and depends on `unusedExports` being `true`)
```
silverwind, line 32:
```suggestion
 - **`ignoreUnusedTypeExports`**: if `true`, type exports without any static usage within other modules are reported (defaults to `false` and has no effect unless `unusedExports` is `true`)
```

silverwind:
This CI failures seem intermittent and the coverage annotation "Added line #L518 was not covered by tests" is incorrect, I verified the line is being hit during tests.

ljharb:
node 22.5 itself is broken rn, so i'll rerun those failing jobs once a fixed 22.5.1 is released.

ljharb (approved):

PR #3012, opened 2024-06-03
[Fix] `no-extraneous-dependencies`: allow incorrect paths in packageDir 

chabb:
This rule can potentially fails ( = reports incorrect errors ) in a monorepo settings; this is due to the fact that the building of the map of the available installed dependencies stops when there is an error ( e.g when the path to the package.json does not point to an existing package.json file )

-  If you pass only one path to a package.json file, then this path should be correct, this is ok 
- If you pass multiple paths, there are some situations when those paths point to a wrong path, this happens typically in a nx monorepo. The problem is that the parsing of the package.json files stop is one of the path is wrong, and then the rule will give us false positive. One example is a nx monorepo with a pre-commit hook that runs eslint
-- NX will run eslint in the projects folder, so we need to grab the root package.json
-- The pre-commit hook will run eslint in the root folder, so one of the path given will be an incorrect path, but we do not want to throw there, otherwise the rull will fail

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`dynamic-import-chunkname`]: add `allowEmpty` option to allow empty leading comments ([#2942], thanks [@JiangWeixian])
 - [`dynamic-import-chunkname`]: Allow empty chunk name when webpackMode: 'eager' is set; add suggestions to remove name in eager mode ([#3004], thanks [@amsardesai])
 
+### Fixed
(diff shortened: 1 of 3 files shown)


danymarques:
Would be nice to have this merged! We also have issues with this.

alaingiller:
Same here, this rule doesn't work in a nx monorepo setup

hedinfaok:
As a workaround, I run `nx run-many -t lint` in the repo root directory. The eslint config file is also in the repo root directory.

```
      'warn',
      {
        packageDir: [
          __dirname, // <root>/package.json
          process.cwd(), // <project>/package.json
        ],
      },
    ],
```

ljharb (approved):

PR #3018, opened 2024-06-19
[New] add support for Flat Config

### LINKED ISSUE #2556: [Feature Request] Support new ESLint flat config
ESLint has a [new experimental config file format](https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new) and this plugin doesn't work with it. The plugin fails at [this line](https://github.com/import-js/eslint-plugin-import/blob/main/utils/parse.js#L60) because the new config format doesn't pass parsers via paths. Instead the [parser object itself is passed](https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new#:~:text=parser%20%2D%20Either%20an%20object%20containing%20a%20parse()%20method%20or%20a%20string%20indicating%20the%20name%20of%20a%20parser%20inside%20of%20a%20plugin%20(i.e.%2C%20%22pluginName/parserName%22).%20(default%3A%20%22%40/espree%22)).

I was able to mostly fix this in my [fork](https://github.com/TomerAberbach/eslint-plugin-import/commit/05a9115f2209817bc643230831482b41c1913f12), but because the `parserPath` doesn't exist anymore for this config format some of the [`keysFromParser` logic won't work anymore](https://github.com/import-js/eslint-plugin-import/blob/main/utils/parse.js#L26-L31).

Anyway, figured I'd open this issue to start discussion on the new config file format!

michaelfaith:
This change adds support for ESLint's new Flat config system.  It maintains backwards compatibility with eslintrc style configs as well.

To achieve this, we're now dynamically creating flat configs on a new `flatConfigs` export.

Example Usage

```js
import importPlugin from 'eslint-plugin-import';
import js from '@eslint/js';
import tsParser from '@typescript-eslint/parser';

export default [
  js.configs.recommended,
  importPlugin.flatConfigs.recommended,
  importPlugin.flatConfigs.react,
  importPlugin.flatConfigs.typescript,
  {
    files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'],
    languageOptions: {
      parser: tsParser,
      ecmaVersion: 'latest',
      sourceType: 'module',
    },
    ignores: ['eslint.config.js'],
    rules: {
      'no-unused-vars': 'off',
      'import/no-dynamic-require': 'warn',
      'import/no-nodejs-modules': 'warn',
    },
  },
];
```

I wasn't able to reproduce any issues with the parser (as mentioned in #2556), so there's nothing here specifically to address that.  And just to be clear, this is only aiming to provide support for flat config (which is a separate issue than supporting eslint v9), and has only been tested with the latest version of v8.  I created two testbeds under a new `examples` folder, one for `legacy` and one for `flat`, each setup with the `@typescript-eslint/parser` and a handful of `ts` and `tsx` files that contain rule violations, to ensure the parsing still works as expected.  I also checked that the parsing workflow is happening properly by scattering some log messages at different points in the logic that resolves the parser, and in both legacy and flat setups, it's getting the parser ok (screenshots below).  It looks like there was already code to navigate the fact that the parsing options have changed shape in the new config format.  So if @TomerAberbach or anyone else that's had issues with the plugin parsing can test this branch out with their use case or give guidance on how to reproduce their issue, that could help.  Otherwise, i think this should satisfy both legacy and flat configs. 

I do think there should be a larger refactor at some point to move away from the parser by name paradigm and embrace the new way of passing a parser object, but it wasn't necessary to do that here.  Maybe something to consider for v9 support (or v10, since some of the deprecated functions will be removed in v10).  

Legacy Config Execution:
legacy

Flat Config Execution:
flat

Closes #2556


michaelfaith:
@ljharb I found this issue you logged about the `no-unused-modules` rule specifically https://github.com/eslint/eslint/issues/18087.  So it may not be possible to land this until that has been addressed.  But maybe this can branch can help provide a testing ground to implement that?  It seemed one of the issues you were having was a good place to test the approach.

michaelfaith:
@ljharb It appears that the `no-unused-modules` rule is still working in 8.x with flat config.  I added that to both of the example tests.  I also added several console logs to the rule to see if all of the files were being populated correctly, and it all seems to behave the same as the legacy config.  For v9+ I think @nzakas' proposed api change (https://github.com/eslint/eslint/issues/18087#issuecomment-1944319008) along with his recommendation of using `@nodelib/fs-walk` for walking the files under cwd ([walkSync](https://www.npmjs.com/package/@nodelib/fs.walk#walksyncpath-optionsorsettings)) would work.  
image

Rule Execution:
image

tl;dr: i think this could be merged and released for 8.x flat config support, unless there's an additional element i'm not considering.

ljharb:
@michaelfaith if we can run the tests in flat config also, and they pass, then that's good enough for me - it's probably a good idea to implement the proposed eslint API (as a fallback prior to whenever eslint ships it) and base it on that, then we can use the built-in API once it's available?

michaelfaith:
> it's probably a good idea to implement the proposed eslint API (as a fallback prior to whenever eslint ships it) and base it on that, then we can use the built-in API once it's available?

Makes sense.  Would you like that part of this change, or a separate PR?

ljharb:
Seems like part of this PR if it’s only going to be used in flat config

controversial:
otoh, there might be some value in merging support for v8 flat config first, if it’s ready and tested, and then releasing eslint v9 support separately

ljharb:
@controversial thats what this PR should be doing.

controversial:
My mistake, I had misremembered and thought the “proposed eslint API” from https://github.com/eslint/eslint/issues/18087 was to cover a v9-removed API, rather than a flat config incompatibility

nzakas:
> it's probably a good idea to implement the proposed eslint API (as a fallback prior to whenever eslint ships it) and base it on that, then we can use the built-in API once it's available?

Just want to flag that there's no guarantee that the API I proposed and prototyped will ship. It still needs to be RFCed based on feedback. It was just an idea to test out feasibility.

michaelfaith:
> > it's probably a good idea to implement the proposed eslint API (as a fallback prior to whenever eslint ships it) and base it on that, then we can use the built-in API once it's available?
> 
> Just want to flag that there's no guarantee that the API I proposed and prototyped will ship. It still needs to be RFCed based on feedback. It was just an idea to test out feasibility.

@nzakas In that sense, is there anything else you need from this project to move forward with the proposed api updates?  Or is the runway clear?

nzakas:
If you can comment back on https://github.com/eslint/eslint/issues/18087 with the results of using the prototype that would help. It would also help to know if you're accessing files that ESLint might not have linted during its lifecycle. (For example, if I run `eslint foo.js`, are you only ever looking at `foo.js` or are you still looking at all the files?)

ljharb:
@nzakas we're still looking at all the files - at a minimum, all the files that `foo.js` directly or transitively imports.

michaelfaith:
> If you can comment back on [eslint/eslint#18087](https://github.com/eslint/eslint/issues/18087) with the results of using the prototype that would help.

Was that poc branch published to a pre-release version?  Or what's the best way to consume that POC here?

nzakas:
@ljharb hmmm okay, then this probably won't work without async rules, which are a ways off. 

@michaelfaith you'll need to check out the branch mentioned in the issue.

ljharb:
@nzakas sorry if that wasn't made clear that that's how we're using FileEnumerator - basically we traverse every lintable file and build up a complete dependency graph, and then go from there.

michaelfaith:
> My mistake, I had misremembered and thought the “proposed eslint API” from [eslint/eslint#18087](https://github.com/eslint/eslint/issues/18087) was to cover a v9-removed API, rather than a flat config incompatibility

I don't think this is actually the case.  I'm seeing the flat config without this additional changes working just fine in v8 with the changes I've already made.  I don't really have a horse in the race as far as whether this should go in with or without the additional changes we've discussed, but purely from a v8 flat-config compatibility perspective, I believe this could be released as is.  The rule in question `no-unused-modules` is still working with the og `FileEnumerator` in the flat config example I added.

controversial:
If this PR includes thorough tests of everything working under eslint v8 flat config, then I don’t see a reason to avoid releasing it!

ljharb:
Indeed, if the FileEnumerator problem only applies in v9 and not in flat config, then it'd be fine - but that wasn't my understanding of the problem. In flat config, does FileEnumerator still respect the eslint config's ignore settings, for example?

michaelfaith:
>  In flat config, does FileEnumerator still respect the eslint config's ignore settings, for example?
In my local, I added a log messages in several places to see how `FileEnumerator` behaved, specifically for the `no-unused-modules` rule.

Here is the list of files obtained under different scenarios (in all cases `exports.ts` has a violation of the `no-unused-modules` rule):

### legacy rc _without_ ignoring `**/exports.ts`
image
image

### legacy rc ignoring `**/exports.ts`
image
image

### flat config without ignoring `**/exports.ts`
image
image

### flat config ignoring `**/exports.ts`
image
image

So, you're right that the files being processed by the rule include files that were ignored, which means the rule is having to do more work than it should (not good from a performance perspective).  Though, interestingly they're not being reported as violations (i.e. not having any obvious user-facing impact).  I'm happy to keep working on this in this PR, but with the violations still reporting as expected, not sure how you want to treat it.

michaelfaith:
I added another `exports-used.ts` to both sets of examples, with each config ignoring the file.  That way `exports.ts` can still demonstrate the rule violation working, while we explore the ignored files difference.

nzakas:
Okay, a little bit of history here to help clear things up. :)

`FileEnumerator` was an internal-only API that we used for finding files inside of ESLint based on the eslintrc configuration system. It takes care of reading in every configuration file and figuring out which files should be returned for linting.

This plugin was only able to use `FileEnumerator` because early on Node.js allowed access to any file in the package whether or not it was mentioned in `package.json`. Once we decided to lock down the API to prevent this from happening, we agreed to leave `FileEnumerator` exposed so as not to break `eslint-plugin-import` in the short term. But we did warn that this wasn't going to be forever and that we didn't plan on creating a replacement for `FileEnumerator` either internally or externally.

All that is to say, `FileEnumerator` does not read flat config files and therefore can't be used effectively when the project uses flat config files.

This use case (crawling into files that aren't part of the lint session) isn't something that ESLint can formally support at this point, so it will require some trickery or imperfect solutions if it's going to work at all.

To that end, though, it may be worth considering pointing people to [Knip](https://knip.dev/), which is its own standalone tool that can solve this same problem without the constraints of running inside of ESLint.

michaelfaith:
> To that end, though, it may be worth considering pointing people to [Knip](https://knip.dev/), which is its own standalone tool that can solve this same problem without the constraints of running inside of ESLint.

The implication being that this plugin phases out the rule entirely?  Seems reasonable, actually.  I.e. use the right tool for the job, rather than trying to force eslint to do something it's not intended for.

guillaumebrunerie:
> This use case (crawling into files that aren't part of the lint session) isn't something that ESLint can formally support at this point

Isn’t it reasonable to expect being able to use ESLint to find unused exports? Unused exports are a potential problem, in exactly the same way as unused variables are: "[most likely an error due to incomplete refactoring. Such [exports] take up space in the code and can lead to confusion by readers](https://eslint.org/docs/latest/rules/no-unused-vars)". And finding such potential issues is the whole point of ESLint. I was actually very surprised when I first used ESLint to learn that this rule is not part of ESLint core but is only available via some third-party plugin.

ljharb:
It's quite reasonable - however we could certainly use knip or something similar inside the rule as an alternative way to find unused files, if it's compatible with the rule. (i used knip as the core of https://www.npmjs.com/package/@ljharb/unused-files, so I'm aware it probably won't work in this project, but that's a possible direction to go)

@nzakas why would `require`ability matter for eslint? I assumed you'd use fs functions, which aren't constrained by the `exports` field.

michaelfaith:
> > This use case (crawling into files that aren't part of the lint session) isn't something that ESLint can formally support at this point
> 
> Isn’t it reasonable to expect being able to use ESLint to find unused exports? Unused exports are a potential problem, in exactly the same way as unused variables are: "[most likely an error due to incomplete refactoring. Such [exports] take up space in the code and can lead to confusion by readers](https://eslint.org/docs/latest/rules/no-unused-vars)". And finding such potential issues is the whole point of ESLint. I was actually very surprised when I first used ESLint to learn that this rule is not part of ESLint core but is only available via some third-party plugin.

Correct me if I'm wrong, but I think it has more to do with the paradigm of ESLint being that rules should be something that can be validated within a single file.  Reaching across multiple files in order to validate a rule, breaks paradigm.  It's not to say that it isn't a worthy thing to care about, but just that there are other tools that are more suited for that purpose. At least that's what I'm reading into it.

nzakas:
> why would `require`ability matter for eslint? I assumed you'd use fs functions, which aren't constrained by the `exports` field.

@ljharb Sorry, I lost the thread and can't find what this is in relation to. Can you expand?

> Correct me if I'm wrong, but I think it has more to do with the paradigm of ESLint being that rules should be something that can be validated within a single file. Reaching across multiple files in order to validate a rule, breaks that paradigm

@guillaumebrunerie @michaelfaith This is correct. The way ESLint works is on one rule at a time, and there is no concept of information from one file being available in another file. Whether or not this should be added is a good discussion to have, but it's a nontrivial architectural change that would require a lot of thought and planning. I think in an ideal world, ESLint is capable of doing this in the core rather than forcing rules to work through this on their own. It's something I've pondered for a bit, and something that the core rewrite will give us an opportunity to explore.

ljharb:
@nzakas 
> This plugin was only able to use FileEnumerator because early on Node.js allowed access to any file in the package whether or not it was mentioned in package.json.

node continues to allow that via `fs.readFile` and friends; it's only the `exports` field that requires `require` or `import` of files, so I wouldn't expect that the `exports` field has any impact on linting.

nzakas:
Gotcha. Minor point - just don't use internal APIs. :)

ljharb:
ahhh you’re saying that’s why we were able to access FileEnumerator - yes, that’s true :-)

michaelfaith:
@ljharb I had some time to put towards this refactor, using the tentative apis that @nzakas introduced in his branch.  Everything looks good with those new functions.  I hooked up the the examples to the local build of that branch, and verified that it gathered the list of files correctly, using `walkSync` as @nzakas suggested.  The one caveat I noticed is that, while ignores were working as expected for most files, it it was _not_ ignoring the `eslint.config.mjs` file itself, despite that being listed in the `ignoreFiles` array.  

image

vs 

image

The other ignore patterns worked fine.  It just seems to be including the config regardless.

michaelfaith:
Also, worth mentioning, when running with the 9.0.0-beta branch, there were issued with other rules that will need to be resolved as part of the v9 update.  Namely, the `importDeclaration` function includes a call to `context.getAncestors()` which doesn't seem to exist anymore.  
``js
export default function importDeclaration(context) {
  const ancestors = context.getAncestors();
  return ancestors[ancestors.length - 1];
}
```

![image](https://github.com/user-attachments/assets/66d8656c-75ae-45e8-a52a-77e12ebc5490)

File: src/rules/no-unused-modules.js [hunk 1]
```diff
@@ -17,51 +18,170 @@ import ExportMapBuilder from '../exportMap/builder';
...
+ * @param {string[]} extensions - list of supported file extensions
+ * @param {object} context - the eslint context object
+ * @returns the list of files that this rule will evaluate.
+ */
+function listFilesToProcess(src, extensions, context) {
+  // If the context object has the new session functions, then prefer those
+  // Otherwise, fallback to using the deprecated `FileEnumerator` for legacy support.
+  // https://github.com/eslint/eslint/issues/18087
+  if (
+    context.session &&
```
michaelfaith, line 167: I wasn't able to use optional chaining for this, like I originally attempted.  I guess this is being built down to a really early node version.

src/rules/no-unused-modules.js
```diff
@@ -174,17 +294,36 @@ const isNodeModule = (path) => (/\/(node_modules)\//).test(path);
...
 
   // prepare list of ignored files
-  const ignoredFilesList = listFilesToProcess(ignoreExports, extensions);
+  const ignoredFilesList = listFilesToProcess(
+    ignoreExports,
+    extensions,
+    context,
+  );
+
+  // The modern api will return a list of file paths, rather than an object
```
michaelfaith, line 306: The shape of the legacy `listFilesToProcess` was an array of objects consisting of `filename` and `ignored`.  
image

Whereas the new api just returns a list of filepaths
image

utils/ignore.js
```diff
@@ -14,7 +14,9 @@ const log = require('debug')('eslint-plugin-import:utils:ignore');
 function makeValidExtensionSet(settings) {
   // start with explicit JS-parsed extensions
   /** @type {Set<import('./types').Extension>} */
-  const exts = new Set(settings['import/extensions'] || ['.js']);
+  const exts = new Set(
```
michaelfaith, line 17: Thought it would be good to add support for `cjs` and `mjs` files

src/rules/no-unused-modules.js
```diff
@@ -17,51 +18,167 @@ import ExportMapBuilder from '../exportMap/builder';
...
+
+  for (let i = 0; i < srcPaths.length; i++) {
+    const src = srcPaths[i];
+    // Use walkSync along with the new session api to gather the list of files
+    const entries = fsWalk.walkSync(src, {
+      deepFilter(entry) {
+        const fullEntryPath = resolvePath(src, entry.path);
+
+        // Include the directory if it's not marked as ignore by eslint
+        return !session.isDirectoryIgnored(fullEntryPath);
```
michaelfaith, line 40: @nzakas These new apis worked well here.

michaelfaith:
Oh, and `@nodelib/fs.walk` has engine declaration of `>=16.14.0`.

ljharb:
> Oh, and `@nodelib/fs.walk` has engine declaration of node `>=16.14.0`.

then we can't use it, unfortunately.

controversial:
Note that eslint@9 requires Node.js `^18.18.0 || ^20.9.0 || >=21.1.0` 

So if `eslint-plugin-import` were willing to **skip** supporting eslint@8 flat config (i.e., declare that flat config will only be supported for `eslint>=9`) then `@nodelib/fs.walk` could be used without breaking compatibility for anyone:
- For `eslint<9` users, this plugin could use FileEnumerator under the old config format
- For `eslint>=9` users (who all have flat configs), this plugin could use `@nodelib/fs.walk` because those users necessarily have a compatible node version.

controversial:
In fact, I don’t think it’s clear whether the (currently still hypothetical) eslint API from https://github.com/eslint/eslint/issues/18087 is being considered at all for a future 8.x release, or whether it’s only being considered to add to a future 9.x ?

If it’s the latter, then the `@nodelib/fs.walk` solution wouldn’t be viable for any 8.x users anyway.

ljharb:
@controversial that would work perfectly fine, except that `engines.node` declarations aren't conditional.

controversial:
What about `optionalDependencies`?
1. `eslint-plugin-import` adds `@nodelib/fs.walk` to `optionalDependencies`
2. We assume that it will be installed for any user on eslint@9 (because otherwise the eslint@9 `engines` declaration will have failed)
3. If this is not the case at runtime (ie `@nodelib/fs.walk` cannot be found for a user on eslint@9) throw an exception due to invariant violation

silverwind:
> engines.node

`engines.node` ultimately does not matter, it's just an advice. You can put anything you want in there, package managers will still install the module just fine. I recommand you put the lowest supported version. anyone running eslint 9 will be using a supported node version anyways.

ljharb:
@silverwind that's not accurate. It will cause `npm ls` to fail, it will fail installs that use `engines.strict`, and in an upcoming version of npm, it will cause that version of a package to not be installed at all.

The issue is that `@nodelib/fs.walk` has `engines.node` set to `'>=16.14.0'`.

silverwind:
> @silverwind that's not accurate. It will cause `npm ls` to fail, it will fail installs that use `engines.strict`, and in an upcoming version of npm, it will cause that version of a package to not be installed at all.
> 
> The issue is that `@nodelib/fs.walk` has `engines.node` set to `'>=16.14.0'`.

I doubt npm will ever go that route. But if this is really a concern, puttting it in `optionalDependencies` as suggested is the way to go.

controversial:
> in an upcoming version of npm, it will cause that version of a package to not be installed at all

Any “upcoming version of npm” requires at least node 18, so we don’t need to worry about whether a future npm version might create problems for node 16.x users

ljharb:
npm's already decided to go that route (in that it'll autoselect a version of the package whose engines declarations match).

either way, adding an optionalDep is a breaking change.

To be clear, I'm not going to add any transitive engines declarations to the runtime deps of any package, that conflict with the package's own engines declarations. That's a nonstarter.

silverwind:
Maybe just use or copy any of the existing directory walking implementations then, there are hundrets of these on npm.

ljharb:
Sure, that's fine if that's the best available option.

controversial:
Why is adding an optionalDep a breaking change?

ljharb:
@controversial because if someone already has it installed, but the version doesn't satisfy the added range, the dep graph will no longer be valid (install may fail, and `ls` will certainly fail).

michaelfaith:
Even if I were to re-implement what `fs.walk`'s sync function is doing, which I wouldn't mind doing, I'd still want to use `fs` apis that were introduced in node v10 (e.g. `Dirent`).  I'm not super interested in working with node v4 APIs.  That seems like an unnecessary constraint.  Happy to hand this off to someone else, if that remains the expectation.

mshima:
> Even if I were to re-implement what `fs.walk`'s sync function is doing, which I wouldn't mind doing, I'd still want to use `fs` apis that were introduced in node v10 (e.g. `Dirent`). 

It’s possible to use FileEnumerator for ESLint < v9 and fall back to lazy load fs api for eslint v9 which requires node >= 18.18.0.
Lazy loading will allow new apis to be used.

mshima:
In current PR if `@nodelib/fs.walk` is lazy loaded CI I think CI will pass.
The missing task is to replace `@nodelib/fs.walk` with in module implementation.

controversial:
@mshima 

@ljharb [discussed](https://github.com/import-js/eslint-plugin-import/pull/3018#issuecomment-2237506411) that he’s not willing to add `fs.walk` as a dependency given its `engines` declaration.

But if @michaelfaith is willing to rebuild part of `fs.walk` functionality, I think using node 10+ APIs there would be fine, since those could be lazy-loaded at runtime for flat config users only (who will have compatible node versions as dictated by their eslint version)

michaelfaith:
Seems like a reasonable approach.  I'll go that route

ljharb:
That’s exactly right; conditionally using modern APIs is perfectly fine; an engines declaration that would break everyone is not.

michaelfaith:
@ljharb I implemented a local version of `walkSync` (src/core/fsWalk.js)  and removed the dependency on `@nodelib/fsWalk`.  I made it in such a way that when / if this library ever does do a major version and is ok with installing the full lib, it'd just be an import path swap in the rule.

controversial:
That’s amazing @michaelfaith; thanks for all your hard work

mohammedhammoud:
Great job on this! ⭐ Do you have any idea when it will be released?

ljharb (approved):
Thanks!

(Rebased, squashed, removed the "stage 0" flat config that nobody should be using anyways, reverted some of the prettier mess in the rule, fixed a bug in the rule in `resolveFiles`, and cleaned up fsWalk a bit)

controversial:
@nzakas It looks like this PR was just merged with an implementation that includes support for `context.session.isFileIgnored` and `context.session.isDirectoryIgnored`!

There’s still a fallback to use FileEnumerator where these APIs aren’t available, but it seems like the proposed APIs from https://github.com/eslint/eslint/issues/18087 are ready to be used by `eslint-plugin-import` if/when eslint is able to ship them (see `listFilesWithModernApi`). Of course, this would also depend on `eslint-plugin-import` gaining support for eslint 9.x, assuming that the `context.session.is*Ignored` APIs wouldn’t be backported to a new 8.x release.

michaelfaith:
> Thanks!
> 
> (Rebased, squashed, removed the "stage 0" flat config that nobody should be using anyways, reverted some of the prettier mess in the rule, fixed a bug in the rule in `resolveFiles`, and cleaned up fsWalk a bit)

Thanks for your attention on this.

vinozganic:
This is how my `eslint.config.mjs` looks like right now:
```js
//@ts-nocheck
import react from 'eslint-plugin-react';
import globals from 'globals';
import path from 'node:path';
import { fileURLToPath } from 'node:url';
import js from '@eslint/js';
import { FlatCompat } from '@eslint/eslintrc';
import * as espree from 'espree';
import importPlugin from 'eslint-plugin-import'; // NEWLY ADDED

const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);
const compat = new FlatCompat({
  baseDirectory: __dirname,
  recommendedConfig: js.configs.recommended,
  allConfig: js.configs.all,
});

export default [
  importPlugin.flatConfigs.recommended, // NEWLY ADDED
  importPlugin.flatConfigs.react, // NEWLY ADDED
  {
    /*
     *** Files and directories which ESLint ignores:
     * - eslint.config.mjs: ESLint configuration file
     * - *.js and *.ts: JavaScript and TypeScript files in the root folder
     */
    ignores: ['eslint.config.mjs', '*.js', '*.ts'],
  },
  ...compat.extends('plugin:react/recommended', 'airbnb', 'plugin:prettier/recommended'),
  {
    plugins: {
      react,
    },

    languageOptions: {
      globals: {
        ...globals.browser,
      },
      ecmaVersion: 'latest',
      sourceType: 'module',
      parser: espree,
      parserOptions: {
        ecmaFeatures: {
          jsx: true,
        },
      },
    },

    rules: {
      'no-restricted-syntax': 0,
      'import/prefer-default-export': 0,
      'react/prop-types': 0,
      'jsx-a11y/label-has-associated-control': [
        'error',
        {
          required: {
            some: ['nesting', 'id'],
          },
        },
      ],
      'no-unused-vars': [
        'error',
        {
          argsIgnorePattern: '^_',
        },
      ],
    },
  },
];
```

I'm using `"eslint": "^8.57.1"` and `"eslint-plugin-import": "^2.30.0"`. React app with JavaScript.

This `eslint.config.mjs` was automatically generated with [ESLint flat config migration tool](https://www.npmjs.com/package/@eslint/migrate-config), plus I've added those 3 **NEWLY ADDED** lines to get rid of the 
`parserPath or languageOptions.parser is required! (undefined:undefined)eslint[import/no-named-as-default](https://github.com/import-js/eslint-plugin-import/blob/v2.30.0/docs/rules/no-named-as-default.md)` 
message.

Now I'm getting this error:
```
ConfigError: Config (unnamed): Key "plugins": Cannot redefine plugin "import".
    at rethrowConfigError (/home/jakov/fireball_prod/fireball-frontend/node_modules/@humanwhocodes/config-array/api.js:230:8)
    at /home/jakov/fireball_prod/fireball-frontend/node_modules/@humanwhocodes/config-array/api.js:1032:5
    at Array.reduce (<anonymous>)
    at FlatConfigArray.getConfig (/home/jakov/fireball_prod/fireball-frontend/node_modules/@humanwhocodes/config-array/api.js:1028:39)
    at FlatESLint.lintText (/home/jakov/fireball_prod/fireball-frontend/node_modules/eslint/lib/eslint/flat-eslint.js:951:32)
    at async /root/.vscode-server/extensions/dbaeumer.vscode-eslint-3.0.10/server/out/eslintServer.js:1:26981
    at async M (/root/.vscode-server/extensions/dbaeumer.vscode-eslint-3.0.10/server/out/eslintServer.js:1:19807)
    at async /root/.vscode-server/extensions/dbaeumer.vscode-eslint-3.0.10/server/out/eslintServer.js:1:234554
    at async /root/.vscode-server/extensions/dbaeumer.vscode-eslint-3.0.10/server/out/eslintServer.js:1:63886
```
when i try to run ESLint on my project. My eslint.config.mjs is invalid.

How can I fix this? What am I doing wrong?

vinozganic:
@michaelfaith @ljharb please check my [comment above](https://github.com/import-js/eslint-plugin-import/pull/3018#issuecomment-2363324250)

michaelfaith:
Can you try removing "airbnb", from the extends to see if you still get the error?

Since it's not yet set up for flat config, and it declares the `import` plugin, that you're ending up with two declarations of import
image

If that's the issue, I can suggest a workaround to use, until that library has updated to support.

vinozganic:
> Can you try removing "airbnb", from the extends to see if you still get the error?
> 
> I suspect since it's not yet set up for flat config, and it declares the `import` plugin, that you're ending up with two declarations of import image
> 
> If that's the issue, I can suggest a workaround to use, until that library has updated to support.

Now I'm getting the same error:
`Parse errors in imported module 'react-router-dom': parserPath or languageOptions.parser is required! (undefined:undefined)eslint[import/namespace](https://github.com/import-js/eslint-plugin-import/blob/v2.30.0/docs/rules/namespace.md)`
but it's from eslint ([import/namespace](https://github.com/import-js/eslint-plugin-import/blob/v2.30.0/docs/rules/namespace.md)).

You know what, I'm having too much issues and im losing too much time on setting the new flat config up. 
I think I'll switch back to old `eslintrc.js` config.

**Do you have any samples or do you know what's the best and most conventional `eslintrc.js` config for a React JavaScript project?**
I want it to cover all .js and .jsx files in the directory when I run: `npx eslint .`

My old (not fully working) .eslintrc.js was looking like this:
```js
module.exports = {
  env: {
    browser: true,
    es2021: true,
  },
  extends: ['plugin:react/recommended', 'airbnb', 'plugin:prettier/recommended'],
  parserOptions: {
    ecmaFeatures: {
      jsx: true,
    },
    ecmaVersion: 'latest',
    sourceType: 'module',
  },
  plugins: ['react'],
  rules: {
    'no-restricted-syntax': 0,
    'import/prefer-default-export': 0,
    'react/prop-types': 0,
    'jsx-a11y/label-has-associated-control': [
      'error',
      {
        required: {
          some: ['nesting', 'id'],
        },
      },
    ],
    'no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
  },
};
```
and it wasn't working for .jsx files - when I would run `npx eslint .` or  `npx eslint --no-ignore .` it would scan all my .js files correctly, but .jsx files wouldn't be scanned. And when I would run e.g. `npx eslint src/components/Calendar/Calendar.jsx` it would scan that single .jsx file successfuly.

So, again, can you check my old `.eslintrc.js` config and see why it's not scanning .jsx files by default and help me fix it or give me some `.eslintrc.js` sample that i could copy paste and it would work for my React JavaScript project?

Thank you!

michaelfaith:
> Now I'm getting the same error: `Parse errors in imported module 'react-router-dom': parserPath or languageOptions.parser is required! (undefined:undefined)eslint[import/namespace](https://github.com/import-js/eslint-plugin-import/blob/v2.30.0/docs/rules/namespace.md)` but it's from eslint ([import/namespace](https://github.com/import-js/eslint-plugin-import/blob/v2.30.0/docs/rules/namespace.md)).

That should be fixed once https://github.com/import-js/eslint-plugin-import/pull/3061 is merged and released.

> when I would run npx eslint . or npx eslint --no-ignore . it would scan all my .js files correctly, but .jsx files wouldn't be scanned. And when I would run e.g. npx eslint src/components/Calendar/Calendar.jsx it would scan that single .jsx file successfuly.

with an rc config you can provide extensions on the command line similar to `npx eslint .{jsx,js,ts,tsx}`

ljharb:
Indeed, you can't use the airbnb config with flat config yet.

Issue #2717, opened 2023-02-14
Suggestion: new rule to enforce usage of `node:` prefix for Node.js built-ins


rdsedmundo:
I searched a bit in the existing issues and didn't see someone asking for anything similar, also couldn't find in the existing rules. I think this could belong to this package.

Benefits (copied this from 3rd party resources, and added a few personal touches):
- More explicit protocol makes it more readable and also makes it clear that a built-in module is being imported. Since there are a lot of built-in ones, this extra information is useful for beginners or for someone who might not know about one of them. I once had a senior engineer with 10 years of experience install [assert](https://www.npmjs.com/package/assert) without knowing that Node provided it.
- It reduces the risk of a module present in node_modules overriding the newer imports.
- If the prefix is used, the built-in module is always returned, completely bypassing the require cache. For instance, require('node:http') will always return the built in HTTP module, even if there is require.cache entry by that name.
- New core modules in the future may only be available if the prefix is used, so enforcing it for everything will keep everything consistent. For instance, the new built-in test module introduced in Node.js 18 can only be used via `node:test`.

ljharb:
Why? What's the advantage to using the `node:` prefix?

rdsedmundo:
I updated the description with a few benefits I found in the web. Let me know if that's sufficient.

ljharb:
The third point is already true for unprefixed core modules.

You're correct that `node:test` is only supported via the prefix, but that was a very unfortunate decision that I hope node won't repeat in the future.

rdsedmundo:
In my opinion, the upside of forcing the prefix is that it allows a more unique name to be chosen, without worrying about naming clashes with a package that is in `npm`.

ljharb:
That's certainly a benefit for node core, but i'm not sure how that'd be a benefit for everyone else? node already only adds core modules when they've reserved the package name on npm.

what1s1ove:
Agreed. Will be looking forward to this rule. Or, if anyone has snippets for `no-restricted-import` that would be nice too 🙏 (but the new rule would be simpler 😁).

I think another plus for forcing the use of the `node:` prefix – https://deno.com/blog/v1.30#support-for-built-in-nodejs-modules
Other platforms will only support node's api packages with a prefix.

ljharb:
This is a node project, as is eslint, and I'm not concerned with deno's incomplete support of node modules.

what1s1ove:
[...]

PR #3023, opened 2024-07-03
feat: added support for node: protocol

GoldStrikeArch:
Fix to this [issue](https://github.com/import-js/eslint-plugin-import/issues/2717)
Added plugin 'prefer-node-builtin-imports'.
The plugin is based on 'prefer-node-protocol' rule from 'eslint-plugin-unicorn' plugin and contains 0 dependencies.
All test cases are taken from [eslint-plugin-unicorn_test](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/prefer-node-protocol.mjs)


ljharb (changes requested):
This doesn't achieve anything for consumers, it just lints *this project*, and we do not actually prefer the node protocol here. The linked issue is about a *rule*, in _this_ plugin, that allows this to be configured.

PR #3024, opened 2024-07-04
[New] add `enforce-node-protocol-usage` rule

### LINKED ISSUE #2717: Suggestion: new rule to enforce usage of `node:` prefix for Node.js built-ins

GoldStrikeArch:
Fixes #2717
Added a new rule `prefer-node-builtin-imports`.
All test cases are taken from [eslint-plugin-unicorn_test](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/prefer-node-protocol.mjs)

### DIFF
--- a/docs/rules/enforce-node-protocol-usage.md
+++ b/docs/rules/enforce-node-protocol-usage.md
@@ -1,0 +1,81 @@
+# import/enforce-node-protocol-usage
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+Enforce either using, or omitting, the `node:` protocol when importing Node.js builtin modules.
+
+## Rule Details
+
+This rule enforces that builtins node imports are using, or omitting, the `node:` protocol.
+
+Determining whether a specifier is a core module depends on the node version being used to run `eslint`.
+This version can be specified in the configuration with the [`import/node-version` setting](../../README.md#importnode-version).
+
+Reasons to prefer using the protocol include:
+
+ - the code is more explicitly and clearly referencing a Node.js built-in module
+
+Reasons to prefer omitting the protocol include:
+
+ - some tools don't support the `node:` protocol
+ - the code is more portable, because import maps and automatic polyfilling can be used
+
+## Options
+
+The rule requires a single string option which may be one of:
+
+ - `'always'` - enforces that builtins node imports are using the `node:` protocol.
+ - `'never'` - enforces that builtins node imports are not using the `node:` protocol.
+
+## Examples
+
+### `'always'`
+
+❌ Invalid
+
+```js
+import fs from 'fs';
+export { promises } from 'fs';
+// require
+const fs = require('fs/promises');
+```
+
+✅ Valid
+
+```js
+import fs from 'node:fs';
+export { promises } from 'node:fs';
+import * as test from 'node:test';
+// require
+const fs = require('node:fs/promises');
+```
+
+### `'never'`
+
+❌ Invalid
+
+```js
+import fs from 'node:fs';
+export { promises } from 'node:fs';
+// require
+const fs = require('node:fs/promises');
+```
+
+✅ Valid
+
+```js
+import fs from 'fs';
+export { promises } from 'fs';
+
+// require
+const fs = require('fs/promises');
+
+// This rule will not enforce not using `node:` protocol when the module is only available under the `node:` protocol.
+import * as test from 'node:test';
+```
+
+## When Not To Use It
+
+If you don't want to consistently enforce using, or omitting, the `node:` protocol when importing Node.js builtin modules.
--- a/tests/src/rules/enforce-node-protocol-usage.js
+++ b/tests/src/rules/enforce-node-protocol-usage.js
@@ -1,0 +1,345 @@
+import { RuleTester } from '../rule-tester';
+import flatMap from 'array.prototype.flatmap';
+import { satisfies } from 'semver';
+
+import { test, testVersion } from '../utils';
+
+const ruleTester = new RuleTester();
+const rule = require('rules/enforce-node-protocol-usage');
+
+const preferUsingProtocol = ['always'];
+const preferNotUsingProtocol = ['never'];
+const useNewerParser = { ecmaVersion: 2021 };
+
+const actualModules = ['fs', 'fs/promises', 'buffer', 'child_process', 'timers/promises'];
+
+const settings = {
+  'import/node-version': '16.0.0', // the node: prefix is only available as of `^14.18 || >= 16`
+};
+
+const invalidTests = [].concat(
+  flatMap(actualModules, (moduleName) => [].concat(
+    {
+      code: `import x from "${moduleName}";`,
+      output: `import x from "node:${moduleName}";`,
+      options: preferUsingProtocol,
+      errors: [
+        { messageId: 'requireNodeProtocol', data: { moduleName } },
+      ],
+    },
+    {
+      code: `export {promises} from "${moduleName}";`,
+      output: `export {promises} from "node:${moduleName}";`,
+      options: preferUsingProtocol,
+      errors: [
+        { messageId: 'requireNodeProtocol', data: { moduleName } },
+      ],
+    },
+    testVersion('>= 7', () => ({
+      code: `
+        async function foo() {
+          const x = await import('${moduleName}');
+        }
+      `,
+      output: `
+        async function foo() {
+          const x = await import('node:${moduleName}');
+        }
+      `,
+      options: preferUsingProtocol,
+      parserOptions: useNewerParser,
+      errors: [
+        { messageId: 'requireNodeProtocol', data: { moduleName } },
+      ],
+    })),
+  )),
+
+  {
+    code: 'import fs from "fs/promises";',
+    output: 'import fs from "node:fs/promises";',
+    options: preferUsingProtocol,
+    errors: [
+      {
+        messageId: 'requireNodeProtocol',
+        data: { moduleName: 'fs/promises' },
+      },
+    ],
+    settings,
+  },
+  {
+    code: 'export {default} from "fs/promises";',
+    output: 'export {default} from "node:fs/promises";',
+    options: preferUsingProtocol,
+    errors: [
+      {
+        messageId: 'requireNodeProtocol',
+        data: { moduleName: 'fs/promises' },
+      },
+    ],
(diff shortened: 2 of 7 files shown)


config/recommended.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
     'import/no-named-as-default': 'warn',
     'import/no-named-as-default-member': 'warn',
     'import/no-duplicates': 'warn',
+    'import/prefer-node-builtin-imports': 'warn',
```
ljharb, line 20: adding things to recommended is a breaking change

docs/rules/prefer-node-builtin-imports.md
```diff
@@ -0,0 +1,39 @@
+# import/prefer-node-builtin-imports
```
ljharb, line 1: I’m not interested in a rule that solely pushes using the protocol - if anything, it’d be a rule that can be configured to require, or forbid, the protocol.

ljharb (changes requested):

docs/rules/prefer-node-builtin-imports.md
GoldStrikeArch: @ljharb What config option should be from user's perspective. Just a boolean to toggle this? If yes then I don't see how is it different from simply have it `on` or `off`

ljharb: "off" is the current state - where you can have `import 'fs'` and `import 'node:path'` in the same file. One rule setting would error on the first one, and require `node:fs`. The other would error on the second one, and require `path`.

Thompson1985 (approved):

ljharb:
@GoldStrikeArch are you still interested in completing this PR?

GoldStrikeArch:
> @GoldStrikeArch are you still interested in completing this PR?

Yeah, forgot about this. I will try to finish it on this weekend

sevenc-nanashi:
Any updates on this PR? I'm very excited about this rule.

(I'd like to take over this PR if needed)

ljharb:
@sevenc-nanashi if you'd like to do so, please do NOT make a new PR - instead, comment here with a link to your branch containing this PR, rebased and updated per https://github.com/import-js/eslint-plugin-import/pull/3024#discussion_r1665739477, and i'll pull in the changes.

sevenc-nanashi:
I rebased and added some changes according to the review:
https://github.com/sevenc-nanashi/eslint-plugin-import/tree/feat/prefer-node-protocol?tab=readme-ov-file

- This rule now accepts `always` or `never`.
- Fixed tests.

I think the name should be `prefer-node-protocol`, as the name `prefer-node-builtin-imports` sounds like preferring using node's builtin over other packages.

ljharb:
@sevenc-nanashi i don't think it should have "prefer" in it at all, because the rule won't be expressing an opinion that using, or avoiding, the protocol is preferred, but I agree that "node protocol" is better wording.

Perhaps `enforce-node-protocol-usage`?

sevenc-nanashi:
@ljharb I renamed to `enforce-node-protocol-usage`.

ljharb:
The rule should either default to “never” - the far more universal behavior - or have no default at all.

The docs will need a lot of work too - deno and hun only support the protocol, i believe, so you’d want to set the rule to “always”; if you’re supporting a node version, bundler, or tool that doesn’t understand the node protocol, you’d want to set it to “never”.

Additionally, prefix-only core modules, like node:test, node:sea, node:sqlite, etc need to remain unaffected by the “never” setting. Although, because module.builtinModules doesn’t include prefix-only modules, you may not need to do anything here but add test cases.

sevenc-nanashi:
@ljharb 
I made these changes:
- The rule will throw error when the option is not specified
- Added test cases that `require("node:test")` is valid in `'never'`

ljharb:
k, i've pulled in and rebased the changes. I also made it use is-core-module since module.builtinModules isn't available in older node versions.

docs/rules/enforce-node-protocol-usage.md
```diff
@@ -0,0 +1,73 @@
...
+## Rule Details
+
+This rule enforces that builtins node imports are using, or omitting, the `node:` protocol.
+
+Reasons to prefer using the protocol include:
+ - the code is more explicitly and clearly referencing a Node.js built-in module
+
+Reasons to prefer omitting the protocol include:
+ - some tools don't support the `node:` protocol
+ - the code is more portable, because import maps and automatic polyfilling can be used
```
ljharb, line 18: feel free to make suggestions here for more reasons in either direction

ljharb:
@sevenc-nanashi i haven't reviewed the tests yet; assuming that every module being tested is tested with both configurations, and that eg `node:test` and `node:sea` are tested, and that tests pass, then this should be good.

sevenc-nanashi:
@ljharb I made some changes:
- [Use object-based option schema](https://eslint.org/docs/latest/extend/custom-rules#:~:text=A%20full%20JSON%20Schema%20object%20that%20will%20validate%20the%20context.options%20array) to create mandatory option
- Added some test cases

ljharb:
hmm, lots of the tests are failing for me locally, but i'm not sure why. the logic seems correct.

sevenc-nanashi:
@ljharb Looks like the issue caused by the guard of visitor.

<details>
<summary>The fix</summary>

```diff
diff --git a/src/rules/enforce-node-protocol-usage.js b/src/rules/enforce-node-protocol-usage.js
index f8f3ee39..548f303b 100644
--- a/src/rules/enforce-node-protocol-usage.js
+++ b/src/rules/enforce-node-protocol-usage.js
@@ -96,11 +96,16 @@ module.exports = {
       url: docsUrl('enforce-node-protocol-usage'),
     },
     fixable: 'code',
-    schema: [
-      {
-        enum: ['always', 'never'],
-      },
-    ],
+    schema: {
+      type: 'array',
+      minItems: 1,
+      maxItems: 1,
+      items: [
+        {
+          enum: ['always', 'never'],
+        },
+      ],
+    },
     messages,
   },
   create(context) {
@@ -115,18 +120,12 @@ module.exports = {
         return checkAndReport(arg, context);
       },
       ExportNamedDeclaration(node) {
-        if (!isStringLiteral(node)) { return; }
-
         return checkAndReport(node.source, context);
       },
       ImportDeclaration(node) {
-        if (!isStringLiteral(node)) { return; }
-
         return checkAndReport(node.source, context);
       },
       ImportExpression(node) {
-        if (!isStringLiteral(node)) { return; }
-
         return checkAndReport(node.source, context);
       },
     };
```

</details>

In my environment, all tests passed with this patch.

sevenc-nanashi:
btw: is it actually worth supporting Node v4?

- No one uses that,
- It's dangerous not to upgrade Node,
- It makes CI slower,
- It makes installation slower and node_modules heavier, (because of polyfills)
- It makes development slower, (I saw spread expression was replaced with `[].concat`; there would be less code changes for backward compatibility like this)
- And users can choose not to upgrade `eslint-plugin-import` (There should be a major version change though)

ljharb:
@sevenc-nanashi yes, people do; it's not dangerous in many contexts; i don't care about CI speed; disk space is infinite and free and npm installation is very fast; the `[].concat` is something i'd do in any codebase, spread is slow and should forever be avoided; doing a major version bump like that means that users on the old major are stuck without features, bugfixes, and security patches, so it's actually the most harmful thing any package can do for the ecosystem.

ljharb:
So one thing to note: the list of core modules is dependent on the node version that eslint is running in.

I think perhaps it might make sense to add a `settings` option for "node version" that this rule uses, that will override the current eslint node version (`isCoreModule` has that API already). Thoughts?

ljharb:
ok, updated. lmk what yall think

docs/rules/enforce-node-protocol-usage.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+✅ Valid
+
+```js
+import fs from 'fs';
+export { promises } from 'fs';
+import * as test from 'node:test';
+// require
+const fs = require('fs/promises');
```
sevenc-nanashi, line 70: I think there should be an explanation about why this is valid.
```suggestion
import fs from 'fs';
export { promises } from 'fs';

// require
const fs = require('fs/promises');

// This rule will not enforce not using `node:` protocol when the module is only available under the `node:` protocol.
import * as test from 'node:test';
```

tests/src/rules/enforce-node-protocol-usage.js
```diff
@@ -0,0 +1,336 @@
...
+
+  invalid: [].concat(
+    // Prefer using the protocol
+    // in node versions without `node:`, the rule should not report
+    satisfies('^14.18 || >= 16') ? invalidTests.map((testCase) => test({
+      ...testCase,
+      errors: testCase.errors.map(({ messageId, data, ...testCase }) => ({
+        ...testCase,
+        message: rule.meta.messages[messageId].replace(/{{moduleName}}/g, data.moduleName),
+      })),
```
sevenc-nanashi, line 320: Cannot this code be replaced using `messageId` and `data`, like above? (There should be a comment why not use that)

ljharb: messageId isnt an option because we support older eslint versions that don’t have that feature. I kept it here because it’ll be a smaller diff when i make improvements in the future.

ljharb (approved):

JounQin:
Just curious.

gshpychka:
> @gshpychka it's pretty valid question to understand when this feature will be released and publicly available.

Yes, your question was, but @JounQin's question was off-topic.

JounQin:
I don't think it's off-topic because the two rules aim to same behavior.

ljharb:
It’s off topic because this repo is about this project, and there’s no reason to discuss alternative projects here.

PR #3029, opened 2024-07-14
webpack resolver: Simplify loop

fregante:
This PR reduces the amount of code run by using a plain for loop with early return. I was hesitant to make the change at first but I noticed that there are a lot of similar loops in the code so it shouldn't be a problem to use the same here

### DIFF
--- a/resolvers/webpack/index.js
+++ b/resolvers/webpack/index.js
@@ -3,7 +3,6 @@
 const findRoot = require('find-root');
 const path = require('path');
 const isEqual = require('lodash/isEqual');
-const find = require('array.prototype.find');
 const interpret = require('interpret');
 const fs = require('fs');
 const isCore = require('is-core-module');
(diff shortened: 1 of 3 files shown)


resolvers/webpack/index.js
```diff
@@ -293,17 +292,20 @@ const MAX_CACHE = 10;
...
-  let cached = find(_cache, function (entry) { return isEqual(entry.key, cacheKey); });
-  if (!cached) {
-    cached = {
-      key: cacheKey,
-      value: createResolveSync(configPath, webpackConfig, cwd),
-    };
-    // put in front and pop last item
-    if (_cache.unshift(cached) > MAX_CACHE) {
-      _cache.pop();
+  for (const entry of _cache) {
```
ljharb, line 295: this uses for-of tho, which is slower than standard array iteration - a standard for loop would be the replacement here

fregante: 2c577cb  can be undone if plain `for` loops are preferable, but I see both `for-of` and plain `for` loops used everywhere in the repo. I'd say `for-of` gives a good balance between performance (still better than polyfilling .find()) and readability (over plain `for` loops)

fregante: Let me know if I should revert `2c577cb` or if for-of is good enough here

fregante: Done, it uses a regular for loop now

ljharb (approved):

PR #3032, opened 2024-08-06
[Fix] `no-named-as-default`: allow named specifier if equivalent to default

### LINKED ISSUE #1594: no-named-as-default warn/error when value is both named export and default export
Example (here I want to export everything from one or merge more files, then set a default export):
```JavaScript
// filename is: ./file/index.js
export * from './file'; // this contains export const file = ...
export * from './anotherfile';
export {file as default} from './file';

// filename is ./main.js
import file from './file';  // import from directory file, resolving to ./file/index.js
/*
 Line above gives warn/error:
 Using exported name 'file' as identifier
 for default export.eslint(import/no-named-as-default)
*/
```

akwodkiewicz:
New behaviour / fix for rule `no-named-as-default`: the rule will **not** show an error if the thing that was imported with `someName` is exported from the source both as `default` and `someName`

There's a big caveat related to how the _thing_ is defined in the imported module. If it's a re-export, then this fix will work. If it's defined in the imported module -- it won't work.

This PR fixes all the issues with 
```ts
import userEvent from '@testing-library/user-event';
```
because, luckily to us, `userEvent` is re-exported there and not defined in the imported file.

Fixes #1594

### DIFF
--- a/src/rules/no-named-as-default.js
+++ b/src/rules/no-named-as-default.js
@@ -15,28 +15,71 @@ module.exports = {
 
   create(context) {
     function checkDefault(nameKey, defaultSpecifier) {
+      /**
+       * For ImportDefaultSpecifier we're interested in the "local" name (`foo` for `import {bar as foo} ...`)
+       * For ExportDefaultSpecifier we're interested in the "exported" name (`foo` for `export {bar as foo} ...`)
+       */
+      const analyzedName = defaultSpecifier[nameKey].name;
+
       // #566: default is a valid specifier
-      if (defaultSpecifier[nameKey].name === 'default') { return; }
+      if (analyzedName === 'default') { return; }
 
       const declaration = importDeclaration(context, defaultSpecifier);
+      /** @type {import('../exportMap').default | null} */
+      const importedModule = ExportMapBuilder.get(declaration.source.value, context);
+      if (importedModule == null) { return; }
 
-      const imports = ExportMapBuilder.get(declaration.source.value, context);
-      if (imports == null) { return; }
-
-      if (imports.errors.length) {
-        imports.reportErrors(context, declaration);
+      if (importedModule.errors.length > 0) {
+        importedModule.reportErrors(context, declaration);
         return;
       }
 
-      if (imports.has('default') && imports.has(defaultSpecifier[nameKey].name)) {
-
-        context.report(
-          defaultSpecifier,
-          `Using exported name '${defaultSpecifier[nameKey].name}' as identifier for default export.`,
-        );
-
+      if (!importedModule.hasDefault) {
+        // The rule is triggered for default imports/exports, so if the imported module has no default
+        // this means we're dealing with incorrect source code anyway
+        return;
       }
+
+      if (!importedModule.has(analyzedName)) {
+        // The name used locally for the default import was not even used in the imported module.
+        return;
+      }
+
+      /**
+       * FIXME: We can verify if a default and a named export are pointing to the same symbol only
+       * if they are both `reexports`. In case one of the symbols is not a re-export, but defined
+       * in the file, the ExportMap structure has no info about what actually is being exported --
+       * the value in the `namespace` Map is an empty object.
+       *
+       * To solve this, it would require not relying on the ExportMap, but on some other way of
+       * accessing the imported module and its exported values.
+       *
+       * Additionally, although `ExportMap.get` is a unified way to get info from both `reexports`
+       * and `namespace` maps, it does not return valid output we need here, and I think this is
+       * related to the "cycle safeguards" in the `get` function.
+       */
+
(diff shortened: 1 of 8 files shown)


src/rules/no-named-as-default.js
```diff
@@ -28,7 +35,15 @@ module.exports = {
...
       }
 
-      if (imports.has('default') && imports.has(defaultSpecifier[nameKey].name)) {
+      if (imports.has('default') && imports.has(analyzedName)) {
+
+        // #1594: the imported module exports the same thing via a default export and a named export
+        const namedExportInImportedModule = imports.reexports.get(analyzedName);
+        const defaultExportInImportedModule = imports.reexports.get('default');
+        if (defaultExportInImportedModule.getImport().path === namedExportInImportedModule.getImport().path
+          && defaultExportInImportedModule.local === namedExportInImportedModule.local) {
```
ljharb, line 44:
```suggestion
        if (
          defaultExportInImportedModule.getImport().path === namedExportInImportedModule.getImport().path
          && defaultExportInImportedModule.local === namedExportInImportedModule.local
        ) {
```

ljharb (commented):
LGTM pending comment

src/rules/no-named-as-default.js
akwodkiewicz: Apologies, but I don't believe that this particular line formatting is particularly better. The file is not Prettified anyway, and even if it was, then this line would be formatted even more differently than you suggest.

akwodkiewicz: Eventually, the comment is not relevant anymore, because of various changes in the final version.

ljharb: it's better because any time a `(` doesn't end a line, it should be on the same line as the corresponding `)`.

src/rules/no-named-as-default.js
```diff
@@ -15,28 +15,70 @@ module.exports = {
...
+      if (analyzedName === 'default') { return; }
 
       const declaration = importDeclaration(context, defaultSpecifier);
+      /** @type {import('../exportMap').default | null} */
+      const importedModule = ExportMapBuilder.get(declaration.source.value, context);
+      if (importedModule == null) { return; }
 
-      const imports = ExportMapBuilder.get(declaration.source.value, context);
-      if (imports == null) { return; }
+      if (importedModule.errors.length) {
```
ljharb, line 32:
```suggestion
      if (importedModule.errors.length > 0) {
```

akwodkiewicz: I did not want to change the existing logic of error reporting for the part I don't fully understand. You sure this change is safe?

akwodkiewicz: Added in fd567c69499b16a4a14017a7bda2b6cb16b612e3

src/rules/no-named-as-default.js
akwodkiewicz: This is the first time I hear such a rule, but I'm glad we don't have any outstanding formatting issues anymore :+1:

src/rules/no-named-as-default.js
ljharb: oh fair, but yeah i think this would always be a safe change :-)

src/rules/no-named-as-default.js
```diff
@@ -15,28 +15,70 @@ module.exports = {
...
-      if (imports == null) { return; }
+      if (importedModule.errors.length > 0) {
+        importedModule.reportErrors(context, declaration);
+        return;
+      }
 
-      if (imports.errors.length) {
-        imports.reportErrors(context, declaration);
+      if (!importedModule.hasDefault) {
+        // Impossible: how could we even get here? Kept in case of unexpected regressions.
```
ljharb, line 38: why would this be impossible? wouldn't a module that only has named exports qualify?

akwodkiewicz: The rule itself is triggered only for `ImportDefaultSpecifier`/`ExportDefaultSpecifier`. So the question is: is it possible for `{Import,Export}DefaultSpecifier` rule to be triggered if there are no default exports in the imported module?

I expect that... no? But I don't know. That would be improper code anyway, that's why we just skip the analysis with `return`

ljharb: it would definitely be code that fails at runtime, but it's entirely possible to write that code and have it parse :-)

akwodkiewicz: Added additional tests for this as well 3a0af7ecd2d2c1c1effbb6fcdbc3165c38e6c026

akwodkiewicz: Added additional tests for this as well in 7e2e78f8b33a19a47b660fba9e8ac2faf8822c6c

ljharb (approved):

PR #3039, opened 2024-08-23
[utils] [fix] `parse`: Also delete parserOptions.projectService

Mysak0CZ:
Recently typescript-eslint released version 8. One of the breaking changes introduced there was renaming `EXPERIMENTAL_useProjectService` to `projectService`.
xref: https://typescript-eslint.io/blog/announcing-typescript-eslint-v8-beta/#project-service

This fix deletes the new property as well to let this plugin be compatible with v8. During my testing I didn't encounter any other issues than this.

Change is based on prior change to get project service option working:
ref #2963

### DIFF
--- a/utils/CHANGELOG.md
+++ b/utils/CHANGELOG.md
@@ -8,6 +8,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Changed
 - [types] use shared config (thanks [@ljharb])
 
+### Fixed
+- `parse`: also delete `parserOptions.projectService` ([#3039], thanks [@Mysak0CZ])
+
 ## v2.8.1 - 2024-02-26
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ljharb:
The change is simple, but we're not testing on v6 of that parser, let alone v8 yet. I think we need to add those tests first.

(related, https://github.com/jsx-eslint/eslint-plugin-react/pull/3629)

Mysak0CZ:
Hi @ljharb !
Do I understand correctly, that you accepted change for getting compatible with a _beta_ of v8, but don't want this plugin to be compatible with full release of v8?

ljharb:
What v6 changes did i accept without testing in v6?

Mysak0CZ:
> What v6 changes did i accept without testing in v6?

The PR I mentioned (#2963) is the exact same as this one - it adds support for projectservice, which was introduced in v6 of ts-eslint by deleting a new property named `EXPERIMENTAL_useProjectService`. This PR does the exact same, only for the new name that property got in v8.
As dev dependencies in package.json define only parser up to major version 5, the `EXPERIMENTAL_useProjectService` was definitely not tested as in version 5 it doesn't exist.

ljharb:
eslint-module-utils v2.8.2 is now published, with this change.

PR #3043, opened 2024-09-01
[New] `import/order`: add support for named imports

### LINKED ISSUE #1787: Allow enforcement of import-specifiers
Imports can have multiple specifiers (objects which are being imported).
Would be awesome if there would be a way to enforce alphabetization there, too:

Example:
```js
import { isNull, isNaN } from "util"; // reports an error: "isNaN" should appear before "isNull"
```

### LINKED ISSUE #2553: [Suggestion]: `import/named-order` : Enforce to sort specifiers of import, export, require
Sorry, It was my first time posting a PR, so I didn't know that I had to create an issue first. 

I experienced a significant decrease in readability as the number of specifiers increased in the named import. So, I made this rule.

I already add a rule to PR https://github.com/import-js/eslint-plugin-import/pull/2552

manuth:
## `import/order`: add support for named imports
Changes made in this PR change the `import/order` rule to allow configuring the order of named specifiers in `import`, `export` and `require` statements.

This will report errors and automatically fix expressions as shown in this example:
***Reports error:***
```ts
import { Bravo, Alpha } from 'foo';
const { Delta, Charlie } = require('foo');
export { Foxtrot, Echo } from 'foo';
export { Alpha as Hotel, India, Alpha as Golf };
```

***Fixed code:***
```ts
import { Alpha, Bravo } from 'foo';
const { Charlie, Delta } = require('foo');
export { Echo, Foxtrot } from 'foo';
export { Alpha as Golf, Alpha as Hotel, India } from 'foo';
```

This feature heavily uses the pre-existing functions used for re-ordering import/export statements and thus does not require a lot of changes.

To address the requirement brought forward by @ronparkdev, there are 3 options `named.import`, `named.export` and `named.require` for individually forcing ordering of named items for the specified category.

Furthermore, in response to the comment written by @JensDll, there is an option `types` which allows to specify the order of the specifiers based on their category:
  - `mixed`: all specifiers are ordered alphabetically
  - `types-first`: `type` imports must occur first
  - `types-last`: value imports must occur first

With this option, one could force `type` imports to occur last by setting the options of the `order` rule to the following:

```json
{
  "named": {
    "enabled": true,
    "types": "types-last"
  },
  "alphabetize": { "order": "asc" }
}
```

## Things to consider
  - ~~Following tiny bit of code makes use of `RegExp`:
    https://github.com/import-js/eslint-plugin-import/blob/60b04933341a4c9eb5f2f3484e9a0bdd685c0df9/src/rules/order.js#L239
    https://github.com/import-js/eslint-plugin-import/blob/60b04933341a4c9eb5f2f3484e9a0bdd685c0df9/src/rules/order.js#L250~~
  - ~~The lines between the `default` and the `type`/`value` groups are a little blurred as a `default` re-export technically still is either a `value` or a `type` export. `default` will take highest priority which means that both `export { type default as Readline } from 'foo'` and ` export { default as Readline } from 'foo'` will have their group set to `default`.~~
  - named specifiers are not only sorted by their source name, but also by their alias
  - When two specifiers have the same source name (like, say, `const { Alpha: Bravo, Alpha } = require('foo');`), error messages will contain both the original name and the alilas for preventing confusion:
    `` `Alpha` import should occur before import of `Alpha as Bravo` ``
  - For enabling forced ordering for all kinds of expressions, the `named` or the `named.enabled` option can be set to `true`. However, `named.all` might be a more suitable option name (?)

## Related
  - #2553, the original proposal
  - #1787 (tagging @andrewbranch, bc this PR might be interesting to you)

Merging this PR will fix #2553 and fix #1787

### DIFF
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -285,6 +285,78 @@ import index from './';
 import sibling from './foo';
 ```
 
+### `named: true|false|{ enabled: true|false, import: true|false, export: true|false, require: true|false, cjsExports: true|false, types: mixed|types-first|types-last }`
+
+Enforce ordering of names within imports and exports:
+
+ - If set to `true`, named imports must be ordered according to the `alphabetize` options
+ - If set to `false`, named imports can occur in any order
+
+`enabled` enables the named ordering for all expressions by default.
+Use `import`, `export` and `require` and `cjsExports` to override the enablement for the following kind of expressions:
+
+ - `import`:
+
+   ```ts
+   import { Readline } from "readline";
+   ```
+
+ - `export`:
+
+   ```ts
+   export { Readline };
+   // and
+   export { Readline } from "readline";
+   ```
+
+ - `require`
+
+   ```ts
+   const { Readline } = require("readline");
+   ```
+
+ - `cjsExports`
+
+   ```ts
+   module.exports.Readline = Readline;
+   // and
+   module.exports = { Readline };
+   ```
+
+The `types` option allows you to specify the order of `import`s and `export`s of `type` specifiers.
+Following values are possible:
+
+ - `types-first`: forces `type` specifiers to occur first
+ - `types-last`: forces value specifiers to occur first
+ - `mixed`: sorts all specifiers in alphabetical order
+
+The default value is `false`.
+
+Example setting:
+
+```ts
+{
+  named: true,
+  alphabetize: {
+    order: 'asc'
+  }
+}
+```
+
+This will fail the rule check:
+
+```ts
+/* eslint import/order: ["error", {"named": true, "alphabetize": {"order": "asc"}}] */
+import { compose, apply } from 'xcompose';
+```
+
+While this will pass:
+
+```ts
+/* eslint import/order: ["error", {"named": true, "alphabetize": {"order": "asc"}}] */
+import { apply, compose } from 'xcompose';
+```
+
 ### `alphabetize: {order: asc|desc|ignore, orderImportKind: asc|desc|ignore, caseInsensitive: true|false}`
 
 Sort the order within each group in alphabetical manner based on **import path**:
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -3,20 +3,25 @@
 import minimatch from 'minimatch';
 import includes from 'array-includes';
 import groupBy from 'object.groupby';
-import { getSourceCode } from 'eslint-module-utils/contextCompat';
+import { getScope, getSourceCode } from 'eslint-module-utils/contextCompat';
+import trimEnd from 'string.prototype.trimend';
 
 import importType from '../core/importType';
 import isStaticRequire from '../core/staticRequire';
 import docsUrl from '../docsUrl';
+
+const categories = {
+  named: 'named',
+  import: 'import',
+  exports: 'exports',
+};
 
 const defaultGroups = ['builtin', 'external', 'parent', 'sibling', 'index'];
 
 // REPORTING AND FIXING
 
 function reverse(array) {
-  return array.map(function (v) {
-    return { ...v, rank: -v.rank };
-  }).reverse();
+  return array.map((v) => ({ ...v, rank: -v.rank })).reverse();
 }
 
 function getTokensOrCommentsAfter(sourceCode, node, count) {
@@ -131,6 +136,26 @@ function findStartOfLineWithComments(sourceCode, node) {
   return result;
 }
 
+function findSpecifierStart(sourceCode, node) {
+  let token;
+
+  do {
+    token = sourceCode.getTokenBefore(node);
+  } while (token.value !== ',' && token.value !== '{');
+
+  return token.range[1];
+}
+
+function findSpecifierEnd(sourceCode, node) {
+  let token;
+
+  do {
+    token = sourceCode.getTokenAfter(node);
+  } while (token.value !== ',' && token.value !== '}');
+
+  return token.range[0];
+}
+
 function isRequireExpression(expr) {
   return expr != null
     && expr.type === 'CallExpression'
@@ -170,6 +195,49 @@ function isPlainImportEquals(node) {
   return node.type === 'TSImportEqualsDeclaration' && node.moduleReference.expression;
 }
 
+function isCJSExports(context, node) {
+  if (
+    node.type === 'MemberExpression'
+    && node.object.type === 'Identifier'
+    && node.property.type === 'Identifier'
+    && node.object.name === 'module'
+    && node.property.name === 'exports'
+  ) {
+    return getScope(context, node).variables.findIndex((variable) => variable.name === 'module') === -1;
+  }
+  if (
+    node.type === 'Identifier'
+    && node.name === 'exports'
+  ) {
+    return getScope(context, node).variables.findIndex((variable) => variable.name === 'exports') === -1;
+  }
+}
+
+function getNamedCJSExports(context, node) {
+  if (node.type !== 'MemberExpression') {
+    return;
+  }
+  const result = [];
+  let root = node;
+  let parent = null;
+  while (root.type === 'MemberExpression') {
+    if (root.property.type !== 'Identifier') {
+      return;
+    }
+    result.unshift(root.property.name);
+    parent = root;
+    root = root.object;
+  }
+
+  if (isCJSExports(context, root)) {
+    return result;
+  }
+
+  if (isCJSExports(context, parent)) {
+    return result.slice(1);
+  }
+}
+
 function canCrossNodeWhileReorder(node) {
   return isSupportedRequireModule(node) || isPlainImportModule(node) || isPlainImportEquals(node);
 }
@@ -190,6 +258,12 @@ function canReorderItems(firstNode, secondNode) {
 }
 
 function makeImportDescription(node) {
+  if (node.type === 'export') {
+    if (node.node.exportKind === 'type') {
+      return 'type export';
+    }
+    return 'export';
+  }
   if (node.node.importKind === 'type') {
     return 'type import';
   }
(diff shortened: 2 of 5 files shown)


docs/rules/order.md
```diff
@@ -285,6 +285,49 @@ import index from './';
...
 ```
 
+### `named: true|false|{enabled: true|false, import: true|false, export: true|false, require: true|false, groups: [array]}`
+
+Enforce ordering of names within imports:
+
+ - If set to `true`, named imports must be ordered according to the `alphabetize` options
+ - If set to `false`, named imports can occur in any order
+
+`import`, `export` and `require` allows forcing ordering of named `import`s/`export`s for the corresponding kind of expression.
```
ljharb, line 295:
```suggestion
`import`, `export`, and `require` allows forcing ordering of named `import`s/`export`s for the corresponding kind of expression.
```

ljharb: also, what about `exports.foo =` or `module.exports = { a, b, c }` (CJS quasi named exports)?

docs/rules/order.md
```diff
@@ -285,6 +285,49 @@ import index from './';
...
+Use `enabled` to force ordering all of the aforementioned groups.
+
+The `groups` option allows to specify the order of the following kind of named `import`/`export` expressions:
+
+ - `value`: Importing/exporting ordinary values
+ - `default`: Exporting the `default` member in an expression like `export { default as Console } from './my-awesome-console.js'`
+ - `type`: Explicitly importing or exporting types: `import { type Readline } from 'node:readline'`
+
+The default value is `false` and the default `groups`-option is `['default']`.
+
```
ljharb, line 305: can you help me understand the use case for wanting name order to only be enforced in specific groups, or in different orders inside the same file?

src/rules/order.js
```diff
@@ -198,58 +204,113 @@ function makeImportDescription(node) {
...
 
-  let newCode = sourceCode.text.substring(secondRootStart, secondRootEnd);
-  if (newCode[newCode.length - 1] !== '\n') {
-    newCode = `${newCode}\n`;
+  if (firstNode.displayName === secondNode.displayName) {
+    for (const node of [firstNode, secondNode]) {
+      if (node.alias) {
+        node.displayName = `${node.displayName} as ${node.alias}`;
+      }
+    }
```
ljharb, line 232: i'm a bit confused here why we're just overwriting the first node's alias when the second node has one - either way, let's avoid for-of here, it's unnecessary for 2 hardcoded items and it's slower than other forms of iteration.

src/rules/order.js
```diff
@@ -198,58 +204,113 @@ function makeImportDescription(node) {
...
-        [secondRootStart, firstRootEnd],
-        sourceCode.text.substring(secondRootEnd, firstRootEnd) + newCode,
-      )),
-    });
+  function trimEnd(text) {
+    let i = text.length - 1;
+    while (i >= 0 && text[i] === ' ') {
+      i--;
+    }
+    return text.substring(0, i + 1);
```
ljharb, line 244:
```suggestion
    return text.slice(0, i + 1);
```

src/rules/order.js
```diff
@@ -198,58 +204,113 @@ function makeImportDescription(node) {
...
+  }
+
+  if (named) {
+    const firstCode = sourceCode.text.substring(firstRootStart, firstRootEnd);
+    const secondCode = sourceCode.text.substring(secondRootStart, secondRootEnd);
+
+    if (order === 'before') {
+      const trimmed = trimEnd(secondCode);
+      const gapCode = sourceCode.text.substring(firstRootEnd, secondRootStart - 1);
+      const whiespaces = secondCode.substring(trimmed.length);
```
ljharb, line 254:
```suggestion
      const whitespaces = secondCode.slice(trimmed.length);
```

src/rules/order.js
```diff
@@ -198,58 +204,113 @@ function makeImportDescription(node) {
...
+        node: secondNode.node,
+        message,
+        fix: (fixer) => fixer.replaceTextRange(
+          [firstRootStart, secondRootEnd],
+          `${trimmed},${firstCode}${gapCode}${whiespaces}`),
+      });
+    } else if (order === 'after') {
+      const trimmed = trimEnd(firstCode);
+      const gapCode = sourceCode.text.substring(secondRootEnd + 1, firstRootStart);
+      const whiespaces = firstCode.substring(trimmed.length);
```
ljharb, line 265:
```suggestion
      const whitespaces = firstCode.slice(trimmed.length);
```

src/rules/order.js
```diff
@@ -638,6 +721,29 @@ module.exports = {
...
+                properties: {
+                  enabled: { type: 'boolean' },
+                  groups: groupsType,
+                },
+                additionalProperties: false,
+              }, {
+                properties: {
+                  import: { type: 'boolean' },
+                  export: { type: 'boolean' },
+                  require: { type: 'boolean' },
```
ljharb, line 740: i think i'd expect here that `enabled` could always be present, and `import`, `export`, and `require` would just override whatever base `enabled` implies?

src/rules/order.js
```diff
@@ -700,6 +823,30 @@ module.exports = {
       return importMap.get(node);
     }
 
+    function makeNamedOrderReport(context, namedImports) {
+      if (namedImports.length > 1) {
+        const imports = namedImports.map(
+          (namedImport) => {
+            const name = namedImport.value;
+            const kind = namedImport.kind === 'type' ? namedImport.kind : name === 'default' ? 'default' : 'value';
+            const rank = named.groups.findIndex((entry) => Array.isArray(entry) ? entry.indexOf(kind) >= 0 : entry === kind);
```
ljharb, line 832:
```suggestion
            const rank = named.groups.findIndex((entry) => [].concat(entry).indexOf(kind) > -1);
```

src/rules/order.js
```diff
@@ -717,6 +864,22 @@ module.exports = {
...
+              node.specifiers.filter(
+                (specifier) => specifier.type === 'ImportSpecifier').map(
+                (specifier) => ({
+                  node: specifier,
+                  value: specifier.imported.name,
+                  type: 'import',
+                  kind: specifier.importKind,
+                  ...specifier.local.range[0] !== specifier.imported.range[0] ? {
+                    alias: specifier.local.name,
+                  } : {},
```
ljharb, line 880:
```suggestion
                  ...specifier.local.range[0] !== specifier.imported.range[0] && {
                    alias: specifier.local.name,
                  },
```

src/rules/order.js
```diff
@@ -771,6 +934,40 @@ module.exports = {
...
+            makeNamedOrderReport(
+              context,
+              node.id.properties.map(
+                (prop) => ({
+                  node: prop,
+                  value: prop.key.name,
+                  type: 'require',
+                  ...prop.key.range[0] !== prop.value.range[0] ? {
+                    alias: prop.value.name,
+                  } : {},
```
ljharb, line 949:
```suggestion
                  ...prop.key.range[0] !== prop.value.range[0] && {
                    alias: prop.value.name,
                  },
```

src/rules/order.js
```diff
@@ -771,6 +934,40 @@ module.exports = {
...
+            context,
+            node.specifiers.map(
+              (specifier) => ({
+                node: specifier,
+                value: specifier.local.name,
+                type: 'export',
+                kind: specifier.exportKind,
+                ...specifier.local.range[0] !== specifier.exported.range[0] ? {
+                  alias: specifier.exported.name,
+                } : {},
```
ljharb, line 967:
```suggestion
                ...specifier.local.range[0] !== specifier.exported.range[0] && {
                  alias: specifier.exported.name,
                },
```

ljharb (changes requested):
There's a number of lines not covered by tests (see the annotations in the diff tab)

docs/rules/order.md
manuth: I probably didn't do much of a great job explaining this option.
This option allows to specify the order of the corresponding groups of specifiers.

So, the default `['default']` expresses "`default` first, everything else second" while `[['default', 'value'], 'type']` expresses "`default` and `value` specifiers first, `type` specifiers second" and `[]` expresses that every group has the same rank.

So this option allows specifying the order of the groups rather than whether or not they are enabled.

src/rules/order.js
manuth: Alright, will do!

docs/rules/order.md
ljharb: ok so specifically, this is about determine the relative order of "default" vs "non-default values" vs "type imports"?

note that type imports can be a default as well (`import type foo from 'bar'`), and `import { default }` is a "named" import of the default, so this might get a bit complex.

In general, I think the simple thing to do here is support an option for types vs values, the choices being "mixed", "types first", or "types last". Would that be sufficient, do you think?

src/rules/order.js
manuth: Sounds like a great idea - I'm not sure how to efficiently aggregate the configuration, then, though
A bunch of `"{key}" in options.named` calls?

docs/rules/order.md
manuth: The `import type foo from 'bar'` would be untouched by the `named`-ordering, as the new logic only affects the named specifiers part (the `{ A, B, C }` part in `import foo, { A, B, C } from 'bar').
I think having the choices `mixed`, `types first` and `types last` would be more simple to use. I thought it might be nice to give the users full control, though.

If you think the `mixed`, `types first/last` option is more suitable, I will work on changing the code accordingly.

ljharb: ahhh. if we're only talking about inside the curly braces, i don't think "default" should get any special treatment. If you want it to follow the semantics of being a default, then you use it in default position and not as a name.

docs/rules/order.md
manuth: I will try to add support for these as well
I totally forgot about these.

src/rules/order.js
ljharb: hm, yeah that'll be tricky. maybe indeed something like:
```js
{
  import: 'import' in options ? options.import : options.enabled,
  export: 'export' in options ? options.export : options.enabled,
  require: 'require' in options ? options.require : options.enabled,
}

File: docs/rules/order.md (hunk 2 above)
Comment (manuth) [reply docs/rules/order.md:305]: Oh - that's true, though, yes
I think I will drop the special treatment of `default` out, then!

File: src/rules/order.js [hunk 12]
```diff
@@ -225,10 +225,11 @@ function fixOutOfOrder(context, firstNode, secondNode, order, named) {
...
   if (firstNode.displayName === secondNode.displayName) {
-    for (const node of [firstNode, secondNode]) {
-      if (node.alias) {
-        node.displayName = `${node.displayName} as ${node.alias}`;
-      }
+    if (firstNode.alias) {
+      firstNode.displayName = `${firstNode.displayName} as ${firstNode.alias}`;
+    }
+    if (secondNode.alias) {
+      secondNode.displayName = `${secondNode.displayName} as ${secondNode.alias}`;
```
ljharb, line 232: maybe even, `if (secondNode.alias) { } else if (firstNode.alias) {` since the first node doesn't matter if the second one is set?

manuth: In my opinion, if I had a code like
```ts
import { Readline as RL, Readline as Interface } from "readline";
```
I would prefer the message `` `Readline as Interface` import should occur before import of `Readline as RL` `` over `` `Readline as Interface` import should occur before import of `Readline` ``

What do you think?

ljharb: ohhhh sorry you're setting each node's displayName separately, i just misread it. this is fine as is :-)

docs/rules/order.md
manuth: not sure what to do about them.
Only act on `module.exports` and `exports` assignments of the shape `{ a }` and `{ a: b }`?

ljharb: i think any `exports.foo` assignment, as well as any time `module.exports` is assigned an object literal.

manuth: This sounds a little different from what the `named` option does.
The `named` option orders import/export specifiers inside `import { … } from ""`, `export { … }`, `export { … } from ""` and `const { … } = require("")` expressions while ordering `exports.{…}` and `module.exports.{…}` assignments work on the entire file.

Do you think it might make sense to introduce a new option (something like `{ "import/order": { "exports": true } }` or whatnot) for this use case?

ljharb: `module.exports.x =` is semantically the same (or arguably, less clear) as `exports.x =`, so both of those patterns need to be recognized.

Similarly, `module.exports = { a: 1, b: 2 }` has two conceptual named exports, "a" and "b".

manuth: That's true, though - but unlike `module.exports` or `exports`-assignments, the aforementioned expressions `const { … } = require("foo");`, `import { … } from "foo";`, `export { … } from "foo";` and `export { … }` are way more strict in their syntax

I have quite a struggle to wrap my head around the `module.exports` and `exports` aspect, tbh

ljharb: I agree those are trickier to check for - but the AST can locate a "global" variable called "module" or "exports", and we have rules that target those already, so it should be doable.

manuth: What are your thoughts on adding an option separate from `named` for this use case?

ljharb: I think that one option for imported names, and another for exported names, would be fine - but the export one would have to include these CJS patterns.

manuth: I will give it a try

manuth: Alright!
I feel like it's working out properly

The new option is called `named.cjsExports`

manuth:
Alright! I guess I'm ready
What can I do about the tests of package versions seemingly unable to detect `type` exports?

ljharb:
I'll take a look at whatever tests fail.

manuth:
I had to switch to draft for an instant here, because I have noticed that the named ordering did not work if they are spread across multiple lines.

I was able to fix this now - thus switching back to `Ready for review`

ljharb:
@manuth any reason for the custom `trimEnd` function instead of the standard one?

manuth:
The `trimEnd` function doesn't seem to exist in older Node.js versions.

ljharb:
That's what the string.prototype.trimend package is for.

manuth:
Ooh weird - I felt like the checks once didn't run through cause of it
Maybe I'm mistaken

I'll try to change it back, then! Thanks 😄

ljharb:
oh no, what i mean is, add that package as a dependency and import it instead of rolling your own

ljharb:
k, i've rebased this and i think it's ready to land, but i'm going to wait til i get a patch release out first.

manuth:
Awesome, thank you so much 😄 
Looking forward to it!

ljharb (approved):

manuth:
Just saw a little inelegance in the JSON schema I have written
Will force push the change

PR #3045, opened 2024-09-02
[Test] add explicit marker for trailing whitespace in cases

G-Rath:
This ensures that IDEs like Webstorm won't trim trailing whitespace that is expected as part of test output


ljharb (approved):
Thanks! this is good test case hygiene anyways :-)

PR #2996, opened 2024-04-07
[New] support eslint 9

G-Rath:
Resolves #2948

PR #3046, opened 2024-09-03
[Test] actually push `namespace` case into `valid` array

G-Rath:
I came across this while working on #2996 - I'm pretty sure it's not intentional 😅

To make it a bit easier to spot, I've also put the closing `)` on a newline for both the `valid` and `invalid` arrays, even though that's not technically required to fix the test itself.

I've also checked the rest of the codebase and found no other instances of comma expressions being "used".


ljharb (approved):
oof, thanks

PR #3049, opened 2024-09-04
[utils] [new] add context compatibility helpers

michaelfaith:
This change adds helper functions to `eslint-module-utils` in order to add eslint v9 support to `eslint-plugin-import` in a backwards compatible way.

Contributes to #2996

### DIFF
--- a/utils/package.json
+++ b/utils/package.json
@@ -7,6 +7,7 @@
   },
   "main": false,
   "exports": {
+    "./contextCompat": "./contextCompat.js",
     "./ModuleCache": "./ModuleCache.js",
     "./ModuleCache.js": "./ModuleCache.js",
     "./declaredScope": "./declaredScope.js",
--- a/utils/resolve.js
+++ b/utils/resolve.js
@@ -5,6 +5,7 @@ exports.__esModule = true;
 const fs = require('fs');
 const Module = require('module');
 const path = require('path');
+const { getPhysicalFilename } = require('./contextCompat');
 
 const hashObject = require('./hash').hashObject;
 const ModuleCache = require('./ModuleCache').default;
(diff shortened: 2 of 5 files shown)


utils/package.json
```diff
@@ -7,6 +7,8 @@
   },
   "main": false,
   "exports": {
+    "./contextCompat": "./contextCompat.js",
+    "./contextCompat.js": "./contextCompat.js",
```
ljharb, line 11:
```suggestion
```
the only reason the other extensioned ones are there was to make it a nonbreaking change; extensions in specifiers are bad :-)

utils/resolve.js
```diff
@@ -229,7 +230,7 @@ const erroredContexts = new Set();
  */
 function resolve(p, context) {
   try {
-    return relative(p, context.getPhysicalFilename ? context.getPhysicalFilename() : context.getFilename(), context.settings);
+    return relative(p, getPhysicalFilename(context), context.settings);
```
ljharb, line 233: this seems like a noop change, is that correct? (iow, the contextCompat getPhysicalFilename is doing this exact same thing)

ljharb (commented):
Can I assume that all the uncovered lines will be covered in a different PR?

michaelfaith:
> Can I assume that all the uncovered lines will be covered in a different PR?

I don't mind adding unit tests in this pr, but there didn't seem to be any in the utils package, so I wasn't sure how you wanted that set up.

utils/resolve.js
michaelfaith: Almost, but not quite.  The `contextCompat` version would also check `context.filename` before falling back on `context.getFileName()`.  And `context.filename` is the go-forward api whereas `context.getFilename()` is deprecated.

utils/package.json
michaelfaith: Done.  And I agree, I was just following the pattern.

ljharb:
@michaelfaith usually tests for the utils package are done by writing tests for the things that use the utils.

utils/resolve.js
ljharb: aha, thanks :-)

ljharb:
This should be fine tho - I'll land it now, and then land a commit on top after it's published that *uses* these helpers.

ljharb (approved):

PR #3052, opened 2024-09-05
[Fix] `ExportMap` / flat config: include `languageOptions` in context

### LINKED ISSUE #3051: Missing documentation on how to setup eslint-plugin-import with flat confiuration and Typescript
Readme only contains information about how to setup typescript for this plugin using the old (non-flat) eslint configuration format.
https://github.com/import-js/eslint-plugin-import?tab=readme-ov-file#typescript

Upgraded to latest eslint-plugin-import and now the compat functions from @eslint/compat cant be used any longer (fixupConfigRules/fixupPluginRules) and i don't manage to lint my ts files any longer.

I get loads of errors like this:

```
 error  Parse errors in imported module '@emotion/cache': parserPath or languageOptions.parser is required! (undefined:undefined)  import/default
 ```
 
 I have a languageOptions.parser defined in my config:
 
 ```
 const tsParser = require('@typescript-eslint/parser');
 ...
   
 languageOptions: {
            parser: tsParser,
            ecmaVersion: 6,
            sourceType: 'module',
            parserOptions: {
                ecmaFeatures: {
                    jsx: true,
                },
            },
        },
```

michaelfaith:
This change fixes a bug with flat config support.  There is a function called `childContext` that's used by the ExportBuilder to "cleanse" the context object.  This function wasn't including the new `languageOptions` object, which contains the parser.  So by the time this cleansed context made it to the parse function, `languageOptions` wasn't there anymore.

Since `parserPath` was still being included in non-flat config scenarios, the parse function made it through ok and used `parserPath`.  However, once you shift to flat config, `parserPath` is no longer defined, and the actual parser object needs to be there.

Fixes #3051

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## [Unreleased]
 
+### Fixed
+- `ExportMap` / flat config: include `languageOptions` in context ([#3052], thanks [@michaelfaith])
+
 ## [2.30.0] - 2024-09-02
 
 ### Added
@@ -1129,6 +1132,7 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#3052]: https://github.com/import-js/eslint-plugin-import/pull/3052
 [#3036]: https://github.com/import-js/eslint-plugin-import/pull/3036
 [#3033]: https://github.com/import-js/eslint-plugin-import/pull/3033
 [#3018]: https://github.com/import-js/eslint-plugin-import/pull/3018
--- a/src/exportMap/childContext.js
+++ b/src/exportMap/childContext.js
@@ -10,7 +10,7 @@ let prevSettings = '';
  * also calculate a cacheKey, where parts of the cacheKey hash are memoized
  */
 export default function childContext(path, context) {
-  const { settings, parserOptions, parserPath } = context;
+  const { settings, parserOptions, parserPath, languageOptions } = context;
 
   if (JSON.stringify(settings) !== prevSettings) {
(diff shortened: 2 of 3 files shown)


ljharb (commented):
Thanks! Any way we can test for this?

michaelfaith:
> Thanks! Any way we can test for this?

For sure.  I can do that.  I'm having issues running the unit test suite locally (I think my version of node is too new for this version of `nyc`), so it may take a few updates to the PR to get things just right, using CI as my feedback loop.

michaelfaith:
Added a couple of tests for `childContext` (and I was able to get the tests running locally by updating the version of `nyc` to the latest, as well as making a few other tweaks to the setup).

rarenatoe:
Do you guys think this will get merged and released today?

ljharb (approved):
Thanks! It'd be really nice to have an actual eslint 9 test that matches the flow that exposed this issue, but this is sufficient for now :-)

michaelfaith:
> Thanks! It'd be really nice to have an actual eslint 9 test that matches the flow that exposed this issue, but this is sufficient for now :-)

Totally agree, and those'll be in @G-Rath's PR.  Their v9 tests are failing right now because they need this change.  So, once they bring this into their branch, those tests will be the v9 integration tests that cover this scenario.

G-Rath:
(fwiw I did a poor cherry-pick before this landed and pushed that up and at least some of the tests were improved, and now locally I've rebased and am just sorting out the handling of `env` before I push again and start digging into the whole failures etc etc 🙂)

ljharb:
I'm waiting for CI, and then I'll be publishing eslint-module-utils, and then updating `main` to require it and use the new helpers, after which you'll need to rebase, so feel free to wait :-)

PR #3057, opened 2024-09-09
[utils] [fix] `parse`: remove unneeded extra backticks

G-Rath:
I won't admit how many times I'd read this warning while working on #2996 before I realised there were these (cosmetic) double backticks 😅

### DIFF
--- a/utils/CHANGELOG.md
+++ b/utils/CHANGELOG.md
@@ -5,6 +5,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## Unreleased
 
+### Fixed
+- `parse`: remove unneeded extra backticks ([#3057], thanks [@G-Rath])
+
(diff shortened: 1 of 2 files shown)


ljharb:
They're intentional, to demarcate a JS property path versus prose words.

G-Rath:
ah ok - that doesn't entirely surprise me, but is that a typical thing? I don't think I've ever seen double backquotes used, and meanwhile the sentence starts with a single backquote of a property which is why I decided to open the PR.

Either way no biggie, feel free to close if you want to keep it as is 🙂

ljharb:
ohhh i get it, no, this is indeed just a typo :-) thanks, good catch

G-Rath:
sorry I should have included an actual snippet or screenshot 😅

ljharb (approved):

PR #3061, opened 2024-09-15
[utils] [fix] `parse`: espree parser isn't working with flat config

michaelfaith:
This change addresses an issue with using the espree parser with flat config.  `keysFromParser` is responsible for finding the `visitorKeys` from a combination of the parser instance, parserPath, and the results of a call to `parseForESLint`.  When using flat config, the `parserPath` will not be on the context object, and when using the espree parser there are no results from `parseForESLint`.  The espree parser _does_ provide visitor keys as a  prop on the parser itself.  However, this logic was only returning it when it found that "espree" was somewhere in the `parserPath` (which doesn't exist on flat config).

I changed it so that it first does the check for the old-school babel parser using parser path, and then just checks the parser instance for the presence of `VisitorKeys`, rather than using parserPath at all. The reason for the re-order is that if the old babel-eslint parser, for some reason has `VisitorKeys`, having the new condition first, would change the behavior.

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -14,6 +14,7 @@
 		"resolvers/*/test",
 		"scripts",
 		"memo-parser",
-		"lib"
+		"lib",
+		"examples"
 	]
 }
--- a/examples/flat/eslint.config.mjs
+++ b/examples/flat/eslint.config.mjs
@@ -20,6 +20,7 @@ export default [
       'import/no-dynamic-require': 'warn',
       'import/no-nodejs-modules': 'warn',
       'import/no-unused-modules': ['warn', { unusedExports: true }],
+      'import/no-cycle': 'warn',
     },
   },
 ];
--- a/examples/flat/src/depth-zero.js
+++ b/examples/flat/src/depth-zero.js
@@ -1,0 +1,3 @@
+import { foo } from "./es6/depth-one-dynamic";
+
+foo();
--- a/examples/flat/src/es6/depth-one-dynamic.js
+++ b/examples/flat/src/es6/depth-one-dynamic.js
@@ -1,0 +1,3 @@
+export function foo() {}
+
+export const bar = () => import("../depth-zero").then(({foo}) => foo);
--- a/examples/legacy/.eslintrc.cjs
+++ b/examples/legacy/.eslintrc.cjs
@@ -20,5 +20,6 @@ module.exports = {
     'import/no-dynamic-require': 'warn',
     'import/no-nodejs-modules': 'warn',
     'import/no-unused-modules': ['warn', { unusedExports: true }],
+    'import/no-cycle': 'warn',
   },
 };
--- a/examples/legacy/src/depth-zero.js
+++ b/examples/legacy/src/depth-zero.js
@@ -1,0 +1,3 @@
+import { foo } from "./es6/depth-one-dynamic";
+
+foo();
--- a/examples/legacy/src/es6/depth-one-dynamic.js
+++ b/examples/legacy/src/es6/depth-one-dynamic.js
@@ -1,0 +1,3 @@
+export function foo() {}
+
+export const bar = () => import("../depth-zero").then(({ foo }) => foo);
--- a/tests/src/core/parse.js
+++ b/tests/src/core/parse.js
@@ -138,4 +138,18 @@ describe('parse(content, { settings, ecmaFeatures })', function () {
     parseStubParser.parse = parseSpy;
     expect(parse.bind(null, path, content, { settings: {}, parserPath: 'espree', languageOptions: { parserOptions: { sourceType: 'module', ecmaVersion: 2015, ecmaFeatures: { jsx: true } } }, parserOptions: { sourceType: 'script' } })).not.to.throw(Error);
   });
+
+  it('passes ecmaVersion and sourceType from languageOptions to parser', () => {
+    const parseSpy = sinon.spy();
+    const languageOptions = { ecmaVersion: 'latest', sourceType: 'module', parserOptions: { ecmaFeatures: { jsx: true } } };
+    parseStubParser.parse = parseSpy;
+    parse(path, content, { settings: {}, parserPath: parseStubParserPath, languageOptions });
+    expect(parseSpy.args[0][1], 'custom parser to clone the parserOptions object').to.not.equal(languageOptions);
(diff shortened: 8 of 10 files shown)


michaelfaith:
Thinking i should maybe revert the addition of `mjs` and `cjs` extensions to the default extensions list that I made in the flat config pr, since that was more or less an after thought / not central to the flat config support, and could be causing inadvertent regression?  What do you think?

G-Rath:
fwiw this doesn't seem to have changed or improved anything in my local tests or in #2996

michaelfaith:
> fwiw this doesn't seem to have changed or improved anything in my local samples or in #2996

Yeah, like I mentioned here https://github.com/import-js/eslint-plugin-import/pull/2996#issuecomment-2351643614 this will need more work for the v9 tests, due to how the v9 `RuleTester` is wrapping the parser object.  Are you able to share your v8 repro?  Maybe updating the flat example in the root of the repo with a reproduction?  I can look closer at it, but I think this change is still needed regardless.

G-Rath:
I've pushed up https://github.com/G-Rath/eslint-plugin-import-debugging - `npm i && bash run.sh` should do the trick

michaelfaith:
> I've pushed up https://github.com/G-Rath/eslint-plugin-import-debugging - `npm i && bash run.sh` should do the trick

I just added the equivalent test to the legacy and flat examples in this repo and they're both reporting error for `no-cycle`.  I can add those files to this PR.

image

G-Rath:
I don't follow - are you saying that with my repository you're getting `no-cycle` errors on both legacy and flat config?

michaelfaith:
> I don't follow - are you saying that with my repository you're getting `no-cycle` errors on both legacy and flat config?

I wanted to test with this branch, so I just recreated the test in the existing examples here, and yes, it's detecting `no-cycle` correctly.

G-Rath:
> it's detecting no-cycle correctly with both.

sorry for the doubt but I want to be extra clear since that doesn't match my results locally - you're saying with a fresh clone of https://github.com/G-Rath/eslint-plugin-import-debugging, running _just_ `npm i && bash run.sh` without any modifications to files, you're getting `no-cycle` erroring four times?

Not saying it's not possible, but it's annoying if that is the case 😅 I assume I've got a local cache issue

michaelfaith:
> > it's detecting no-cycle correctly with both.
> 
> sorry for the doubt but I want to be extra clear since that doesn't match my results locally - you're saying with a fresh clone of https://github.com/G-Rath/eslint-plugin-import-debugging, running _just_ `npm i && bash run.sh` without any modifications to files, you're getting `no-cycle` erroring four times?
> 
> Not saying it's not possible, but it's annoying if that is the case 😅 I assume I've got a local cache issue

That wouldn't test it with this code change.  Your repo is pulling the release version.  
image

Check out the examples folders.  It's a very simple setup, just like yours.  But it's testing this code.

G-Rath:
Sure but either way we need to establish a baseline - happy to have this conversation over on #2996 if you think it's confusing to have here, but either way my request is the same.

fwiw I have applied your fix here to the repo manually and it didn't change the output.

My whole point with that repo is it's a behaviour that does not seem to be being caught by the test suite, so making changes here is pretty moot because I would expect you to not have matching behaviour.

I've now run my repo on two different laptops which give the same result (only legacy config gives `no-cycle` errors), which'd indicate this isn't a caching issue

michaelfaith:
> Sure but either way we need to establish a baseline - happy to have this conversation over on #2996 if you think it's confusing to have here, but either way my request is the same.
> 
> fwiw I have applied your fix here to the repo manually and it didn't change the output.
> 
> My whole point with that repo is it's a behaviour that does not seem to be being caught by the test suite, so making changes here is pretty moot because I would expect you to not have matching behaviour.
> 
> I've now run my repo on two different laptops which give the same result (only legacy config gives `no-cycle` errors), which'd indicate this isn't a caching issue

It occured to me that the examples in this repo are both set up with the TS parser (oops)😅 So, i pulled your repo, and discovered the issue there.  For one, I had to add both the changes from this PR _and_ this PR: https://github.com/import-js/eslint-plugin-import/pull/3052 since it's not released yet.  After doing that, it still wasn't reporting the error.  But after debugging I found that the call to `espree`'s parse function was returning this error
`"sourceType 'module' is not supported when ecmaVersion < 2015. Consider adding `{ ecmaVersion: 2015 }` to the parser options."`.  And sure enough, I moved `ecmaVersion` into `languageOptions.parserOptions` instead of just `languageOptions` and it parses ok, and reports `no-cycle`.  That seems like it might be a bug with the v8 espree, since `ecmaVersion` is supposed to be under `languageOptions` now (unless that wasn't changed in espree until the latest major?).   But that isn't anything this plugin should have to navigate I don't think, since adjusting the config allows the parse to work correctly.

G-Rath:
hmm ok that's interesting - something still seems off there to me.

Pulling together a few things:
  - I had forgotten that https://github.com/import-js/eslint-plugin-import/pull/3052 hadn't been released, and either way that's an important bug fix we shouldn't be forgetting about
  - I had discovered a similar sounding bug which I fixed in my v9 PR: https://github.com/import-js/eslint-plugin-import/pull/2996/commits/b80bd59d7077ad8c91829cc9aa5c7740295362f1
  - according to the docs `ecmaVersion` should always be passed to parsers
    -  https://github.com/import-js/eslint-plugin-import/pull/2996#discussion_r1746456781 & https://eslint.org/docs/latest/use/configure/parser
  - so I think there's a bug _somewhere_ be it this plugin, eslint, or espree...

michaelfaith:
> * according to the docs `ecmaVersion` should always be passed to parsers
>   
>   * [feat: support ESLint v9 #2996 (comment)](https://github.com/import-js/eslint-plugin-import/pull/2996#discussion_r1746456781) & https://eslint.org/docs/latest/use/configure/parser
> * so I think there's a bug _somewhere_ be it this plugin, eslint, or espree...

Ah, in that case you're right. We aren't passing ecmaVersion to the parse function. That's an easy fix.

michaelfaith:
Updated with that change. And verified that it works in your debugging repo without needing to change the config.  
image

michaelfaith:
Added a unit test to cover the `ecmaVersion` and `sourceType` fix

ljharb (approved):

PR #3062, opened 2024-09-15
[Fix] `exportMap`: export map cache is tainted by unreliable parse results

michaelfaith:
This change addresses an issue observed with the v9 upgrade, where the `ExportMap` Cache is being tainted with a bad export map, if the parse doesn't yield a `visitorKeys` (which can happen if an incompatible parser is used (e.g. og babel eslint)).  If one run of the no-cycle rule uses an incompatible parser and taints the cache and a subsequent test is run with a compatible parser, then the bad export map will be found in the cache and used instead of attempting to proceed with the parse.   

I don't think this is likely to occur in the wild, but in the event that it does, this should help. One thing to note, though: if someone is using an incompatible parser, then this change will have a performance impact, since it will attempt to create an export map every time, until it yields one that has `visitorKeys` attached.

I also updated the `getExports` test to use a valid `parserPath`, rather than a mocked one, so the tests are acting on a valid `parserPath`.  Otherwise the export map won't be cached following this change and the test for it caching would fail.

Additional context: https://github.com/import-js/eslint-plugin-import/pull/2996#issuecomment-2351204273

### DIFF
--- a/tests/src/core/getExports.js
+++ b/tests/src/core/getExports.js
@@ -1,15 +1,18 @@
 import { expect } from  'chai';
+import fs from 'fs';
 import semver from 'semver';
 import sinon from 'sinon';
 import eslintPkg from 'eslint/package.json';
+import { test as testUnambiguous } from 'eslint-module-utils/unambiguous';
 import typescriptPkg from 'typescript/package.json';
 import * as tsConfigLoader from 'tsconfig-paths/lib/tsconfig-loader';
+
 import ExportMapBuilder from '../../../src/exportMap/builder';
-
-import * as fs from 'fs';
-
 import { getFilename } from '../utils';
-import { test as testUnambiguous } from 'eslint-module-utils/unambiguous';
+
+const babelPath = require.resolve('babel-eslint');
+const hypotheticalLocation = babelPath.replace('index.js', 'visitor-keys.js');
+const isVisitorKeysSupported = fs.existsSync(hypotheticalLocation);
 
 describe('ExportMap', function () {
   const fakeContext = Object.assign(
@@ -21,7 +24,7 @@ describe('ExportMap', function () {
     },
     {
       settings: {},
-      parserPath: 'babel-eslint',
+      parserPath: require.resolve('babel-eslint'),
     },
   );
 
@@ -36,9 +39,18 @@ describe('ExportMap', function () {
 
   });
 
-  it('returns a cached copy on subsequent requests', function () {
+  (isVisitorKeysSupported ? it : it.skip)('returns a cached copy on subsequent requests', function () {
     expect(ExportMapBuilder.get('./named-exports', fakeContext))
       .to.exist.and.equal(ExportMapBuilder.get('./named-exports', fakeContext));
+  });
+
+  it('does not return a cached copy if the parse does not yield a visitor keys', function () {
+    const mockContext = {
+      ...fakeContext,
+      parserPath: 'not-real',
+    };
+    expect(ExportMapBuilder.get('./named-exports', mockContext))
+      .to.exist.and.not.equal(ExportMapBuilder.get('./named-exports', mockContext));
   });
 
   it('does not return a cached copy after modification', (done) => {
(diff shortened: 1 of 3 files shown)


michaelfaith:
The test failures are from eslint 2 and 3, which I'm guessing doesn't yield a `visitorKeys` from the parse, causing the cache to be bypassed.

michaelfaith:
Another idea for how to approach this, to limit the impact to just the rules that need it:  
we could add an additional optional param to `ExportMapBuilder.get` that takes a skip cache function, which takes an export map and returns a boolean for whether that instance of the map should be cached.  `no-cycle` could then pass in `(exportMap) => !exportMap.visitorKeys`, which has the same effect, but limits it to only that rule.

tests/src/core/getExports.js
```diff
@@ -41,6 +41,15 @@ describe('ExportMap', function () {
       .to.exist.and.equal(ExportMapBuilder.get('./named-exports', fakeContext));
   });
 
+  it('does not return a cached copy if the parse does not yield a visitor keys', function () {
+    const mockContext = {
+      ...fakeContext,
+      parserPath: 'not-real',
+    }
```
G-Rath, line 48:
```suggestion
    };
```

G-Rath:
Confirming that this looks to address some of the test failures in #2996 - definitely the ones from `no-cycle` and probably others.

> The test failures are from eslint 2 and 3

Could you just also check what version of ESLint is being used? as this seems to only take effect when using flat config and/or ESLint v9 which neither of those versions will support, and it feels like it could be more of a footgun to have rules control the cache

michaelfaith:
> Could you just also check what version of ESLint is being used? as this seems to only take effect when using flat config and/or ESLint v9 which neither of those versions will support, and it feels like it could be more of a footgun to have rules control the cache

Yeah, I like that too.

ljharb (approved):
This approach seems fine to me. If a version check is the cleanest way to handle this, that's fine, but it'd be much nicer to feature-test instead.

tests/src/core/getExports.js
michaelfaith: Fixed

michaelfaith:
> This approach seems fine to me. If a version check is the cleanest way to handle this, that's fine, but it'd be much nicer to feature-test instead.

Updated the tests to only run the caching test if we determine `babel-eslint` has visitor-keys.

tests/src/core/getExports.js
```diff
@@ -11,6 +11,12 @@ import * as fs from 'fs';
 import { getFilename } from '../utils';
 import { test as testUnambiguous } from 'eslint-module-utils/unambiguous';
 
+const doesBabelHaveVisitorKeys = () => {
+  const babelPath = require.resolve('babel-eslint');
```
ljharb, line 15: is there a reason not to do this at module level, and also use it on line 30?

tests/src/core/getExports.js
```diff
@@ -36,11 +44,20 @@ describe('ExportMap', function () {
 
   });
 
-  it('returns a cached copy on subsequent requests', function () {
+  isVisitorKeysSupported && it('returns a cached copy on subsequent requests', function () {
```
ljharb, line 47:
```suggestion
  (isVisitorKeysSupported ? it : it.skip)('returns a cached copy on subsequent requests', function () {
```

ljharb (changes requested):

tests/src/core/getExports.js
michaelfaith: Not a strong reason.  I generally default to keeping control variables like that at the lowest level needed and not adding a bunch to the module scope.  But I don't have a strong opinion in this case, and can change it if you prefer.

tests/src/core/getExports.js
michaelfaith: Updated

tests/src/core/getExports.js
ljharb: It should have a slight performance benefit, and it eliminates the chance that the resolution could be different, so i think it'd be good to make the change.

michaelfaith: Updated

ljharb (approved):

PR #3065, opened 2024-09-19
[Fix] `export`: False positive for exported overloaded functions in TS

liuxingbaoyu:
This currently fails when there is only one `TSDeclareFunction`, this PR fixes that.

### DIFF
--- a/src/rules/export.js
+++ b/src/rules/export.js
@@ -2,7 +2,6 @@ import ExportMapBuilder from '../exportMap/builder';
 import recursivePatternCapture from '../exportMap/patternCapture';
 import docsUrl from '../docsUrl';
 import includes from 'array-includes';
-import flatMap from 'array.prototype.flatmap';
 
 /*
 Notes on TypeScript namespaces aka TSModuleDeclaration:
@@ -27,42 +26,25 @@ const rootProgram = 'root';
 const tsTypePrefix = 'type:';
 
 /**
- * Detect function overloads like:
+ * remove function overloads like:
  * ```ts
  * export function foo(a: number);
  * export function foo(a: string);
- * export function foo(a: number|string) { return a; }
  * ```
  * @param {Set<Object>} nodes
- * @returns {boolean}
  */
-function isTypescriptFunctionOverloads(nodes) {
-  const nodesArr = Array.from(nodes);
-
-  const idents = flatMap(
-    nodesArr,
-    (node) => node.declaration && (
-      node.declaration.type === 'TSDeclareFunction' // eslint 6+
-      || node.declaration.type === 'TSEmptyBodyFunctionDeclaration' // eslint 4-5
-    )
-      ? node.declaration.id.name
-      : [],
-  );
-  if (new Set(idents).size !== idents.length) {
-    return true;
-  }
-
-  const types = new Set(nodesArr.map((node) => node.parent.type));
-  if (!types.has('TSDeclareFunction')) {
-    return false;
-  }
-  if (types.size === 1) {
-    return true;
-  }
-  if (types.size === 2 && types.has('FunctionDeclaration')) {
-    return true;
-  }
-  return false;
+function removeTypescriptFunctionOverloads(nodes) {
+  nodes.forEach((node) => {
+    const declType = node.type === 'ExportDefaultDeclaration' ? node.declaration.type : node.parent.type;
+    if (
+      // eslint 6+
+      declType === 'TSDeclareFunction'
+      // eslint 4-5
+      || declType === 'TSEmptyBodyFunctionDeclaration'
+    ) {
+      nodes.delete(node);
+    }
+  });
 }
 
 /**
@@ -227,9 +209,11 @@ module.exports = {
       'Program:exit'() {
         for (const [, named] of namespace) {
           for (const [name, nodes] of named) {
+            removeTypescriptFunctionOverloads(nodes);
+
             if (nodes.size <= 1) { continue; }
 
-            if (isTypescriptFunctionOverloads(nodes) || isTypescriptNamespaceMerging(nodes)) { continue; }
+            if (isTypescriptNamespaceMerging(nodes)) { continue; }
 
             for (const node of nodes) {
               if (shouldSkipTypescriptNamespace(node, nodes)) { continue; }
--- a/tests/src/rules/export.js
+++ b/tests/src/rules/export.js
@@ -56,6 +56,15 @@ ruleTester.run('export', rule, {
       `,
       parser,
     })),
+    getTSParsers().map((parser) => ({
+      code: `
+        export default function foo(param: string): boolean;
+        export default function foo(param: string, param1?: number): boolean {
+          return param && param1;
+        }
+      `,
+      parser,
+    })),
   ),
 
   invalid: [].concat(
@@ -153,6 +162,19 @@ ruleTester.run('export', rule, {
       parserOptions: {
         ecmaVersion: 2022,
       },
+    })),
+
+    getTSParsers().map((parser) => ({
+      code: `
+        export default function a(): void;
+        export default function a() {}
+        export { x as default };
+      `,
+      errors: [
+        'Multiple default exports.',
+        'Multiple default exports.',
+      ],
+      parser,
     })),
   ),
 });
@@ -510,7 +532,7 @@ context('TypeScript', function () {
           }),
           test({
             code: `
-              export function Foo();
+              export function Foo() { };
               export class Foo { }
               export namespace Foo { }
             `,
@@ -529,7 +551,7 @@ context('TypeScript', function () {
           test({
             code: `
               export const Foo = 'bar';
-              export function Foo();
+              export function Foo() { };
               export namespace Foo { }
             `,
             errors: [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Fixed
 - `ExportMap` / flat config: include `languageOptions` in context ([#3052], thanks [@michaelfaith])
 - [`no-named-as-default`]: Allow using an identifier if the export is both a named and a default export ([#3032], thanks [@akwodkiewicz])
+- [`export`]: False positive for exported overloaded functions in TS ([#3065], thanks [@liuxingbaoyu])
 
 ### Changed
 - [Docs] [`no-relative-packages`]: fix typo ([#3066], thanks [@joshuaobrien])
@@ -1140,6 +1141,7 @@ for info on changes for earlier releases.
 [`memo-parser`]: ./memo-parser/README.md
 
 [#3066]: https://github.com/import-js/eslint-plugin-import/pull/3066
+[#3065]: https://github.com/import-js/eslint-plugin-import/pull/3065
 [#3052]: https://github.com/import-js/eslint-plugin-import/pull/3052
 [#3043]: https://github.com/import-js/eslint-plugin-import/pull/3043
 [#3036]: https://github.com/import-js/eslint-plugin-import/pull/3036
@@ -1875,6 +1877,7 @@ for info on changes for earlier releases.
 [@lilling]: https://github.com/lilling
 [@ljharb]: https://github.com/ljharb
 [@ljqx]: https://github.com/ljqx
+[@liuxingbaoyu]: https://github.com/liuxingbaoyu
 [@lo1tuma]: https://github.com/lo1tuma
 [@loganfsmyth]: https://github.com/loganfsmyth
 [@luczsoma]: https://github.com/luczsoma


src/rules/export.js
```diff
@@ -39,17 +38,21 @@ const tsTypePrefix = 'type:';
...
-      || node.declaration.type === 'TSEmptyBodyFunctionDeclaration' // eslint 4-5
-    )
-      ? node.declaration.id.name
-      : [],
-  );
-  if (new Set(idents).size !== idents.length) {
-    return true;
+  if (nodesArr[0].type === 'ExportDefaultDeclaration') {
+    let num = 0;
+    for (const node of nodesArr) {
```
ljharb, line 43: for..of is slow and best avoided, and i vastly prefer the Set approach because it deduplicates - unless the deduplication is the problem?

ljharb (changes requested):

src/rules/export.js
liuxingbaoyu: From my understanding, it seems that the results of deduplicates node names will not be used, and different names will cause TS errors.
The original logic seems to only check if there are two or more overloaded functions.

ljharb: Gotcha. Let's go back to the flatMap approach then, but omit the Set and use the `.length` instead?

liuxingbaoyu: I tried it, and it seems that `for of` is faster.
But if you still prefer `flatMap`, I can revert it. :)
[perf](https://perf.link/#eyJpZCI6ImJ4M3ZxY3Uyc2c5IiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IG5vZGVzQXJyID0gW3tcInR5cGVcIjpcIkV4cG9ydERlZmF1bHREZWNsYXJhdGlvblwiLFwiZGVjbGFyYXRpb25cIjp7XCJ0eXBlXCI6XCJUU0RlY2xhcmVGdW5jdGlvblwiLFwiaWRcIjp7XCJ0eXBlXCI6XCJJZGVudGlmaWVyXCIsXCJuYW1lXCI6XCJiXCIsXCJyYW5nZVwiOlsyNCwyNV19LFwiZ2VuZXJhdG9yXCI6ZmFsc2UsXCJleHByZXNzaW9uXCI6ZmFsc2UsXCJhc3luY1wiOmZhbHNlLFwicGFyYW1zXCI6W10sXCJyYW5nZVwiOlsxNSwzM10sXCJyZXR1cm5UeXBlXCI6e1widHlwZVwiOlwiVFNUeXBlQW5ub3RhdGlvblwiLFwicmFuZ2VcIjpbMjcsMzJdLFwidHlwZUFubm90YXRpb25cIjp7XCJ0eXBlXCI6XCJUU0FueUtleXdvcmRcIixcInJhbmdlXCI6WzI5LDMyXX19fSxcInJhbmdlXCI6WzAsMzNdLFwiZXhwb3J0S2luZFwiOlwidmFsdWVcIn0se1widHlwZVwiOlwiRXhwb3J0RGVmYXVsdERlY2xhcmF0aW9uXCIsXCJkZWNsYXJhdGlvblwiOntcInR5cGVcIjpcIkZ1bmN0aW9uRGVjbGFyYXRpb25cIixcImlkXCI6e1widHlwZVwiOlwiSWRlbnRpZmllclwiLFwibmFtZVwiOlwiYVwiLFwicmFuZ2VcIjpbNTgsNTldfSxcImdlbmVyYXRvclwiOmZhbHNlLFwiZXhwcmVzc2lvblwiOmZhbHNlLFwiYXN5bmNcIjpmYWxzZSxcInBhcmFtc1wiOltdLFwiYm9keVwiOntcInR5cGVcIjpcIkJsb2NrU3RhdGVtZW50XCIsXCJib2R5XCI6W10sXCJyYW5nZVwiOls2Miw2OF19LFwicmFuZ2VcIjpbNDksNjhdfSxcInJhbmdlXCI6WzM0LDY4XSxcImV4cG9ydEtpbmRcIjpcInZhbHVlXCJ9XTsiLCJ0ZXN0cyI6W3sibmFtZSI6ImZvciBvZiIsImNvZGUiOiJsZXQgbnVtID0gMDtcbmZvciAoY29uc3Qgbm9kZSBvZiBub2Rlc0Fycikge1xuICBpZiAoXG4gICAgLy8gZXNsaW50IDYrXG4gICAgbm9kZS5kZWNsYXJhdGlvbi50eXBlID09PSAnVFNEZWNsYXJlRnVuY3Rpb24nXG4gICAgLy8gZXNsaW50IDQtNVxuICAgIHx8IG5vZGUuZGVjbGFyYXRpb24udHlwZSA9PT0gJ1RTRW1wdHlCb2R5RnVuY3Rpb25EZWNsYXJhdGlvbidcbiAgKSB7XG4gICAgbnVtKys7XG4gIH1cbn0iLCJydW5zIjpbNzA3ODAwMCwxMjM1MDAwLDEwMzE3MDAwLDIwMDMwMDAsODM2OTAwMCw3MDA2MDAwLDgwMDAsMTE3MDAwMCwxMjE2MDAwLDE4ODcwMDAsODUyNTAwMCw5ODgwMDAwLDc2MjYwMDAsNjk4OTAwMCwyODM5MDAwLDc1MzAwMCwyMDUzMDAwLDQ1MTUwMDAsMTM5MzAwMCwxNjIwMDAsNTUzMTAwMCwxMDQyMjAwMCw2MDMwMDAsMTY2MDAwMCwxMjU4MDAwLDQ5NDEwMDAsMjcwNTAwMCw3NzMwMDAsMTAzMTcwMDAsMTAzMTcwMDAsOTY3NzAwMCw4MDAwLDk5OTkwMDAsNjQzNTAwMCw5NDAxMDAwLDEwMDMwMDAsOTUwNjAwMCwxMTk0MDAwLDQ0MzAwMDAsMTM4OTAwMCwxMjc1MDAwLDE0MzAwMDAsNTUwMTAwMCwyNjUwMDAsMTExNzAwMCwyOTYwMDAsNjM0MjAwMCw0NTkwMDAsMjA1MzAwMCwxMDEwNDAwMCw0NTkwMDAsMzYxNzAwMCw0ODgwMDAsNjk0MDAwLDEwNzUwMDAsMzQ5MTAwMCwxNzQxMDAwLDU4MzcwMDAsODk4MjAwMCwxNDMwMDAwLDMzMTcwMDAsNDgzMDAwMCwxMTM1MDAwLDMyMDkwMDAsMTA1ODAwMCw4MTM0MDAwLDI4OTQwMDAsMjg5NDAwMCwyODYwMDAsNjg2MTAwMCwxMDMxNzAwMCwxMDY0MDAwLDY0ODAwMDAsNTk2NjAwMCw1NDgwMDAwLDkyMjAwMCwxOTU3MDAwLDI0MzMwMDAsNTIyOTAwMCw2MDYwMDAsMzYzNzAwMCw0OTYwMDAsMTkzMTAwMCw2OTQ5MDAwLDE5ODAwMDAsMTAzMTEwMDAsODc4MjAwMCw1OTUwMDAsMjMzOTAwMCwxMDMxNzAwMCw2OTQ3MDAwLDc3MDAwLDUwNjkwMDAsNDU5MDAwLDk3MDQwMDAsMzA0NzAwMCwxNjg3MDAwLDM2MTkwMDAsNDQ5MjAwMCwxMDMxNzAwMF0sIm9wcyI6NDExMDc2MH0seyJuYW1lIjoiZmxhdE1hcCIsImNvZGUiOiIgIGNvbnN0IGlkZW50cyA9IG5vZGVzQXJyLmZsYXRNYXAoXG4gICAgKG5vZGUpID0%2BIG5vZGUuZGVjbGFyYXRpb24gJiYgKFxuICAgICAgbm9kZS5kZWNsYXJhdGlvbi50eXBlID09PSAnVFNEZWNsYXJlRnVuY3Rpb24nIC8vIGVzbGludCA2K1xuICAgICAgfHwgbm9kZS5kZWNsYXJhdGlvbi50eXBlID09PSAnVFNFbXB0eUJvZHlGdW5jdGlvbkRlY2xhcmF0aW9uJyAvLyBlc2xpbnQgNC01XG4gICAgKVxuICAgICAgPyBub2RlLmRlY2xhcmF0aW9uLmlkLm5hbWVcbiAgICAgIDogW10sXG4gICk7IiwicnVucyI6WzMyMDEwMDAsMjk4MDAwLDI5NjMwMDAsMTE0MjAwMCwyMDgzMDAwLDI5NzYwMDAsMTgzNTAwMCwyNTYwMDAsMzQ3MDAwLDI1NjAwMCw0MDIwMDAwLDMxODcwMDAsMjUwNDAwMCwyMjIzMDAwLDE0ODAwMDAsNDY1ODAwMCwzMTUwMDAsODc2MDAwLDI5NjAwMDAsMjM3OTAwMCwxMTIwMDAwLDQ0NDIwMDAsNDkzNjAwMCwyNTYwMDAsNzgyMDAwLDEyNDMwMDAsMjY0MDAwLDM2ODYwMDAsNDg2NTAwMCwzMjAzMDAwLDMzNjUwMDAsMjkwOTAwMCwzMTAwMDAwLDM5NjUwMDAsMzY1NTAwMCw1MDI1MDAwLDI3OTMwMDAsMzIwMDAsOTYzMDAwLDI3MjMwMDAsNTAwMCw1MDAwMCwxMzY4MDAwLDE5MDAwLDUwMTgwMDAsNDE1NDAwMCwxNDI2MDAwLDM4NzMwMDAsODM5MDAwLDI4MzMwMDAsMjU2MDAwLDc2NjAwMCwzNjg5MDAwLDQ5MDcwMDAsNTAwMCwxNTYyMDAwLDQ4MjIwMDAsMTU1NzAwMCwyNzkyMDAwLDQ5MjQwMDAsNDc4MDAwLDczNjAwMCw0NjU2MDAwLDU3NjAwMCw0Njk3MDAwLDE4OTYwMDAsNjg3MDAwLDMzNTgwMDAsMTg4MzAwMCwxNTk0MDAwLDM3NDMwMDAsMjA4MDAwLDE3OTUwMDAsMTA4MTAwMCw5MDEwMDAsNDE2MzAwMCwzMjAwMDAsNDA5NTAwMCwxMzg2MDAwLDI1NjAwMCwzNTgwMDAsNDQ2OTAwMCwxMjEwMDAsMTU2ODAwMCwyNTYwMDAsMzk2ODAwMCwyMzg3MDAwLDQ3OTEwMDAsNTA3MTAwMCw0Njg1MDAwLDIzNDcwMDAsMzc4MjAwMCwxMjI2MDAwLDI1NjAwMCwzMDM5MDAwLDQ5NzAwMCwyMzAwMCw4OTcwMDAsMTA0NzAwMCwzNjQ5MDAwXSwib3BzIjoyMjEwOTcwfV0sInVwZGF0ZWQiOiIyMDI0LTA5LTIzVDIzOjM0OjUxLjEwMloifQ%3D%3D)

ljharb: Fair enough, but I still prefer avoiding the iterator protocol if possible.

src/rules/export.js
```diff
@@ -39,17 +38,22 @@ const tsTypePrefix = 'type:';
...
-    (node) => node.declaration && (
-      node.declaration.type === 'TSDeclareFunction' // eslint 6+
-      || node.declaration.type === 'TSEmptyBodyFunctionDeclaration' // eslint 4-5
-    )
-      ? node.declaration.id.name
-      : [],
-  );
-  if (new Set(idents).size !== idents.length) {
-    return true;
+  if (nodesArr[0].type === 'ExportDefaultDeclaration') {
```
ljharb, line 41: wait, why are we only checking the first item? what if it's a mix of an overload and something different?

liuxingbaoyu:
```
export { x as default };
export default function a() {}
```
Oh, I hadn't thought of this example before.
But TS will complain in this case, do we need to deal with it? If so I will update this PR.

ljharb: Yes, we still should deal with it.

liuxingbaoyu:
What do you think of the new changes? :)
This greatly simplifies the implementation.
It also additionally avoids reporting overloads `Multiple default exports.`, which could confuse people.
But now overloads and other types of conflicts will not be reported, which I think is reasonable, because anyway overloads cannot appear alone, must be followed by an implementation to be legal TS code.

tests/src/rules/export.js
```diff
@@ -510,7 +532,7 @@ context('TypeScript', function () {
           }),
           test({
             code: `
-              export function Foo();
+              export function Foo() { };
```
ljharb, line 535: why is this change needed?

liuxingbaoyu: Now we will ignore overloading in order to report two errors instead of three in this case.
```
export default function a(): void;
export default function a() {}
export { x as default };
```
And this test was testing duplicate export of overloading and class.
I changed it to export of method and class to fix the test.
Also I did not find a test for method and class, the new one might be the original intention.

ljharb: hmm - i'm concerned with changing an existing test being a breaking change.

We certainly should be only reporting two here, but why are the curly braces needed?

liuxingbaoyu: Without the curly braces, its node type would be `TSDeclareFunction`, which is just an overload and not a real function.
I guess this was not the original intention of this test, since there is no test for functions. We would definitely prefer to test other legal code besides export duplication.
This should not be affected in the real world, because a lone `TSDeclareFunction` will produce the error `Function implementation is missing or not immediately following the declaration.(2391)`.
The `FunctionDeclaration` that it must follow will throw our eslint error.

ljharb: let's leave the original test unchanged, and add a new test with the curly braces?

liuxingbaoyu: The original test currently fails, because we no longer report errors for overriding. :)
But in the real world having overridden methods without implementations is illegal anyway, so this is probably not a breaking change.

ljharb (approved):

PR #3072, opened 2024-09-26
[Fix] `exportMap`: improve `cacheKey` when using flat config 

michaelfaith:
Discovered this issue in https://github.com/import-js/eslint-plugin-import/pull/2996#issuecomment-2372522774.  

This change improves the logic for generating the cache key used for both the ExportMap cache and the resolver cache when using flat config.  Prior to this change, the cache key was a combination of the `parserPath`, a hash of the `parserOptions`, a hash of the plugin settings, and the path of the file.  When using flat config, `parserPath` isn't provided.  So, there's the possibility of incorrect / stale cache objects being used if someone ran with this plugin using different parsers within the same lint execution, if `parserOptions` and settings are the same.  

For flat config, we can achieve similar results by constructing the cacheKey using `languageOptions` as a component of the key, rather than `parserPath` and `parserOptions`.  One caveat is that the `parser` property of `languageOptions` is an object that oftentimes has the same two functions (`parse` and `parseForESLint`).  This won't be reliably distinct when using the base `JSON.stringify` function to detect changes.  So, this implementation uses a replacer function along with `JSON.stringify` to stringify function properties along with other property types in order to detect different parsers.

To ensure that this works properly with v9, I also tested this against https://github.com/import-js/eslint-plugin-import/pull/2996 with v9 installed.  Not only does it pass all tests in that branch, but it also removes the need to add this exception: https://github.com/import-js/eslint-plugin-import/pull/2996#discussion_r1774135785

### DIFF
--- a/src/exportMap/childContext.js
+++ b/src/exportMap/childContext.js
@@ -1,9 +1,17 @@
 import { hashObject } from 'eslint-module-utils/hash';
 
-let parserOptionsHash = '';
-let prevParserOptions = '';
+let optionsHash = '';
+let prevOptions = '';
 let settingsHash = '';
 let prevSettings = '';
+
+// Replacer function helps us with serializing the parser nested within `languageOptions`.
+function stringifyReplacerFn(_, value) {
+  if (typeof value === 'function') {
+    return String(value);
+  }
+  return value;
+}
 
 /**
  * don't hold full context object in memory, just grab what we need.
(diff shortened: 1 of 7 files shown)


michaelfaith:
Note: I broke the change to `utils/hash` out as its own commit, so that can be released separately.

src/exportMap/childContext.js
```diff
@@ -1,10 +1,18 @@
...
-let prevParserOptions = '';
+let optionsHash = '';
+let prevOptions = '';
 let settingsHash = '';
 let prevSettings = '';
 
+// Replacer function helps us with serializing the parser nested within `languageOptions`.
+function stringifyReplacerFn(_, value) {
+  if (typeof value === 'function') {
+    return value.toString();
```
ljharb, line 11:
```suggestion
    return String(value);
```
identical, since it's a known function, but safer in general :-)

src/exportMap/childContext.js
```diff
@@ -17,13 +25,28 @@ export default function childContext(path, context) {
...
+      prevOptions = JSON.stringify(parserOptions);
+    }
+    // When not using flat config, we use a combination of the hashed parserOptions
+    // and parserPath as the token
+    optionsToken = String(parserPath) + optionsHash;
   }
 
   return {
-    cacheKey: String(parserPath) + parserOptionsHash + settingsHash + String(path),
+    cacheKey: optionsToken + settingsHash + String(path),
```
ljharb, line 49: would it make more sense to just hash a single object composed of all these pieces, rather than multiple hashes?

utils/hash.js
```diff
@@ -17,6 +17,8 @@ function hashify(value, hash) {
 
   if (Array.isArray(value)) {
     hashArray(value, hash);
+  } else if (typeof value === 'function') {
+    hash.update(value.toString());
```
ljharb, line 21:
```suggestion
    hash.update(String(value));
```

src/exportMap/childContext.js
michaelfaith: We could. I think the one downside is that if only settings change or if only one of the options changes, we'd have to generate a hash for a larger object, rather than hashing just settings or just the options object (the part of it that changed).  So, two hashes initially and smaller incremental hashes as things change, vs one bigger hash that would have to be regenerated when any portion of it changes.  Which option would be more performant would likely depend on how we expect people to interact with the cache.

utils/hash.js
michaelfaith: Updated

src/exportMap/childContext.js
michaelfaith: Updated

src/exportMap/childContext.js
ljharb: that's a solid argument.

ljharb (approved):
Thanks!

ljharb:
Published v2.12.0 of eslint-module-utils already; this PR will be merged once CI passes.

PR #3089, opened 2024-10-21
FIx  Key "plugins": Cannot redefine plugin "import". when used in fla…

offspring:
Fix for:

ESLint: 9.12.0

ConfigError: Config (unnamed): Key "plugins": Cannot redefine plugin "import".
    at rethrowConfigError (/Users/atrzeciak/workspace/uber/uber-paste/uber-paste/node_modules/@eslint/config-array/dist/cjs/index.cjs:303:8)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -82,7 +82,7 @@ const importPlugin = {
 const createFlatConfig = (baseConfig, configName) => ({
   ...baseConfig,
   name: `import/${configName}`,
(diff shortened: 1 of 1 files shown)


ethaizone:
I want to share something. I got same error too but later I found out it's error because eslint-config-next included this plugin when include with eslintrc.

so it's just FYI for who faced this issue and it doesn't cause from this package. It should cause from other packages.

PR #3091, opened 2024-10-25
feat: drop Node 4, 6, 8, 10, 12, 14 & 16 support

MichaelDeBoey:
BREAKING CHANGE: Requires Node@^18.18.0 || ^20.9.0 || >=21.1.0


MichaelDeBoey:
@ljharb Settings should be updated to make `node 18+` required instead of `node 4+`

PR #3092, opened 2024-10-25
feat: drop ESLint v2, v3, v4, v5, v6 & v7 support

MichaelDeBoey:
BREAKING CHANGE: Requires ESLint@^8.57.0 || ^9.0.0


jaguiarjardim-riparian (approved):
LGTM

ljharb:
@andrzej-erli the project isn't dead at all, although there are a number of forks people have made in the past. "unnecessary" and "bloating" are opinions that I don't share, but certainly dropping eslint < 8 will make a lot of the dependencies no longer BE necessary.

If there were an update, it would be posted here, so you can assume there's no update (but the lack of an update doesn't say anything about the project)

PR #3096, opened 2024-11-06
Allow to disable resolver

### LINKED ISSUE #3095: Allow to disable default resolver
By default eslint merges the config, making it impossible to use a custom resolver with option without the default resolver node being added.

###  Try 1

```js
settings: {
  'import/resolver': {
    [path.resolve(__dirname, './resolver')]: { /* some options */ },
  },
},
```

Then the following resolvers are used:
```js
{
  '/home/.../eslint-config/resolver': {},
  node: { extensions: [ '.js', '.jsx', '.json' ] }
}
```

###  Try 2:

```js
settings: {
  'import/resolver': {
    [path.resolve(__dirname, './resolver')]: { /* some options */ },
    node: false,
  },
},
```

Then the following resolvers are used:
```js
{
  '/home/.../eslint-config/resolver': {},
  node: false,
}
```

And the node resolver is still used, same with null, undefined or other

###  Try 3:

I can only disable resolver node by doing this, but I can't pass any options:

```js
settings: {
  'import/resolver': path.resolve(__dirname, './resolver'),
},
```

---

I propose as a solution not to use a resolver if its configuration is equal to false.

bertho-zero:
Fix #3095

### DIFF
--- a/utils/resolve.js
+++ b/utils/resolve.js
@@ -84,7 +84,7 @@ function resolverReducer(resolvers, map) {
 
   if (typeof resolvers === 'object') {
     for (const key in resolvers) {
-      if (resolvers[key] === false) continue;
+      if (resolvers[key] === false) { continue; }
       map.set(key, resolvers[key]);
     }
     return map;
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
This seems reasonable - although I think `undefined` is a better choice - and it will need tests.

PR #3097, opened 2024-11-08
[New] add TypeScript types

### LINKED ISSUE #3090: Could not find a declaration file for module 'eslint-plugin-import'
Using latest eslint with flat config in a TS project.  Need type declarations.

G-Rath:
For now I've just typed these out manually based on what we're using internally - while it would be good to have some tests at some point, I'm hoping/recommending not blocking on that for now as I think the types are small enough to be easy to maintain manually for the time being.

Resolves #3090

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,33 @@
+import { ESLint, Linter, Rule } from 'eslint';
+
+declare const plugin: ESLint.Plugin & {
+  meta: {
+    name: string;
+    version: string;
+  };
+  configs: {
+    'recommended': Linter.LegacyConfig;
+    'errors': Linter.LegacyConfig;
+    'warnings': Linter.LegacyConfig;
+    'stage-0': Linter.LegacyConfig;
+    'react': Linter.LegacyConfig;
+    'react-native': Linter.LegacyConfig;
+    'electron': Linter.LegacyConfig;
+    'typescript': Linter.LegacyConfig;
+  };
+  flatConfigs: {
+    'recommended': Linter.FlatConfig;
+    'errors': Linter.FlatConfig;
+    'warnings': Linter.FlatConfig;
+    'stage-0': Linter.FlatConfig;
+    'react': Linter.FlatConfig;
+    'react-native': Linter.FlatConfig;
+    'electron': Linter.FlatConfig;
+    'typescript': Linter.FlatConfig;
+  };
+  rules: {
+    [key: string]: Rule.RuleModule;
+  };
+};
+
+export = plugin;
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "node": ">=4"
   },
   "main": "lib/index.js",
+  "types": "index.d.ts",
   "directories": {
     "test": "tests"
   },
@@ -16,7 +17,8 @@
     "docs",
     "lib",
     "config",
-    "memo-parser/{*.js,LICENSE,*.md}"
+    "memo-parser/{*.js,LICENSE,*.md}",
+    "index.d.ts"
   ],
   "scripts": {
     "prebuild": "rimraf lib",
@@ -34,6 +36,7 @@
     "test-examples": "npm run build && npm run test-example:legacy && npm run test-example:flat",
     "test-example:legacy": "cd examples/legacy && npm install && npm run lint",
     "test-example:flat": "cd examples/flat && npm install && npm run lint",
+    "test-types": "npx --package typescript@latest tsc --noEmit index.d.ts",
     "prepublishOnly": "safe-publish-latest && npm run build",
     "prepublish": "not-in-publish || npm run prepublishOnly",
     "preupdate:eslint-docs": "npm run build",
@@ -63,6 +66,7 @@
     "@angular-eslint/template-parser": "^13.5.0",
     "@eslint/import-test-order-redirect-scoped": "file:./tests/files/order-redirect-scoped",
     "@test-scope/some-module": "file:./tests/files/symlinked-module",
+    "@types/eslint": "^8.56.12",
     "@typescript-eslint/parser": "^2.23.0 || ^3.3.0 || ^4.29.3 || ^5.10.0",
     "babel-cli": "^6.26.0",
     "babel-core": "^6.26.3",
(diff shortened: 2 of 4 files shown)


ljharb:
Can we add a tsc run, at least?

index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+    'errors': ESLint.Linter.LegacyConfig;
+    'warnings': ESLint.Linter.LegacyConfig;
+    'stage-0': ESLint.Linter.LegacyConfig;
+    'react': ESLint.Linter.LegacyConfig;
+    'react-native': ESLint.Linter.LegacyConfig;
+    'electron': ESLint.Linter.LegacyConfig;
+    'typescript': ESLint.Linter.LegacyConfig;
+  };
+  flatConfigs: {
+    'recommended': ESLint.Linter.LegacyConfig;
```
OlivierZal, line 19: for `flatConfigs`, isn't it rather something like:

```js
  flatConfigs: {
    errors: Linter.Config
    typescript: Linter.Config & {
      settings: Record<string, unknown> & {
        'import/resolver': Record<string, unknown>
      }
    }
  }
```

index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+    'warnings': ESLint.Linter.LegacyConfig;
+    'stage-0': ESLint.Linter.LegacyConfig;
+    'react': ESLint.Linter.LegacyConfig;
+    'react-native': ESLint.Linter.LegacyConfig;
+    'electron': ESLint.Linter.LegacyConfig;
+    'typescript': ESLint.Linter.LegacyConfig;
+  };
+  rules: {
+    [key: string]: ESLint.Rule.RuleModule;
+  };
```
OlivierZal, line 30: We could also add types for `settings`, something like:

```js
settings: Record<string, unknown> & {
  // we can add there any known property
  'import/resolver'?: Record<string, unknown>
  typescript?: { alwaysTryTypes?: boolean }
  'extensions'?: ...
}
```

G-Rath: Yup in the long-run that'd be ideally and I'm happy to help work on that but I'd prefer to not block this on that since I'm not super familiar with all the settings and really right now their types won't get used out of the box since the new config system doesn't result in the types getting automatically applied.

i.e. they'd be useful for letting you manually type an object via `pluginImport['settings']` but that's a more complex setup

G-Rath: (I actually only happened to include `rules` at all as I had that type on hand from what we've got in `eslint-plugin-jest` - you could make the same point that ideally the rules could be typed with all their settings, but again that'd be a lot of work for something that right now wouldn't be that usable)

G-Rath:
> Can we add a tsc run, at least?

Done - note that it requires using `typescript` v5 since types like `@types/node` require that version, which led to a weird "works locally but not in CI", but I think I've got that resolved and should be good enough for now 🤞

.github/workflows/node-4+.yml
```diff
@@ -10,6 +10,19 @@ permissions:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ljharb/actions/node/install@main
+        name: 'npm install'
+        with:
+          skip-ls-check: true
+      # for some reason we've got to force typescript to install here
+      # even though the npm script has `typescript@latest`
+      - run: npm i --force typescript@latest
+      - run: npm run test-types
```
ljharb, line 24: this is "fine", but i would assume it'd be easier to add a prelint or postlint script that runs tsc?

package.json
```diff
@@ -34,6 +36,7 @@
     "test-examples": "npm run build && npm run test-example:legacy && npm run test-example:flat",
     "test-example:legacy": "cd examples/legacy && npm install && npm run lint",
     "test-example:flat": "cd examples/flat && npm install && npm run lint",
+    "test-types": "npx --package typescript@latest tsc --noEmit index.d.ts",
```
ljharb, line 39: shouldn't noEmit be set in tsconfig, and tsc be run on the whole project?

also, let's add typescript to the dev deps rather than using npx.

G-Rath: The issue is typescripts already a dev dependency but supporting versions from 2 through to 4, whereas this check has to use v5 because `@types` requires that as a min. version and I assume that the older versions are required for a reason.

I didn't create a tsconfig.json as that will be picked up by any running of typescript, and having it defined can change other aspects of linting such as what files are included by default.

.github/workflows/node-4+.yml
G-Rath: I'm not sure what you mean by this?

package.json
G-Rath: > shouldn't noEmit be set in tsconfig, and tsc be run on the whole project?

That'd mean the whole codebase would get type checked, requiring it to be correctly typed - while I personally would support that, I think it's out of scope for this pr

.github/workflows/node-4+.yml
ljharb: what i mean is, type checking is a form of linting, and there's no need to make any workflow changes if a `prelint` or `postlint` script is defined in package.json, since it'll just automatically run as part of the pretest workflow.

G-Rath: Ok but there's not a `lint` script (ESLint is run in `posttest`).

I can add this to `posttest` if that's what you mean/want, otherwise could you elaborate further on what you'd like?

package.json
G-Rath: So I did have a go at running `tsc` across the whole codebase just to see how much we'd get, and well with `strict` mode its 3498 errors and without it its 857 😅

Technically, we could have a `tsconfig.json` that includes just the `.d.ts` (and even `.js` because `checkJs` is false by default), but it feels weird to me having that as that only config option and for just a single small file...

.github/workflows/node-4+.yml
ljharb: oh lol, true, this package is weird. in that case let's just move these lines to node-pretest.yml instead of the matrix workflow.

ljharb (approved):
I'll merge this without verifications in place, but it'll be very very important to add that soon.

michaelfaith:
Nice work on this!

anmiles:
@ljharb would you mind to get this published on npmjs.com, please? The latest version 2.31.0 still doesn't have this change.

ljharb:
https://github.com/import-js/eslint-plugin-import/releases/tag/v2.32.0 has been released.

PR #3100, opened 2024-11-13
Add cache parameter into webpack resolver

seiyab:


resolvers/webpack/index.js
```diff
@@ -311,6 +311,29 @@ function getResolveSync(configPath, webpackConfig, cwd) {
   return cached.value;
 }
 
+const _evalCache = [];
```
ljharb, line 314: is there a reason not to use a null object here, or a Map?

seiyab: Just referred to `getResolveSync()`.
https://github.com/import-js/eslint-plugin-import/blob/67cc79841fc823ad4af2532af2dc6704e4b3b03a/resolvers/webpack/index.js#L294-L312

First I considered to implement it with WeakMap. Reading `getResolveSync()`, I thought it uses array because it is easier to remove old entries. I'm ready to use (Weak)Map here.

resolvers/webpack/index.js
```diff
@@ -311,6 +311,22 @@ function getResolveSync(configPath, webpackConfig, cwd) {
   return cached.value;
 }
 
+const _evalCache = new Map();
+function evaluateFunctionConfigCached(configPath, webpackConfig, env, argv) {
+  const cacheKey = JSON.stringify({ configPath, args: [env, argv] });
```
seiyab, line 316: This key is key ordering sensitive in contrast to `hashify()`. https://github.com/import-js/eslint-plugin-import/blob/67cc79841fc823ad4af2532af2dc6704e4b3b03a/utils/hash.js#L15

I consider should be fine because:
- In most practical cases, key ordering will be stable.
- Even when ordering is not stable, it will just increase cache miss. Just get almost as fast as `cache: false` at worst. And won't cause invalid result.
- Even if it cause unexpected problems, existing users won't have problem immediately since cache feature is opt-in.

resolvers/webpack/index.js
seiyab: I changed it to Map. https://github.com/import-js/eslint-plugin-import/pull/3100/commits/c65f9116c6ee7753250fa347036b08615ea4fc2e

resolvers/webpack/index.js
```diff
@@ -398,7 +415,11 @@ exports.resolve = function (source, file, settings) {
   }
 
   if (typeof webpackConfig === 'function') {
-    webpackConfig = webpackConfig(env, argv);
+    if (shouldCacheFunctionConfig) {
+      webpackConfig = evaluateFunctionConfigCached(configPath, webpackConfig, env, argv);
+    } else {
+      webpackConfig = webpackConfig(env, argv);
```
ljharb, line 421:
```suggestion
    webpackConfig = shouldCacheFunctionConfig
      ? evaluateFunctionConfigCached(configPath, webpackConfig, env, argv)
      : webpackConfig(env, argv);
```

ljharb (approved):
Seems pretty good!

Issue #2347, opened 2022-01-03
Type imports do not follow rules from import/order


ImRodry:
I have the import/order rule set up with the following configuration:
```json
		"import/order": ["error", {
			"groups": [	"builtin", "index", "external", "internal", "sibling", "parent", "type"],
			"newlines-between": "always",
			"alphabetize": { "order": "asc", "caseInsensitive": true }
		}],
```
And I expected the groups rule to be applied to type imports, however this doesn't seem the case and these imports seem to just get sorted alphabetically. Here's an example:
```ts
// Sorted type imports by import/order
import type { Command } from "../../lib/imports"
import type { WithId } from "mongodb"
```

```ts
Expected sorting following the order external -> internal
import type { WithId } from "mongodb"
import type { Command } from "../../lib/imports"
```

As I understand this may be seen as expected behavior for some people, I believe this fix could be introduced as a boolean option for the rule that would come disabled by default.

ljharb:
They should be sorted by specifier. The import names have no bearing on the sorting.

I definitely agree that it makes sense to, within the group "type", sort everything by `groups` as well. It'd probably be best to do this under an option, yes.

ImRodry:
Ah sorry for my misunderstanding! Is there any documentation on what each one of the groups mean so that I can try to configure my rule better?
Glad to know you agree with this though, looking forward to seeing this in a future release!

ljharb:
I don't think there's any way to configure it for what you want, at the moment. The best path forward is a PR adding the option you want :-)

Issue #2172, opened 2021-08-01
[import-order] order within `import type`


alvis:
Now, `import type` can be a separate group.
I also find the ability to alphabetize very useful.

However, I find it very annoying that they don't work well together.
The problem is that the order of `import type` statement doesn't follow the order we define for other module types.
Let's say we have the following options

```json
"import/order": [
  "warn",
  {
    "groups": [
      "builtin",
      "external",
      "type"
    ],
    "newlines-between": "always-and-inside-groups",
    "alphabetize": { "order": "asc", "caseInsensitive": true }
  }
],
```

e.g. I'd like to have the following order
```ts
// buildin modules before external modules

// for ordinary import it's ok, builtin first, then external
import { createServer } from 'http'

import { graphql } from 'gatsby';

// but for `import type` it warns that the gatsby import should be put in front of the http one
// the order is right here since http is a buildin and gatsby is an external
import type { RequestOptions } from 'http'

import type { PageProps } from 'gatsby';
```

ljharb:
Seems like to do what you want, the schema would need some way of providing alternative config options for a specific group.

I suppose we could add `groupOverrides: { [groupName]: groupConfiguration }`, but that seems like a lot of complexity for such a specific use case.

alvis:
or can we take the order from `groups` excluding `type` when we analyze the order within `import type` statements?

ljharb:
I'm not sure what you mean. Can you elaborate?

Exelenz9:
[This answer](https://github.com/import-js/eslint-plugin-import/issues/2088#issuecomment-848590569) helped me to solve similar problem

kaysonwu:
I have the same issue. @ljharb the `type` group should also follow the basic order, e.g. : "builtin", "external", "internal", "parent", "Sibling ","index", "object"

danranVm:
> Seems like to do what you want, the schema would need some way of providing alternative config options for a specific group.
> 
> I suppose we could add , but that seems like a lot of complexity for such a specific use case.`groupOverrides: { [groupName]: groupConfiguration }`

Hope to support this configuration, thank you.

PR #3104, opened 2024-11-19
feat(import/order): enable advanced spacing and sorting of type-only imports

### LINKED ISSUE #2441: import/order: "type" group is not enough
Currently, the rule offers the following groups:
"builtin", "external", "internal", "unknown", "parent", "sibling", "index", "object", "type"

To me, having single "type" group for all type imports is not enough.

I can import types from "builtin" modules, from "external" module and so on.

So, it would be logical to have type counterpart for each group, i.e.:
"build" -> "built-type"
"external" -> "external-type"
and so on

Xunnamius:
Resolves #2441, #2615, #2347, #2172 \
Related to #2912

First, thanks for making eslint-plugin-import! It has saved me a lot of time and headache.

This PR implements in `import/order` (1) intragroup sorting of type-only imports (to match intergroup sorting of normal imports), (2) newline controls for type-only imports, and (3) a new option to consolidate newlines and save space where possible.

Along the way, I also encountered a bug where a `NaN` rank would get passed around in some edge cases (such as using an import with an absolute specifier under certain configurations). I fixed it (applied to `computeRank`) and added a test to catch it, both of which are included in this PR.

The documentation updates associated with the changes in this PR are part of [a separate PR](#) since those changes probably warrant further discussion. This PR also contains additional tests that cover all examples in the documentation, including for the other settings; several of the old "passing" rule examples failed and were corrected.

If this PR is too large (most of the additions are in the test files), I still have access to the atomic commits and can split the features up in whatever way is easiest. Otherwise, this PR contains the implementation and tests for the following:

## Allow intragroup sorting of type-only imports

This is implemented via `sortTypesAmongThemselves`. The proposed documentation corresponding to this new feature can be previewed [here](https://github.com/Xunnamius/eslint-plugin-import/blob/main/docs/rules/order.md#sorttypesamongthemselves).

Given this code (which is already correctly ordered):

```typescript
import type A from 'fs';
import type B from 'path';
import type C from '../foo.js';
import type D from './bar.js';
import type E from './';

import a from 'fs';
import b from 'path';
import c from '../foo.js';
import d from './bar.js';
import e from './';
```

And the following settings, the rule check will fail:

```jsonc
{
  "groups": ["type", "builtin", "parent", "sibling", "index"],
  "alphabetize": { "order": "asc" }
}
```

With `--fix` yielding:

```typescript
import type C from '../foo.js'; // ???
import type A from 'fs';
import type B from 'path';
import type D from './bar.js';
import type E from './';

import a from 'fs';
import b from 'path';
import c from '../foo.js';
import d from './bar.js';
import e from './';
```

This is currently the technically correct behavior, but the wrong outcome in my opinion.

However, with the following settings, the rule check will succeed instead:

```diff
{
  "groups": ["type", "builtin", "parent", "sibling", "index"],
  "alphabetize": { "order": "asc" },
+ "sortTypesAmongThemselves": true
}
```

Of course, achieving the reverse order (type-only imports after normal imports) is possible by moving `"type"` to the end of `groups`:

```diff
{
+ "groups": ["builtin", "parent", "sibling", "index", "type"],
  "alphabetize": { "order": "asc" },
  "sortTypesAmongThemselves": true
}
```

</details>

It was only after implementing this feature, when I started reviewing the issues literature, that I saw #2441, which lead me to eslint-plugin-perfectionist (neat) and its implementation of "builtin-type," "external-type," etc. I much prefer this PR's implementation, which takes a similar approach to [this comment](https://github.com/import-js/eslint-plugin-import/issues/2441#issuecomment-1110101003). I don't see how allowing type-only builtin/external/etc imports to be sorted in a different order than non-type builtin/external/etc imports makes things _less_ confusing. And that's before we consider custom `pathGroups`.

Instead, the goal of `sortTypesAmongThemselves` is to allow the "type" group to be sorted among itself in a backward-compatible way without the added complexity (and potential inconsistency) of adding regexp or "\*-type" groups to `groups`. I'm already writing too much configuration. I just want to flick a switch 😅.

Additionally, the code that makes `sortTypesAmongThemselves` work could be extended to make something like #2912 work too (perhaps through another option), though that use case is not of interest to me currently.

## Allow controlling intragroup spacing of type-only imports

This is implemented via `newlines-between-types`. The proposed documentation corresponding to this new feature can be previewed [here](https://github.com/Xunnamius/eslint-plugin-import/blob/main/docs/rules/order.md#newlines-between-types).

Given this code:

```typescript
import type A from 'fs';
import type B from 'path';
import type C from '../foo.js';
import type D from './bar.js';
import type E from './';

import a from 'fs';
import b from 'path';

import c from '../foo.js';

import d from './bar.js';

import e from './';
```

And the following settings, the rule check will fail:

```jsonc
{
  "groups": ["type", "builtin", "parent", "sibling", "index"],
  "sortTypesAmongThemselves": true,
  "newlines-between": "always"
}
```

With `--fix` yielding:

```typescript
import type A from 'fs';
import type B from 'path';

import type C from '../foo.js';

import type D from './bar.js';

import type E from './';

import a from 'fs';
import b from 'path';

import c from '../foo.js';

import d from './bar.js';

import e from './';
```

However, with the following settings, the rule check will succeed instead:

```diff
{
  "groups": ["type", "builtin", "parent", "sibling", "index"],
  "sortTypesAmongThemselves": true,
  "newlines-between": "always",
+ "newlines-between-types": "ignore"
}
```

</details>

`sortTypesAmongThemselves` allows sorting type-only and normal imports separately. By default, `newlines-between` will govern all newlines between import statements like normal.

I generally want my type-only imports to be sorted for ease of reference but never have newlines between them (save space) while I want my normal imports (which I tend to visually peruse more often) to be aesthetically pleasing, grouped, and sorted. `newlines-between` is too coarse-grained for this, so this PR introduces `newlines-between-types`, a setting identical to `newlines-between` except it only applies to type-only imports, and only when `sortTypesAmongThemselves` is enabled (i.e. it is backward-compatible).

When `newlines-between` and `newlines-between-types` conflict, `newlines-between-types` takes precedence for type-only imports. For normal imports, `newlines-between-types` is ignored entirely.

One issue that might warrant further discussion is which setting governs the newline separating type-only imports from normal imports. Right now, I have it so `newlines-between-types` controls this space, but perhaps it should be its own setting.

## Collapse excess spacing for aesthetically pleasing imports

This is implemented via `consolidateIslands`. The proposed documentation corresponding to this new feature can be previewed [here](https://github.com/Xunnamius/eslint-plugin-import/blob/main/docs/rules/order.md#consolidateislands).

Given this code (which could be the output of a previous `--fix` pass):

```typescript
var fs = require('fs');
var path = require('path');
var { util1, util2, util3 } = require('util');

var async = require('async');

// Ugly but technically valid

var relParent1 = require('../foo');
var {
  relParent21,
  relParent22,
  relParent23,
  relParent24,
} = require('../');
var relParent3 = require('../bar');

var { sibling1,
  sibling2, sibling3 } = require('./foo');
var sibling2 = require('./bar');
var sibling3 = require('./foobar');
```

And the following settings, the rule check will _pass_:

```jsonc
{
  "newlines-between": "always-and-inside-groups"
}
```

However, when given the following instead, the rule check will _fail_:

```diff
{
  "newlines-between": "always-and-inside-groups",
+ "consolidateIslands": "inside-groups"
}
```

With `--fix` yielding:

```typescript
var fs = require('fs');
var path = require('path');
var { util1, util2, util3 } = require('util');

var async = require('async');

// Pretty

var relParent1 = require('../foo');

var {
  relParent21,
  relParent22,
  relParent23,
  relParent24,
} = require('../');

var relParent3 = require('../bar');

var { sibling1,
  sibling2, sibling3 } = require('./foo');

var sibling2 = require('./bar');
var sibling3 = require('./foobar');
```

Note how the intragroup "islands" of grouped single-line imports, as well as multi-line imports, are surrounded by new lines.

</details>

Essentially, I was looking for a `newlines-between`-like setting somewhere between `"never"` and `"always-and-inside-groups"`. I want newlines separating `groups`/`pathGroups` imports from one another (like `"always-and-inside-groups"`), newlines separating imports that span multiple lines from other imports (this is the new thing), and any remaining newlines deleted or "consolidated" (like `"never"`). The example above demonstrates this use case.

Right now, this is achievable with `newlines-between` set to `"always-and-inside-groups"` if you add additional newlines around multi-line imports to every file by hand. The goal of `consolidateIslands` is to allow `eslint --fix` to take care of the tedium in a backward-compatible way.

There was a slight complication with `consolidateIslands` though: while testing across a few mid-sized repos, I discovered my naive implementation caused a conflict when enabled alongside `sortTypesAmongThemselves: true`, `newlines-between: "always-and-inside-groups"`, and `newlines-between-types: "never"`... and then only when a normal import was followed by a multi-line type-only import. This conflict makes sense, since `newlines-between-types: "never"` wants no newlines ever and demands no newline separate type-only imports from normal imports (since, currently, `newlines-between-types` governs that space), yet `consolidateIslands` demands a newline separate _all_ multi-line imports from other imports.

To solve this, the current implementation has `newlines-between-types` yield to `consolidateIslands` whenever they conflict. I've also added a test to catch any regressions around this edge case.

---

A demo package containing these features is available in the registry for easy testing:

```bash
```

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -513,22 +513,38 @@ function computePathRank(ranks, pathGroups, path, maxPosition) {
   }
 }
 
-function computeRank(context, ranks, importEntry, excludedImportTypes) {
+function computeRank(context, ranks, importEntry, excludedImportTypes, isSortingTypesAmongThemselves) {
   let impType;
   let rank;
+
+  const isTypeGroupInGroups = ranks.omittedTypes.indexOf('type') === -1;
+  const isTypeOnlyImport = importEntry.node.importKind === 'type';
+  const isExcludedFromPathRank = isTypeOnlyImport && isTypeGroupInGroups && excludedImportTypes.has('type')
+
   if (importEntry.type === 'import:object') {
     impType = 'object';
-  } else if (importEntry.node.importKind === 'type' && ranks.omittedTypes.indexOf('type') === -1) {
+  } else if (isTypeOnlyImport && isTypeGroupInGroups && !isSortingTypesAmongThemselves) {
     impType = 'type';
   } else {
     impType = importType(importEntry.value, context);
   }
-  if (!excludedImportTypes.has(impType)) {
+
+  if (!excludedImportTypes.has(impType) && !isExcludedFromPathRank) {
     rank = computePathRank(ranks.groups, ranks.pathGroups, importEntry.value, ranks.maxPosition);
   }
-  if (typeof rank === 'undefined') {
+
+  if (rank === undefined) {
     rank = ranks.groups[impType];
+
+    if(rank === undefined) {
+      return -1;
+    }
   }
+
+  if (isTypeOnlyImport && isSortingTypesAmongThemselves) {
+    rank = ranks.groups['type'] + rank / 10;
+  }
+
   if (importEntry.type !== 'import' && !importEntry.type.startsWith('import:')) {
     rank += 100;
   }
@@ -536,10 +552,20 @@ function computeRank(context, ranks, importEntry, excludedImportTypes) {
   return rank;
 }
 
-function registerNode(context, importEntry, ranks, imported, excludedImportTypes) {
-  const rank = computeRank(context, ranks, importEntry, excludedImportTypes);
+function registerNode(context, importEntry, ranks, imported, excludedImportTypes, isSortingTypesAmongThemselves) {
+  const rank = computeRank(context, ranks, importEntry, excludedImportTypes, isSortingTypesAmongThemselves);
   if (rank !== -1) {
-    imported.push({ ...importEntry, rank });
+    let importNode = importEntry.node;
+
+    if(importEntry.type === 'require' && importNode.parent.parent.type === 'VariableDeclaration') {
+      importNode = importNode.parent.parent;
+    }
+
+    imported.push({
+      ...importEntry,
+      rank,
+      isMultiline: importNode.loc.end.line !== importNode.loc.start.line
+    });
   }
 }
 
@@ -665,7 +691,7 @@ function removeNewLineAfterImport(context, currentImport, previousImport) {
   return undefined;
 }
 
-function makeNewlinesBetweenReport(context, imported, newlinesBetweenImports, distinctGroup) {
+function makeNewlinesBetweenReport(context, imported, newlinesBetweenImports, newlinesBetweenTypeOnlyImports_, distinctGroup, isSortingTypesAmongThemselves, isConsolidatingSpaceBetweenImports) {
   const getNumberOfEmptyLinesBetween = (currentImport, previousImport) => {
     const linesBetweenImports = getSourceCode(context).lines.slice(
       previousImport.node.loc.end.line,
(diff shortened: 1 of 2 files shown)


Xunnamius:
> I definitely would prefer to see 3 of those peeled off into separate PRs, so that each thing can be easily reviewed separately. Thanks!

Gotcha, I'll slice it up this weekend.

> (also, could you please add parens to ensure proper precedence is conveyed when there's a combination of && and ||?)

I thought I ran my modifications through Prettier, but I probably missed a few blocks. I'll take a look

ljharb:
I think prettier *removes* those parens, which is one of its many flaws :-)

jftanner:
For the record, I definitely would like this feature. As evidenced by the fact that I tried to use it per [the docs on `main`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md#sorttypesamongthemselves). :)

ljharb:
@Xunnamius any progress? i'd love to get this merged for the next release.

Xunnamius:
Sorry, something I'm working on is, of course, taking longer than I anticipated. This PR is up next 😊

src/rules/order.js
```diff
@@ -781,6 +793,10 @@ module.exports = {
               'never',
             ],
           },
+          sortTypesAmongThemselves: {
```
ljharb, line 796: this is a bit wordy; let's rename it throughout to just `sortTypes`?

Xunnamius: I considered the shorter name at first, but I don't believe it's descriptive enough. The `import/order` rules already sort type-only imports; `"type"` is already a valid `groups` name, and its `groups` that actually controls if type-only imports are sorted or not sorted, right? With `sortTypes`, it wouldn't be immediately clear what turning off "sorting types" is supposed to mean with respect to the other rule options. You'd have scenarios where `sortTypes: false` but types are still being sorted...? And would `sortTypes: false` disable `named.types` sorting? This kind of ambiguity would confuse future me, at least, and would cause an unnecessary trip to the docs, docs which would have to become even more verbose to explain why `sortTypes` doesn't control if types are sorted.

Imo, `sortTypes` doesn't communicate the effect of flipping this switch and can even be confusing. We'd have `sortTypes`, `"type"` in `group`, `"type"` in `pathGroups.group`, `"type" ` in `pathGroupExcludedImportTypes` (which is already wordy and already somewhat-confusingly uses the word "Types" in its own name different to how the word is used in other option names like `named.types` and `newlines-between-types`), etc.

Whereas `sortTypesAmongThemselves` makes it clear that we're not interfering with the other "type"-related options, and we're not controlling if type-only imports are sorted vs not sorted but if they're sorted _among themselves_ vs if they're sorted _among the normals_.

Xunnamius: I think that works 😊

docs/rules/order.md
```diff
@@ -80,35 +80,35 @@ var path = require('path');
...
+- [`pathGroups`][8]
+- [`pathGroupsExcludedImportTypes`][9]
+- [`distinctGroup`][32]
+- [`newlines-between`][20]
+- [`alphabetize`][30]
+- [`named`][33]
+- [`warnOnUnassignedImports`][5]
+- [`sortTypesGroup`][7]
+- [`newlines-between-types`][27]
+- [`consolidateIslands`][25]
```
ljharb, line 111:
```suggestion
 - [`groups`][18]
 - [`pathGroups`][8]
 - [`pathGroupsExcludedImportTypes`][9]
 - [`distinctGroup`][32]
 - [`newlines-between`][20]
 - [`alphabetize`][30]
 - [`named`][33]
 - [`warnOnUnassignedImports`][5]
 - [`sortTypesGroup`][7]
 - [`newlines-between-types`][27]
 - [`consolidateIslands`][25]
```
this seems like an unrelated formatting change

docs/rules/order.md
```diff
@@ -129,19 +129,22 @@ together at the end.
...
+        // Then sibling and parent imports. They can be mingled together
+        ["sibling", "parent"],
+        // Then index file imports
+        "index",
+        // Then any arcane TypeScript imports
+        "object"
+        // Then the omitted imports: internal, external, type, unknown
+      ]
+    }
+  ]
```
ljharb, line 147:
```suggestion
  "import/order": ["error", {
    "groups": [
      // Imports of builtins are first
      "builtin",
      // Then sibling and parent imports. They can be mingled together
      ["sibling", "parent"],
      // Then index file imports
      "index",
      // Then any arcane TypeScript imports
      "object"
      // Then the omitted imports: internal, external, type, unknown
    ],
  }],
```
this seems like an unrelated formatting change

docs/rules/order.md
Xunnamius: Yep yep, I pushed the fix for this and a few other stealth formatting issues  right as you posted this 😄.

ljharb (commented):
This PR adds 3 options but only seems to be adding 1 to the schema. Should the schema be amended, or should 2 of the options be removed so they can land in different PRs?

Xunnamius:
@ljharb Hmm, no this PR should still only contain the changes relevant to `sortTypesGroup`, but I think I see what happened with [docs/rules/order.md](https://github.com/import-js/eslint-plugin-import/pull/3104/files#diff-a0c6b5c921e0e12d6581db31823f00f4a015b2e92a9a0b8bde1474299d6ee4de). [752f00d](https://github.com/import-js/eslint-plugin-import/commit/752f00d4e38549651094d49f31a659512d834f35) is the result of rebasing this PR by hand (cherry-pick), which was required due to the rename to `sortTypesGroup` (the `main` branch has the docs PR already merged in, which was using the old name). Seems the force pushes after `752f00d` tripped over what my cherry-pick was trying to avoid

ljharb:
If that’s the case then my bad :-)

Xunnamius:
I haven't forgotten about this :) So long as these fires don't reach my work station, I aim to wrap these PRs up this weekend!

Xunnamius:
Alrighty, this should be ready for review. Also includes relevant documentation updates. Though codecov says coverage decreased, perhaps due to some deleted lines that were previously covered, codecov's [commit report](https://app.codecov.io/gh/import-js/eslint-plugin-import/commit/c004b59ea5bd64a4e48920b5bf43d5e26a5872a1) claims 100% coverage.

ljharb (approved):

PR #3105, opened 2024-11-19
[New] `extensions`: allow enforcement decision overrides based on specifier

Xunnamius:
**This PR implements in `import/extensions`: a new option enabling developers to override the decision to enforce or ignore extensions for imports with matching specifiers.**

```bash
```

## Prevent false negatives when checking the extensions of bespoke imports

I already use `import/order` with [custom `pathGroups`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md#pathgroups-array-of-objects). This allows me to control the grouping of certain paths, such as custom aliases from `tsconfig.json`'s `paths`. This in turn ensures all imports are sorted properly, even when the project is using syntactically sound but otherwise bespoke custom aliases like `my-project:api.js` (in lieu of `../../../../src/api.js`).

Unfortunately, if I mistype `my-project:api` (i.e. I forgot the extension), the `import/extensions` rule will still pass; the enforcement decision is _"ignore"_ when it should have been _"enforce"_. This is because, like `import/order`, the grouping strategy employed by this rule is quite coarse-grained, but unlike `import/order`, this rule is missing a `pathGroups`-like setting for finer-grained sorting. The goal of this PR is to fill that gap.

The documentation updates associated with the changes in this PR are part of [a separate PR](https://github.com/import-js/eslint-plugin-import/pull/3106) and can be previewed [here](https://github.com/Xunnamius/eslint-plugin-import/blob/main/docs/rules/extensions.md#rule-details) (towards the bottom of the section). This PR also includes tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ### Added
 - add [`enforce-node-protocol-usage`] rule and `import/node-version` setting ([#3024], thanks [@GoldStrikeArch] and [@sevenc-nanashi])
 - add TypeScript types ([#3097], thanks [@G-Rath])
+- [`extensions`]: add `pathGroupOverrides to allow enforcement decision overrides based on specifier ([#3105], thanks [@Xunnamius])
 
 ### Fixed
 - [`no-unused-modules`]: provide more meaningful error message when no .eslintrc is present ([#3116], thanks [@michaelfaith])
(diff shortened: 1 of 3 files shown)


ljharb (commented):
What happens if two contradictory pathGroupOverrides options end up applying to the same file?

Xunnamius:
> What happens if two contradictory pathGroupOverrides options end up applying to the same file?

It mirrors the behavior of `pathGroups` from import/order: the first path group with a matching minimatch pattern wins.

Xunnamius:
@ljharb Did this PR need any further work?

ljharb:
@Xunnamius there's a few lines in the diff that aren't covered by tests.

Xunnamius:
@ljharb I'm not seeing any uncovered lines [on codecov](https://app.codecov.io/gh/import-js/eslint-plugin-import/pull/3105), though I did see some failing eslint tests. They are now passing.

ljharb (approved):

alexandre-le-borgne:
When can we expect this feature to be released?

Xunnamius:
@alexandre-le-borgne Just a guess, but might be waiting on https://github.com/import-js/eslint-plugin-import/pull/3152, which I should find the time to finish up either today or tomorrow :)

Issue #3040, opened 2024-08-30
How to make it recognise internal modules as not external?


RichMatthews:
I've got this:

```
        "pathGroups": [
          {
            "pattern": "react",
            "group": "external",
            "position": "before"
          },
          {
            "pattern": "{app,components,constants,context,generated,images,pages,shared,testing}/**",
            "group": "internal",
            "position": "after"
          }
        ]
```
but for certain modules like `shared` I import like so `import { Button } from "shared"` so it think it's an external module. How can I get it to put all `shared` stuff within the internal category? I also have other packages like this (e.g. context) so I need a way of handling multiple.
The ultimate order I want is:

import react

import emotion

import components/something
import shared/button

import ./currentDir

RichMatthews:
@ljharb any ideas?

Issue #2736, opened 2023-03-15
`import/order` ignores `pathGroups`


TheZoker:
Hi there,

I'm trying to use `import/order` with this setting:
```json
"import/order": [
    "error",
    {
        "alphabetize": {
            "order": "asc",
            "caseInsensitive": false
        },
        "groups": [
            ["builtin", "external"],
            ["internal", "parent", "sibling", "index"]
        ],         
        "newlines-between": "always",
        "pathGroups": [
            {
                "pattern": "app/**",
                "group": "internal"
            }
        ],
        "distinctGroup": false
    }
],
```

When I run the code, the linter orders the imports like this:
```ts
import { NgModule } from '@angular/core';
import { RouterModule } from '@angular/router';
import { ExternalUserPasswordResetModalComponent } from 'app/account/password-reset/external/external-user-password-reset-modal.component';
import { ArtemisSharedModule } from 'app/shared/shared.module';

import { accountState } from './account.route';
import { ActivateComponent } from './activate/activate.component';
```

When instead it should order it like this:
```ts
import { NgModule } from '@angular/core';
import { RouterModule } from '@angular/router';

import { accountState } from './account.route';
import { ActivateComponent } from './activate/activate.component';
import { ArtemisSharedModule } from 'app/shared/shared.module';
import { ExternalUserPasswordResetModalComponent } from 'app/account/password-reset/external/external-user-password-reset-modal.component';
```

Can you tell me, why my `pathGroups` and my `groups` setting to not work?

Thanks!

inglec-arista:
I had to go digging through old issues, but the reason is that `builtin` and `external` are excluded from `pathGroups` by default:
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md#pathgroupsexcludedimporttypes-array

You'll want to set `pathGroupsExcludedImportTypes: []`

Issue which helped:
https://github.com/import-js/eslint-plugin-import/issues/2363

ackvf:
I found another weirdness. Using `internal` with `pathGroups` will not work, but it will work with `import/internal-regex` setting.

Before:
```js
import { X } from 'package/src/ui'
import { useCallback } from 'react'
import { Y } from 'package/src/utils'
import { something } from './something'
```
Expected:
```js
import { useCallback } from 'react'
import { X } from 'package/src/ui'
import { Y } from 'package/src/utils'
import { something } from './something'
```

This **will not** work.
```js
  rules: {
[...]

Issue #2708, opened 2023-02-07
import/order: Allow distinct grouping among same group types only for specific patterns


ZaeemKhaliq:
I have following config: -

 ```
pathGroups: [
    {
      pattern: 'react**',
      group: 'external',
      position: 'before',
    },
    {
      pattern: 'gatsby**',
      group: 'external',
      position: 'before',
    },
    {
      pattern: '@chakra-ui/**',
      group: 'external',
      position: 'after',
    },
    {
      pattern: '@emotion/**',
      group: 'external',
      position: 'after',
    },
  ],
```

If `distinctGroup` is false, I get this: -

```
import React, { useState, useEffect } from 'react';
import { Link as GatsbyLink } from 'gatsby';
import { CloseIcon, ChevronDownIcon, ArrowBackIcon, ChevronRightIcon } from '@chakra-ui/icons';
import styled from '@emotion/styled';
```

If `distinctGroup` is true, I get this: -

```
import React, { useState, useEffect } from 'react';

import { Link as GatsbyLink } from 'gatsby';

import { CloseIcon, ChevronDownIcon, ArrowBackIcon, ChevronRightIcon } from '@chakra-ui/icons';

import styled from '@emotion/styled';
```

What I want to achieve: -

```
import React, { useState, useEffect } from 'react';
import { Link as GatsbyLink } from 'gatsby';

import { CloseIcon, ChevronDownIcon, ArrowBackIcon, ChevronRightIcon } from '@chakra-ui/icons';
import styled from '@emotion/styled';
```

Basically, I want to group style related libraries together, however all should belong to same `external` group.
Maybe an option like this can be included: -

 ```
pathGroups: [
    {
      pattern: 'react**',
      group: 'external',
      position: 'before',
    },
    {
      pattern: 'gatsby**',
      group: 'external',
      position: 'before',
    },
    {
      pattern: '@chakra-ui/**',
      group: 'external',
      position: 'after',
      isDistinct: true,  // This flag would indicate breaking into separate group for this and subsequent patterns for the same group type.
    },
    {
      pattern: '@emotion/**',
      group: 'external',
      position: 'after',
    },
  ],
```

Or anything else. 
TIA for any help.

Issue #2540, opened 2022-09-02
[question] Enforce internal/parent import order with alias


andreafspeziale:
Hello! 
I can't find a way to enforce import order to internal / parent using aliases. Maybe it is just not possible.

Example:

```
--- src
     |--- folderA
           |--- fileA1.ts
           |--- fileA2.ts
     |--- folderB
           |--- folderC
                 |--- fileC1.ts
           |--- fileB1.ts
           |--- fileB2.ts
```

Set an alias in my `tsconfig.json`

```json
{
  "compilerOptions": {
    ...
    "baseUrl": "./",
    "paths": {
      "@src/*": ["src/*"],
    }
  }
}
```

In `fileC1.ts`

```ts
import { foo } from '@src/folderA/fileA1'
import { bar } from '@src/folderB/fileB1'
```

I would really love to be able to enforce the import from `B` (parent) to happen after import `A` (internal) but I'm not sure it is possible since I believe that `alias` would "hide" it.

Thanks in advance!

ljharb:
I'm unclear on what you're asking. Wouldn't folderA and folderB both be internal?

andreafspeziale:
Maybe I'm just totally wrong but based on my understanding being in `fileC1.ts`, file `fileB1` is a parent and `fileA1.ts` is internal. So given a configuration like

```js
'import/order': [
      2,
      {
        groups: ['builtin', 'external', 'internal', 'parent', 'sibling', 'index'],
        'newlines-between': 'never'
      },
    ],
```

I would like to enforce 

`import { bar } from '@src/folderB/fileB1'` after `import { foo } from '@src/folderA/fileA1'`

With the above given config and (I believe) since I'm using alias the linter doesn't care about the order so:

```
import { bar } from '@src/folderB/fileB1'
import { foo } from '@src/folderA/fileA1'
```

is still valid.

ljharb:
oh right, you're in fileC at that point. Yes, folderB would be a parent and folderA would be internal.

So what you're saying is, you WANT them to be sorted so that folderA comes first (because it's internal), but because it's using a TS alias, it's recognizing it as a parent?

Can you share your eslint config? Hopefully you're using the TS eslint resolver.

andreafspeziale:
You got it! (btw I don't really know if it's `folderA` recognize as parent or is it viceversa - so it's `folderB` recognize as internal, what I know is that switching their order would not trigger any error)

`eslintrc.js`

```js
module.exports = {
[...]

PR #3106, opened 2024-11-19
[Docs] `extensions`, `order`: improve documentation

### LINKED ISSUE #2897: `import/order` `pathGroups` and `pathGroupsExcludedImportTypes` is too confusing
There is some convoluted logic involved in how `import/order` with `pathGroups` and `pathGroupsExcludedImportTypes` work.

I described that in detail here https://github.com/import-js/eslint-plugin-import/issues/2736#issuecomment-1760649315, but in short, the gist is that if you have an import such as `import { X } from 'package/src/ui'` that looks as if it was an `external` import, it will get ignored by `pathGroups` pattern and will not even get tested.

This seems counterintuitive as the point of `pathGroups` in the first place is to match an import and assign it to a different group than implicit. But if it is first implicitly assigned to an _excluded_ group, it will not even get matched against `pathGroups` an will not get assigned to the correct group at all.

From my point of view, `pathGroups` should be invoked first. Those matched imports should then be ignored for the next step where all unmatched imports are assigned implicit groups, after which all imports (now assigned to groups) are ordered according to expectations.

If for some reason you decide against this flow of reason, you should improve the documentation to better explain what is going on.

Xunnamius:
Resolves #2897 \
Related to #3104 and #3105 \
Potentially related to #2708, #2304, #2540

I can slice and dice this PR in whatever way is most useful.

The changes to the `import/extensions` documentation ([preview](https://github.com/Xunnamius/eslint-plugin-import/blob/main/docs/rules/extensions.md)) are not so massive and should merge cleanly.

The changes to the `import/order` documentation ([preview](https://github.com/Xunnamius/eslint-plugin-import/blob/main/docs/rules/order.md)), however, are much more extensive. I've been referring back to these docs on and off for several months as I've tweaked my linter configurations, and every time I do I pretty much leave confused and fall back on stepping through eslint-plugin-import's source to see what's actually happening. Along with documenting the new functionality introduced in #3104, this PR attempts to demystify `import/order`'s documentation in the following ways:

- **There is no documentation describing how an import is grouped.** What classifies an import as "external" vs "internal" vs "builtin"? What does "sibling" or "parent" actually mean? Every time I come back to these docs, I find myself looking for a technical description that doesn't exist and so I ultimately guess, which leads to problems 😅. For this PR, I've looked through the rule's source and have attempted to describe the grouping algorithm with specificity. Since I've only spent a day or so with this repo's source code, it may not be totally accurate and should be scrutinized for mistakes.

- **Some of the examples don't work.** They've been fixed (and #3104 adds corresponding tests for them).

- **The options headings contain their type description alongside their names.** This means references into this document from the wider web will break whenever one of the settings is tweaked. I've encountered several of these "dead links" over time. Worse, they make for truly ugly headings. I believe it best to instead use short simple headings that are unlikely to change, such as each option's full name alone.

- **The type descriptions of the options (e.g. `[array of this or that]`) and their defaults are vague and confusing and are documented inconsistently.** Some of the descriptions of the functionality of each option are similarly confusing, such as the description for `pathGroups`. In my review of the issues literature, I found I [was](https://github.com/import-js/eslint-plugin-import/issues/3040) [not](https://github.com/import-js/eslint-plugin-import/issues/2736#issuecomment-1760649315) [alone](https://github.com/import-js/eslint-plugin-import/issues/2897) in this. This is my stab at making this aspect of the documentation more approachable.

- **There are several "gotchas" and "footguns" when using some of these options.** In this PR, these are now shouted out using GFM alert syntax.

- **Option sections are formatted inconsistently.** Several of the examples and their descriptions are formatted differently than the others. This PR ensures all sections are using consistent formatting and structure.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,5 +1,8 @@
 {
   "line-length": false,
+  "no-duplicate-heading": {
+    "siblings_only": true
+  },
   "ul-indent": {
     "start_indent": 1,
     "start_indented": true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ## [Unreleased]
 
+### Changed
+- [Docs] `extensions`, `order`: improve documentation ([#3106], thanks [@Xunnamius])
+
 ## [2.31.0] - 2024-10-03
 
 ### Added
@@ -1152,6 +1155,7 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#3106]: https://github.com/import-js/eslint-plugin-import/pull/3106
 [#3073]: https://github.com/import-js/eslint-plugin-import/pull/3073
 [#3072]: https://github.com/import-js/eslint-plugin-import/pull/3072
 [#3071]: https://github.com/import-js/eslint-plugin-import/pull/3071
@@ -2018,6 +2022,7 @@ for info on changes for earlier releases.
 [@wtgtybhertgeghgtwtg]: https://github.com/wtgtybhertgeghgtwtg
 [@xM8WVqaG]: https://github.com/xM8WVqaG
 [@xpl]: https://github.com/xpl
+[@Xunnamius]: https://github.com/Xunnamius
 [@yesl-kim]: https://github.com/yesl-kim
 [@yndajas]: https://github.com/yndajas
 [@yordis]: https://github.com/yordis
--- a/docs/rules/extensions.md
+++ b/docs/rules/extensions.md
@@ -10,13 +10,13 @@ In order to provide a consistent use of file extensions across your code base, t
 
 This rule either takes one string option, one object option, or a string and an object option. If it is the string `"never"` (the default value), then the rule forbids the use for any extension. If it is the string `"always"`, then the rule enforces the use of extensions for all import statements. If it is the string `"ignorePackages"`, then the rule enforces the use of extensions for all import statements except package imports.
 
-```json
+```jsonc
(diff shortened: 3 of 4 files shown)


Xunnamius:
> @Xunnamius just curious; did you use an LLM in any part of your PR description?

Haha, no 🤭. I write the docs for the fancy linear algebra, it doesn't madlib my docs for me. I hope my writing doesn't read like slop!

And thanks for taking a look at the PR 👍

ljharb:
@Xunnamius glad to hear it, it'd have been the first time it wasn't immediately obvious to me :-p your writing is fine :-)

jftanner:
Should this have been merged before [the implementation](https://github.com/import-js/eslint-plugin-import/pull/3104)?

I just spent entirely too much time trying to figure out why my new ESLint config throws `Unexpected property "sortTypesAmongThemselves".`. 🤣

ljharb:
@jftanner this commit isn't released yet - you always need to check the docs on the version tag you're using, not on master/main :)

jftanner:
Ah. My bad. Google took me to `main`, and I'm used to living in a semantic-release world where `main` == `latest`.

Carry on. :)

ljharb:
actually, @Xunnamius, i think #3106 erroneously included `sortTypesAmongThemselves` documentation.

Xunnamius:
@ljharb My goal was to isolate all documentation changes to a documentation-only PR, mostly because I thought they warranted further discussion on their own merits, but also because I find it's easier for me to write good documentation (wrt section order, refs between sections, etc) in one pass

ljharb:
Understood, but I prefer each feature to land complete, with its own docs :-) it should be pretty easy to revert my commit during your rebase.

Issue #3107, opened 2024-11-20
eslint-plugin-import silently fails if the module is minified


RPGillespie6:
I have 2 files:

1. test.js
```
import { bogus } from './bogus.js';

bogus()
```

2. bogus.js
```
function foo() {
    console.log('foo');
}

export {
    foo
}
```

If `bogus.js` is minified, eslint-plugin-import silently fails:

```
$ npx eslint js/test.js        # Working case

/home/user/test/js/test.js
  1:10  error  bogus not found in './bogus.js'  import/named

✖ 1 problem (1 error, 0 warnings)

$ npx minify js/bogus.js > bogus2.js
$ mv bogus2.js js/bogus.js
$ npx eslint js/test.js        # Silent failure case
$
```

Here's my eslint config:
```
import globals from "globals";
import pluginJs from "@eslint/js";
import importPlugin from "eslint-plugin-import";

/** @type {import('eslint').Linter.Config[]} */
export default [
  importPlugin.flatConfigs.recommended,
  {languageOptions: { globals: globals.browser }},
  pluginJs.configs.recommended,
];
```

Tested with eslint 8.X and 9.X, issue is present in both. Using `eslint-plugin-import:2.31.0`

Edit: Actually, some types of minification work. For example:
```
function a(){console.log("foo")};export{a}; // works
function a(){console.log("foo")}export{a}; // doesn't work
```

Is this library using regex or something to detect exports instead of the file's AST? That's the only thing I can think of that would make it so that the absence of an optional semicolon causes silent failures

ljharb:
That’s very strange. (altho minifying a package never makes sense, ever - minification is something that should only ever be done by a top-level app)

I’ll look into it.

ljharb:
Why would it need to be minified, though? Transpiling is fine, but minification is something else altogether.

RPGillespie6:
> Why would it need to be minified, though? Transpiling is fine, but minification is something else altogether.

So that devs don't attempt to modify it. It's a hint to the developer that it's a build artifact of upstream source, and that they should be modifying the upstream source. 

This is a very standard and common practice for distributing libraries. For example: https://pixijs.download/v8.5.2/pixi.min.mjs

This is PixiJS as an esmodule. If I'm using this in my project, I would hope:

```
import { bogus } from "./pixi.min.mjs"
```

Would get flagged by `eslint-plugin-import` because that's not a valid symbol exported by pixijs

ljharb:
[...]

PR #3115, opened 2024-12-14
[meta] issue template with guidance for bug reports

soryy708:
In relation to a bug report about the plugin failing silently in some edge case: https://github.com/import-js/eslint-plugin-import/issues/3107

It's unclear how to make `eslint-plugin-import` show its detailed logs, which would have made the failure not silent, and saved contributor time in triage and debugging.

Lets add an issue template, with instructions that will guide the issue author to:
- Enable debug logging for the plugin, to narrow down what failed
- Provide an [MCVE](https://stackoverflow.com/help/minimal-reproducible-example)


ljharb (approved):
(replaced the windows line endings with LF)

PR #3116, opened 2024-12-15
[Fix] `no-unused-modules`: don't error out when running with flat config and an eslintrc isn't present

### LINKED ISSUE #3079: Following the ESLint 9 support, `import/no-unused-modules` only works in flat config if an `.eslintrc` file exists as well
I have seen various discussions about this rule, but not quite in the exact same context, so I am unsure if that could be merged with an existing issue: 

I upgraded to `eslint-plugin-import` v2.31.0, which officially supports ESLint V9. Enabling the `import/no-unused-modules` with the existing `unusedExports` option results in the ESLint CLI not recognizing the `eslint.config.js` unless there's a `.eslintrc` file next to it (the `.eslintrc` file can be empty; it just ignores it).

```

ESLint: 9.11.1

ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:

    npm init @eslint/config@latest

ESLint looked for configuration files in <project rootDir> and its ancestors. If it found none, it then looked in your home directory.

If you think you already have a configuration file or if you need more help, please stop by the ESLint Discord server: https://eslint.org/chat
```

As long as the `.eslintrc` exists, the rule works as expected.

michaelfaith:
This change adjusts what we're doing if an error is thrown while attempting to enumerate lintable files in the `no-used-modules` rule, when users are running with flat config.  Now, if `FileEnumerator` throws due to a lack of eslintrc _and_ the user is running with flat config, we catch the error and rethrow with a more informative message.

I also cleaned up the original aspects of the implementation that was using the proposed eslint context functions, since that proposal was walked back and the API was never introduced.

Note: This isn't an ideal state, since this rule still relies on the legacy api to understand what needs to be ignored.  Remaining tethered to the legacy config system is going to need to be solved at some point.

Fixes #3079


.gitignore
```diff
@@ -52,3 +52,6 @@ npm-shrinkwrap.json
 
 # macOS
 .DS_Store
+
+# vscode settings
+.vscode
```
ljharb, line 57: this should probably go in a global gitignore rather than in every individual project

src/rules/no-unused-modules.js
```diff
@@ -51,21 +50,81 @@ function requireFileEnumerator() {
...
-    extensions,
-  });
+  // We need to know whether this is being run with flat config in order to
+  // determine whether we need to customize the FileEnumerator or not.
+
+  // This condition is sufficient to test in v8, since the environment variable
+  // is necessary to turn on flat config
+  let isUsingFlatConfig =
+    process.env.ESLINT_USE_FLAT_CONFIG &&
+    process.env.ESLINT_USE_FLAT_CONFIG !== 'false';
```
ljharb, line 67:
```suggestion
  let isUsingFlatConfig = process.env.ESLINT_USE_FLAT_CONFIG &&
    process.env.ESLINT_USE_FLAT_CONFIG !== 'false';
```
linebreaks after the `=` is a signature prettier thing :-(

ljharb (commented):
Looks pretty good, thanks!

.gitignore
michaelfaith: Some projects I work in do keep some workspace settings in source control.  This is the only one where I need to actively turn off formatting.  I can remove it, if you prefer.  it's just a bit of a pain to always have to remember to turn it off.

ljharb: hmm - i'd expect those projects would explicitly *un*ignore it in their gitignore since so many people globally gitignore vscode settings already. certainly we can keep it in here if it's that important to you.

src/rules/no-unused-modules.js
michaelfaith: Updated

michaelfaith:
Addressed lint errors

ljharb (approved):

michaelfaith:
Updated, based on feedback in #3079 to not attempt to avoid the error, but to instead rethrow with a more informative message.  I updated the PR description to reflect the new approach.

ljharb:
I suppose we could have a test make a temp dir (using `tmp`), copy an example into it, run `npm install` in it, and then invoke eslint inside it?

michaelfaith:
> I suppose we could have a test make a temp dir (using `tmp`), copy an example into it, run `npm install` in it, and then invoke eslint inside it?

Added a test that does the following:
- copies the v9 example folder
- runs exec to install the plugin from cwd
- run the lint command
- catches the error and validates that the message includes our new verbiage

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - add [`enforce-node-protocol-usage`] rule and `import/node-version` setting ([#3024], thanks [@GoldStrikeArch] and [@sevenc-nanashi])
 - add TypeScript types ([#3097], thanks [@G-Rath])
 
+### Fixed
+- [`no-unused-modules`]: don't error out when running with flat config and an eslintrc isn't present ([#3116], thanks [@michaelfaith])
```
michaelfaith, line 14: I think this changelog entry will need to be revised, but I didn't want to adjust that since you added it.

ljharb: totally fine either way; the format should hopefully be self-explanatory based on the commit message

michaelfaith: Updated

michaelfaith:
@ljharb I got this about where I want it, except for one question to you.  I added the test to the existing set of unit tests just because all the CI plumbing was already in place to add it there.  Though, it doesn't really feel right for it to be there, since it's more of an integration test.  Would you rather it be where I have it now, or add a new workflow to CI for doing this kind of integration testing?

ljharb:
I think that it's fine where it is to land this, but it might make sense to move it in a followup?

michaelfaith:
> I think that it's fine where it is to land this, but it might make sense to move it in a followup?

Sounds good

> … although i do see there's 2 uncovered lines in the diff

Do you mean the catch block?  If so, that's because the new test isn't really contributing to the coverage report, since it's building the plugin and running lint as a separate exec command.

ljharb (approved):

PR #3124, opened 2024-12-20
[utils] [fix] `unambiguous`: detect modules exported from minified code

### LINKED ISSUE #3107: eslint-plugin-import silently fails if the module is minified

michaelfaith:
This change adjusts the regex pattern used to detect modules to support detection on minified code.

image

Closes #3107

### DIFF
--- a/tests/files/minified/no-newline.js
+++ b/tests/files/minified/no-newline.js
@@ -1,0 +1,3 @@
+function y() {
+    console.log("y");
+}export {y};
--- a/tests/files/minified/one-line-no-semi-renamed.js
+++ b/tests/files/minified/one-line-no-semi-renamed.js
@@ -1,0 +1,1 @@
+function a(){console.log('foo')}export{a as foo};
--- a/tests/files/minified/one-line-no-semi.js
+++ b/tests/files/minified/one-line-no-semi.js
@@ -1,0 +1,1 @@
+function a(){return true}export{a};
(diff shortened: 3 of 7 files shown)


ljharb (approved):
added some more test cases from the linked issue

PR #3127, opened 2024-12-23
[New] `order`: allow controlling intragroup spacing of type-only imports via `newlines-between-types`

Xunnamius:
Depends on #3104

**This PR implements in `import/order`: a new option analogous to `newlines-between` but specifically for controlling newlines between type-only imports.**

```bash
```

## Allow controlling intragroup spacing of type-only imports

This is implemented via `newlines-between-types`. The proposed documentation corresponding to this new feature can be previewed [here](https://github.com/Xunnamius/eslint-plugin-import/blob/main/docs/rules/order.md#newlines-between-types).

Given this code:

```typescript
import type A from 'fs';
import type B from 'path';
import type C from '../foo.js';
import type D from './bar.js';
import type E from './';

import a from 'fs';
import b from 'path';

import c from '../foo.js';

import d from './bar.js';

import e from './';
```

And the following settings, the rule check will fail:

```jsonc
{
  "groups": ["type", "builtin", "parent", "sibling", "index"],
  "sortTypesAmongThemselves": true,
  "newlines-between": "always"
}
```

With `--fix` yielding:

```typescript
import type A from 'fs';
import type B from 'path';

import type C from '../foo.js';

import type D from './bar.js';

import type E from './';

import a from 'fs';
import b from 'path';

import c from '../foo.js';

import d from './bar.js';

import e from './';
```

However, with the following settings, the rule check will succeed instead:

```diff
{
  "groups": ["type", "builtin", "parent", "sibling", "index"],
  "sortTypesAmongThemselves": true,
  "newlines-between": "always",
+ "newlines-between-types": "ignore"
}
```

</details>

`sortTypesAmongThemselves` allows sorting type-only and normal imports separately. By default, `newlines-between` will govern all newlines between import statements like normal.

I generally want my type-only imports to be sorted for ease of reference but never have newlines between them (save space) while I want my normal imports (which I tend to visually peruse more often) to be aesthetically pleasing, grouped, and sorted. `newlines-between` is too coarse-grained for this, so this PR introduces `newlines-between-types`, a setting identical to `newlines-between` except it only applies to type-only imports, and only when `sortTypesAmongThemselves` is enabled (i.e. it is backward-compatible).

When `newlines-between` and `newlines-between-types` conflict, `newlines-between-types` takes precedence for type-only imports. For normal imports, `newlines-between-types` is ignored entirely.

One issue that might warrant further discussion is which setting governs the newline separating type-only imports from normal imports. Right now, I have it so `newlines-between-types` controls this space, but perhaps it should be its own setting.

### DIFF
--- a/docs/rules/order.md
+++ b/docs/rules/order.md
@@ -107,6 +107,7 @@ This rule supports the following options (none of which are required):
  - [`named`][33]
  - [`warnOnUnassignedImports`][5]
  - [`sortTypesGroup`][7]
+ - [`newlines-between-types`][27]
 
 ---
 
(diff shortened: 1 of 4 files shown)


src/rules/order.js
```diff
@@ -678,35 +690,72 @@ function makeNewlinesBetweenReport(context, imported, newlinesBetweenImports, di
...
+      currentImport,
+      previousImport,
+    );
+
+    const isTypeOnlyImport = currentImport.node.importKind === 'type';
+    const isPreviousImportTypeOnlyImport = previousImport.node.importKind === 'type';
+
+    const isNormalImportFollowingTypeOnlyImportAndRelevant =      !isTypeOnlyImport && isPreviousImportTypeOnlyImport && isSortingTypesGroup;
+
+    const isTypeOnlyImportAndRelevant =      isTypeOnlyImport && isSortingTypesGroup;
```
Xunnamius, line 708: I believe `eslint --fix` is adding these spaces and moving around parentheses elsewhere. If this isn't desired I can revert the auto-fix. If this is cool, then all good

Xunnamius:
Like with the other PR, codecov is [reporting](https://app.codecov.io/gh/import-js/eslint-plugin-import/commit/d472b801bba109bc7a798d1ad0c2639dff7c8877) patch coverage at 100% with no uncovered lines.

When I look at "indirect changes" I'm seeing the same `src/rules/order.js` file (among a bunch of seemingly unrelated files) but with uncovered lines that have nothing to do with the tests added by this PR. This PR also doesn't remove any tests, so I'm not quite sure what's going on there. Could be that some formerly-covered source was deleted. Everything else seems to have passed though :)

docs/rules/order.md
```diff
@@ -592,6 +593,135 @@ This happens because [type-only imports][6] are considered part of one global
...
 The same example will pass.
 
+### `newlines-between-types`
+
+Valid values: `"ignore" | "always" | "always-and-inside-groups" | "never"` \
+Default: the value of [`newlines-between`][20]
+
+> \[!NOTE]
+>
+> This setting is only meaningful when [`sortTypesGroup`][7] is enabled.
```
ljharb, line 603: should we use the schema to ensure that this option can't be set to `true` unless `sortTypesGroup` is also set to true?

ljharb (approved):

docs/rules/order.md
Xunnamius: Sounds like a good idea to me. I was not aware that was a thing :)

ljharb: it often requires a bit of gymnastics, but it's usually worth it for the DX.

ljharb: @Xunnamius is this change something you're planning to make?

Xunnamius: @ljharb Yep, got a bit of free time just now

src/rules/order.js
```diff
@@ -890,6 +890,14 @@ module.exports = {
...
         },
         additionalProperties: false,
+        dependencies: {
+          'newlines-between-types': {
+            properties: {
+              sortTypesGroup: { enum: [true] },
+            },
+            required: ['sortTypesGroup'],
+          },
+        },
```
Xunnamius, line 900: @ljharb I came up with this after reading the json schema docs. I'm not sure if ESLint's rule tester can test for schema validation errors? I did some ad-hoc testing and it appears functional

Xunnamius: This got me thinking: should a dependency also be added between `sortTypesGroup` and `groups` (ensuring `groups` contains `'types'`)? Should we add a dependency between `consolidateIslands` and `newlines-between`/`newlines-between-types` (ensuring at least one is set to `'always-and-inside-groups'`)?

ljharb: Generally, anything that doesn’t make sense should be prohibited by the schema.

Xunnamius: Just making sure, since the `sortTypesGroup` PR was already accepted :)

I'll open a new PR for that change

ljharb: hmm, i'm confused - what is `dependencies`? i'm not familiar with that field in eslint rule schemas.

Xunnamius: I just discovered it the other day. It's the Draft-04 version of [`dependentSchemas`](https://json-schema.org/understanding-json-schema/reference/conditionals#dependentSchemas) and was the best way I could find to conditionally require `sortTypesGroups: true` only when `newlines-between-types` was present

PR #3128, opened 2024-12-23
[Fix] `order`: ensure arcane imports do not cause undefined behavior

Xunnamius:
Depends on #3127

**This PR implements in `import/order`: a fix to ensure `NaN` is never passed into rank-computing functions, which can result in undefined behavior.**

A demo package containing this fix is temporarily available for easy testing:

```bash
```

## Ensure strange imports do not cause strange behavior

There are certain edge cases where a `NaN` rank gets passed around, such as using an import with an absolute specifier under certain configurations. The fix concerns the `computeRank` function.

This PR includes a unit test to catch potential regressions.

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -535,6 +535,10 @@ function computeRank(context, ranks, importEntry, excludedImportTypes, isSorting
 
   if (typeof rank === 'undefined') {
     rank = ranks.groups[impType];
+
+    if (typeof rank === 'undefined') {
+      return -1;
+    }
   }
 
   if (isTypeOnlyImport && isSortingTypesGroup) {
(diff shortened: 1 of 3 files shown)


src/rules/order.js
```diff
@@ -526,8 +526,13 @@ function computeRank(context, ranks, importEntry, excludedImportTypes) {
   if (!excludedImportTypes.has(impType)) {
     rank = computePathRank(ranks.groups, ranks.pathGroups, importEntry.value, ranks.maxPosition);
   }
-  if (typeof rank === 'undefined') {
+
+  if (rank === undefined) {
     rank = ranks.groups[impType];
+
+    if(rank === undefined) {
```
ljharb, line 533:
```suggestion
  if (typeof rank === 'undefined') {
    rank = ranks.groups[impType];

    if (typeof rank === 'undefined') {
```

ljharb (commented):
seems good overall, assuming the test fails without the fix :-)

src/rules/order.js
Xunnamius: I hoped I could factor this change out from contrib-sort-types-among-themselves, but I've noticed the test needs `sortTypesGroup` to trigger the bad behavior :)

Xunnamius:
It's a little interesting that codecov believes this PR does not reduce coverage, this PR that includes the changes from the two preceding PRs... both of which codecov claims have reduced coverage 😄

ljharb (approved):

PR #3129, opened 2024-12-23
[New] `order`: add `consolidateIslands` option to collapse excess spacing for aesthetically pleasing imports

Xunnamius:
Depends on #3128

**This PR implements in `import/order`: a new option enabling the manipulation/consolidation of newlines around imports.**

```bash
```

## Collapse excess spacing for aesthetically pleasing imports

This is implemented via `consolidateIslands`. The proposed documentation corresponding to this new feature can be previewed [here](https://github.com/Xunnamius/eslint-plugin-import/blob/main/docs/rules/order.md#consolidateislands).

Given this code (which could be the output of a previous `--fix` pass):

```typescript
var fs = require('fs');
var path = require('path');
var { util1, util2, util3 } = require('util');

var async = require('async');

// Ugly but technically valid

var relParent1 = require('../foo');
var {
  relParent21,
  relParent22,
  relParent23,
  relParent24,
} = require('../');
var relParent3 = require('../bar');

var { sibling1,
  sibling2, sibling3 } = require('./foo');
var sibling2 = require('./bar');
var sibling3 = require('./foobar');
```

And the following settings, the rule check will _pass_:

```jsonc
{
  "newlines-between": "always-and-inside-groups"
}
```

However, when given the following instead, the rule check will _fail_:

```diff
{
  "newlines-between": "always-and-inside-groups",
+ "consolidateIslands": "inside-groups"
}
```

With `--fix` yielding:

```typescript
var fs = require('fs');
var path = require('path');
var { util1, util2, util3 } = require('util');

var async = require('async');

// Pretty

var relParent1 = require('../foo');

var {
  relParent21,
  relParent22,
  relParent23,
  relParent24,
} = require('../');

var relParent3 = require('../bar');

var { sibling1,
  sibling2, sibling3 } = require('./foo');

var sibling2 = require('./bar');
var sibling3 = require('./foobar');
```

Note how the intragroup "islands" of grouped single-line imports, as well as multi-line imports, are surrounded by new lines.

</details>

Essentially, I was looking for a `newlines-between`-like setting somewhere between `"never"` and `"always-and-inside-groups"`. I want newlines separating `groups`/`pathGroups` imports from one another (like `"always-and-inside-groups"`), newlines separating imports that span multiple lines from other imports (this is the new thing), and any remaining newlines deleted or "consolidated" (like `"never"`). The example above demonstrates this use case.

Right now, this is achievable with `newlines-between` set to `"always-and-inside-groups"` if you add additional newlines around multi-line imports to every file by hand. The goal of `consolidateIslands` is to allow `eslint --fix` to take care of the tedium in a backward-compatible way.

There was a slight complication with `consolidateIslands` though: while testing across a few mid-sized repos, I discovered my naive implementation caused a conflict when enabled alongside `sortTypesAmongThemselves: true`, `newlines-between: "always-and-inside-groups"`, and `newlines-between-types: "never"`... and then only when a normal import was followed by a multi-line type-only import. This conflict makes sense, since `newlines-between-types: "never"` wants no newlines ever and demands no newline separate type-only imports from normal imports (since, currently, `newlines-between-types` governs that space), yet `consolidateIslands` demands a newline separate _all_ multi-line imports from other imports.

To solve this, the current implementation has `newlines-between-types` yield to `consolidateIslands` whenever they conflict. I've also added a test to catch any regressions around this edge case.

### DIFF
--- a/tests/src/rules/order.js
+++ b/tests/src/rules/order.js
@@ -530,7 +530,7 @@ ruleTester.run('order', rule, {
         },
       ],
     }),
-    // Option newlines-between: 'always' with multiline imports #1
+    // Option newlines-between: 'always' with multi-line imports #1
     test({
       code: `
         import path from 'path';
@@ -546,7 +546,7 @@ ruleTester.run('order', rule, {
       `,
       options: [{ 'newlines-between': 'always' }],
     }),
-    // Option newlines-between: 'always' with multiline imports #2
+    // Option newlines-between: 'always' with multi-line imports #2
     test({
       code: `
         import path from 'path';
@@ -557,7 +557,7 @@ ruleTester.run('order', rule, {
       `,
       options: [{ 'newlines-between': 'always' }],
     }),
-    // Option newlines-between: 'always' with multiline imports #3
+    // Option newlines-between: 'always' with multi-line imports #3
     test({
       code: `
         import foo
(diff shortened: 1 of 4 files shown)


tests/src/rules/order.js
```diff
@@ -4228,6 +4228,172 @@ context('TypeScript', function () {
               },
             ],
           }),
+          // Ensure consolidateIslands: 'inside-groups', newlines-between: 'never', and newlines-between-types: 'always-and-inside-groups' do not fight for dominance
```
Xunnamius, line 4231: These two tests also cover the uncovered line codecov reported from the previous commit

ljharb (approved):

PR #3138, opened 2025-01-08
[meta] Update flatConfig example to include typescript config

intellix:
Was having an issue for a few hours trying to get import/no-unresolved not picking up relative imports that don't include the extension like so:
```ts
import { add } from './utils';
```

After asking inside Discord, kkapp helped me by pointing out that I may need to configure the resolver to look for "ts" extensions. After adding it I went on a crusade to figure out how I had managed to miss that when migrating to flat configs.

The example given for non-flat configs tells you to include the typescript config, but the flatConfig variant does not, so for an out-of-the-box experience when migrating, it can cause a headache.


ljharb (approved):
Thanks!

PR #3146, opened 2025-01-17
fix: import/order named not working with spread operator

### LINKED ISSUE #3145: import/order with named fails when exporting uses spread operator.
Rule:
```js
'import/order': [
  'warn',
  {
    named: true,
  },
],
```

minimal repro case:
```js
const test = {
  a: 1,
  browser: 2,
};

module.exports = {
  ...test,
  target: 'node',
};
```

UnderKoen:
Fixes #3145

### DIFF
--- a/tests/src/rules/order.js
+++ b/tests/src/rules/order.js
@@ -3648,6 +3648,24 @@ flowRuleTester.run('order', rule, {
         import typeof {foo} from 'common';
         import {bar} from 'common';
       `,
+    }),
+    test({
+      options: [
+        {
+          named: true,
+        },
+      ],
+      code: `
+        const test = {
+          a: 1,
+          browser: 2,
+        };
+
+        module.exports = {
+          ...test,
+          platform: 'node',
+        };
+      `,
     })],
   invalid: [
     test({
--- a/tests/src/rules/order.js
+++ b/tests/src/rules/order.js
@@ -23,6 +23,11 @@ const rule = require('rules/order');
 
 ruleTester.run('order', rule, {
   valid: [
+    test({
+      code: `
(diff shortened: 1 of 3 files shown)


ljharb:
We should only do this if node's cjs-module-lexer also recognizes it. Let's discuss in the issue.

tests/src/rules/order.js
```diff
@@ -3648,6 +3648,24 @@ flowRuleTester.run('order', rule, {
...
+        const test = {
+          a: 1,
+          browser: 2,
+        };
+
+        module.exports = {
+          ...test,
+          platform: 'node',
+        };
+      `,
```
ljharb, line 3668: i'd like to also include a test that demonstrates that we only recognize `platform` as a named export from this file.

UnderKoen: Im not that familiar with this codebase. How could I test this?

ljharb: you'd want to make a fixture file with this test code in it, and then make two test cases that import this test file - one that imports `platform` and works, and one that imports `a` and/or `browser` and fails.

UnderKoen: Im very sorry but can't figure it out. I added the cases but it doesn't fail? I suspect that I'm doing something wrong but cant figure out what.

ljharb: Can you push up what you have? I'll take a look.

UnderKoen: I tried looking at #3032 as it had fixture files I believe

UnderKoen: @ljharb did you have a chance to look?

UnderKoen: ?

PR #3152, opened 2025-01-26
[Fix] `order`: codify invariants from docs into config schema

Xunnamius:
Updating schema wrt https://github.com/import-js/eslint-plugin-import/pull/3127#discussion_r1926534364

EDIT: depends on #3129 for its added tests

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -590,18 +590,14 @@ function getRequireBlock(node) {
 
 const types = ['builtin', 'external', 'internal', 'unknown', 'parent', 'sibling', 'index', 'object', 'type'];
 
-// Creates an object with type-rank pairs.
-// Example: { index: 0, sibling: 1, parent: 1, external: 1, builtin: 2, internal: 2 }
-// Will throw an error if it contains a type that does not exist, or has a duplicate
+/**
+ * Creates an object with type-rank pairs.
+ *
+ * Example: { index: 0, sibling: 1, parent: 1, external: 1, builtin: 2, internal: 2 }
+ */
 function convertGroupsToRanks(groups) {
   const rankObject = groups.reduce(function (res, group, index) {
     [].concat(group).forEach(function (groupItem) {
-      if (types.indexOf(groupItem) === -1) {
-        throw new Error(`Incorrect configuration of the rule: Unknown type \`${JSON.stringify(groupItem)}\``);
-      }
-      if (res[groupItem] !== undefined) {
-        throw new Error(`Incorrect configuration of the rule: \`${groupItem}\` is duplicated`);
-      }
       res[groupItem] = index * 2;
     });
     return res;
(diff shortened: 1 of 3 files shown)


ljharb:
This PR seems to have too much stuff in it - but either way, if a rule or option is already released, its schema can’t be tightened further.

Xunnamius:
It seems like it has a lot because it's on top of the `consolidateIslands` PR (rather than main so the schema changes can run against the latest tests), but the two new commits are tiny and only touch the schema and two tests. I can pull the lines around `dependencies/consolidateIslands` one commit up into the `consolidateIslands` PR if desired :) 

As for the tightening of existing options, I assume you're talking about lines 865 and 866 for `pathGroupsExcludedImportTypes`? I can revert those. I did add a comment to the `groups` schema warning that additional validation happens elsewhere, and added two newlines to the `named` schema for consistency, but those changes should be just cosmetic.

ljharb:
In that case, I'll wait to review this since it needs to wait on #3129

src/rules/order.js
```diff
@@ -858,9 +858,12 @@ module.exports = {
         properties: {
           groups: {
             type: 'array',
+            // Verified manually in the convertGroupsToRanks function
```
ljharb, line 861: no need for the comment. but this can be verified to have numbers in it, and no duplicates, yes?

src/rules/order.js
```diff
@@ -918,27 +921,28 @@ module.exports = {
...
-                types: {
-                  type: 'string',
-                  enum: [
-                    'mixed',
-                    'types-first',
-                    'types-last',
-                  ],
+            oneOf: [
+              { type: 'boolean' },
+              {
```
ljharb, line 926: it'd be easier to see the diff without the formatting change :-)

ljharb (commented):
these changes are non-breaking _iff_ they exclude things that the rule already checks.

… which would also imply there's some code that can be removed, since the checks are not guaranteed to be done by eslint?

src/rules/order.js
Xunnamius: The [secondary validation](https://github.com/import-js/eslint-plugin-import/blob/a8ab6461544c3d908df3dafaccff823a29e683fe/src/rules/order.js#L599-L604) has its own error messages. I believe you're saying to move the duplicates/existence validation logic up to the schema? That would change the error messages and perhaps other behavior. For instance, there are currently [unit tests](https://github.com/import-js/eslint-plugin-import/blob/4f145a2c64af4931f4bf3ae951c8b719b544718f/tests/src/rules/order.js#L1646-L1702) that check for the current error messages.

Would that count as tightening the schema of a released option?

ljharb: Changed error messages are fine; test cases that always error, changing to "removed test cases because it's now unreachable" is also fine, I think.

Xunnamius: Gotcha, I'll take a look

ljharb:
I just rebased this, and took a look - can you remind me why removing tests isn't a breaking change?

Xunnamius:
IIRC, those tests were removed because the validation code they were testing was removed (at your request), because we're relying on eslint and json schema for validation previously provided by that code.

The error messages will change to eslint's.

ljharb:
Perfect, thank you :-)

ljharb (approved):

PR #3167, opened 2025-03-18
[Refactor] `order`: remove unnecessary negative check

JounQin:
`distinctGroup` is always `false` after `||`.

### DIFF
--- a/src/rules/order.js
+++ b/src/rules/order.js
@@ -770,7 +770,7 @@ function makeNewlinesBetweenReport(context, imported, newlinesBetweenImports_, n
 
       if (shouldAssertNewlineBetweenGroups) {
         if (currentImport.rank !== previousImport.rank && emptyLinesBetween === 0) {
-          if (distinctGroup || !distinctGroup && isStartOfDistinctGroup) {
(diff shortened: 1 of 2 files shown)


src/rules/order.js
```diff
@@ -770,7 +770,7 @@ function makeNewlinesBetweenReport(context, imported, newlinesBetweenImports_, n
 
       if (shouldAssertNewlineBetweenGroups) {
         if (currentImport.rank !== previousImport.rank && emptyLinesBetween === 0) {
-          if (distinctGroup || !distinctGroup && isStartOfDistinctGroup) {
+          if (distinctGroup && isStartOfDistinctGroup) {
```
ljharb, line 773: This would cause breakage unless it's:
```suggestion
          if (distinctGroup || isStartOfDistinctGroup) {
```

JounQin, line 773:
```suggestion
          if (distinctGroup || isStartOfDistinctGroup) {
```

JounQin: Hah, you're right.

ljharb (approved):
thanks!

PR #3194, opened 2025-06-25
fix(no-duplicate): fix no-duplicate false positive on side-effect + type imports

### LINKED ISSUE #3130: [no-duplicates] prefer-inline breaks unassigned and top-level type imports
```
"import/no-duplicates": ["error", { "prefer-inline": true }],
```

Will detect an error in the following typescript code:

```ts
import type { X } from "xyz";

import "xyz";

// [...]
```

It will auto-fix by removing the `import "xyz";` statement. This is not the desired behavior. I would expect it to stay as is.

todor-a:
Fixes #3130, #3195.

### DIFF
--- a/src/rules/no-duplicates.js
+++ b/src/rules/no-duplicates.js
@@ -259,33 +259,41 @@ function getFix(first, rest, sourceCode, context) {
   };
 }
 
-function shouldSkipDuplicateCheckForInlineTypes(nodes) {
+function shouldSkipDuplicateCheckForInlineTypes(nodes, preferInline = false) {
   const importTypes = {
     hasType: false,
     hasSideEffect: false,
     hasDefault: false,
-    hasTypeSpecifier: false,
+    hasNamespaceTypeImport: false,
+    hasTypeImportSpecifier: false,
     hasOther: false,
   };
 
   for (const node of nodes) {
     if (node.importKind === 'type') {
       importTypes.hasType = true;
+      if (node.specifiers.length === 1 && node.specifiers[0].type === 'ImportNamespaceSpecifier') {
+        importTypes.hasNamespaceTypeImport = true;
+      }
     } else if (node.specifiers.length === 0) {
       importTypes.hasSideEffect = true;
     } else if (node.specifiers.length === 1 && node.specifiers[0].type === 'ImportDefaultSpecifier') {
       importTypes.hasDefault = true;
     } else if (node.specifiers.some((spec) => spec.importKind === 'type')) {
-      importTypes.hasTypeSpecifier = true;
+      importTypes.hasTypeImportSpecifier = true;
     } else {
       importTypes.hasOther = true;
       break;
     }
   }
 
+  if (!preferInline && importTypes.hasNamespaceTypeImport && importTypes.hasTypeImportSpecifier) {
+    return true;
+  }
+
   return !importTypes.hasOther
     && importTypes.hasType
-    && !importTypes.hasTypeSpecifier
+    && !importTypes.hasTypeImportSpecifier
     && (importTypes.hasSideEffect || importTypes.hasDefault);
 }
 
@@ -295,7 +303,7 @@ function checkImports(imported, context) {
 
   for (const [module, nodes] of imported.entries()) {
     if (nodes.length > 1) {
-      if (preferInline && shouldSkipDuplicateCheckForInlineTypes(nodes)) {
+      if (shouldSkipDuplicateCheckForInlineTypes(nodes, preferInline)) {
         continue;
       }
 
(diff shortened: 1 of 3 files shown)


src/rules/no-duplicates.js
```diff
@@ -261,8 +261,20 @@ function getFix(first, rest, sourceCode, context) {
...
     if (nodes.length > 1) {
+      if (preferInline) {
+        const typeImports = nodes.filter((node) => node.importKind === 'type');
+        const sideEffectImports = nodes.filter((node) => node.specifiers.length === 0);
+        const valueImports = nodes.filter((node) => !typeImports.includes(node) && !sideEffectImports.includes(node));
+
+        if (typeImports.length > 0 && sideEffectImports.length > 0 && valueImports.length === 0) {
+          continue;
+        }
+      }
```
ljharb, line 276: the way this is written, it is literally always true, because nodes.length is > 1, and everything in it is in one of the three groups (since the "value" condition is an "else"). is that intentional?

tests/src/rules/no-duplicates.js
```diff
@@ -552,6 +552,14 @@ context('TypeScript', function () {
...
           ...parserConfig,
         }),
+        test({
+          code: `
+            import type { A } from 'a';
+            import 'a';
+          `,
+          options: [{ 'prefer-inline': true }],
+          ...parserConfig,
+        }),
```
ljharb, line 562: let's also add a similar invalid test, with output?

ljharb (changes requested):

tests/src/rules/no-duplicates.js
todor-a: do you mean this:
```
import { A } from 'a';
import 'a';
```

?

todor-a: ah, got it, will add it

src/rules/no-duplicates.js
todor-a: > (since the "value" condition is an "else")

not sure i got this. my thinking was - if there are duplicates and one of them is a type import, while the other is a side effect, do not err

tests/src/rules/no-duplicates.js
todor-a: This fails but I think for the wrong reasons and is a different issue. It currently suggests:
```
import B, type { A } from 'a';
```

which is syntactically wrong

todor-a: (sorry this is moving so slow)

i pushed a fix that allows for this condition, let me know what you think

src/rules/no-duplicates.js
todor-a: nice, done

tests/src/rules/no-duplicates.js
todor-a: pushed a fix for both issues in #3195 

had to disable one of the tests if the parser is babel because apparently it does not support `import type * as`

https://astexplorer.net/#/gist/a676569326155f55fa15da7fd791ffda/latest

src/rules/no-duplicates.js
todor-a: bump

todor-a: Rebased onto main and rewrote this as the plain loop you suggested — two booleans, early return, no deps, no extra arrays. It also no longer needs the `prefer-inline` option threaded into `checkImports`.

tests/src/rules/no-duplicates.js
todor-a: #3195 was fixed in the meantime by #3253, so the namespace/default handling I had here is gone and one of my tests is already on main.

That also settles this thread: the remaining `import type { A } from 'a'; import B from 'a'` case now takes the same route #3253 chose for its sibling cases (not reported), rather than being an invalid test with output. Happy to switch it to a real autofix (`import B, { type A } from 'a'`) instead if you would rather have that — say the word and I will do it here.

todor-a:
Rebased onto main and squashed to a single commit — sorry for the long gap.

What changed since the last review:

- #3195 landed via #3253, so the namespace/default-type handling I had here is dropped as dead code, and one of my tests (`import type * as A from 'a'; import { type B } from 'a'`) is already covered on main.
- The helper is now the plain loop @ljharb suggested: two booleans, early return, no deps, no extra arrays, and `prefer-inline` no longer needs threading into `checkImports`.
- Added the CHANGELOG entry I had missed.

What is left is only #3130, which still reproduces on current main with `prefer-inline`:

```ts
import type { A } from 'a';
import 'a';
// reported, and the autofix deletes the side-effect import entirely
```

```ts
import type { A } from 'a';
import B from 'a';
// reported, and the autofix emits `import B, type { A } from 'a'` — invalid syntax
```

Net diff is now 26 lines of `src`, 11 lines of tests, plus the changelog. Reverting just the `src` change fails the two new tests on both parsers.

PR #3196, opened 2025-06-27
feat(order): add orderBySplitPaths option

tw3:


docs/rules/order.md
```diff
@@ -367,8 +367,8 @@ import index from './';
 
 ### `alphabetize`
 
-Valid values: `{ order?: "asc" | "desc" | "ignore", orderImportKind?: "asc" | "desc" | "ignore", caseInsensitive?: boolean }` \
-Default: `{ order: "ignore", orderImportKind: "ignore", caseInsensitive: false }`
+Valid values: `{ order?: "asc" | "desc" | "ignore", orderImportKind?: "asc" | "desc" | "ignore", caseInsensitive?: boolean, orderBySplitPaths?: boolean }` \
+Default: `{ order: "ignore", orderImportKind: "ignore", caseInsensitive: false, orderBySplitPaths: true }`
```
ljharb, line 371: booleans should never default to true; can we invert the naming here so it can default to false?

ljharb (changes requested):
Looks like there's some uncovered lines, so we'll need some more test cases.

docs/rules/order.md
```diff
@@ -385,6 +385,8 @@ Valid properties and their values include:
 
  - **`caseInsensitive`**: use `true` to ignore case and `false` to consider case when sorting
 
+ - **`orderByFullPathString`**: use `false` to split by paths and sort by each one and `true` to sort by the full un-split path string
```
ljharb, line 388: is there a better way to phrase this?

like, perhaps one way sorts the way certain operating systems and IDEs do, and the other sorts the way other ones do? (with an example list of each)

docs/rules/order.md
```diff
@@ -385,6 +385,9 @@ Valid properties and their values include:
 
  - **`caseInsensitive`**: use `true` to ignore case and `false` to consider case when sorting
 
+ - **`orderByFullPathString`**: use `true` to sort by the full un-split path string and `false` to split by paths and sort by each one
+   - enabling this flag may better align with the sort algorithm used by certain operating systems and IDE's, e.g. the Organize Imports action in JetBrains IDE's
```
ljharb, line 389: sorry to ask for more work, but i'd love some research on lists for *both* of the options :-)

tw3: so far I'm only aware of the impact to Organize Imports in WebStorm

not sure how i would find other impacted examples

tw3: Here's a video demo of the behavior in WebStorm IDE:

https://github.com/user-attachments/assets/a3216e74-554d-49b8-b390-f1d6d0ba7501

Ideally Jetbrains would make this behavior configurable from their end, but unfortunately that isn't possible.

So without a change in this plugin my team will either need to:
* Stay on version 2.27, or
* Always sort according to Eslint and no longer use the action in WebStorm, or
* Choose a different/simpler import plugin that "vibes" with WebStorm

tw3: This is what I see on Windows

image

Or were you looking for something different?

robmisasi:
Hello @ljharb and @tw3! It looks like this PR hasn't been worked on for some time and I wanted to address the outstanding issues. I forked the repository and copied the changes from this PR into a branch [here](https://github.com/robmisasi/eslint-plugin-import/tree/issue/2722).

It looks like there was a problem in the added unit tests where the new `orderByFullPathString` was incorrectly specified preventing the new tests from running. The latest commit on the linked branch has changes moving `orderByFullPathString` into the test's `alphabetize` options in alignment with the **import/order** spec.

ljharb:
Thanks! I'll take a look and pull in the changes.

PR #3198, opened 2025-07-09
feat(no-extraneous-dependencies): add exclude option to rule

sf0rman:
Encountered an issue when using `redux-ui` and importing the nested dependencies `@redux-ui/react-<component>` and wanted a way to exclude anything like `@redux-ui/react-*` without having to do it per line or per file which would have other side effects.

Implemented an `exclude` option to the `no-extraneous-dependencies` rule.

### DIFF
--- a/src/rules/no-extraneous-dependencies.js
+++ b/src/rules/no-extraneous-dependencies.js
@@ -177,6 +177,17 @@ function checkDependencyDeclaration(deps, packageName, declarationStatus) {
   }), newDeclarationStatus);
 }
 
+function isInExcludeList(packageName, exclude) {
+  if (!exclude) {
+    return false;
+  }
+
+  if (Array.isArray(exclude)) {
+    return exclude.some((pattern) => minimatch(packageName, pattern));
+  }
+  return minimatch(packageName, exclude);
+}
+
 function reportIfMissing(context, deps, depsOptions, node, name) {
   // Do not report when importing types unless option is enabled
   if (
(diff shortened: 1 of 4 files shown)


src/rules/no-extraneous-dependencies.js
```diff
@@ -177,6 +177,17 @@ function checkDependencyDeclaration(deps, packageName, declarationStatus) {
...
 
+function isInExcludeList(packageName, exclude) {
+  if (!exclude) {
+    return false;
+  }
+
+  if (Array.isArray(exclude)) {
+    return exclude.some((pattern) => minimatch(packageName, pattern));
+  }
+  return minimatch(packageName, exclude);
```
ljharb, line 188:
```suggestion
  return [].concat(exclude || []).some((pattern) => minimatch(packageName, pattern));
```

src/rules/no-extraneous-dependencies.js
```diff
@@ -277,6 +292,7 @@ module.exports = {
           packageDir: { type: ['string', 'array'] },
           includeInternal: { type: ['boolean'] },
           includeTypes: { type: ['boolean'] },
+          exclude: { type: ['string', 'array'] },
```
ljharb, line 295: for the array, let's ensure it's unique, and that the items can only be of type string

ljharb (changes requested):
I'm confused what you mean by "nested dependencies". If you're using `@redux-ui/react-<component>`, then it should be in your package.json, no?

ljharb:
Is this no longer a problem?

PR #3210, opened 2025-09-02
Add a `prefer-top-level-if-only-type-imports` option for the `consistent-type-specifier-style` rule

aldeed:
Adds a `prefer-top-level-if-only-type-imports` option for the `consistent-type-specifier-style` rule. See docs. This is useful when you generally prefer inline but you are using TypeScript `verbatimModuleSyntax` and want all-type imports to be omitted by bundlers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 
 ### Added
 - [`no-deprecated`]: detect `@deprecated` on default-exported identifier declarations ([#3247], thanks [@mixelburg] [@etyrrell22])
+- [`consistent-type-specifier-style`]: add `prefer-top-level-if-only-type-imports` option ([#3210], thanks [@aldeed])
 
 ### Fixed
 - [`no-duplicates`]: fix `prefer-inline` autofix producing invalid syntax when an identifier named `from` is imported ([#3236], thanks [@DukeDeSouth] [@Quantaly])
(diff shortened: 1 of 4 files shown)


graham-fisher-ii:
FWIW, I think you can achieve something similar to this today via the following cocktail of rules:

**`import/no-duplicates` + `'prefer-inline': true`**
Forbids separate `import` and `import type` statements, essentially enforcing inline type imports unless you are _only_ importing types from a module. Autofixable. Implements the "prefer inline type imports unless only type imports" part of the `prefer-top-level-if-only-type-imports` option.

**`@typescript-eslint/no-import-type-side-effect`**
Forbids `import` statements that exclusively import `type` bindings - enforces using an `import type` statement instead. Autofixable. Implements the "prefer top-level imports if only type imports" part of the `prefer-top-level-if-only-type-imports` option.

With that said, it may still be useful to have this combined behavior available under a single rule.

P.S. for strictness in usage of `type` imports, I also like to throw in
**`@typescript-eslint/consistent-type-imports` + `fixStyle: 'inline-type-imports'`** 
Enforces that bindings used as types only are imported as types. Allows both top-level (`import type`) and inline (`import { type }`) styles. However, the rules above will force it into one or the other based on whether the module is used only for type imports. Autofixable. `fixStyle: 'inline-type-imports'` means that IDE auto-importing a type (for me the binding is imported as a value) + autofixing = inline type import.

ljharb:
@aldeed unfortunately, since you made this PR from a fork you don't own (never a good idea), I can't force push to this branch and thus can't merge the PR. would you please add me as a contributor to the fork, so I can finish landing this?

aldeed:
@ljharb Added you. Let me know if that works now.

ljharb:
Amazing, thank you - once tests pass this should be good to go, and once this is merged you are more than welcome to remove me from the fork.

ljharb (approved):

Issue #2289, opened 2021-11-04
False "Multiple exports of name" error when two "export * as ..." contain same name [import/export] 


Stijnkool:
The issue I am having seems related to https://github.com/import-js/eslint-plugin-import/issues/1834
That specific issue is fixed. But I am having the same issue when using an extra level of indirection.
TypeScript does not seem to have a problem with the added level. So I guess this plugin should allow it...?

## Scenario described in https://github.com/import-js/eslint-plugin-import/issues/1834
./A/A1.ts
```ts
export type X = "A1";
```
./A/A2.ts
```ts
export type X = "A2";
```
./A/index.ts
```ts
export * as A1 from "./A1";
export * as A2 from "./A2";
```
In this scenario, as described in https://github.com/import-js/eslint-plugin-import/issues/1834 there would be an error `Multiple exports of name 'X'`.
There is no error. This is fixed.

## My scenario
I have the same files as above. With the following added:
./B/B.ts
```ts
export type X = "B";
```
./B/index.ts
```ts
export * as B from "./B";
```
./index.ts
```ts
export * as A from "./A";
```

ljharb:
Yes, I agree this should work.

darrenvong:
Not sure how much this helps, but this issue was fixed for me from `2.25.2` up to `2.28.2` using eslint 8. `2.29.0` onwards is where I'm seeing the same issue again

ljharb:
Fixed by #3250, now on `main` (ships in the next release). The false `import/export` "Multiple exports of name" error for two `export * as` re-exports sharing a name no longer reproduces with that change.

Issue #2002, opened 2021-03-11
[export] incorrect report on `export * as`


JounQin:
```js
// processors/remark.js
export const remark = {}

// processors/index.js
export * as processors from './remark'

// rules/remark.js
export const remark = {}

// rules/index.js
import { remark } from './remark'

export { remark }
export const rules = { remark }

// index.js
export * from './processors'
export * from './rules'
```

Unexpected error:

```log
Multiple exports of name 'remark'
```

But `processors` doesn't export `remark` at all.

---

Workaround:

```js
// processors/index.js
import { remark } from './remark'

export const processors = { remark }
```

jsg2021:
From your snippet we can clearly see an export const remark, and a little below an `export { remark }`... that's what the error is catching.

ljharb:
Indeed, this looks like a correct error. Thanks @jsg2021 for identifying that!

jsg2021:
@ljharb sorry for the noise. i deleted my comment after realizing the snippet is of several files not one.

JounQin:
😂 Please reopen it.

jsg2021:
@JounQin if you could make a sample repo with the problem it would help or a failing test in a PR.

JounQin:
@jsg2021 Here you are: https://github.com/JounQin/test/tree/eslint-plugin-import

Simply run: `yarn lint`

devfservant:
Having the same “Multiple exports” false positive issue with a slightly different situation:  
(where `remark` is not even exported from *rules/index.js*)

```
root
│   index.ts
│
└───bar
│   │   bar.utils.ts
│   │   index.ts
│
└───foo
    │   foo.utils.ts
    │   index.ts
```


```ts
// bar/bar.utils.ts
export const getOptions = (): any => {};
```

```ts
// bar/index.ts
export * as barUtils from './bar.utils';
```

```ts
// foo/foo.utils.ts
export const getOptions = (): any => {};
```

```ts
// foo/index.ts
export * as fooUtils from './foo.utils';
```

```ts
// index.ts
export * from './bar';
export * from './foo';
```

```
/../root/index.ts
  1:1  error  Multiple exports of name 'getOptions'  import/export
  2:1  error  Multiple exports of name 'getOptions'  import/export
✖ 2 problems (2 errors, 0 warnings)
```

kristiannotari:
Having the same issue as @devfservant stated. The conflicting names are behind a named export, hence they should not be conflicting at all

Issue #1845, opened 2020-07-02
import/named and import/namespace errors after upgrading Babel


eliw00d:
I was on @babel/core 7.8.x and upgraded to 7.10.x and suddenly started seeing errors for import/named and import/namespace.

I have something similar to:

./alpha/abc.js
```jsx
export const A = 'A'
export const B = 'B'
export const C = 'C'
```

./alpha/def.js
```jsx
export const D = 'D'
export const E = 'E'
export const F = 'F'
```

./alpha/index.js
```jsx
export * as './abc
export * as './def
```

And then in files that need it:
```jsx
import * as alpha from './path/to/alpha
```

I have `babel-eslint` and `@babel/plugin-proposal-logical-assignment-operators` installed. Any thoughts?

ljharb:
What exact errors do you see?

eliw00d:
```
'A' not found in imported namespace 'abc'      import/namespace
'B' not found in imported namespace 'abc'      import/namespace
'C' not found in imported namespace 'abc'      import/namespace
'D' not found in imported namespace 'def'      import/namespace
'E not found in imported namespace 'def'      import/namespace
'F' not found in imported namespace 'def'      import/namespace
```

ljharb:
Thanks. Is there any chance you could make a PR with a failing test case? That's the fastest way I can make a fix and release it :-)

eliw00d:
I was able to get a minimum repro of the issue [here](https://github.com/eliw00d/minimum-repro). I hope that helps!

eliw00d:
Did the minimum repro help or do you still need a PR?

ljharb:
Sorry for the delay; the repro repo is *perfect* - now I just need to figure out how to fix it :-D

ljharb:
@eliw00d however, what i can't repro is everything *working* with babel/core 7.8. Any chance you can reproduce everything _passing_ with the same code (but different dep versions)?

VladimirPal:
Hi! Same issue :(
```
"@babel/core": "7.12.10",
"@babel/eslint-parser": "7.12.1",
"@babel/plugin-proposal-export-namespace-from": "7.12.1",

"eslint": "7.16.0",
"eslint-plugin-import": "2.22.1",
```
In the file where import ns - import/named: ns not found in module
```
export * as ns from 'mod';
```

ljharb:
@VladimirPal i can reproduce the failure, but i can't reproduce it ever working in the first place.

[...]

PR #3230, opened 2026-02-19
ESLint v10 support

### LINKED ISSUE #640: Upgrade to ESLint v10
# Summary

With our TypeScript rewrite in #579, ESLint was upgraded from v6 to v9.

Since then, v10 has been released as stable:
https://eslint.org/blog/2026/02/eslint-v10.0.0-released/

We should migrate to v10 as soon as practical:
https://eslint.org/docs/latest/use/migrate-to-10.0.0

This will only be possible once each of our ESLint plugin dependencies have been updated to support v10 as well.

# Plugin Upgrade Tracker

- [ ] **eslint-plugin-import**
  - https://github.com/import-js/eslint-plugin-import/issues/3227
  - https://github.com/import-js/eslint-plugin-import/pull/3230

- [x] **eslint-plugin-jest**
  - https://github.com/jest-community/eslint-plugin-jest/pull/1929

- [ ] **eslint-plugin-jsx-a11y**
  - https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/1075
  - https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/1081
    _(unofficial)_

- [x] **eslint-plugin-prettier**
  - https://github.com/prettier/eslint-plugin-prettier/pull/779

- [ ] **eslint-plugin-react**
  - https://github.com/jsx-eslint/eslint-plugin-react/issues/3977
  - https://github.com/jsx-eslint/eslint-plugin-react/pull/3979

- [x] **eslint-plugin-react-hooks**
  - https://github.com/facebook/react/pull/35720

- [x] **eslint-plugin-testing-library**
  - No ticket or PR as it apparently Just Works <sup>TM</sup>

- [x] **eslint-plugin-unicorn**
  - https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2823

- [x] **typescript-eslint**
  - https://github.com/typescript-eslint/typescript-eslint/issues/11762
  - https://github.com/typescript-eslint/typescript-eslint/pull/11914

### LINKED ISSUE #3079: Following the ESLint 9 support, `import/no-unused-modules` only works in flat config if an `.eslintrc` file exists as well

rasmi:
## Summary

Add ESLint v10 support while maintaining backward compatibility with ESLint v2–v9. The v10 compat changes in shipped code use feature detection (e.g. checking if `context.languageOptions` exists) rather than version checks.

Closes:
* #3227 
* #3079

**Split out into separate PRs** (independent / pre-existing issues revealed during v10 update; this branch rebases onto them and becomes v10-only):

- #3250: for issues #2002, #1883, #1834, #1845, #2289 
- #3251 
- #3252: (maybe)

### Removed API replacements

ESLint v10 removed several deprecated `context` properties and `SourceCode` methods. Each call site now feature-detects the new API and falls back to the old one:

- **`context.parserOptions`** → `context.languageOptions.parserOptions` (affects `sourceType.js`, `childContext.js`, `typescript.js`, `scc.js`, `parse.js`)
- **`context.parserPath`** → `context.languageOptions` used for cache hashing instead (`scc.js`)
- **`context.getPhysicalFilename()`** → `context.physicalFilename` property (`contextCompat.js`)
- **`context.getScope()`** → already had a compat wrapper, but one call site in `declaredScope.js` bypassed it — now fixed
- **`sourceCode.getTokenOrCommentAfter/Before()`** → `sourceCode.getTokenAfter/Before(node, { includeComments: true })` (`order.js`)

### `no-unused-modules` file enumeration fallback

ESLint v10 removed `FileEnumerator` and the internal `glob-utils` module. The existing cascade (`FileEnumerator` → legacy `glob-utils`) now has a third tier: a Node.js `fs` + `minimatch` fallback that walks directories and matches globs.

- Only reached when both prior methods throw `MODULE_NOT_FOUND` or `ERR_PACKAGE_PATH_NOT_EXPORTED`
- Uses `minimatch` and `is-glob` (existing direct dependencies)
- On v10, this resolves [#3079](https://github.com/import-js/eslint-plugin-import/issues/3079) (flat config requiring a dummy `.eslintrc`) since `FileEnumerator` is no longer used. The issue remains on v9.

### Test infrastructure (test files only)

On v10, the unmaintained `babel-eslint` (Babel 6) is replaced with `@babel/eslint-parser` v8 for test coverage. This restores ~475 babel-related tests that would otherwise be skipped on v10.

- `FlatCompatRuleTester` extended to auto-inject `@babel/eslint-parser` config (`requireConfigFile: false`, `babelrc: false`, syntax plugins) and handle v10 RuleTester strictness
- 5 `no-duplicates` tests skipped: duplicate import identifiers correctly rejected by `@babel/eslint-parser`
- TypeScript syntax tests gated for `@typescript-eslint/parser` v8 (`import type Default, { Named }` removed in TS 5.0)
- Angular template parser upgraded to v21 on v10 (v13 does not support ESLint 10; [angular-eslint#2903](https://github.com/angular-eslint/angular-eslint/pull/2903))
- #1645 runs on v10 via a new fixture in #3252

### Dependency changes

| Package | Change | Why |
|---|---|---|
| `eslint` | `\|\| ^10` added to peerDeps and devDeps | v10 support |

v10-specific parser deps are installed via `dep-time-travel.sh` (not in devDeps) to avoid peer dep conflicts with `typescript@4.5.5` during `npm install`:

- `@typescript-eslint/parser@8` — on all ESLint 10 Node versions
-`@angular-eslint/template-parser@21` — on all ESLint 10 Node versions (v13 does not support ESLint 10)
- `@babel/eslint-parser@8` + `@babel/core@8` (ESM, RC) — on Node >= 20
- No babel parser on Node 18/19/21 (v8 is ESM-only, engines `^20.19.0 || >=22.12.0`; [v7 will not get ESLint 10 support](https://github.com/babel/babel/pull/17796)). Babel tests are skipped; all other tests run.

### CI changes

- `eslint-8+.yml`: added ESLint 10 to the matrix, excluded Node < 18
- `dep-time-travel.sh`: installs v10-specific parser deps (`@typescript-eslint/parser@8` and `@angular-eslint/template-parser@21` on all nodes, `@babel/eslint-parser@8` on Node >= 20 only)
- See also #3237 to fix tests failing in main.


StephanTLavavej:
@rasmi Do you want to disclose your usage of LLMs to generate this PR description and/or PR, which @ljharb objected to in another repo https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/1077 ?

rasmi:
Was not aware, thanks @StephanTLavavej! I will delete this PR.

rasmi:
@ljharb -- point well-taken! I did go through a more rigorous process here -- it really did take me hours of focused work, thinking about implementation approaches, iterating, fixing tests, etc. It is not my intention to submit slop, I wanted to submit a PR that I hoped would meet the standard of the library, at least as a first pass.  I lack decision-making insight/context on things like "Is it okay to just use glob, if not let's do a custom file search" and "exactly how should node vs. eslint-parser vs. core library version compatibility be handled" -- but this is what the review process is for! 

I also understand that big changes like this may be simpler for the maintainer to just take on directly, as you have all the proper context and expertise on all the nuances of the library and ecosystem, to say the least. Apologies again, I didn't mean to burden you at all, I know it is hard to maintain core libraries like this. Please do with this what you wish!

tests/src/cli.js
```diff
@@ -64,7 +64,8 @@ describe('CLI regression tests', function () {
     let eslint;
     let cli;
     beforeEach(function () {
-      if (semver.satisfies(eslintPkg.version, '< 6')) {
+      if (semver.satisfies(eslintPkg.version, '< 6') || semver.satisfies(eslintPkg.version, '>= 10')) {
+        // TODO: re-enable when eslint-plugin-json supports v10 (https://github.com/azeemba/eslint-plugin-json/issues/97)
```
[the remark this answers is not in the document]
fernandopasik: good idea, last commit in https://github.com/azeemba/eslint-plugin-json was 17 months ago

src/exportMap/visitor.js
```diff
@@ -57,9 +57,12 @@ export default class ImportExportVisitorBuilder {
       },
       ExportAllDeclaration() {
         const getter = captureDependency(astNode, astNode.exportKind === 'type', this.remotePathResolver, this.exportMap, this.context, this.thunkFor);
-        if (getter) { this.exportMap.dependencies.add(getter); }
         if (astNode.exported) {
+          // `export * as ns from './mod'` — named namespace, not a star-export
           processSpecifier(astNode, astNode.exported, this.exportMap, this.namespace);
+        } else if (getter) {
```
ljharb, line 63: Previously, `export * as ns from './mod'` would both add the dependency *and* process the specifier. Now it only processes the specifier, skipping `dependencies.add(getter)`. This means `export * as ns from './mod'` won't be tracked as a dependency of the current module, which could affect cycle detection or re-export tracking.

Is this intentional? If so, does the lazy getter in `specifier.js` fully compensate?

tests/src/rule-tester.js
```diff
@@ -3,24 +3,96 @@ import { version as eslintVersion } from 'eslint/package.json';
...
+      return;
+    }
+
+    // Pass constructor parser so per-test configs can inherit babel options
+    const ctorParser = this._constructorConfig && this._constructorConfig.parser;
+
     this._tester.run(ruleName, rule, {
-      valid: tests.valid.map((t) => FlatCompatRuleTester._flatCompat(t)),
-      invalid: tests.invalid.map((t) => FlatCompatRuleTester._flatCompat(t)),
+      valid: tests.valid
```
ljharb, line 40: Valid test cases shouldn't have `errors` or `output` to begin with. If they do, it likely indicates a test bug that should be surfaced, not silently papered over - let's fix them instead.

src/exportMap/typescript.js
```diff
@@ -7,7 +7,10 @@ const tsconfigCache = new Map();
 
 function readTsConfig(context) {
   const tsconfigInfo = tsConfigLoader({
-    cwd: context.parserOptions && context.parserOptions.tsconfigRootDir || process.cwd(),
+    cwd: (context.parserOptions
+      ? context.parserOptions.tsconfigRootDir
+      : context.languageOptions && context.languageOptions.parserOptions && context.languageOptions.parserOptions.tsconfigRootDir
+    ) || process.cwd(),
```
ljharb, line 13: Subtle behavior change: if `context.parserOptions` is truthy but doesn't have `tsconfigRootDir`, the old code (`context.parserOptions && context.parserOptions.tsconfigRootDir || process.cwd()`) would evaluate to falsy and fall through to `process.cwd()`. The new code takes the truthy `context.parserOptions` branch and returns `undefined` for `tsconfigRootDir`, skipping `languageOptions.parserOptions` entirely.

This is probably fine since both shouldn't be set simultaneously, but it's worth noting as a semantic change from the original.

src/rules/no-unused-modules.js
```diff
@@ -147,6 +148,86 @@ function listFilesWithLegacyFunctions(src, extensions) {
...
+  extensions = extensions.map((ext) => ext.startsWith('.') ? ext : `.${ext}`);
+  const results = [];
+
+  src.forEach((pattern) => {
+    if (isGlob(pattern)) {
+      // For glob patterns, walk the base directory and filter with minimatch
+      // Extract the base directory from the glob (everything before the first glob character)
+      const base = pattern.replace(/[*?{[].*/g, '').replace(/\/[^/]*$/, '') || '.';
+      const resolvedBase = resolve(base);
+      const allFiles = walkDirectory(resolvedBase, extensions, [], fs, join, extname);
```
ljharb, line 207: The glob base extraction `pattern.replace(/[*?{[].*/g, '').replace(/\/[^/]*$/, '') || '.'` could be slow for patterns like `*.js` or `{a,b}/*.js` - it falls back to `.` which walks the entire project tree.

Also, `minimatch` doesn't handle all glob edge cases (negation, brace expansion) the same way ESLint's `FileEnumerator` did. Has this been tested with real-world `src` patterns?

tests/dep-time-travel.sh
```diff
@@ -50,3 +50,18 @@ if [ "${ESLINT_VERSION}" = '8' ]; then
...
+if [[ "$ESLINT_VERSION" -ge "10" ]]; then
+  echo "Installing @typescript-eslint/parser v8 for ESLint 10..."
+  npm i --no-save @typescript-eslint/parser@8
+
+  echo "Installing @angular-eslint/template-parser v21.3.0+ for ESLint 10..."
+  npm i --no-save '@angular-eslint/template-parser@^21.3.0'
+
+  # @babel/eslint-parser v8 requires Node ^20.19.0 || >=22.12.0; v7 doesn't support ESLint 10
+  if [[ "$TRAVIS_NODE_VERSION" -eq "20" ]] || [[ "$TRAVIS_NODE_VERSION" -ge "22" ]]; then
+    echo "Installing @babel/eslint-parser v8 for ESLint 10..."
```
ljharb, line 72: The `^8.0.0-rc.2` range won't match `@babel/core` 8.0.0 stable when it releases - semver pre-release ranges don't satisfy stable versions. This will silently stop installing the package once 8.0.0 drops.

Let's use `^8.0.0-0`.

tests/src/utils.js
```diff
@@ -9,8 +9,36 @@ import 'babel-eslint';
...
+      }
+      return require.resolve('@typescript-eslint/parser');
+    } catch (e) {
+      return false;
+    }
+  })(),
+  // babel-eslint is unmaintained; on ESLint v10+ use @babel/eslint-parser instead
+  BABEL_OLD: (() => {
+    if (semver.major(eslintPkg.version) < 10) { return require.resolve('babel-eslint'); }
+    try { return require.resolve('@babel/eslint-parser'); } catch (e) { return false; }
```
ljharb, line 27: On v10, `BABEL_OLD` silently becomes `@babel/eslint-parser`, which is the same value as `BABEL_NEW`. This makes the name misleading - downstream tests like `no-duplicates.js` already have to check `parser === parsers.BABEL_NEW` to work around behavioral differences.

I think it would be clearer to have `BABEL_OLD` return `false` on v10 (like unavailable parsers), and let tests that need it use `BABEL_NEW` explicitly.

tests/dep-time-travel.sh
mgol: Wait, why would it not match? And how would `-0` be different as the prerelease part?

I tested this on the npm semver calculator at https://semver.npmjs.com/ and for `jquery` the range `^4.0.0-beta.2` matches `4.0.0-beta.2`, `4.0.0-rc.1`, `4.0.0-rc.2` and `4.0.0`.

ljharb: ah ok cool, my bad on that one then :-)

tests/src/cli.js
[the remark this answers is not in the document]
ljharb: I'm fine with that switch, in its own commit.

ljharb:
@rasmi i've got some feedback waiting for you :-)

ma-st-dt:
Are there any updates on this? :) 
Merging this would be highly appreciated ❤️

TheGiraffe3:
It looks like it's waiting on @rasmi  to resolve maintainer feedback.

manzoorwanijk:
Can we somehow help speed this up?

lake2:
Eslint-plugin-import is not working because the Vscode eslint plugin appears to have upgraded to version 10.

lorand-horvath:
@rasmi Any progress?

li-jia-nan:
Are there any updates ?

tests/src/rule-tester.js
rasmi: Done.

tests/src/utils.js
rasmi: Okay -- I fixed the `BABEL_NEW` vs. `BABEL_OLD` distinction, but in terms of coverage, what is your preference? Should I just add:
```
test({ ..., parser: parsers.BABEL_NEW })
```
After each `BABEL_OLD` test? Or just do this for cases that actually exercise `BABEL_NEW` functionality / are testing regressions? 

Or, allow `rule-tester` to accept a list of multiple parsers for `parser: ...`? Or have a unified `parsers.BABEL` that always runs both? Or perhaps a secret other thing...?

src/rules/no-unused-modules.js
rasmi: Is it okay to just use `glob` here for the sake of reliability, or do you want to avoid the additional dependency/chaos/requirements imposed?

rasmi: Can also use `glob-parent` if preferred?

src/exportMap/visitor.js
rasmi: Cycle detection reads [`m.imports`](https://github.com/rasmi/eslint-plugin-import/blob/eslint-update/src/rules/no-cycle.js#L128) (still populated by [`captureDependency`](https://github.com/rasmi/eslint-plugin-import/blob/eslint-update/src/exportMap/captureDependency.js#L22-L30)), not [`m.dependencies`](https://github.com/rasmi/eslint-plugin-import/blob/eslint-update/src/exportMap/index.js#L7-L16), so the import is still tracked. `m.dependencies` is only populated for `export * from`, not `export * as ns`.  The [lazy getter](https://github.com/rasmi/eslint-plugin-import/blob/eslint-update/src/exportMap/specifier.js#L16-L20) resolves `ns.namespace` to `./mod`'s ExportMap on access [(matching the existing pattern)](https://github.com/rasmi/eslint-plugin-import/blob/eslint-update/src/exportMap/specifier.js#L11-L15).

LMK if I should comment inline or update the "ExportMap export * as fix" section in the PR to clarify/note this!

tests/src/cli.js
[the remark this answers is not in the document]
rasmi: Done in ed9fd9282e5f919f1cf99586f12b2f6fb853bdab!

rasmi:
> @rasmi i've got some feedback waiting for you :-)

Sincere apologies for the delay, can commit to a 1-day SLO on feedback/revisions until this is merged!

tests/src/cli.js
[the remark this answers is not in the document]
rasmi: (~Failing on older versions, will fix this eve~) fixed!

rcantin-w:
Thank you for working on this! Just realised I'd have to lose my clever ordering, but I'll suffer the buggy linting until this gets merged 🤞

Julien-Marcou:
@rasmi I think you need to re-ask @ljharb for a review if your PR is ready

src/exportMap/specifier.js
```diff
@@ -14,7 +14,9 @@ export default function processSpecifier(specifier, astNode, exportMap, namespac
       }));
       return;
     case 'ExportAllDeclaration':
-      exportMap.namespace.set(specifier.exported.name || specifier.exported.value, namespace.add(exportMeta, specifier.source.value));
```
ljharb: this means it's not longer doing `namespace.add` - can you elaborate on this?

src/rules/no-unused-modules.js
ljharb: You can use glob if you use a version that's old enough to support the same node versions as us.

src/rules/no-unused-modules.js
```diff
@@ -147,6 +148,86 @@ function listFilesWithLegacyFunctions(src, extensions) {
...
+}
+
+/**
+ * List files using Node.js fs and minimatch as a fallback when FileEnumerator
+ * and legacy ESLint APIs are unavailable (ESLint v10+).
+ * @param {string[]} src - list of file paths, directories, or glob patterns
+ * @param {string[]} extensions - list of supported file extensions
+ * @returns {string[]} list of matched file paths
+ */
+function listFilesWithNodeFs(src, extensions) {
```
ljharb: why wouldn't this function work on eslint 9?

tests/src/rules/named.js
```diff
@@ -165,12 +165,10 @@ ruleTester.run('named', rule, {
 
     test({
       code: 'const { baz } = require("./bar")',
-      errors: [error('baz', './bar')],
```
ljharb: why are these two no longer errors?

tests/src/rules/no-dynamic-require.js
```diff
@@ -100,15 +100,13 @@ ruleTester.run('no-dynamic-require', rule, {
         }),
         _test({
           code: 'import("../" + name)',
-          errors: [dynamicImportError],
```
ljharb: and why are these no longer errors?

package.json
```diff
@@ -91,6 +91,8 @@
     "eslint-plugin-eslint-plugin": "^2.3.0",
     "eslint-plugin-import": "2.x",
     "eslint-plugin-json": "^2.1.2",
+    "eslint-plugin-jsonc": "^2.21.1",
+    "espree": "^9",
```
ljharb: why is this pin needed?

ljharb (changes requested):
I've rebased this. Can you pull out the pre-existing bug fixes into their own commits (ideally stacked before the eslint 10 changes)? Would also be nice to extract out the `parsers` refactors, including `typescriptEslintParserSatisfies`, to separate commits.

Also, can you elaborate on the eslint-plugin-jsonc stuff? I do see that eslint-plugin-json doesn't work on eslint 10 yet.

src/rules/no-unused-modules.js
rasmi: Fixed the base extraction (and brace expansion / recursion) so it no longer falls back to `.` and walks the whole tree (using the same `minimatch` options as `FileEnumerator`). It's not 1:1 with `FileEnumerator` as it will skip dotfile dirs and symlinks, is this important to address here?

src/rules/no-unused-modules.js
rasmi: It would work, just not reached as a fallback. Updated comment to clarify!

src/exportMap/specifier.js
rasmi: `namespace.add` was a no-op here -- it keys off `identifier.name`, but was passed `specifier.source.value` (the module string), so it never set a getter, just returned the bare `{}`. Updated code sets the lazy getter directly, same as [`ExportNamespaceSpecifier`](https://github.com/rasmi/eslint-plugin-import/blob/eslint-update/src/exportMap/specifier.js#L11-L15), so `export * as ns` resolves.

tests/src/rules/named.js
rasmi: This is the [fix from earlier](https://github.com/import-js/eslint-plugin-import/pull/3230#discussion_r3030886367). These never errored -- `checkRequire` early-returns unless `commonjs: true`, which these don't set. The `errors:` was presumably copied from the adjacent case in 54d86c8a. Previous `RuleTester` just ignored it, it now flags it which is why the test failed / required a workaround in the earlier commit/comment. Will split this out in the bugfix commit!

tests/src/rules/no-dynamic-require.js
rasmi: Same [issue as above](https://github.com/import-js/eslint-plugin-import/pull/3230#discussion_r3255642584)! `no-dynamic-require` only flags `import()` with `esmodule: true`, which these don't set, so they never errored. These were added in 7163824d and the `errors` was ignored by the previous `RuleTester`.

rasmi: (can move these into a fix commit as well)

package.json
rasmi: `espree` is normally hoisted transitively from eslint, but installing `eslint-plugin-jsonc@^3` for v10 via `dep-time-travel.sh` de-hoists it, breaking `require.resolve('espree')` in `utils.js` (see failed test run [here](https://github.com/import-js/eslint-plugin-import/actions/runs/25606777902/job/75169971002)). I used `^9` to avoid v11, which uses `Array.prototype.at()` and dies on node 12/14 ([node 12 failure](https://github.com/import-js/eslint-plugin-import/actions/runs/25620179543/job/75205228912), [node failure](https://github.com/import-js/eslint-plugin-import/actions/runs/25620179543/job/75205228909)). v9 still works for the tests that use `parsers.ESPREE`.

rasmi: These become unnecessary if we use a fixture as in https://github.com/import-js/eslint-plugin-import/pull/3252

src/exportMap/visitor.js
rasmi: Broken out into #3250

rasmi:
@ljharb -- ready!

* Fixed the glob issue (not showing above but comment is [here](https://github.com/import-js/eslint-plugin-import/pull/3230#discussion_r3255620140)). LMK if this is sufficiently robust / non-disruptive of other versions.

* Pulled out the `export * as ns` fix (#3250) and test cleanup (#3251) into separate PRs. This branch will be cleaner after a rebase on top of those.

* Curious for your thoughts [here](https://github.com/import-js/eslint-plugin-import/pull/3230#discussion_r3213346261)

* Will pull the parsers refactor into a separate commit (alongside whatever the right approach is for the previous point/question)

The jsonc stuff is messy, we could use a fixture (see #3252) or skip the `cli.js` test for #1645 for v10+ (not ideal), or do these version-pinning workarounds (or cleaner workarounds if you have any ideas). The context for the jsonc switch (originally proposed [here](https://github.com/import-js/eslint-plugin-import/pull/3230#discussion_r2840783267)) is: v2 works on eslint 6-9, but eslint 10 needs v3 (ESM-only, newer Node). No single version covers every eslint, so v3 is only installed for the `eslint-10` runs in `dep-time-travel.sh`.

I kept `eslint-plugin-json` for the old `.eslintrc.json` path instead of switching everything to jsonc. jsonc v2 pulls in `synckit`, which uses syntax Node 12 can't parse ([failed run](https://github.com/import-js/eslint-plugin-import/actions/runs/25606777902/job/75169971083)). `no-unused-modules` loads that `.eslintrc.json`, so switching it would break unrelated tests on Node 12. The flat-config path (eslint 9+) uses jsonc, and [`cli.js`](https://github.com/rasmi/eslint-plugin-import/blob/eslint-update/tests/src/cli.js#L63-L183) asserts differently depending on which one is active.

If we can use a fixture to reproduce the issue in #1645 instead of `eslint-plugin-json` or `eslint-plugin-jsonc` directly, these workarounds can be avoided entirely (see https://github.com/import-js/eslint-plugin-import/pull/3252).

src/exportMap/specifier.js
rasmi: Moved to #3250

tests/src/rules/no-dynamic-require.js
rasmi: Moved to #3251

tests/src/rules/named.js
rasmi: Moved to #3251

src/core/sourceType.js
```diff
@@ -3,10 +3,10 @@
...
 export default function sourceType(context) {
-  if ('sourceType' in context.parserOptions) {
-    return context.parserOptions.sourceType;
-  }
   if ('languageOptions' in context && context.languageOptions) {
     return context.languageOptions.sourceType;
   }
+  if (context.parserOptions && 'sourceType' in context.parserOptions) {
+    return context.parserOptions.sourceType;
+  }
```
ljharb, line 11: reverting this change doesn't fail any tests, so it either needs to be reverted, or a test added.

src/rules/no-unused-modules.js
ljharb: then should we fix that, so that eslint 9 doesn't require an eslintrc for this functionality?

Julien-Marcou:
Is there a point to try & keep supporting old versions of nodejs?

Eslint v10 has dropped the support for nodejs versions < v20.19

I understand the backward compatibility needs, but I guess it would be nice to release a new major version (at some point in a future PR), so that people using old eslint/nodejs versions can stay on the v2 of eslint-plugin-import, while people keeping up with latest versions of eslint/nodejs could use the v3.

I guess it would make the maintaining of this project easier as eslint & nodejs have come a long way since eslint v2 & nodejs v6 😅

ljharb:
@Julien-Marcou yes, there's a point - the most harmful thing for the ecosystem by a massive margin, for any package, is a major version bump, because it strands users on previous release lines. I do intend to do a breaking change and drop older eslints and nodes at some point, but it's not a decision to make lightly, and it's _absolutely not_ justified by personal maintainer/contributor pain nor by "but old bad and new good!".

ClementValot:
Genuine question though, for the specific case of linters, which are meant to be executed on dev/CI machines, and perform static analysis thus are totally de-correlated from the production runtime: Is there a realistic case where one would be stuck with an older version of node and/or ESLint, and would at the same time be actively maintaining the ESLint plugins?

tests/src/rules/namespace.js
```diff
@@ -343,6 +344,7 @@ const invalid = [].concat(
 // deep dereferences //
 //////////////////////
 [['deep', require.resolve('espree')], ['deep-es7', parsers.BABEL_OLD]].forEach(function ([folder, parser]) { // close over params
```
rasmi: This was previously `parsers.ESPREE`, I think reverted in a commit @ljharb  -- I don't mind one way or the other.

rasmi:
@ljharb -- rebased onto main and broke out the parsers into a separate commit as requested! I left a small comment above but it's inconsequential. One more close look at everything would be appreciated. 

And this [open question on BABEL_NEW test coverage](https://github.com/import-js/eslint-plugin-import/pull/3230#discussion_r3213346261) as well -- we could:
* Add `parsers.BABEL_NEW` to the `export * as` regression cases (#1834 / #1883 / #1845)
* Ensure all possible tests are run against BABEL_NEW (duplicate BABEL_OLD counterparts, or a multi-parser list or `parsers.BABEL` helper that runs both?).
* Something else?

Thanks for your patience and persistence!

tests/src/rules/namespace.js
ljharb: let's restore that in the parsers commit.

ljharb:
@ClementValot it's true that usage may be low, but even eslint 7 still gets [4.3 _million_ downloads a week](https://majors.nullvoxpopuli.com/q?minors=false&old=false&packages=eslint), and a TON of companies are stuck on very old node versions.

ljharb (commented):
sorry, didn't realize this review wasn't submitted from the other day

ClementValot:
> @ClementValot it's true that usage may be low, but even eslint 7 still gets [4.3 _million_ downloads a week](https://majors.nullvoxpopuli.com/q?minors=false&old=false&packages=eslint), and a TON of companies are stuck on very old node versions.

My point was, if you're a company that is stuck with node 6 for your prod runtime (which is, indeed, still a reasonably common occurrence to account for), and your development tooling is also stuck with node 6 (which is already a more questionable choice), is "Aw shoot, I can't upgrade eslint-plugin-import to v3" really a concern? 

I think I remember a comment from yourself about how there's no urgency in upgrading your linter :D

ljharb:
Yes, because while there's no urgency in updating _eslint_, updating your plugins gives you important bugfixes :-) (you could certainly argue that eslint's own rules need bugfixes too, which would be a pretty compelling counterargument to my past statement)

src/core/sourceType.js
rasmi: Added: [tests/src/core/sourceType.js](https://github.com/import-js/eslint-plugin-import/pull/3230/changes#diff-70d25154a2d21900b54b3b73abd0c20237c27a1cb905a6cf549d5f2d0b62edb4)

src/rules/no-unused-modules.js
rasmi: Done!

rasmi:
@ljharb comments addressed!

Any thoughts on:
* BABEL_NEW test coverage approach from the comment above?
* For `no-unused-modules` enabling `listFilesWithNodeFs` as a fallback in v9 (233fe0469f8e2cd92faec2ad5cc50427aa1ff6bf) -- the same caveat applies with the v10 path around not honoring `.eslintignore` / flat `ignores` compared to `FileEnumerator` / earlier versions. However, on v9 you can still do ignore-filtering with `.eslintrc`+`ignorePatterns`, on v10 you MUST use `src` or `ignoreExports`. Is this behavior okay? Should it be better documented, or is it expected behavior at this point?

ljharb:
@rasmi ok, so - the fallback commit should have some tests (and be sorted before the eslint 10 commit); re https://github.com/import-js/eslint-plugin-import/pull/3230#discussion_r3213346261, tests should ideally only specify a parser when it's testing something unique to that parser.

As to your question - maybe i'm not understanding the matrix here. in eslint <= 8, `no-unused-modules` works without any explicit `src` or `ignoreExports` configuration. Let's forget about the weird broken current behavior in v9 - with the fallback commit, what exactly is the behavior in eslint >= 9? If it differs, why is it not possible to make it work the same as in <= 8?

ljharb:
ping @rasmi

rasmi:
Updated to split out `listFilesWithNodeFs`+ tests from the v10 update!

As for behavior on eslint >=9:
* v9 with `.eslintrc.*` keeps using `FileEnumerator`, same as v8.
* For v9 with `eslint.config.js` and no `.eslintrc`, and for v10:
  * `listFilesWithNodeFs` walks cwd, filters by `extensions`, and skips `node_modules` and dot-prefixed entries.
  * It does NOT handle `.eslintignore`, `ignorePatterns`, or `eslint.config.js` `ignores`.
* Setting `src` or `ignoreExports` explicitly in the rule config gives v8-equivalent behavior.
  * (`ignoreExports` is not EXACTLY equivalent as ignored files' imports still [count toward the "used" graph](https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/no-unused-modules.js#L370-L374), but for the common case of excluding `dist/`/`build/` outputs that don't import from `src/`, it works.)

I am not sure what the "right" way to load/respect `eslint.config.js` would be as `no-unused-modules` runs sync, but loading `eslint.config.js` / consuming it via eslint APIs isn't sync without some workaround -- js is not my strongest language and I'm not sure what the best practice is for this type of operation in a library that needs to be robust!

src/core/listFilesWithNodeFs.js
```diff
@@ -0,0 +1,100 @@
...
+        while (baseParts.length < segments.length && typeof segments[baseParts.length] === 'string') {
+          baseParts.push(segments[baseParts.length]);
+        }
+        const base = baseParts.join('/') || '/';
+        const globPart = segments.slice(baseParts.length);
+        // `src/*.js` stays in `src/`; `src/**/*.js` recurses.
+        const recursive = globPart.length > 1 || globPart.indexOf(GLOBSTAR) !== -1;
+        const allFiles = walkDirectory(base, normalizedExts, [], recursive);
+        allFiles.forEach((file) => {
+          if (mm.match(file)) {
```
ljharb: This is the `codecov/patch` failure. The false branch of `if (mm.match(file))` is never exercised (nyc: 90% branch on this file, counts `[5,0]`) — every test glob (`src/*.js`, `**/*.js`) matches everything `walkDirectory` returns, so `mm.match` never returns false. Adding a case where the walk yields a same-extension sibling the pattern rejects — e.g. glob `src/a*.js` with both `src/a.js` and `src/b.js` present, asserting only `a.js` comes back — covers the branch and takes patch coverage to 100%.

src/core/listFilesWithNodeFs.js
```diff
@@ -0,0 +1,100 @@
...
+export default function listFilesWithNodeFs(src, extensions) {
+  const normalizedExts = extensions.map((ext) => ext.startsWith('.') ? ext : `.${ext}`);
+  const results = [];
+
+  src.forEach((pattern) => {
+    if (isGlob(pattern)) {
+      // Expand braces, then take the base from the parsed pattern's leading
+      // literal segments, mirroring how ESLint's FileEnumerator resolves globs.
+      minimatch.braceExpand(pattern).forEach((expanded) => {
+        const mm = new Minimatch(resolve(expanded), minimatchOpts);
```
ljharb: Windows: `resolve(expanded)` (and the `path.join` paths from `walkDirectory`) emit `\`-separated paths, which minimatch v3 treats as escape characters — so glob matching silently breaks on Windows. This only affects the v10 fallback tier (FileEnumerator + the legacy APIs both unavailable), so it's a v10-on-Windows regression for `no-unused-modules` with glob srcs. Suggest normalizing separators to `/` before constructing the matcher and before `mm.match` (e.g. `.split(path.sep).join('/')`). (Verified by code analysis + minimatch semantics; not executed on Windows.)

utils/declaredScope.js
```diff
@@ -6,7 +6,7 @@ const { getScope } = require('./contextCompat');
 
 /** @type {import('./declaredScope').default} */
 exports.default = function declaredScope(context, name, node) {
-  const references = (node ? getScope(context, node) : context.getScope()).references;
+  const references = getScope(context, node).references;
```
ljharb, line 9: This drops the `node ? … : context.getScope()` fallback and now always calls `getScope(context, node)`. `contextCompat.getScope` doesn't guard a falsy `node`, so a 2-arg external caller of the published `eslint-module-utils/declaredScope` on eslint 8/9 flat config now throws `Missing required argument: node` (reproduced on 8.57.1) where it previously worked. All internal callers pass a node, so the plugin itself is unaffected, and `declaredScope.d.ts` still types `node?` as optional. Let's restore the no-node guard.

jacobeturpin:
Appreciate the input from all here -- both contributors and reviewers/maintainers. I understand a significant effort is put into maintaining this in addition to your days jobs.

That being said, I'm passing along that my organization is looking to actively move off this plugin in short order from the lack of resolution in order to meet compliance requirements. I can't imagine we're the only ones given the level of activity between comments and linked PRs in the conversation history.

utils/declaredScope.js
rasmi: Fixed.

src/core/listFilesWithNodeFs.js
rasmi: I think this is already handled properly by minimatch? [Here](https://github.com/isaacs/minimatch/blob/v3.1.5/minimatch.js#L139-L142) for the pattern and [here](https://github.com/isaacs/minimatch/blob/v3.1.5/minimatch.js#L742-L745) for the input.

src/core/listFilesWithNodeFs.js
rasmi: Done!

src/rules/no-unused-modules.js
```diff
@@ -85,25 +86,13 @@ function listFilesUsingFileEnumerator(FileEnumerator, src, extensions) {
...
-Due to the exclusion of certain internal ESLint APIs when using flat config,
-the import/no-unused-modules rule requires an .eslintrc file to know which
-files to ignore (even when using flat config).
-The .eslintrc file only needs to contain "ignorePatterns", or can be empty if
-you do not want to ignore any files.
-
-See https://github.com/import-js/eslint-plugin-import/issues/3079
-for additional context.
-`);
+      return listFilesWithNodeFs(src, extensions);
```
ljharb: This `listFilesWithNodeFs` fallback — plus the second one in the `listFilesToProcess` catch (~line 158) and the `getTokenAfter` v10 shim in `order.js:29/35` — is what `codecov/patch` is now failing on. These new v10/flat-config fallback paths aren't covered by any test leg (codecov shows `no-unused-modules.js` −23, `order.js` −6).

This is **not** an upload artifact like #3258: all coverage workflows for this head succeeded and codecov's report is `state: complete`, so it's a genuine gap. It traces back to the v10 test-leg issues from the earlier review — `@babel/eslint-parser` / `BABEL_NEW` isn't reliably installed on the v10 legs, so the modern-parser fallbacks never execute. To clear patch honestly: add tests that drive these fallbacks under eslint v10 (flat config without `.eslintrc`, and the legacy-APIs-absent path), and make sure the v10 legs actually run + upload coverage.

ljharb (commented):
Only one item left, i think

src/rules/no-unused-modules.js
rasmi: Okay, I extracted the new file helpers to `src/core/listFilesToProcess.js` so that they can be tested properly. I think codecov is otherwise failing to report some branches because of [429s/rate-limiting](https://github.com/import-js/eslint-plugin-import/actions/runs/26915958165/job/79408761745#step:6:32) (needs CODECOV_TOKEN)?

A5rocks: Could the runs be retried? https://app.codecov.io/gh/import-js/eslint-plugin-import/pull/3230 says the latest commit from the PR uploaded 6 coverage reports, whereas the one on main uploaded 18.

(also I see a toplevel `.coveralls.yml` file that I think should be deleted?)

lorand-horvath:
@rasmi Thanks for the work you did, this hurdle seems to be a really difficult one to pass. Hopefully @ljharb manages to proceed now...

captaindonald:
@ljharb I opened [rasmi/eslint-plugin-import#1](https://github.com/rasmi/eslint-plugin-import/pull/1) against @rasmi's `eslint-update` to clear the failing codecov/patch + codevov/project here - no targets, thresholds, or ignores changed, just covers the two new v10 spots that were uncovered: the `getTokenOrComment` feature-detect arms in `order`, and the `ERR_PACKAGE_PATH_NOT_EXPORTED` fallback in `listFilesToProcess`. Merge it into the branch, or pull the commits straight into this PR - whichever is easier.

captaindonald:
Ok, this one should clear that -0.04%.

[https://github.com/rasmi/eslint-plugin-import/pull/2](https://github.com/rasmi/eslint-plugin-import/pull/2)

captaindonald:
Codecov being codecov again. Failed uploads.

ljharb:
Codecov isn't a blocker, don't worry.

ljharb (approved):

src/rules/no-unused-modules.js
ljharb: I can address this in a followup.

manzoorwanijk:
Thank you @ljharb

TheGiraffe3:
And thanks @captaindonald!

ThomasAFink:
Thank you for this!

rcantin-w:
Thanks so much for keeping up the work on this! 
Might be my lack of understanding of how releases work for packages, but do we know when we might expect this to be released (I assume as a patch)?

ljharb:
Adding things is a minor; and there's no concrete dates this project commits to. However, I'm hoping to release it soon.

aprendendofelipe:
For anyone looking to test ESLint v10 compatibility in their projects right now, a preview build is available via [pkg.pr.new](https://pkg.pr.new):

```bash
npm i -D https://pkg.pr.new/aprendendofelipe/eslint-plugin-import@main
```

Or add it directly to your `package.json`:

```json
"devDependencies": {
  "eslint-plugin-import": "https://pkg.pr.new/aprendendofelipe/eslint-plugin-import@main"
}
```

Fork repository: https://github.com/aprendendofelipe/eslint-plugin-import

*Note for maintainers: If interested, I'd be happy to open a PR to configure `pkg.pr.new` directly in this repository for automatic preview builds on future PRs.*

ljharb:
It would be very harmful for unreleased permutations of software to gain broad adoption; if anything, I'd prefer to actively block all of my repos' usage with that service.

PR #3243, opened 2026-04-14
fix: namespaced wildcard exports should not be flagged as duplicate identifiers

### LINKED ISSUE #3220: Namespaced exports reported as duplicate identifiers
```ts
// AB.ts
export type A = {};
export type B = {};
```

```ts
// ABFeature.ts
export type * as AB from "./AB.ts";
```

```ts
// foo.ts
export type A = {};
```

```ts
// index.ts
export * from "ABFeature.ts"; // Error: duplicate identifier `A`
export * from "foo.ts";       // Error: duplicate identifier `A`
```

There are no duplicate identifiers. Symbol `A` can't be accessed through `ABFeature.ts`, and can only be accessed as `AB.A`.

whatfontisthis:
## Summary

Fixes #3220.

When a module uses `export * as NS from '...'`, its internal names are bound under the `NS` namespace (i.e. `NS.A`, `NS.B`), not re-exported into the flat namespace of the current module. However, the `ExportMap` builder was unconditionally adding the dependency to `exportMap.dependencies` even for namespaced wildcard re-exports. Since `dependencies` is iterated as flat star-exports in `ExportMap.forEach`, this caused the `export` rule to see the internal names (e.g. `A`, `B`) as if they were directly re-exported — leading to false-positive "duplicate identifier" errors.

**Example that was incorrectly reported as an error:**
```ts
// AB.ts
export type A = {};
export type B = {};

// ABFeature.ts
export type * as AB from "./AB";  // A and B are under AB.*, not flat

// foo.ts
export type A = {};

// index.ts (was incorrectly flagged: "Multiple exports of name 'A'")
export * from "./ABFeature";
export * from "./foo";
```

## Root Cause

In `/src/exportMap/visitor.js`, `ExportAllDeclaration` always added the getter to `this.exportMap.dependencies`. For `export * as NS from '...'`, this is wrong — the module is aliased as a namespace, not star-re-exported. The namespace binding is correctly handled separately by `processSpecifier`.

## Fix

Only add to `dependencies` when there is no `exported` alias on the `ExportAllDeclaration` node. When `astNode.exported` is set, it's a namespace re-export handled by `processSpecifier` and should not be added to flat star-export dependencies.

## Test Plan

- Added fixture files: `tests/files/issue-3220-namespace-re-export/{AB.ts,ABFeature.ts,foo.ts}`
- Added a valid test case in the TypeScript context of `tests/src/rules/export.js` that re-exports a file containing `export type * as AB from "./AB"` alongside another file exporting `type A` — previously this triggered a false positive, now it passes cleanly
- All 3013 existing tests continue to pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/exportMap/visitor.js
+++ b/src/exportMap/visitor.js
@@ -57,7 +57,16 @@ export default class ImportExportVisitorBuilder {
       },
       ExportAllDeclaration() {
         const getter = captureDependency(astNode, astNode.exportKind === 'type', this.remotePathResolver, this.exportMap, this.context, this.thunkFor);
-        if (getter) { this.exportMap.dependencies.add(getter); }
+        if (getter) {
+          // `export * as NS from '...'` aliases the entire module as a namespace binding.
+          // Its internal names are NOT re-exported into the flat namespace of the current
+          // module, so we must not add it to `dependencies` (which are iterated as
+          // flat star-exports). The namespace binding is handled separately by
+          // processSpecifier below. See: https://github.com/import-js/eslint-plugin-import/issues/3220
+          if (!astNode.exported) {
+            this.exportMap.dependencies.add(getter);
+          }
+        }
         if (astNode.exported) {
           processSpecifier(astNode, astNode.exported, this.exportMap, this.namespace);
         }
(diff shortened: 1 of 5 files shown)


ljharb:
This needs to be rebased; please mark it as ready for review when it is.

PR #3244, opened 2026-04-14
fix: no-unused-modules correctly handles `export { X as default }` syntax

### LINKED ISSUE #3216: `no-unused-modules` erroneously flags `as default` exports
## Issue Description

The `no-unused-modules` rule incorrectly reports exports as unused when they use the `export { X as default }` syntax, even when the default export is properly imported and used by other modules.

This appears to be a bug in how the plugin resolves default exports that are created via the re-export syntax rather than inline default export declarations.

## Minimal Reproduction

**test-export.ts**
```typescript
export function testHandler() {
  return 'result';
}

export { testHandler as default };
```

**test-import.ts**
```typescript
import handler from './test-export';
console.log(handler());
```

**Current behavior**: `eslint-plugin-import` reports `export function testHandler` in `test-export.ts` as unused.

**Expected behavior**: The rule should recognize that:
1. `testHandler` is re-exported as the default export
2. The default export is imported and used in `test-import.ts`
3. Therefore, neither the named export nor the re-export as default should be flagged as unused

## Configuration

```json
{
  "import/no-unused-modules": ["error", { "unusedExports": true }]
}
```

## Environment

- **eslint-plugin-import version**: 2.31.0
- **Node version**: 22.13.1
- **npm version**: 10.9.2

whatfontisthis:
## Summary

Fixes #3216.

When a file uses `export { testHandler as default }` syntax (exporting a local binding as the default export), the `no-unused-modules` rule incorrectly flagged the default export as unused even when it was imported correctly via `import handler from './module'` in another file.

- **Root cause:** In `updateExportUsage`, when scanning `ExportNamedDeclaration` specifiers, the raw exported name `"default"` was added to `newExportIdentifiers`. However, the exports map uses `IMPORT_DEFAULT_SPECIFIER` (`"ImportDefaultSpecifier"`) as the key for default exports. This mismatch meant the existing `whereUsed` data (populated during `doPreparation`) was dropped from the map on the first lint pass, and the export was stored under the wrong key `"default"`. On the next lint of the same file, `checkUsage` looked up `IMPORT_DEFAULT_SPECIFIER` and found nothing, incorrectly reporting the export as unused.
- **Fix:** Normalize `"default"` → `IMPORT_DEFAULT_SPECIFIER` when building `newExportIdentifiers` in `updateExportUsage`, consistent with how `EXPORT_DEFAULT_DECLARATION` is already handled on the line above.

## Changes

- `src/rules/no-unused-modules.js`: Normalize `"default"` exported specifier name to `IMPORT_DEFAULT_SPECIFIER` in `updateExportUsage`.
- `tests/src/rules/no-unused-modules.js`: Add regression test that lints the file twice (second lint triggers the bug — `doPreparation` is skipped due to caching, exposing the corrupted export map).
- `tests/files/no-unused-modules/export-as-default/`: New fixture files (`test-export.js`, `usage.js`) for the test.

## Test plan

- [x] New test fails before the fix and passes after
- [x] Full `no-unused-modules` test suite passes: `BABEL_ENV=test node_modules/.bin/nyc node_modules/.bin/mocha tests/src/rules/no-unused-modules.js` → 157 passing, 1 pending, 0 failing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/rules/no-unused-modules.js
+++ b/src/rules/no-unused-modules.js
@@ -723,7 +723,11 @@ module.exports = {
           if (specifiers.length > 0) {
             specifiers.forEach((specifier) => {
               if (specifier.exported) {
-                newExportIdentifiers.add(specifier.exported.name || specifier.exported.value);
+                const exportedName = specifier.exported.name || specifier.exported.value;
+                // `export { X as default }` uses the string "default" as the exported name,
+                // but the exports map stores default exports under IMPORT_DEFAULT_SPECIFIER.
+                // Normalize here so that the map key lookup is consistent.
+                newExportIdentifiers.add(exportedName === DEFAULT ? IMPORT_DEFAULT_SPECIFIER : exportedName);
               }
             });
           }
--- a/tests/files/no-unused-modules/export-as-default/test-export.js
+++ b/tests/files/no-unused-modules/export-as-default/test-export.js
@@ -1,0 +1,2 @@
+function testHandler() {}
+export { testHandler as default };
--- a/tests/files/no-unused-modules/export-as-default/usage.js
+++ b/tests/files/no-unused-modules/export-as-default/usage.js
@@ -1,0 +1,1 @@
+import handler from './test-export';
--- a/tests/src/rules/no-unused-modules.js
+++ b/tests/src/rules/no-unused-modules.js
@@ -742,6 +742,45 @@ describe('renameDefault', () => {
   });
 });
 
+describe('export { X as default } syntax (issue #3216)', () => {
+  const exportAsDefaultOptions = [{
+    unusedExports: true,
+    src: [testFilePath('./no-unused-modules/export-as-default/**/*.js')],
+  }];
+
+  // First lint: establishes the export/import maps via doPreparation
+  ruleTester.run('no-unused-modules', rule, {
+    valid: [
+      test({
+        options: exportAsDefaultOptions,
+        code: `import handler from '${testFilePath('./no-unused-modules/export-as-default/test-export.js')}'`,
+        filename: testFilePath('./no-unused-modules/export-as-default/usage.js'),
+      }),
+      test({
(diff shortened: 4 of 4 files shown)


whatfontisthis:
Note on the `codecov/project` check: `codecov/patch` passes — all lines introduced by this fix are covered by the new regression test. The project-level coverage gate appears to be a pre-existing threshold issue unrelated to this change. Happy to add more tests if the maintainers think that would help.

LaishGlenberg:
Can we please get this merged? I am getting this problem on non default module exports as well. No amount of configuration or testing could get it to work.

nrps9909 (approved):
Reviewed `5a6c4b0e1f52c883d688586e783a49631651fbd4`. Normalizing the exported `default` name to the same map key used during preparation preserves `whereUsed` across repeated lint passes. I reproduced the new regression on base `3a99e4c8d3bfd2cd466353d11784eb06dad9b166` (152 passing, 1 pending, 1 failing); the focused head run passes 153 tests, with 1 pending.

An independent Linter probe passes all 13 observations on head, versus 6 failures on base: three consecutive lint passes for a used default export, an unused named export beside a used default, an unused default beside a used named export, and a quoted `"default"` specifier. Removing the consumer import also correctly makes the default export unused. This supports the scoped default-export cache fix; it does not establish that every named-export report in the linked issue is resolved.

With ESLint 8.57.1, `npm test` passes 3,015 tests (1 pending), including posttest lint, generated documentation checks, markdownlint and build. With ESLint 9.39.5, the test phase passes 3,011 tests (1 pending); its posttest is blocked by the repository's legacy lint configuration/development plugin compatibility.

The upstream check list still contains two failed old-ESLint matrix jobs and cancelled Windows jobs. Their run log endpoint returns HTTP 410, so I have not established their causes and am not claiming an entirely green CI matrix.

Review and local validation performed with Codex.

PR #3246, opened 2026-04-23
[Fix] `no-duplicates`: remove extra space when merging import specifiers

### LINKED ISSUE #3222: no-duplicates rule adds extra space
The no-duplicates rule auto-fix logic is adding an extra space in some cases.

Example case:
```
import { Observable, Subscription } from 'rxjs';
import { of } from 'rxjs';
```

Auto-fixing this rule turns into:
```
import { Observable, Subscription , of } from 'rxjs';
```

Note the extra space after "Subscription" before the comma.

Some feedback I've received is to use the `@stylistic/comma-spacing` rule to fix it, and that does work in a vacuum. However, this doesn't work for my project because I have other rules that affect imports. Specifically, I'm using the `imports` rule from the plugin `eslint-plugin-simple-import-sort` which sorts the imports alphabetically, turning it into this:
```
import { Observable, of, Subscription  } from 'rxjs';
```
And then the `comma-spacing` rule doesn't clean up the double spaces before the closing curly brace.

Using `comma-spacing` as a fix is a workaround anyway to fix the extra space that was introduced by the `no-duplicates` rule. The cleanest solution then is to fix the original bug in eslint-plugin-import.

mixelburg:
Fixes #3222

## Problem

The `no-duplicates` rule auto-fix inserts extra spaces when merging import specifiers. For example:

```js
// Before fix
import { Observable, Subscription } from 'rxjs';
import { of } from 'rxjs';

// After auto-fix (broken)
import { Observable, Subscription , of } from 'rxjs';
//                               ^ extra space before comma
```

## Root Cause

`specifier.identifiers` is created by splitting the source text between `{` and `}` on commas (line 123):

```js
identifiers: sourceCode.text.slice(openBrace.range[1], closeBrace.range[0]).split(',')
```

This preserves leading/trailing whitespace: `' Subscription '` → `[' Observable', ' Subscription ']` → after split, identifiers have spaces.

When building the merged specifier text (line 174), `curWithType` used `cur` (untrimmed) instead of `trimmed`, causing extra spaces in the output.

## Fix

One-line change on line 174: use `trimmed` instead of `cur` in `curWithType`.

```diff
- const curWithType = ... ? `type ${cur}` : cur;
+ const curWithType = ... ? `type ${trimmed}` : trimmed;
```

All existing tests pass.

### DIFF
--- a/src/rules/no-duplicates.js
+++ b/src/rules/no-duplicates.js
@@ -2,6 +2,7 @@ import { getSourceCode } from 'eslint-module-utils/contextCompat';
 import resolve from 'eslint-module-utils/resolve';
 import semver from 'semver';
 import flatMap from 'array.prototype.flatmap';
+import trimEnd from 'string.prototype.trimend';
 
 import docsUrl from '../docsUrl';
 
@@ -171,7 +172,21 @@ function getFix(first, rest, sourceCode, context) {
         // Add *only* the new identifiers that don't already exist, and track any new identifiers so we don't add them again in the next loop
         const [specifierText, updatedExistingIdentifiers] = specifier.identifiers.reduce(([text, set], cur) => {
           const trimmed = cur.trim(); // Trim whitespace before/after to compare to our set of existing identifiers
-          const curWithType = trimmed.length > 0 && preferInline && isTypeSpecifier ? `type ${cur}` : cur;
+          const hasLineComment = (/\/\/[^\n]*$/).test(trimmed);
+          let curWithType;
+          if (trimmed.length > 0 && preferInline && isTypeSpecifier) {
+            curWithType = `type ${trimmed}`;
+          } else if (hasLineComment && trimmed.length > 0) {
+            // Preserve a trailing newline after line comments so the closing brace
+            // is not accidentally commented out when the specifier is spliced in.
+            curWithType = `${trimmed}\n`;
+          } else if (cur.includes('\n')) {
+            // Preserve leading newline+indentation for multiline import specifiers
+            // so merged specifiers stay on their own lines.
+            curWithType = trimEnd(cur).replace(/^[^\S\n]+/, '');
+          } else {
+            curWithType = trimmed;
+          }
           if (existingIdentifiers.has(trimmed)) {
             return [text, set];
           }
@@ -217,7 +232,22 @@ function getFix(first, rest, sourceCode, context) {
       fixes.push(fixer.insertTextAfter(firstToken, ` ${defaultImportName},`));
       if (shouldAddSpecifiers) {
         // `import def, {...} from './foo'` → `import def, {..., ...} from './foo'`
-        fixes.push(fixer.insertTextBefore(closeBrace, specifiersText));
+        const textBeforeClose2 = sourceCode.text.slice(openBrace.range[1], closeBrace.range[0]);
+        const trailingMatch2 = textBeforeClose2.match(/(\s+)$/);
+        const trailingWhitespace2 = trailingMatch2 ? trailingMatch2[1] : '';
+        if (textBeforeClose2.trim() === '') {
+          fixes.push(fixer.replaceTextRange(
+            [openBrace.range[1], closeBrace.range[0]],
+            specifiersText,
+          ));
+        } else if (trailingWhitespace2) {
+          fixes.push(fixer.replaceTextRange(
+            [closeBrace.range[0] - trailingWhitespace2.length, closeBrace.range[1]],
+            `${specifiersText}${trailingWhitespace2}}`,
+          ));
+        } else {
+          fixes.push(fixer.insertTextBefore(closeBrace, specifiersText));
+        }
       }
     } else if (!shouldAddDefault && openBrace == null && shouldAddSpecifiers) {
       if (first.specifiers.length === 0) {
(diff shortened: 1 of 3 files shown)


tests/src/rules/no-duplicates.js
```diff
@@ -402,7 +403,7 @@ import {x,y} from './foo'
         ${''}
         export default TestComponent;
       `,
-      output: `\n        import {DEFAULT_FILTER_KEYS,BULK_ACTIONS_ENABLED} from '../constants';\n        import React from 'react';\n                ${''}\n        const TestComponent = () => {\n          return <div>\n          </div>;\n        }\n        ${''}\n        export default TestComponent;\n      `,
+      output: `\n        import {\n          DEFAULT_FILTER_KEYS,\n          BULK_DISABLED,\n        BULK_ACTIONS_ENABLED} from '../constants';\n        import React from 'react';\n                ${''}\n        const TestComponent = () => {\n          return <div>\n          </div>;\n        }\n        ${''}\n        export default TestComponent;\n      `,
```
ljharb: i'm confused, this now has a newline after the `{` but not one before the `}`? this seems unrelated to the purpose of the PR.

tests/src/rules/no-duplicates.js
```diff
@@ -424,7 +425,7 @@ import {x,y} from './foo'
         } from 'bar';
 
       `,
-      output: `\n        import {A1,B1,C1} from 'foo';\n                ${''}\n        import {A2,B2,C2} from 'bar';\n                ${''}\n      `,
+      output: `\n        import {A1,B1,C1} from 'foo';\n                ${''}\n        import {\n          A2,\n        B2,C2} from 'bar';\n                ${''}\n      `,
```
ljharb: same question here

src/rules/no-duplicates.js
```diff
@@ -171,7 +171,15 @@ function getFix(first, rest, sourceCode, context) {
         // Add *only* the new identifiers that don't already exist, and track any new identifiers so we don't add them again in the next loop
         const [specifierText, updatedExistingIdentifiers] = specifier.identifiers.reduce(([text, set], cur) => {
           const trimmed = cur.trim(); // Trim whitespace before/after to compare to our set of existing identifiers
-          const curWithType = trimmed.length > 0 && preferInline && isTypeSpecifier ? `type ${cur}` : cur;
+          const hasLineComment = /\/\/[^\n]*$/.test(trimmed);
+          let curWithType;
+          if (trimmed.length > 0 && preferInline && isTypeSpecifier) {
+            curWithType = `type ${trimmed}`;
+          } else if (hasLineComment && trimmed.length > 0) {
+            curWithType = `${trimmed}\n`;
```
ljharb, line 182: why the newlines? that's not behavior the rule has had before.

tests/src/rules/no-duplicates.js
```diff
@@ -53,20 +53,20 @@ ruleTester.run('no-duplicates', rule, {
   invalid: [
     test({
       code: "import { x } from './foo'; import { y } from './foo'",
-      output: "import { x , y } from './foo'; ",
+      output: "import { x ,y} from './foo'; ",
```
ljharb: this seems wrong to me - either both x and y should have spaces between them and the closest curly brace, or neither should.

tests/src/rules/no-duplicates.js
```diff
@@ -53,20 +53,20 @@ ruleTester.run('no-duplicates', rule, {
...
-      output: "import {x,y, z } from './foo';  ",
+      output: "import {x,y,z} from './foo';  ",
       errors: ['\'./foo\' imported multiple times.', '\'./foo\' imported multiple times.', '\'./foo\' imported multiple times.'],
     }),
 
     // ensure resolved path results in warnings
     test({
       code: "import { x } from './bar'; import { y } from 'bar';",
-      output: "import { x , y } from './bar'; ",
+      output: "import { x ,y} from './bar'; ",
```
ljharb: also here

ljharb (changes requested):

tests/src/rules/no-duplicates.js
mixelburg: Addressed in commit b7c39594

tests/src/rules/no-duplicates.js
mixelburg: Addressed in commit b7c39594

src/rules/no-duplicates.js
mixelburg: Addressed in commit b7c39594

tests/src/rules/no-duplicates.js
mixelburg: Addressed in commit b7c39594

tests/src/rules/no-duplicates.js
mixelburg: Addressed in commit b7c39594

tests/src/rules/no-duplicates.js
```diff
@@ -168,7 +168,7 @@ ruleTester.run('no-duplicates', rule, {
 
     test({
       code: "import { } from './foo'; import {x} from './foo'",
-      output: "import { x} from './foo'; ",
+      output: "import {x } from './foo'; ",
```
ljharb: why is this affecting the spacing directly inside the braces?

tests/src/rules/no-duplicates.js
```diff
@@ -186,7 +186,8 @@ ruleTester.run('no-duplicates', rule, {
 
     test({
       code: "import './foo'; import { /*x*/} from './foo'; import {//y\n} from './foo'; import {z} from './foo'",
-      output: "import { /*x*///y\nz} from './foo';   ",
+      output: "import {/*x*///y\nz} from './foo';   ",
```
ljharb, line 189: same question here?

tests/src/rules/no-duplicates.js
```diff
@@ -405,9 +406,7 @@ import {x,y} from './foo'
       output: `
         import {
           DEFAULT_FILTER_KEYS,
-          BULK_DISABLED,
-        ${''}
-          BULK_ACTIONS_ENABLED
+          BULK_DISABLED,BULK_ACTIONS_ENABLED
```
ljharb: i don't think this should be removing newlines here - i don't think there's another rule that will clean this up properly as is.

tests/src/rules/no-duplicates.js
```diff
@@ -443,10 +442,8 @@ import {x,y} from './foo'
         import {A1,B1,C1} from 'foo';
                 ${''}
         import {
-          A2,
-        ${''}
-          B2,
-          C2} from 'bar';
+          A2,B2,C2
```
ljharb: same here

ljharb (changes requested):

tests/src/rules/no-duplicates.js
mixelburg: Addressed in commit 76f848a0486d2749eeee9ab2205b3cf4f148eca1 — when the first import has only whitespace between its braces (empty import like `{ }`), we now replace the entire brace content rather than preserving trailing space, giving `{x}` instead of `{x }`.

tests/src/rules/no-duplicates.js
mixelburg: Addressed in commit 76f848a0486d2749eeee9ab2205b3cf4f148eca1

tests/src/rules/no-duplicates.js
mixelburg: Addressed in commit 76f848a0486d2749eeee9ab2205b3cf4f148eca1 — specifiers from multiline imports now preserve their leading newline+indentation, so `BULK_ACTIONS_ENABLED` stays on its own line.

tests/src/rules/no-duplicates.js
mixelburg: Addressed in commit 76f848a0486d2749eeee9ab2205b3cf4f148eca1 — same fix: `A2`, `B2`, `C2` each stay on their own lines.

ljharb (approved):

Issue #1834, opened 2020-06-23
[import/export] False "Multiple exports of name" error when two "export * as ..." contain same name. 


zimmed:
*Apologies if this is a duplicate -- it's difficult to search terms with an asterisk.*

Found issue using:
  `eslint` 6.6.0, 6.8.0, 7.3.1
  `eslint-plugin-import` 2.20.2, 2.21.2
(tried various versions to see if it had already been fixed)

## Setup:
```javascript
// A.js
export const FOO = 'a-foobar';
```
```javascript
// B.js
export const FOO = 'b-foobar';
```
```javascript
// C.js
export * as A from './A';
export * as B from './B';
```

## Expected:
No `import/export` error in `C.js`, since both instances of `FOO` are under their own namespaces (A/B).
```javascript
import * as C from './C';

console.log(C.A.FOO); // 'a-foobar'
console.log(C.B.FOO); // 'b-foobar'
console.log(C.FOO); // undefined
```

## Result:
Error in C.js:
`Multiple exports of name 'FOO'. eslint(import/export)`

## Workaround

The following equivalent method works without errors, but annoying because ESLint is preventing me from taking advantage of newer ES syntax:

```javascript
import * as A from './A';
import * as B from './B';

export { A, B };
```

ljharb:
This might be a duplicate of #518?

zimmed:
> This might be a duplicate of #518?

That seems to be related to shadowing, whereas this example has no actual variable shadowing. If it was...
```javascript
export * from './A';
export * from './B';
```
Then that would likely be the issue, but since the original code posted is equivalent to:
```javascript
import * as A from './A';
import * as B from './B';

export { A, B };
```
There are no shadowed variable names, since `FOO` only exists within the individual namespaces. (This works without lint errors, by the way).

PR #3250, opened 2026-05-18
Resolve `export * as ns` re-exports under modern parsers

### LINKED ISSUE #2002: [export] incorrect report on `export * as`

rasmi:
Fixes two pre-existing `ExportAllDeclaration` bugs that affect modern parsers (espree ES2020+, `@babel/eslint-parser`) but not `babel-eslint`. Extracted from [#3230](https://github.com/import-js/eslint-plugin-import/pull/3230) (eslint v10 support) to make that PR cleaner.

## Root cause

`babel-eslint` represents `export * as ns from './mod'` as `ExportNamedDeclaration` + `ExportNamespaceSpecifier` (a working code path), so the bug was latent. Modern parsers emit `ExportAllDeclaration` with `exported`, which hit two broken spots:

- **`specifier.js`**: passed a string (`specifier.source.value`) to `namespace.add()`, which expects an AST node — so the namespace getter was never set up and `ns` didn't resolve. Fixed by defining the lazy getter directly, matching the existing [`ExportNamespaceSpecifier`](https://github.com/rasmi/eslint-plugin-import/blob/export-star-as-fix/src/exportMap/specifier.js#L11-L15) case.
- **`visitor.js`**: `export * as ns` was unconditionally added to the star-export `dependencies` set, incorrectly flattening the target module's exports into the current one. Fixed by only adding to `dependencies` for plain `export *` (no `as`).

## Tests

The included regression test covers the false negative: the `namespace` rule now correctly flags invalid deep references through `export * as` re-exports that were previously silently missed (fails on `main`, passes with the fix).

The false-positive symptoms #1834, #1883, #1845, #2002, #2289 ("errors after upgrading Babel") are resolved by the same change, but only reproduce with `@babel/eslint-parser`, which is introduced in #3230.

Fixes #2002, #1883, #1834, #1845, #2289.

### DIFF
--- a/src/exportMap/specifier.js
+++ b/src/exportMap/specifier.js
@@ -9,16 +9,27 @@ export default function processSpecifier(specifier, astNode, exportMap, namespac
       local = 'default';
       break;
     case 'ExportNamespaceSpecifier':
-      exportMap.namespace.set(specifier.exported.name, Object.defineProperty(exportMeta, 'namespace', {
+      Object.defineProperty(exportMeta, 'namespace', {
         get() { return namespace.resolveImport(nsource); },
-      }));
+      });
+      exportMap.namespace.set(specifier.exported.name, exportMeta);
       return;
     case 'ExportAllDeclaration':
-      exportMap.namespace.set(specifier.exported.name || specifier.exported.value, namespace.add(exportMeta, specifier.source.value));
+      Object.defineProperty(exportMeta, 'namespace', {
+        get() { return namespace.resolveImport(specifier.source.value); },
+      });
+      exportMap.namespace.set(
+        specifier.exported.name || specifier.exported.value,
+        exportMeta,
+      );
       return;
     case 'ExportSpecifier':
       if (!astNode.source) {
-        exportMap.namespace.set(specifier.exported.name || specifier.exported.value, namespace.add(exportMeta, specifier.local));
+        namespace.add(exportMeta, specifier.local);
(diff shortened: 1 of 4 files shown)


src/exportMap/specifier.js
```diff
@@ -14,7 +14,9 @@ export default function processSpecifier(specifier, astNode, exportMap, namespac
       }));
       return;
     case 'ExportAllDeclaration':
-      exportMap.namespace.set(specifier.exported.name || specifier.exported.value, namespace.add(exportMeta, specifier.source.value));
+      exportMap.namespace.set(specifier.exported.name || specifier.exported.value, Object.defineProperty(exportMeta, 'namespace', {
+        get() { return namespace.resolveImport(specifier.source.value); },
+      }));
```
ljharb: let's do the defineProperty on a separate line before the namespace.set, for clarity

ljharb (approved):

Themoor1 (approved):

Issue #1645, opened 2020-02-05
[import/no-unused-modules] - "No files matching the pattern" error


EvgenyOrekhov:
I'm using both eslint-plugin-import and [eslint-plugin-json](https://github.com/azeemba/eslint-plugin-json). When there are only `*.json` files in my project, I get "No files matching the pattern" error from the `import/no-unused-modules` rule, **and ESLint crashes**.

Steps to reproduce:

1. Clone this minimal reproducible case: https://github.com/EvgenyOrekhov/eslint-config-hardcore/tree/no-files-matching-the-pattern-bug
2. `npm install`
3. `npm test`

Actual result:

```text
> no-files-matching-the-pattern-bug@0.0.0 test C:\eslint-config-hardcore
> eslint --ext .js,.json .


Oops! Something went wrong! :(

ESLint: 6.8.0.

No files matching the pattern "C:\eslint-config-hardcore" were found.
Please check for typing mistakes in the pattern.

npm ERR! Test failed.  See above for more details.
```

Expected result: there should be no "No files matching the pattern" error, there should be only one warning from eslint-plugin-json

```text
> no-files-matching-the-pattern-bug@0.0.0 test C:\eslint-config-hardcore
> eslint --ext .js,.json .


C:\eslint-config-hardcore\invalid.json
  1:2  error  Expected a JSON object, array or literal  json/*

✖ 1 problem (1 error, 0 warnings)

npm ERR! Test failed.  See above for more details.
```

When I remove the `import/no-unused-modules` rule from `.eslintrc.json`, the problem goes away.

EvgenyOrekhov:
Note that it's not about `*.json` files. It's about the absence of `*.js` files. If I try to use the `import/no-unused-modules` rule in a TypeScript project, and if there are no `*.js` files, I get the same error.

ljharb:
This is an error coming from eslint itself. By default, if you're linting "no javascript files", it's probably a bug.

If you have no `.foo` files, don't pass `.foo` to `--ext`.

EvgenyOrekhov:
@ljharb I have a [config](https://github.com/EvgenyOrekhov/eslint-config-hardcore) that has ~500 rules from 15 different plugins, and only `import/no-unused-modules` is causing this error.

ljharb:
Indeed; it's the only rule where we're using eslint's FileEnumerator logic (which is what the command-line eslint uses as well).

[...]

PR #3252, opened 2026-05-18
Replace eslint-plugin-json with a fixture processor

rasmi:
While updating to support eslint v10 in #3230, `eslint-plugin-json` (and specifically `eslint-plugin-jsonc` for v10+) caused a series of interrelated version/dependency issues which required various `dep-time-travel.sh` and version-pinning workarounds (see [here](https://github.com/import-js/eslint-plugin-import/pull/3230#discussion_r3255667807)). 

The #1645 regression test only needs *some* `.json` processor in the pipeline so `no-unused-modules` is exercised against a JSON-only directory (the bug was the import rule crashing there, not the JSON diagnostic). The fixture does that and removes the JSON-plugin dependency entirely.

(potential alternative to #3230's `eslint-plugin-jsonc` switch https://github.com/import-js/eslint-plugin-import/commit/d8d72941d0e5b52a88359ba73a8f7876728ff803)

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,6 @@
     "eslint-module-utils": "file:./utils",
     "eslint-plugin-eslint-plugin": "^2.3.0",
     "eslint-plugin-import": "2.x",
-    "eslint-plugin-json": "^2.1.2",
     "find-babel-config": "=1.2.0",
     "fs-copy-file-sync": "^1.1.1",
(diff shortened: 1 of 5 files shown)


ljharb (approved):

rasmi:
Will remove the workarounds in #3230 today and rebase on main!

PR #3256, opened 2026-05-22
[Tests] regression tests for `export * as` bug class fixed by #3250

rasmi:
Adds load-bearing regression tests for the `export * as`-related bugs fixed in #3250. Each was verified by temp-reverting `src/exportMap/specifier.js` + `src/exportMap/visitor.js` to the pre-`#3250` state and confirming each test fails:

- **#2289** (`tests/src/rules/export.js`): `export * from` of barrels that use `export * as` must not false-flag same-named exports in unrelated sibling chains. 
  - Pre-`#3250` flagged `Multiple exports of name 'X'`.
- **#2002** (`tests/src/rules/export.js`): Same bug class, slightly different shape mixing barrels (one using `export * as`).
  - Pre-`#3250` flagged `Multiple exports of name 'remark'`.
- **#1845** (`tests/src/rules/namespace.js`): Deep access (`alpha.abc.A` / `alpha.def.D`) through multiple `export * as` re-exports must resolve real properties.
  - Pre-`#3250` flagged both as not found.
- **Named-import-side complement to `#3250`** (`tests/src/rules/namespace.js`): `import { myName } from "./middle"; console.log(myName.b)` invalid case. Fills in the named-import entry point alongside `#3250`'s existing `middle.myName.b` test (which uses `import * as middle`).

Three new fixture directories (`tests/files/issue-{1845,2002,2289}`).

### DIFF
--- a/tests/files/issue-1845/alpha/abc.js
+++ b/tests/files/issue-1845/alpha/abc.js
@@ -1,0 +1,3 @@
+export const A = 'A';
+export const B = 'B';
+export const C = 'C';
--- a/tests/files/issue-1845/alpha/def.js
+++ b/tests/files/issue-1845/alpha/def.js
@@ -1,0 +1,3 @@
+export const D = 'D';
+export const E = 'E';
+export const F = 'F';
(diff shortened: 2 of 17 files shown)


rasmi:
To clarify -- #1834 was fixed long before #3250, and #1883 was not clearly reproducible or has otherwise since been fixed (but not by #3250).

ljharb (approved):

PR #3258, opened 2026-06-01
Report qs instead of @types/qs for extraneous deps

### LINKED ISSUE #3208: [no-extraneous-import] Error message reports the "@types/blah" package instead of the "blah" package
Here's a minimal repro: https://github.com/cakoose/no-extraneous-dependencies-reporting-types-package-bug

## Summary of repo contents

index.ts

```
import qs from 'qs';
...
```

package.json (I forget to add a dependency on "qs", but it gets included indirectly via "express")

```
    "dependencies": {
        "express": "^4.21.2"
    },
    "devDependencies": {
        "@types/qs": "^6.9.7",
        ...
    }
```

eslint.config.mjs:

```
    rules: {
        'import/no-extraneous-dependencies': ['error', {
            devDependencies: false,
            peerDependencies: false,
        }],
    },
```

## To reproduce

npm install
npm run lint

```
 .../index.ts
   1:1  error  '@types/qs' should be listed in the project's dependencies, not devDependencies
 import/no-extraneous-dependencies
```

Problem: It's reporting "@types/qs" instead of "qs".

This slight reporting issue really threw me off. I was trying all kinds of weird things to figure out why it was reporting the "@types/qs" package, only to realize that I didn't include "qs" in my dependencies list.

morgan-coded:
`no-extraneous-dependencies` could point at `@types/qs` even though the package a user needs to declare is `qs`.
When resolution lands in an `@types/*` package, the report now uses the runtime package name.
Checked it with the focused mocha run, the red/green case, eslint, and markdownlint.
Closes #3208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - ExportMap: resolve export * as ns re-exports under modern parsers ([#3250], thanks [@rasmi] [@JounQin] [@butterybread])
 - [`order`]: make the alphabetize comparator transitive for sibling/parent imports in the same group, so the autofix converges under Node 25's updated V8 sort ([#3235], thanks [@Sekhmet])
 - [`no-duplicates`]: remove extra space when merging import specifiers ([#3246], thanks [@mixelburg] [@shastaxc])
+- [`no-extraneous-dependencies`]: report runtime package name, not `@types/*`, when an import resolves via a `@types` devDependency ([#3208], thanks [@morgan-coded] [@cakoose] [@JounQin])
 
 ### Changed
 - [Docs] `no-unused-modules`: Fix docs of `ignoreUnusedTypeExports` option ([#3233], thanks [@ej612])
@@ -1201,6 +1202,7 @@ for info on changes for earlier releases.
 [#3236]: https://github.com/import-js/eslint-plugin-import/pull/3236
 [#3235]: https://github.com/import-js/eslint-plugin-import/issues/3235
 [#3233]: https://github.com/import-js/eslint-plugin-import/pull/3233
+[#3208]: https://github.com/import-js/eslint-plugin-import/issues/3208
 [#3195]: https://github.com/import-js/eslint-plugin-import/issues/3195
 [#3191]: https://github.com/import-js/eslint-plugin-import/pull/3191
 [#3173]: https://github.com/import-js/eslint-plugin-import/pull/3173
@@ -1863,6 +1865,7 @@ for info on changes for earlier releases.
 [@brendo]: https://github.com/brendo
 [@brettz9]: https://github.com/brettz9
 [@butterybread]: https://github.com/butterybread
+[@cakoose]: https://github.com/cakoose
 [@chabb]: https://github.com/chabb
 [@Chamion]: https://github.com/Chamion
 [@charlessuh]: https://github.com/charlessuh
@@ -2018,6 +2021,7 @@ for info on changes for earlier releases.
 [@mixelburg]: https://github.com/mixelburg
 [@mlenser]: https://github.com/mlenser
 [@moeriki]: https://github.com/moeriki
+[@morgan-coded]: https://github.com/morgan-coded
 [@mpalmer685]: https://github.com/mpalmer685
 [@mpint]: https://github.com/mpint
 [@mplewis]: https://github.com/mplewis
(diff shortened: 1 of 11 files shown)


src/rules/no-extraneous-dependencies.js
```diff
@@ -233,17 +233,24 @@ function reportIfMissing(context, deps, depsOptions, node, name, moduleSystem) {
     }
   }
 
+  // when the import only resolves through an `@types/*` package (e.g. importing
+  // `qs` resolves to `@types/qs`), the missing runtime package is the imported
+  // name, not the `@types/*` package the resolver landed on
+  const reportedPackageName = realPackageName && realPackageName.startsWith('@types/')
```
ljharb: For the `@types/*` case this reports `importPackageName`. That's correct for direct imports, but it leans on the imported name - and the comment just above notes `importPackageName` "can be misinterpreted" for aliased imports (an alias resolving into `@types/qs` would report the alias, not `qs`).

The runtime name is deterministically derivable from `realPackageName` via DefinitelyTyped's scope-mangling formula - the same transform TypeScript itself uses (`getPackageNameFromTypesPackageName` → `unmangleScopedPackageName` in `src/compiler/moduleNameResolver.ts`): `@types/qs` → `qs`, `@types/babel__core` → `@babel/core`.

```js
// mirrors TS's unmangleScopedPackageName
function runtimePackageFromTypes(name) {
  const inner = name.slice('@types/'.length);
  return inner.includes('__') ? `@${inner.replace('__', '/')}` : inner;
}
```

`@definitelytyped/utils` exports this as `unmangleScopedPackage`, but that dep's engines range is too restrictive. Deriving from `realPackageName` handles scoped `@types` packages and aliased imports robustly, rather than trusting the imported string.

src/rules/no-extraneous-dependencies.js
```diff
@@ -233,17 +233,24 @@ function reportIfMissing(context, deps, depsOptions, node, name, moduleSystem) {
...
+    ? importPackageName
+    : realPackageName || importPackageName;
```
ljharb: This `|| importPackageName` fallback is the `codecov/project` failure: it's the one uncovered branch (`[34,0]` - `realPackageName` is always truthy in the suite, so the right operand never runs), which drops branch coverage below base. Adopting the `realPackageName`-derived formula above removes this fallback entirely; otherwise a test exercising a falsy `realPackageName` is needed to cover it.

src/rules/no-extraneous-dependencies.js
```diff
@@ -233,17 +233,24 @@ function reportIfMissing(context, deps, depsOptions, node, name, moduleSystem) {
...
+
   if (declarationStatus.isInDevDeps && !depsOptions.allowDevDeps) {
-    context.report(node, devDepErrorMessage(realPackageName || importPackageName));
+    context.report(node, devDepErrorMessage(reportedPackageName));
```
ljharb: Sanity check on the message category: for the `qs` case this emits *"'qs' should be listed in the project's dependencies, **not devDependencies**"*, but `qs` isn't in devDependencies - only `@types/qs` is. Since the substituted runtime package is genuinely undeclared, `missingErrorMessage` (*"…Run `npm i -S qs`…"*) reads more accurately than the devDeps message here.

I think it's important to report on both dependencies being in the proper category.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - ExportMap: resolve export * as ns re-exports under modern parsers ([#3250], thanks [@rasmi] [@JounQin] [@butterybread])
 - [`order`]: make the alphabetize comparator transitive for sibling/parent imports in the same group, so the autofix converges under Node 25's updated V8 sort ([#3235], thanks [@Sekhmet])
 - [`no-duplicates`]: remove extra space when merging import specifiers ([#3246], thanks [@mixelburg] [@shastaxc])
+- [`no-extraneous-dependencies`]: report the imported runtime package name instead of the `@types/*` package when an import only resolves through an `@types` devDependency ([#3208], thanks [@cakoose] [@JounQin])
```
```suggestion
- [`no-extraneous-dependencies`]: report the imported runtime package name instead of the `@types/*` package when an import only resolves through an `@types` devDependency ([#3208], thanks [@morgan-coded] [@cakoose] [@JounQin])
```

ljharb (changes requested):

ljharb (approved):

PR #3261, opened 2026-06-04
extensions: treat relative package roots as packages for ignorePackages

### LINKED ISSUE #2844: [import/extensions] ".." and "." import paths are incorrectly flagged as requiring an extension, when either is al local package
In #2778, we determined that this plugin incorrectly flags ".." and "." import paths, when those imports refer to a local package (with a package.json).

morgan-coded:
Problem I saw:
`import/extensions` could still report `.` or `..` as missing an extension even when `ignorePackages` was enabled.
Those relative imports can point at a directory that is itself a package root, but the rule only gave package treatment to external or scoped imports. Thanks @benasher44 for filing the follow-up from #2778.

What I changed:
I added a narrow package-root check for relative `.` and `..` imports.
When one of those specifiers resolves to a package root, it now follows the same `ignorePackages` behavior as package imports. I kept the scope to `ignorePackages` and did not broaden the ambiguous `always` plus relative-import case.

How I checked it:
I moved the existing TODO cases into valid first and got the expected red run: 69 passing, 2 failing.
After the fix, the `extensions` rule suite passed with 71 passing and 0 failing. The full repo suite also passed with 3029 passing, 0 failing, and 1 pending. Lint passed on the changed files, and pre-push hygiene is clean.

Closes #2844

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`consistent-type-specifier-style`]: add `prefer-top-level-if-only-type-imports` option ([#3210], thanks [@aldeed])
 
 ### Fixed
+- Fixed `extensions` false positives for relative package-root imports when `ignorePackages` is enabled. ([#2844], thanks [@benasher44])
 - [`no-duplicates`]: fix `prefer-inline` autofix producing invalid syntax when an identifier named `from` is imported ([#3236], thanks [@DukeDeSouth] [@Quantaly])
 - [`no-duplicates`]: avoid false positives for TypeScript namespace and default type imports that cannot be merged ([#3195], thanks [@sjh9714] [@robyoder])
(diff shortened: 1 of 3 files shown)


ljharb (changes requested):
I'm a bit skeptical of non-breaking-change PRs that change existing tests. can you try to minimize the diff on existing tests?

morgan-coded:
That makes sense. I reworked the tests so the existing TODO cases are just moved verbatim from invalid to valid, with the extra fallback coverage added separately.

Focused suite, full tests-only, lint, diff-check, and hygiene are all clean.

morgan-coded:
Closing this one as well — it's fallen behind `main` and stalled after the last round of review, so I'll take it off my queue rather than let it linger. Happy to revisit with a rebased version if it's still useful. Thanks for the look.

ljharb:
same as the other one; please leave PRs open even if they're stale.

PR #3262, opened 2026-06-27
[actions] update `codecov/codecov-action` to v7

captaindonald:
Follow-up to #3230 (ESLint v10) coverage work.

`v3.1.5` is the last v3 release. v4 made the upload token effectively required and broke tokenless coverage uploads from fork PRs of public repos — the reason this action was pinned back. v5 restored reliable tokenless fork-PR uploads; v7 is the current release.

Fork PRs cannot read repository secrets, so they have always relied on tokenless uploads. Against the sunset v3 endpoint those uploads were frequently rate-limited/dropped, so version-specific code covered only by certain matrix jobs appeared uncovered in the merged head report (~19 upload sessions on the head vs ~40 on the base in #3230). Upgrading makes uploads reliable.

This bumps `codecov/codecov-action@v3.1.5` -> `v7.0.0` in all four coverage-uploading workflows.

### DIFF
--- a/.github/workflows/eslint-8+.yml
+++ b/.github/workflows/eslint-8+.yml
@@ -68,7 +68,7 @@ jobs:
           skip-ls-check: true
       - run: npm run pretest
       - run: npm run tests-only
-      - uses: codecov/codecov-action@v3.1.5
+      - uses: codecov/codecov-action@v7.0.0
 
   node:
     name: 'eslint 8+'
--- a/.github/workflows/eslint-old.yml
+++ b/.github/workflows/eslint-old.yml
@@ -102,7 +102,7 @@ jobs:
           skip-ls-check: true
       - run: npm run pretest
       - run: npm run tests-only
-      - uses: codecov/codecov-action@v3.1.5
+      - uses: codecov/codecov-action@v7.0.0
 
   node:
     name: 'eslint 2 - 7'
--- a/.github/workflows/native-wsl.yml
+++ b/.github/workflows/native-wsl.yml
@@ -151,7 +151,7 @@ jobs:
         }
 
     - name: codecov
-      uses: codecov/codecov-action@v3.1.5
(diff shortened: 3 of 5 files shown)


ljharb:
Are you saying that as of v5, the tokenless behavior is as good or better than with 3.1.5?

captaindonald:
> Are you saying that as of v5, the tokenless behavior is as good or better than with 3.1.5?

Yeah, the v5 wrapper/CLI path is what Codecov points to for "[more reliable tokenless uploads.](https://docs.codecov.com/docs/codecov-tokens#legacy-upload-methods)" This repo does tokenless uploads via the legacy uploader, which shares Codecov's global rate limit — almost certainly why we see these false 15%+ coverage drops on PRs that actually add coverage (a rate-limited upload just fails, so that job's coverage never posts).

The two big things v7 would provide over v5.5.5 are node24 and VULN-1652 template-injection hardening. v5 is more proven though so I see value in pinning at v5.5.5 for now as well.

lorand-horvath:
Is this the last step before an actual new version of eslint-plugin-import can be released?

captaindonald:
No, this is just QOL to make the codecov checks more reliable. It's not blocking anything.

ljharb:
What would we gain by pinning at 5.5.5 vs going all the way to 7?

captaindonald:
Nothing that matters for this repo, honestly. Just conservatism. v5 is more proven, v7 is very new. But there is no upload/coverage benefit to holding back. The keybase/GPG key rotation that came with v7 also got back ported to v5.5.5. The upload path is identical, so tokenless reliability is the same across v5, v6, and v7.

If anything the edge is with v7 because of the VULN hardening. I just wanted to offer another option that solves the primary issue we face with v3.1.5 but without going "bleeding edge" (if you'd call it that)

ljharb (approved):

PR #3264, opened 2026-06-30
[New] add `requireResolve` option to check `require.resolve()` paths

### LINKED ISSUE #585: Support for checking `require.resolve()`
Sometimes I use `require.resolve()` just to resolve paths to other resources than javascript modules.

Okay, this is not a regular _import_, but it would be nice if this plugin could also check these paths.

Maybe it could be a new setting like `'import/check-require-resolve': true`, whereas various rules related to CommonJS imports also apply here.

manzoorwanijk:
Closes #585. Supersedes #1217.

## Motivation

While working on a [PR in Gutenberg](https://github.com/WordPress/gutenberg/pull/79703#issuecomment-4846406458) (which powers the WordPress block editor), I noticed that `require.resolve()` calls are not considered by these rules. Looking for existing discussion I found #585, so this PR implements it.

## Summary

`require.resolve()` resolves a module path the same way `require()` does, but until now the resolution rules ignored it. This PR adds an opt-in **`requireResolve`** boolean option (default `false`) to the shared `moduleVisitor`. When enabled, the first string-literal argument of a `require.resolve("<path>")` call is visited as a `require` module path, so the rules that resolve specifiers can check it too.

```js
/*eslint import/no-unresolved: [2, { requireResolve: true }]*/
const x = require.resolve('./foo'); // reported if './foo' is not found
require.resolve('./foo', { paths: [dir] }); // ignored (`paths` changes the resolution base)

require.resolve(0); // ignored (non-string)
require['resolve']('./foo'); // ignored (computed access)
```

## Relation to #1217

This supersedes the earlier #1217, which also added a `requireResolve` option but stalled in review. The main differences here: the option is a plain boolean independent of `commonjs` (instead of the `boolean | { commonjs }` `oneOf`), the 2-arg `{ paths }` form is ignored since resolvers cannot honor a custom resolution base, `no-cycle` explicitly ignores `require.resolve`, and the option is wired into several additional resolution rules — all with tests and docs.

## Why opt-in (off by default)

This mirrors the conclusion of the earlier attempts (#1216 / #1217): emitting new warnings on existing code is a breaking change, and `require.resolve` is sometimes used to resolve non-module assets (e.g. build-generated files) that a resolver can't find. The option therefore defaults to `false` and is independent of `commonjs`.

> Note on `require.resolve(0)`: like `require()` and `import()`, the visitor only forwards static string paths, so non-string / dynamic arguments are skipped rather than reported. (In the #1217 review @ljharb noted a warning there could be useful; that is arguably a separate "invalid argument type" concern and is intentionally left out of scope here.)

## Affected rules

The change lives in `moduleVisitor`, so every resolution rule can opt in:

- **Enabled via the shared options schema** (no rule change needed): [`no-unresolved`], [`no-absolute-path`], [`no-relative-packages`], [`no-relative-parent-imports`].
- **[`no-cycle`]**: explicitly **ignores** `require.resolve` — it computes a path without loading the module, so it cannot form a runtime cycle. This prevents false cycle reports when the option is enabled elsewhere.
- **Explicitly wired** (these have their own schema and/or hardcode the visitor options): [`no-extraneous-dependencies`], [`no-self-import`], [`no-webpack-loader-syntax`], [`no-useless-path-segments`], [`no-internal-modules`].

Deliberately **not** wired: `max-dependencies` (a resolved path is not a load-time dependency edge), `extensions` (`require.resolve` is often used specifically to resolve paths _with_ an extension), and `no-nodejs-modules` (marginal value).

## Behaviour / edge cases

- The 2-arg `require.resolve(path, { paths })` form is ignored: `paths` overrides the resolution base, which resolvers cannot honor, so checking the first argument would produce false positives. (This deviates from the "check the first argument in every case" suggestion in the #1217 review, which predates the `paths` semantics being raised.)
- Only non-computed `require.resolve(...)` member calls match; `require['resolve'](...)`, `require[resolve](...)`, and `foo.require.resolve(...)` are ignored.
- Non-string / dynamic first arguments are ignored.
- The visitor reports with `moduleSystem: 'require'`, so resolver `moduleSystem` configuration applies consistently with `require()`.

## Checklist

- [x] write tests — valid (off-by-default) and invalid (option-on) cases for every wired rule, across the existing parser matrix where applicable
- [x] implement feature
- [x] update docs (`docs/rules/*`)
- [x] make a note in the change log

## Testing

- `npm run tests-only` for the affected suites — all passing
- `eslint .`, `tsc --noEmit index.d.ts`, `npm run update:eslint-docs -- --check`, and `markdownlint` all clean

## AI tool disclosure

This change was prepared with AI assistance. The plan, implementation, tests, and docs were drafted with Claude Code, and the plan and final diff were independently reviewed with Codex. All output was reviewed by me before submission, and I take responsibility for the contents of this PR.

### DIFF
--- a/docs/rules/no-absolute-path.md
+++ b/docs/rules/no-absolute-path.md
@@ -41,6 +41,7 @@ You may provide an options object providing true/false for any of
  - `esmodule`: defaults to `true`
  - `commonjs`: defaults to `true`
  - `amd`: defaults to `false`
+ - `requireResolve`: defaults to `false`
 
 If `{ amd: true }` is provided, dependency paths for AMD-style `define` and `require`
 calls will be resolved:
--- a/docs/rules/no-extraneous-dependencies.md
+++ b/docs/rules/no-extraneous-dependencies.md
@@ -40,6 +40,12 @@ There are 2 boolean options to opt into checking extra imports that are normally
 "import/no-extraneous-dependencies": ["error", {"includeInternal": true, "includeTypes": true}]
 ```
 
+The `requireResolve` boolean option (off by default) additionally checks the path argument of `require.resolve()` calls:
+
+```js
+"import/no-extraneous-dependencies": ["error", {"requireResolve": true}]
+```
+
 Also there is one more option called `packageDir`, this option is to specify the path to the folder containing package.json.
 
 If provided as a relative path string, will be computed relative to the current working directory at linter execution time. If this is not ideal (does not work with some editor integrations), consider using `__dirname` to provide a path relative to your configuration.
--- a/docs/rules/no-internal-modules.md
+++ b/docs/rules/no-internal-modules.md
@@ -11,6 +11,8 @@ This rule has two mutally exclusive options that are arrays of [minimatch/glob p
  - `allow` that include paths and import statements that can be imported with reaching.
  - `forbid` that exclude paths and import statements that can be imported with reaching.
 
+It also accepts a `requireResolve` boolean option (default `false`) that, when enabled, checks the path argument of `require.resolve()` calls.
+
 ### Examples
 
 Given the following folder structure:
--- a/docs/rules/no-self-import.md
+++ b/docs/rules/no-self-import.md
@@ -30,3 +30,13 @@ import bar from './bar';
 
 const bar = require('./bar');
 ```
+
+## Options
+
+### requireResolve
+
+When set to `true`, this rule also checks the path argument of `require.resolve()` calls. Defaults to `false`.
+
+```js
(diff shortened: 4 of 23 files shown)


manzoorwanijk:
@ljharb, a gentle request: is it possible to wrap this up before the next release?

manzoorwanijk:
Some CI failures are due to git being blocked by npm v12 by default - https://github.com/import-js/eslint-plugin-import/actions/runs/29027018826/job/86149376901?pr=3264

manzoorwanijk:
@ljharb what do you think about this?

PR #3266, opened 2026-07-13
[Deps] let user know `eslint-import-resolver-node` needs to be in root of node_modules (important when using `install-strategy=linked`)

nikolawork:
Using `install-strategy=linked` only puts the packages from `dependencies` and `devDependencies` in the root of node_modules. This means that `eslint-import-resolver-node` is not in the root of node_modules which leads to the following error:

```
Resolve error: unable to load resolver "node"
```

This is fixed by adding `eslint-import-resolver-node` to your `package.json`. We signal this to the user by adding `eslint-import-resolver-node` to `peerDependencies`.

`install-strategy=linked` is stable since `npm@11.18.0`: https://docs.npmjs.com/cli/v11/using-npm/config#install-strategy

npm even recommends using it when developing. From the above link:

> We recommend that package authors use `--install-strategy=linked` during development to catch undeclared ("phantom") dependencies before publishing: the isolated layout only exposes a package's declared dependencies, so an `import` of a package that was never added to `package.json` can fail instead of resolving by accident and shipping broken. See [Catching undeclared ("phantom") dependencies](https://docs.npmjs.com/cli/v11/using-npm/developers#catching-undeclared-phantom-dependencies).

There is a closed (but unresolved) issue for this: #828

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,8 @@
     "typescript-eslint-parser": "^15 || ^20 || ^22"
   },
   "peerDependencies": {
-    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8 || ^9 || ^10"
+    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8 || ^9 || ^10",
+    "eslint-import-resolver-node": "^0.4.0"
   },
   "dependencies": {
     "@rtsao/scc": "^1.1.0",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 ## [Unreleased]
 
 ### Added
+- [Deps] let user know `eslint-import-resolver-node` needs to be in root of node_modules (important when using `install-strategy=linked`) ([#3266], thanks [@nikolawork])
 - support eslint v10 ([#3230], thanks [@rasmi])
 - [`no-deprecated`]: detect `@deprecated` on default-exported identifier declarations ([#3247], thanks [@mixelburg] [@etyrrell22])
 - [`consistent-type-specifier-style`]: add `prefer-top-level-if-only-type-imports` option ([#3210], thanks [@aldeed])


nikolawork:
This fix can also be added to the fix in #2283 and then we can close this PR.

manzoorwanijk:
I see the dependency is already being declared correctly.

https://github.com/import-js/eslint-plugin-import/blob/bf2d2aab796116103ca2b9fbefa7dc11ab2dae88/package.json#L125

nikolawork:
> I see the dependency is already being declared correctly.

The problem is the way the dependency is being resolved here:

https://github.com/import-js/eslint-plugin-import/blob/bf2d2aab796116103ca2b9fbefa7dc11ab2dae88/utils/resolve.js#L101-L103

The only way this works is if `eslint-import-resolver-node` is in the root of `node_modules`. Which it isn't when using `install-strategy=linked` if the dependency is not declared in the end user's `package.json` (which you should definitely know haha).

manzoorwanijk:
> The problem is the way the dependency is being resolved here:

Yes, that is what I was going to write. That require logic may need to be updated to ensure it works in isolated dependencies layout.

ljharb:
This is intentional, and is why the "linked" strategy (and yarn's PNP, and similar approaches) is fundamentally flawed.

Adding it to peer deps explicitly increases maintenance burden and could be a breaking change.

manzoorwanijk:
I have created a dedicated bug report for it - #3268 

Also created a PR to fix it - #3269

manzoorwanijk:
But the CI is broken because of npm v12.

ljharb:
the npm 12 thing is fixed, if you rebase.

package.json
```diff
@@ -112,7 +112,8 @@
     "typescript-eslint-parser": "^15 || ^20 || ^22"
   },
   "peerDependencies": {
-    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8 || ^9 || ^10"
+    "eslint": "^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8 || ^9 || ^10",
+    "eslint-import-resolver-node": "^0.4.0"
```
ljharb, line 116: note that this is a breaking change, because it will fail if someone has a newer version installed at a higher level.

manzoorwanijk:
> the npm 12 thing is fixed, if you rebase.

Your commit seems to have used another failure - https://github.com/import-js/eslint-plugin-import/actions/runs/29535690274/job/87764990067

nikolawork:
> I have created a dedicated bug report for it - #3268
> 
> Also created a PR to fix it - #3269

Thanks for the more thorough fix, @manzoorwanijk. If that PR is merged, we can close my PR.

PR #3269, opened 2026-07-16
[Fix] load the default `node` resolver under an isolated `node_modules` layout

### LINKED ISSUE #3268: [BUG] `Resolve error: unable to load resolver "node"` under npm's isolated install layout (`--install-strategy=linked`)
Installing with npm's isolated layout breaks the plugin's **default** resolver. No custom resolver is configured, and every import — including core modules — is reported unresolved.

This is the "npm isolated mode" case anticipated in [#2283](https://github.com/import-js/eslint-plugin-import/pull/2283) (details at the bottom).

## Minimal reproducible example

Four files, stock config:

`.npmrc`

```ini
install-strategy=linked
```

`package.json`

```json
{
  "type": "module",
  "scripts": {
    "lint": "eslint src"
  },
  "devDependencies": {
    "eslint": "^9.39.5",
    "eslint-plugin-import": "^2.32.0"
  }
}
```

`eslint.config.mjs`

```js
import importPlugin from "eslint-plugin-import";

export default [
  {
    files: ["src/**/*.js"],
    languageOptions: { ecmaVersion: 2022, sourceType: "module" },
    plugins: { import: importPlugin },
    rules: { "import/no-unresolved": "error" },
  },
];
```

`src/index.js`

```js
import path from "path";

export default path;
```

## Expected

No errors. `path` is a core module and the config is stock.

## Actual

```
src/index.js
  1:1   error  Resolve error: unable to load resolver "node"  import/no-unresolved
  1:18  error  Unable to resolve path to module 'path'        import/no-unresolved

✖ 2 problems (2 errors, 0 warnings)
```

The second error is a cascade of the first, not a separate issue: `requireResolver` throws, `resolve()` catches it and returns `undefined`, and `no-unresolved` treats `undefined` as "not found". Because core-module classification lives inside the resolver that never loaded, **every** import is reported — builtins included. `Resolve error` is deduped once per context via `erroredContexts`, while `Unable to resolve` fires once per import, so a real project gets one cause line and an error on every import in every file.

## Debug output

`DEBUG=eslint-plugin-import:*` produces **no output** for this failure — `utils/resolve.js` contains no `debug` calls, so the resolver-loading path is entirely unlogged. Widening the namespace yields only:

```
$ DEBUG='eslint-plugin-import:*,eslint-module-utils:*' npx eslint src
2026-07-16T10:37:59.295Z eslint-module-utils:ModuleCache cache miss for .../src44136fa...path
```

## Cause

`eslint-import-resolver-node` is a dependency of **`eslint-plugin-import`**, but the code that loads it — `tryRequire` in **`eslint-module-utils/resolve.js`** — lives in a different package that does not declare it. `tryRequire` gets two chances, and under an isolated layout both miss:

| require anchored at                             | finds `eslint-import-resolver-node`?       |
| ----------------------------------------------- | ------------------------------------------ |
| the linted source file (`src/index.js`)         | no — the app doesn't depend on it          |
| `eslint-module-utils/resolve.js` (the fallback) | no — utils doesn't declare it              |
| `eslint-plugin-import/lib/index.js`             | **yes** — it's the plugin's own dependency |

Verified directly against the repro's tree:

```
FAIL sourceFile (src/index.js)  [tryRequire arg 2] -> MODULE_NOT_FOUND
FAIL resolve.js itself          [tryRequire fallback] -> MODULE_NOT_FOUND
OK   plugin lib/index.js -> .../eslint-import-resolver-node/index.js
```

A hoisted layout flattens every package into one top-level `node_modules`, so any package can require any other and the missing declaration is invisible. `--install-strategy=linked` gives each package only its declared dependencies, which exposes it.

## Why this has not been reported via pnpm or yarn

They patch it at install time. Yarn maintains a compatibility database of packages with undeclared dependencies ([`@yarnpkg/extensions`](https://github.com/yarnpkg/berry/blob/0a230c14e71247576f6b51fa811ae08edb6608aa/packages/yarnpkg-extensions/sources/index.ts#L745)); pnpm bundles the same list. It contains this entry, which cites this very PR:

```js
// https://github.com/import-js/eslint-plugin-import/pull/2283
[`eslint-module-utils@*`, {
  peerDependenciesMeta: {
    "eslint-import-resolver-node": optionalPeerDep,
    "eslint-import-resolver-typescript": optionalPeerDep,
    "eslint-import-resolver-webpack": optionalPeerDep,
    "@typescript-eslint/parser": optionalPeerDep
  }
}],
```

Those peer deps do not exist in the published `eslint-module-utils@2.14.0` package.json (`peerDependencies` is `undefined`); pnpm injects them during resolution, so the resolver is symlinked into `eslint-module-utils`'s own `node_modules`. This is not about hoisting — with `hoistPattern: []` (no `.pnpm/node_modules` at all) pnpm still passes. **pnpm and yarn cannot reproduce this**; npm has no such database, so it surfaces only there.

## Relationship to #2283

[#2283](https://github.com/import-js/eslint-plugin-import/pull/2283) (Oct 2021, still open) proposed declaring these as optional peer dependencies and was rejected, with the note:

> For one, anything npm doesn't do isn't something I think needs to be catered to - npm is the standard. That said, npm will be adding a pnpm-like "isolated mode" soon, at which point I'd be obliged to address it somehow.

npm's isolated mode now exists — `--install-strategy=linked` is it, and this report is that case.

Two points distinguish this from the argument in that thread:

1. **This is not a consumer-config problem.** The rejection reasoning was that `eslint-module-utils` dynamically requires resolvers based on config, so consumers must install what their config references. Here nothing references a resolver: `node` is the default, and the plugin already ships `eslint-import-resolver-node` as its own dependency.
2. **A fix need not involve peer dependencies at all,** avoiding the objection that peer deps aren't for declaring "allowed dependencies to access". Since `eslint-plugin-import` is the package that actually depends on the resolver, it can pass its own `require` (or a resolved path) down to `tryRequire`, so the lookup is anchored at the package that owns the dependency.

Happy to open a PR along the lines of (2) if that direction is agreeable.

## Environment

|                             |              |
| --------------------------- | ------------ |
| eslint-plugin-import        | 2.32.0       |
| eslint-module-utils         | 2.14.0       |
| eslint-import-resolver-node | 0.3.10       |
| eslint                      | 9.39.5       |
| node                        | 24.17.0      |
| npm                         | 11.18.0      |
| OS                          | macOS 26.5.2 |

(ESLint 10 is untested here — `eslint-plugin-import@2.32.0`'s peer range stops at `^9`.)

manzoorwanijk:
Fixes #3268.

Under an isolated `node_modules` layout (`npm install --install-strategy=linked`), a stock config fails with `Resolve error: unable to load resolver "node"`, and every import - core modules included - is then reported unresolved.

`eslint-import-resolver-node` is our dependency, but it is `require`d from `eslint-module-utils`, which does not declare it. A hoisted layout hides that, since every package can see every other; an isolated one does not. pnpm and yarn paper over it via Yarn's [compat DB](https://github.com/yarnpkg/berry/blob/0a230c14e71247576f6b51fa811ae08edb6608aa/packages/yarnpkg-extensions/sources/index.ts#L745), which injects optional peer deps citing #2283 - npm has no such thing, so it surfaces there.

## Approach

`eslint-module-utils/resolve` gains `registerResolverRequire`; we lend it our own `require`, so the resolver is loaded from the package that actually depends on it. Passing a `require` rather than a path anchors the lookup at this package whatever the layout, so it does not rely on the compat DB continuing to cover for us - if that entry ever goes away, pnpm and yarn are already handled.

No peer dependencies involved, per #2283. `eslint-module-utils` stays generic - it learns no consumer's name.

## Backwards compatibility

Registered requires are consulted only after every existing lookup has failed, so resolution is bit-for-bit unchanged wherever it already succeeds; only previously-failing cases can start resolving. Relative and absolute names skip them entirely, as those belong to the source file's package. Load errors propagate rather than being swallowed, matching the existing lookups.

Two notes:

- The call is guarded with `typeof registerResolverRequire === 'function'`, since `eslint-module-utils` is released separately and our declared range still admits versions that predate this API. It can go once the range is bumped.
- Needs an `eslint-module-utils` release plus the usual `[Drps]` bump before it takes effect for consumers.

## Verification

Against a real `--install-strategy=linked` repro, lint goes from the error above to clean, with `eslint-module-utils` still genuinely unable to see the resolver (`MODULE_NOT_FOUND`) - and a bogus import is still reported, so the resolver is doing real work. Also verified where the plugin is a transitive dep of a shared config, which the app itself does not declare. Under yarn PnP the compat DB means the bug does not arise; checked that this changes nothing there.

Five tests added, each confirmed to fail without its fix. Full suite green.

Out of scope: a third-party resolver declared only by a shared config (not by the app or by us) is still unreachable under an isolated layout - nothing in the chain can see it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - export a top-level `meta` object, matching the declared types, and strengthen the type tests ([#3169], thanks [@ljharb])
 
 ### Fixed
+- load the default `node` resolver under an isolated `node_modules` layout, where `eslint-module-utils` cannot see our dependencies ([#3269], thanks [@manzoorwanijk])
 - [`no-duplicates`]: fix `prefer-inline` autofix producing invalid syntax when an identifier named `from` is imported ([#3236], thanks [@DukeDeSouth] [@Quantaly])
 - [`no-duplicates`]: avoid false positives for TypeScript namespace and default type imports that cannot be merged ([#3195], thanks [@sjh9714] [@robyoder])
 - ExportMap: resolve export * as ns re-exports under modern parsers ([#3250], thanks [@rasmi] [@JounQin] [@butterybread])
@@ -1214,6 +1215,7 @@ for info on changes for earlier releases.
 [`memo-parser`]: ./memo-parser/README.md
 
 [#3284]: https://github.com/import-js/eslint-plugin-import/pull/3284
+[#3269]: https://github.com/import-js/eslint-plugin-import/pull/3269
 [#3283]: https://github.com/import-js/eslint-plugin-import/pull/3283
 [#3282]: https://github.com/import-js/eslint-plugin-import/pull/3282
 [#3276]: https://github.com/import-js/eslint-plugin-import/pull/3276
@@ -2035,6 +2037,7 @@ for info on changes for earlier releases.
 [@malykhinvi]: https://github.com/malykhinvi
 [@manovotny]: https://github.com/manovotny
 [@manuth]: https://github.com/manuth
+[@manzoorwanijk]: https://github.com/manzoorwanijk
 [@marcusdarmstrong]: https://github.com/marcusdarmstrong
 [@mastilver]: https://github.com/mastilver
 [@mathieudutour]: https://github.com/mathieudutour
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,14 @@
+import { registerResolverRequire } from 'eslint-module-utils/resolve';
+
 import { name, version } from '../package.json';
+
+/**
+ * Lend our `require` to `eslint-module-utils`, which can not see our
+ * `eslint-import-resolver-node` under an isolated `node_modules` layout.
+ */
+if (typeof registerResolverRequire === 'function') { // older versions lack this API
+  registerResolverRequire(require);
+}
 
 export const rules = {
   'no-unresolved': require('./rules/no-unresolved'),
--- a/tests/src/core/resolve.js
+++ b/tests/src/core/resolve.js
@@ -184,6 +184,17 @@ describe('resolve', function () {
       expect(registered.requested).to.have.lengthOf(0);
     });
 
+    it('registers each `require` once, and ignores anything else', function () {
+      const registered = fakeRequire({});
+      registerResolverRequire(registered);
+      registerResolverRequire(registered);
+      registerResolverRequire(null);
+      registerResolverRequire('not a require');
+
+      expect(reportsFor({ 'import/resolver': 'nope-not-here' }).resolved).to.equal(undefined);
+      expect(registered.requested).to.eql(['eslint-import-resolver-nope-not-here', 'nope-not-here']);
+    });
+
     it('surfaces a load exception from a registered resolver', function () {
       registerResolverRequire(fakeRequire({
         'eslint-import-resolver-throws-on-load'() { throw new SyntaxError('TEST SYNTAX ERROR'); },
--- a/tests/src/core/resolve.js
+++ b/tests/src/core/resolve.js
@@ -2,7 +2,7 @@ import { expect } from 'chai';
 import eslintPkg from 'eslint/package.json';
 import semver from 'semver';
 
-import resolve, { CASE_SENSITIVE_FS, fileExistsWithCaseSync } from 'eslint-module-utils/resolve';
+import resolve, { CASE_SENSITIVE_FS, fileExistsWithCaseSync, registerResolverRequire } from 'eslint-module-utils/resolve';
 
 import * as path from 'path';
 import * as fs from 'fs';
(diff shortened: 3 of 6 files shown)


ljharb:
This is clever, but maybe another way would be requiring the node resolver based on a fully resolved path from the location of eslint-plugin-import itself? meaning, don't expand the API of the node resolver.

manzoorwanijk:
> but maybe another way would be requiring the node resolver based on a fully resolved path from the location of eslint-plugin-import itself? meaning, don't expand the API of the node resolver.

That'd be simpler, and I did think about that in the beginning, but I it only covers the case where the linted project depends on `eslint-plugin-import` directly.

The plugin is very often a transitive dep of a shared config (app → `eslint-config-foo` → `eslint-plugin-import`). For example, we [use it as such in Gutenberg](https://github.com/WordPress/gutenberg/blob/9faa7de501a61628a90b2fddc2219c8798adc5df/packages/eslint-plugin/package.json#L53) - the monorepo that powers the WordPress block editor.
Under an isolated layout, it's then absent from the app's `node_modules`, so there's nothing to anchor to from the source file. I built a repro to check that and I could confirm that `eslint-plugin-import/package.json` resolves to `MODULE_NOT_FOUND` from the linted file, so the resolver stays unreachable and the error is unchanged. Lending the plugin's own `require` works there because it doesn't care where the plugin lives - and it's guaranteed to be the copy actually running, rather than whichever one the source file happens to see.

manzoorwanijk:
This is now ready and CI is happy. I will update from the main branch and resolve conflicts once approved.

PR #3270, opened 2026-07-16
[Docs] clarify that the `typescript` config (legacy and flat) does not set up the TypeScript resolver

### LINKED ISSUE #3170: Missing typescript setting in `flatConfigs.typescript`
With the recent official support of ESLint TypeScript config ([ref](https://eslint.org/docs/latest/use/configure/configuration-files#typescript-configuration-files)), when trying to setup the [example config in the README file](https://github.com/import-js/eslint-plugin-import?tab=readme-ov-file#config---flat-with-config-in-typescript-eslint), I get the following issue:

`Unable to resolve path to module 'typescript-eslint'.eslint[import/no-unresolved](https://github.com/import-js/eslint-plugin-import/blob/v2.31.0/docs/rules/no-unresolved.md)`

I managed to fix this by adding this extra setting:

```js
    settings: {
      'import/resolver': {
        typescript: true,
      },
    },
```

When I look at what settings `flatConfigs.typescript` has, it only has the following settings:

```js
    'import/resolver': {
      node: {
        extensions: allExtensions,
      },
    },
```

I presume we should add `typescript: true` in `flatConfigs.typescript` or otherwise maybe mentionning this somewhere in the doc? The only way I was able to fix this was because I had this in a previous config.

KAMRONBEK:
## Problem

Users extending `importPlugin.flatConfigs.typescript` (or `plugin:import/typescript`) assume TypeScript-aware resolution works out of the box, then hit `Unable to resolve path to module 'typescript-eslint'` (`import/no-unresolved`) or `Resolve error: typescript with invalid interface loaded as resolver`. The README's flat-config example (`### Config - Flat with config() in typescript-eslint`) — the snippet the issue author copied — never mentions the resolver at all.

## Root cause

The `typescript` config only adds the TS extensions and parser settings; its `import/resolver` setting is just `node`. It does not enable the TypeScript resolver, because `eslint-import-resolver-typescript` is a separate package that is not a dependency of this plugin (shipping `typescript: true` in the shared config would break every user who doesn't have it installed — the "invalid interface loaded as resolver" error). Per the maintainer answers in the thread, the resolver must be installed manually and enabled via the `import/resolver` setting; imports resolved through `tsconfig.json` `paths` or a package's `exports` field (like `typescript-eslint` itself, since this plugin doesn't support `exports` yet) fail without it.

## Fix (docs only)

- **README `## TypeScript`**: state explicitly that `plugin:import/typescript` / `flatConfigs.typescript` does not set up the TypeScript resolver, and that `eslint-import-resolver-typescript` must be installed and enabled via `import/resolver`.
- **README `### Config - Flat with config() in typescript-eslint`**: add the same note and extend the minimal working example with the missing `settings: { 'import/resolver': { typescript: true, node: true } }` block (mirroring the legacy jsonc example above it, including its install/configure comment).
- **`config/typescript.js`** (backs both `configs.typescript` and `flatConfigs.typescript` via `src/index.js`): extend the header comment with the same clarification and a link to the resolver's configuration docs.
- **CHANGELOG.md**: entry under Unreleased → Changed, following the existing `[Docs]` format.

## Tests

Docs-only; no unit tests. `markdownlint "**/*.md"` (repo config) and `eslint config/typescript.js` (repo .eslintrc) both pass.

Fixes #3170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [Tests] cover ESLint 10 `getTokenOrComment` and `listFilesToProcess` API fallbacks ([#3230], thanks [@captaindonald])
 - [Docs] `no-unused-modules`: Fix docs of `ignoreUnusedTypeExports` option ([#3233], thanks [@ej612])
 - [actions] update `codecov/codecov-action` to v7, for reliable tokenless coverage uploads from fork PRs ([#3262], thanks [@captaindonald])
+- [Docs] clarify that the `typescript` config (legacy and flat) does not set up the TypeScript resolver ([#3170], thanks [@KAMRONBEK])
 
 ## [2.32.0] - 2025-06-20
 
@@ -1220,6 +1221,7 @@ for info on changes for earlier releases.
 [#3191]: https://github.com/import-js/eslint-plugin-import/pull/3191
 [#3173]: https://github.com/import-js/eslint-plugin-import/pull/3173
 [#3172]: https://github.com/import-js/eslint-plugin-import/pull/3172
+[#3170]: https://github.com/import-js/eslint-plugin-import/issues/3170
 [#3167]: https://github.com/import-js/eslint-plugin-import/pull/3167
 [#3166]: https://github.com/import-js/eslint-plugin-import/pull/3166
 [#3152]: https://github.com/import-js/eslint-plugin-import/pull/3152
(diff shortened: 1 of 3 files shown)


KAMRONBEK:
@ljharb TL;DR (docs-only): per the maintainer answers in #3170, `plugin:import/typescript` / `flatConfigs.typescript` only adds TS extensions + parser settings and cannot ship the TypeScript resolver, but the README never says so — and the flat-config example (the snippet the issue author copied) omits the `import/resolver` settings entirely, so users hit `import/no-unresolved` on `typescript-eslint` itself. This states the requirement in both README TypeScript sections, adds the missing `settings: { 'import/resolver': { typescript: true, node: true } }` block to the flat example (mirroring the legacy jsonc example), notes it in the `config/typescript.js` header comment, and adds the CHANGELOG entry in the existing `[Docs]` format.

ljharb (approved):

PR #3271, opened 2026-07-16
[Fix] correct `flatConfigs` types so flat-config usage type-checks

### LINKED ISSUE #3169: Invalid `flatConfigs` types
With the recent official support of ESLint TypeScript config ([ref](https://eslint.org/docs/latest/use/configure/configuration-files#typescript-configuration-files)), when trying to setup the [example config in the README file](https://github.com/import-js/eslint-plugin-import?tab=readme-ov-file#config---flat-with-config-in-typescript-eslint), I get the following typings issue:

`'importPlugin.flatConfigs' is possibly 'undefined'.`

I did the following temporary workaround for now which is not ideal:

```ts
import tseslint from 'typescript-eslint'
import * as rawImportPlugin from 'eslint-plugin-import'
import js from '@eslint/js'

const importPlugin = rawImportPlugin as typeof rawImportPlugin & {
  flatConfigs: NonNullable<typeof rawImportPlugin.flatConfigs>
};

export default tseslint.config(
  js.configs.recommended,
  // other configs...
  {
    files: ['**/*.{ts,tsx}'],
    extends: [importPlugin.flatConfigs.recommended, importPlugin.flatConfigs.typescript],
    // other configs...
  }
)
```

Also it would be nice to have proper type for `importPlugin.flatConfigs.recommended` as it currently uses `any`

KAMRONBEK:
**Problem**

#3169 reports that following the README's TypeScript flat-config example yields `'importPlugin.flatConfigs' is possibly 'undefined'` and `any`-typed configs.

**Root cause / investigation**

The issue was filed 2025-03-22, when the latest release was v2.31.0 — which shipped **no types at all** (the `index.d.ts` added in #3097 first shipped in v2.32.0). That explains the reported errors and the reporter's `declare module 'eslint-plugin-import'` workaround (an ambient override only takes effect when the package ships no declarations).

I verified the shipped 2.32.0 `index.d.ts` against the current ecosystem: both README flat-config examples (`export default [...]` and `tseslint.config(...)`), plus `defineConfig(...)` and `plugins: { import: importPlugin }` usage, type-check cleanly with `tsc --strict` (`skipLibCheck: false`) against eslint 9.39.5 and 10.7.0, TypeScript 5.9.3 and 7.0.2, so the originally-reported errors are already resolved on `main`.

One genuine mismatch remains: `index.d.ts` declares `flatConfigs['stage-0']`, but the runtime `flatConfigs` object in `src/index.js` has no `stage-0` key (it exists only in the legacy `configs`). TypeScript reports a defined `Linter.FlatConfig` for a value that is `undefined` at runtime, so spreading it into a config crashes despite a clean type-check.

**Fix**

- Remove `'stage-0'` from the `flatConfigs` declaration in `index.d.ts` (legacy `configs['stage-0']` is kept — it exists at runtime).
- Add a type-test suite at `tests/types/` (compiled by `npm run test-types` alongside the existing bare `index.d.ts` check) asserting: `flatConfigs`/`configs` and every entry are non-optional and not `any`; every flat config is assignable to `Linter.FlatConfig` and every legacy config to `Linter.LegacyConfig`; the plugin object is assignable to `ESLint.Plugin`; and `flatConfigs['stage-0']` stays a type error (`@ts-expect-error`).

**Tests**

`tsc -p tests/types` passes with the fix and fails (TS2578 unused `@ts-expect-error`) with the `stage-0` removal reverted — verified under both TypeScript 5.9.3 and 7.0.2 (the latter matching CI's `typescript@latest`). The tests/types tsconfig avoids options removed in TS 7 (`baseUrl`, `moduleResolution: node10`), so `test-types` keeps working as CI's `typescript@latest` advances.

**Independence from #3213**

#3213 changes runtime exports in `src/index.js` (default-export identity / `meta`) and does not touch `index.d.ts`; its new package test explicitly expects `stage-0` to be absent from flat configs, consistent with this change. The two merge cleanly in either order.

Fixes #3169.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`no-deprecated`], [`namespace`]: route `declaredScope` through the `getScope` compat shim instead of the removed `context.getScope()` (ESLint 9+) ([#3230], thanks [@ljharb])
 - [`newline-after-import`]: ignore comments between consecutive imports when `considerComments` is enabled ([#2673], thanks [@raisulchowdhury])
 - [`no-unresolved`], [`no-useless-path-segments`]: handle template literal requires ([#3276], thanks [@pratyushsinghal7])
+- [types] remove `stage-0` from the `flatConfigs` types, since it does not exist at runtime, and add type tests for the plugin's exports ([#3169], thanks [@KAMRONBEK])
 
 ### Changed
 - [Refactor] [`order`]: extract the ESLint 10 token/comment compatibility shims into a reusable `getTokenOrComment` util ([#3230], thanks [@captaindonald])
@@ -1228,6 +1229,7 @@ for info on changes for earlier releases.
 [#3191]: https://github.com/import-js/eslint-plugin-import/pull/3191
 [#3173]: https://github.com/import-js/eslint-plugin-import/pull/3173
 [#3172]: https://github.com/import-js/eslint-plugin-import/pull/3172
+[#3169]: https://github.com/import-js/eslint-plugin-import/issues/3169
 [#3167]: https://github.com/import-js/eslint-plugin-import/pull/3167
 [#3166]: https://github.com/import-js/eslint-plugin-import/pull/3166
 [#3152]: https://github.com/import-js/eslint-plugin-import/pull/3152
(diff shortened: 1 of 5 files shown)


KAMRONBEK:
@ljharb TL;DR for #3169: the shipped root `index.d.ts` leaves `flatConfigs` possibly-undefined/any-typed, so the README's own TS flat-config example fails to type-check ('importPlugin.flatConfigs' is possibly 'undefined'). This fixes the declarations so `flatConfigs.recommended`/`typescript`/etc. are non-optional, properly typed `Linter.Config` values, with a tsc probe reproducing the reported errors before the fix. Independent of #3213 (which touches src/index.js default-export identity, not the declarations). CHANGELOG entry included.

ljharb (approved):

Issue #1281, opened 2019-02-06
WIldcard support in `import/core-modules`


penx:
## Feature request:

Allow wildcards in the import/core-modules settings

```json
      "settings": {
        "import/core-modules": [
          '@my-monorepo-namespace/*',
        ]
      }
```

... so that

```
        "import/no-extraneous-dependencies": ["error", {"devDependencies": true}]
```

...does not throw an error for modules within a certain namespace.

## Why?

I have a monorepo structure where we can safely assume that anything within a certain namespace is available and doesn't need to be listed in devDependencies. I would rather not maintain lists of packages and versions manually.

ljharb:
It'd only be names, not versions; how many packages do you expect to have under the namespace that listing them manually is difficult?

(Also note, you can always dynamically create a config that builds up the list of names for you)

penx:
We have [one package per React component](https://github.com/govuk-react/govuk-react). There are reasons we don't want to maintain a list beyond maintenance/size of the list - we don't want these libraries in devDependencies due to causing lerna publishing issues, and this error would encourage a developer of a new component to add the dependency in to devDependencies rather than edit the list in eslint config. I'd rather there wasn't an eslint config maintenance step when a new component is added.

Dynamically generating a list would probably be a better option for us, which I had considered but I was concerned about reading package names from many package.json files on disk whenever eslintrs.js is loaded, and whether this would have a noticeable effect on some IDEs/plugins (I haven't looked in to caching/how often IDEs would load from disk).

I figured wildcard support would be an easy way to solve this and may be a useful feature to have regardless. We should be able to find another solution if not.

penx:
On a side note, we previously worked around our publishing issue by maintaining the list in the root package.json

```json
    "@govuk-react/error-text": "*",
	    "@govuk-react/file-upload": "*",
	    "@govuk-react/grid-col": "*",
	    "@govuk-react/grid-row": "*",
	    "@govuk-react/header": "*",
	    "@govuk-react/hidden-text": "*",
	
```

But then we came to do an alpha release, realised * doesn't map to -alpha and the whole system broke down 😄

https://github.com/govuk-react/govuk-react/pull/536/commits/a90540a7b2612ab6fa5e35970b9100d564666f8d

ljharb:
Thanks for explaining. I think I could get behind allowing a single asterisk at the end, to cover this use case.

davidbarral:
The same problem arises when having something like this:

```
// eslint config
[...]

PR #3275, opened 2026-07-19
[New] support wildcard patterns in `import/core-modules` setting

andymai:
## Summary

Addresses #1281.

Re-submission of #3274, which was auto-closed when its source fork was accidentally deleted. The commits here are identical to that PR's reviewed final state (head `3510938`), restored from `refs/pull/3274/head` — no code changes. #3274 was itself the post-review recreation of #3200; both prior threads are linked below so reviewers keep the full history.

Both review comments from #3200 were already addressed and carried into these commits:

- Uses `minimatch` (already a dependency) instead of dynamic regex construction, per [review feedback](https://github.com/import-js/eslint-plugin-import/pull/3200#discussion_r2216629001)
- The matching is inlined directly in `isBuiltIn` — no wrapper function, per [review feedback](https://github.com/import-js/eslint-plugin-import/pull/3200#discussion_r2218951225)

## Changes

- `src/core/importType.js`: patterns in `import/core-modules` containing `*` are matched with `minimatch` (exact matches still use `indexOf` first)
- Tests for basic wildcards (`@my-monorepo/*`), multiple wildcards (`@my-*/*`), mixed exact + wildcard lists, and subpaths of wildcard-matched modules
- `no-extraneous-dependencies` rule tests for wildcard core-modules
- README example of wildcard usage

## Test plan

- [x] All existing tests pass
- [x] New tests in `tests/src/core/importType.js` and `tests/src/rules/no-extraneous-dependencies.js`

## Notes

- Non-string entries in the setting are ignored (previously `indexOf` tolerated them; `pattern.includes` would have thrown), and minimatch's `!` negation / `#` comment semantics are disabled (`nonegate`/`nocomment`) so a pattern like `!@foo/*` can't accidentally match everything.
- **Known limitation**: `eslint-module-utils`' `resolve` still does an exact-match `Set.has()` on `import/core-modules` ([`utils/resolve.js`](https://github.com/import-js/eslint-plugin-import/blob/main/utils/resolve.js#L155)), so wildcard entries are honored by `importType`-based rules but not by resolver-based ones like `no-unresolved`. Happy to extend this to `eslint-module-utils` in this PR or a follow-up if that's the preferred direction.

<sub>Prior history: #3274 (auto-closed, fork deleted) · #3200 (review discussion)</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - support eslint v10 ([#3230], thanks [@rasmi])
 - [`no-deprecated`]: detect `@deprecated` on default-exported identifier declarations ([#3247], thanks [@mixelburg] [@etyrrell22])
 - [`consistent-type-specifier-style`]: add `prefer-top-level-if-only-type-imports` option ([#3210], thanks [@aldeed])
+- support wildcard patterns in the `import/core-modules` setting ([#3274], thanks [@andymai])
 
 ### Fixed
 - [`no-duplicates`]: fix `prefer-inline` autofix producing invalid syntax when an identifier named `from` is imported ([#3236], thanks [@DukeDeSouth] [@Quantaly])
@@ -1205,6 +1206,7 @@ for info on changes for earlier releases.
 
 [`memo-parser`]: ./memo-parser/README.md
 
+[#3274]: https://github.com/import-js/eslint-plugin-import/pull/3274
 [#3262]: https://github.com/import-js/eslint-plugin-import/pull/3262
 [#3250]: https://github.com/import-js/eslint-plugin-import/pull/3250
 [#3230]: https://github.com/import-js/eslint-plugin-import/pull/3230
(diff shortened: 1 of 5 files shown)


andymai:
A note on the red checks, since none of them stem from this change:

**`Tests: eslint (old versions)` (the `majors …` matrix) — pre-existing failure on `main`.** These jobs fail to install the ancient ESLint/Node combinations (`npm ERR! Invalid Version` / a `SyntaxError` during setup) before any test runs. The same workflow is red on the base commit this PR targets (`1a39fb8`): https://github.com/import-js/eslint-plugin-import/actions/runs/29535690274 — so it's independent of this diff. The functional suites (`Tests: eslint 8+`, `Tests: packages`, `Native and WSL`, `pretest/posttest`) are green.

**`codecov/patch` 0% is an artifact of that broken job.** Because the old-versions job crashes before executing tests, it reports 0% coverage for `src/core/importType.js` and uploads that, which drags the patch number to 0. Running the suite normally shows the changed file fully covered:

```
BABEL_ENV=test nyc mocha tests/src/core/importType.js
  42 passing
  File          | % Stmts | % Branch | % Funcs | % Lines
  importType.js |     100 |      100 |     100 |     100
```

The new wildcard branch in `isBuiltIn` is exercised by the added `importType(...)` specs (basic/multiple wildcards, subpaths, mixed exact+wildcard, non-string entries, and the `nonegate`/`nocomment` behavior). Happy to adjust if you'd prefer the old-versions matrix addressed separately.

PR #3276, opened 2026-07-24
[Fix] no-unresolved: handle template literal requires

### LINKED ISSUE #1575: `import/no-unresolved` ignores require statements with backticks
First of all, thanks for this plugin! It's great!

We're using the `backtick` option with the [`quotes` rule](https://eslint.org/docs/rules/quotes#options) in our codebase. About a week ago, we realized that our linter wasn't catching bad imports. After scratching our heads for a bit, we realized that `import/no-unresolved` doesn't check require statements that use backticks.

```javascript
// index.js
// Assume a, b, and c do not exist

// Does not result in a linting error
const sweetPotato = require(`a`);

// Shows linting error
const biscuit = require("b");

// Shows linting error
import babyYoda from "c";
```

pratyushsinghal7:
## What changed

Treat no-substitution template literals passed to CommonJS `require()` as static module paths in `moduleVisitor`.

The template's cooked string value is exposed on a shallow source-node copy, so existing resolution logic and visitor-based rules can consume it without mutating the parser AST. Template literals containing expressions remain dynamic and are ignored.

Regression coverage verifies that:

- ``require(`./bar`)`` resolves successfully.
- ``require(`./baz`)`` reports `import/no-unresolved`.
- Both the node and webpack resolver test configurations exercise the cases.

Fixes #1575.

## Validation

- `BABEL_ENV=test ./node_modules/.bin/mocha --require babel-register tests/src/rules/no-unresolved.js --reporter dot` — 157 passing
- `npm test` — full tests, ESLint, generated documentation checks, and Markdown lint passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - [`no-unused-modules`]: honor the flat config's global `ignores` in the `listFilesWithNodeFs` fallback, restoring parity with the file set ESLint lints ([#3230], thanks [@ljharb])
 - [`no-deprecated`], [`namespace`]: route `declaredScope` through the `getScope` compat shim instead of the removed `context.getScope()` (ESLint 9+) ([#3230], thanks [@ljharb])
 - [`newline-after-import`]: ignore comments between consecutive imports when `considerComments` is enabled ([#2673], thanks [@raisulchowdhury])
+- [`no-unresolved`], [`no-useless-path-segments`]: handle template literal requires ([#3276], thanks [@pratyushsinghal7])
 
 ### Changed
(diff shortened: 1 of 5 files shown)


pratyushsinghal7:
CI baseline note: the failing `Tests: eslint (old versions)` matrix is unchanged from `main`. The latest `main` run at https://github.com/import-js/eslint-plugin-import/actions/runs/29535690274 and this PR run each have the same 50 failing jobs (the Node 10–15 / ESLint 2–6 combinations), while all 194 other matrix jobs pass.

The focused rule test also passes locally with `npm run mocha -- tests/src/rules/no-unresolved.js`. The old-version matrix failures appear to be pre-existing dependency/runtime failures rather than failures introduced by this two-file change.

ljharb (approved):
this is a good fix, but i hope nobody's actually using no-substitution template literals anywhere

(added additional test coverage)

PR #3285, opened 2026-08-01
[Fix] `no-deprecated`: detect `@deprecated` on decorated TypeScript exports

### LINKED ISSUE #3094: no-deprecated rule does not work if there are decorators
Normally, you put the docblock at the top of the function or class. To make it detect the @deprecated tag, the docblock needs to be after the decorators. I attached an image below

Image

raisulchowdhury:
## Summary

- associate leading documentation before decorators with named and default exported declarations
- report deprecated decorated TypeScript classes through `import/no-deprecated`
- add focused regression fixtures and rule tests for named and default decorated exports

Fixes #3094.

## Validation

- targeted no-deprecated suite: 57 passing (`BABEL_ENV=test BABEL_DISABLE_CACHE=1 ./node_modules/.bin/mocha --require babel-register tests/src/rules/no-deprecated.js`)
- ESLint on changed JavaScript and TypeScript files
- `git diff --check`

The full repository suite was not run because the isolated checkout encountered disk pressure during dependency installation; the targeted suite and changed-file lint pass using the repository dependency set.

## AI assistance

This PR was prepared with AI assistance. The source change, regression tests, and changelog entry were reviewed against the existing export-map documentation contract.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This change log adheres to standards from [Keep a CHANGELOG](https://keepachange
 - support eslint v10 ([#3230], thanks [@rasmi])
 - [`no-deprecated`]: detect `@deprecated` on default-exported identifier declarations ([#3247], thanks [@mixelburg] [@etyrrell22])
 - [`consistent-type-specifier-style`]: add `prefer-top-level-if-only-type-imports` option ([#3210], thanks [@aldeed])
+- [`no-deprecated`]: detect `@deprecated` on decorated TypeScript exports ([#3094], thanks [@raisulchowdhury])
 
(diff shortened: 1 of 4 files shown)


raisulchowdhury:
Follow-up pushed in commit `a125f4b` after an independent review.

- decorated default exports now capture documentation attached to decorators as well as named exports
- added a default-decorated TypeScript regression and the required Unreleased changelog entry
- validation: targeted no-deprecated suite (57 passing), changed-file ESLint, and `git diff --check`

The issue’s `@deprecated` behavior is now covered for both named and default decorated exports.

ljharb (approved):

Issue #3221, opened 2025-11-27
import/no-cycle crashing with external ESM modules (TypeError: Cannot read properties of null)


ernestostifano:
Minimum reproduction [here](https://github.com/ernestostifano/eslint-plugin-import-issue-1) (see README).

### The Issue

***(Please note that `eslint-import-resolver-exports` has been patched in this project to provide some logs)***

1. Run `yarn lint:ok`
2. Linting is completed successfully
3. Run `yarn lint:not-ok`
4. Linting fails with the following error:
```text
Oops! Something went wrong! :(

ESLint: 8.57.1

TypeError: Cannot read properties of null (reading '/Users/username/repositories/eslint-plugin-import-issue-1/src/not-ok.js')
Occurred while linting /Users/username/repositories/eslint-plugin-import-issue-1/src/not-ok.js:1
Rule: "import/no-cycle"
    at checkSourceValue (/Users/username/repositories/eslint-plugin-import-issue-1/node_modules/eslint-plugin-import/lib/rules/no-cycle.js:116:59)
    at checkSourceValue (/Users/username/repositories/eslint-plugin-import-issue-1/node_modules/eslint-module-utils/moduleVisitor.js:32:5)
    at checkSource (/Users/username/repositories/eslint-plugin-import-issue-1/node_modules/eslint-module-utils/moduleVisitor.js:38:5)
    at ruleErrorHandler (/Users/username/repositories/eslint-plugin-import-issue-1/node_modules/eslint/lib/linter/linter.js:1076:28)
    at /Users/username/repositories/eslint-plugin-import-issue-1/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (/Users/username/repositories/eslint-plugin-import-issue-1/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/Users/username/repositories/eslint-plugin-import-issue-1/node_modules/eslint/lib/linter/node-event-generator.js:297:26)
    at NodeEventGenerator.applySelectors (/Users/username/repositories/eslint-plugin-import-issue-1/node_modules/eslint/lib/linter/node-event-generator.js:326:22)
    at NodeEventGenerator.enterNode (/Users/username/repositories/eslint-plugin-import-issue-1/node_modules/eslint/lib/linter/node-event-generator.js:340:14)
```

### The Details
[...]

PR #3287, opened 2026-08-02
[Fix] `no-cycle`: don't throw when the linted path is not on disk

### LINKED ISSUE #3286: no-cycle: TypeError aborts the lint run when the linted path is not on disk (stdin, lintText)
`import/no-cycle` throws a `TypeError` and aborts the entire lint run whenever the linted path does
not exist on disk — which is the normal case for an unsaved buffer, generated code, or
`ESLint#lintText`.

```
$ cat src/foo.js | npx eslint --stdin --stdin-filename=src/not-yet-written.js

ESLint: 9.39.5

TypeError: Cannot read properties of null (reading '/…/src/not-yet-written.js')
Occurred while linting /…/src/not-yet-written.js:1
Rule: "import/no-cycle"
```

Exit code 2, and no results are reported for any file — not just the one that triggered it.

Reproduced on a clean install of `eslint-plugin-import@2.32.0` with `eslint@9.39.5`, and present at
`main` (`1a39fb8`).

### Cause

`src/rules/no-cycle.js:73`

```js
const scc = options.disableScc ? {} : StronglyConnectedComponentsBuilder.get(myPath, context);
```

`StronglyConnectedComponentsBuilder.get` returns `null` when the path does not resolve
(`src/scc.js:16` — `if (path == null) { return null; }`). Line 116 then does `scc[myPath]`.

Two conditions are needed together, which is why it is easy to miss: the linted path must not
resolve, **and** at least one import in the file must resolve — otherwise `checkSourceValue` returns
early at the `imported == null` check before reaching line 116.

The `<text>` guard on line 63 shows non-file input was anticipated. That sentinel is the eslintrc
one; under flat config the buffer keeps whatever `--stdin-filename` supplies, so the guard no longer
covers this.

### Suggested fix

Defaulting the map to `{}` restores exactly the `disableScc` semantics: every lookup is `undefined`,
so each pair compares equal and everything is traversed. The SCC map is only an optimisation for
skipping unreachable imports, so this is safe rather than merely defensive.

```diff
-    const scc = options.disableScc ? {} : StronglyConnectedComponentsBuilder.get(myPath, context);
+    const scc = options.disableScc ? {} : StronglyConnectedComponentsBuilder.get(myPath, context) || {};
```

Happy to open a PR — I have one ready with a regression test. Confirmed the test fails without the
change (`1 failing`, same `TypeError`) and passes with it; full rule suite 2877 passing, lint clean.

Possibly related: #3221 reports the same crash site with a resolver-specific trigger, attributed
there to `eslint-import-resolver-exports`. This reproduction uses the default resolver and no extra
plugins, so the root cause may be shared.

Environment: `eslint-plugin-import@2.32.0`, `eslint@9.39.5`, Node 24.16.0, macOS.

theRizwan:
Fixes #3286.

`import/no-cycle` threw a `TypeError` and aborted the whole lint run whenever the linted path did
not exist on disk:

```
$ cat src/foo.js | eslint --stdin --stdin-filename=src/new-file.js
TypeError: Cannot read properties of null (reading '/…/src/new-file.js')
Rule: "import/no-cycle"
```

Exit code 2, and no results for any file. That is the normal case for an unsaved buffer, generated
code, or `ESLint#lintText`.

## Cause

`StronglyConnectedComponentsBuilder.get` returns `null` when the path does not resolve
(`src/scc.js:16`), and `scc[myPath]` at `no-cycle.js:116` dereferenced it.

Two conditions are needed together, which is why it has gone unnoticed: the linted path must not
resolve, **and** at least one import must resolve — otherwise `checkSourceValue` returns at the
`imported == null` check before line 116 is reached.

The `<text>` guard at line 63 anticipated non-file input, but that is the eslintrc sentinel; under
flat config the buffer keeps whatever `--stdin-filename` supplies, so it no longer covers this.

## Why `{}` is the right default

It restores exactly the `disableScc` semantics. Every lookup is `undefined`, so
`scc[myPath] === scc[imported.path]` is true and the `continue` at line 130 never fires — the rule
traverses everything, as it already does when the optimisation is switched off by config. The SCC
map only skips imports that cannot reach the current file, so an empty one is conservative rather
than merely non-crashing.

Behaviour for paths that do resolve is unchanged.

## Tests

One valid case added, using `./es6/depth-one` because both conditions above must hold — a fixture
that does not resolve, or one with no exports, returns before the crash site and would pass either
way.

Verified it genuinely catches the bug:

```
without the fix:  141 passing, 1 failing
                  TypeError: Cannot read properties of null (reading '…/does-not-exist-on-disk.js')
with the fix:     142 passing
```

Full rule suite: **2877 passing**. `eslint` clean on both changed files.

## Note

#3221 reports the same crash site with a resolver-specific trigger, attributed there to
`eslint-import-resolver-exports`. This reproduction uses the default resolver and no extra plugins,
so the root cause may be shared — worth a look when triaging that one.

### DIFF
--- a/src/rules/no-cycle.js
+++ b/src/rules/no-cycle.js
@@ -70,7 +70,13 @@ module.exports = {
       context,
     );
 
-    const scc = options.disableScc ? {} : StronglyConnectedComponentsBuilder.get(myPath, context);
+    // `StronglyConnectedComponentsBuilder.get` returns null when `myPath` does not
+    // resolve, which happens whenever the linted text has no file on disk — e.g.
+    // `eslint --stdin --stdin-filename=not-yet-written.js`, or `ESLint#lintText`.
+    // The SCC map is only an optimisation for skipping unreachable imports, so an
+    // empty one is the correct fallback: every lookup is `undefined`, which makes
+    // each pair compare equal and traverses everything, exactly as `disableScc` does.
+    const scc = options.disableScc ? {} : StronglyConnectedComponentsBuilder.get(myPath, context) || {};
 
     function checkSourceValue(sourceNode, importer, moduleSystem) {
       if (ignoreModule(sourceNode.value, moduleSystem)) {
--- a/tests/src/rules/no-cycle.js
+++ b/tests/src/rules/no-cycle.js
@@ -39,6 +39,15 @@ const cases = {
       code: 'var bar = require("./bar")',
       filename: '<text>',
     }),
+    // A filename that does not resolve to a file on disk, which is the normal
+    // case for `eslint --stdin --stdin-filename=not-yet-written.js` and for
+    // `ESLint#lintText`. `<text>` above only covers the eslintrc sentinel; flat
+    // config passes the given name through, so this used to throw a TypeError
+    // out of the rule and abort the whole run.
+    test({
+      code: 'import { foo } from "./es6/depth-one"',
+      filename: testFilePath('./cycles/does-not-exist-on-disk.js'),
+    }),
     test({
       code: 'import { foo } from "cycles/external/depth-one"',
       options: [{ ignoreExternal: true }],


theRizwan:
The failing checks all sit in the `majors` (old ESLint) matrix, and I believe they're pre-existing rather than from this change — flagging what I checked so it's less to triage.

**The same workflow is red on `main`.** Run [29535690274](https://github.com/import-js/eslint-plugin-import/actions/runs/29535690274) (`Tests: eslint (old versions)`, `main`, 16 July) has 20 failing `majors` jobs with the same name pattern, and the run before it concluded `failure` too.

**The failing set isn't stable.** Comparing that run against this PR, two jobs fail here that didn't there, and three failed there that haven't here — which looks more like flake or environment than a deterministic break.

**No failures outside that matrix.** Currently 341 passing / 30 failing, and every failure is a `majors (…)` job.

I also ran `tests/src/rules/no-cycle.js` locally against each ESLint major in that matrix:

| eslint | result |
|---|---|
| 2.13.1 | 132 passing |
| 3.19.0 | 132 passing |
| 4.19.1 | 134 passing |
| 6.8.0 | 142 passing |
| 8.57.1 | 142 passing |

All green including the new case. The differing counts are just the version-gated tests being skipped.

Happy to rebase, split the test out, or adjust the approach if you'd prefer something different.

ljharb (approved):

ljharb:
not sure what happened here or why you force pushed over my changes, but i landed them and now i can't push to the branch, so i'll just close this as merged in 7828a5f7d98f4d49e39bd6b8028f98fd0d562b48

